PR #970, opened 2014-03-05
That's what you get from not having tests.

doertedev:
With the previous solution, the filter always marked the entrance page
url link as active. Off course this should only happen if the page being
called is actually the entrance page. Either tests or coffe help here.
Coffee it was in my case!


GawainLynch:
So a year and a half later I am trying to write tests to cover the current() function… and what is the one bit I can't get to test properly! :stuck_out_tongue_winking_eye:

Issue #1725, opened 2014-09-14
pager for taxonomy does stays at first page 


ArdKuijpers:
The pager for taxonomy emits the wrong hrefs. For example with tags, currently `pager()` emits: `http://[domain]/tags/<tagname>?page_tags.<tagname>=2`. But this doesn't bring me to page two. I found out that `http://[domain]/tags/<tagname>?page=2` does bring me to the second page.

This problem also occurs with other taxonomies (e.g. categories).

mr-klaatu:
This seems to have returned...

1. http://bolt/categories/accommodation?page_category_accommodation=2 : FAILED - 1st page showing

2. http://bolt/categories/accommodation?page=2 : SUCCESS - paging works as expected.

The pager only generates the 1st type of link, so paging fails for me.

I am using Bolt V2.02

PR #1741, opened 2014-09-17
#1725 make it pager to work with general listings like taxonomies

rixbeck:
#1725


mr-klaatu:
This seems to have returned...

http://bolt/categories/accommodation?page_category_accommodation=2 : FAILED - 1st page showing

http://bolt/categories/accommodation?page=2 : SUCCESS - paging works as expected.

The pager only generates the 1st type of link, so paging fails for me.

I am using Bolt V2.0.2

PR #2295, opened 2014-12-08
Define permissions at content level

mathias213:
This patch is to add permissions set at content level in addition to those set at content type level.

To limit the complexity inside Bolt's code it was decided to use a user defined select field added to each content types you want to protect. This select will contain all groups of roles you want to be able to set up on these content type's contents.

A new role keyword is defined: "viewers". This role keyword is used in permissions.yml to protect some content types.
Ex: 

<pre>    pages:
        frontend: [ viewers ]</pre>

Then in contenttypes.yml the select is created:
Ex: this is an extract of "pages" content type declaration with addition of "viewers" select field

<pre>        template:
            type: templateselect
            filter: '*.twig'
        viewers:
            type: select
            default: anonymous
            values: [ anonymous, everyone, owner, 'admin, editor' ]
    taxonomy: [ chapters ]</pre>

Here each content of this content type would be grant-able to:
- role "anonymous": all users, a public content.
- role "everyone": only connected users
- role "owner": only owner could see the content in frontend
- roles "admin" or "editor": you would need to belong to admin OR editor to access that content.

So here was the way to extract from Bolt  a part of permissions complexity.

As most complexity is pushed on user side now the point - IMHO - would be to help users using this system. The code modification is meant to that purpose.
This modification applies only on frontend content - or at least is meant to apply only there - and it applies in two ways:
- the checkPermission() function in Permissions.php check also per content permissions when needed to add all roles contained in database field "viewers" in available roles array then these available roles are compared to current user's roles
- the executeGetContentQueries() in Storage.php is also modified to add a SQL filter when retrieving multiple contents. This filter avoid breaking limit parameter sent to requests.

So once the patch is added to the code user would need to:
- modify some content type in contenttypes.yml to add "viewers" select.
- modify this content type permissions in permissions.yml to add this internal "viewers" role to "frontend"
- set to "true" flag "frontend_permission_checks" in config.yml
  And contents will be filtered.

At this point Bolt will complain the database must be updated, if user do it Bolt will add an empty "viewers" field to this content type table structure. To avoid the whole site becomes unavailable in frontend as no "viewers" roles are defined in our empty field the default behaviour is to consider the content as public and so granted to ROLE_ANONYMOUS in case "viewers" is empty.

How listing views work:

Considering the modified content type is still "pages" with "frontend: [ 'viewers' ]" in permissions.yml.

For now all fallback behaviour (I was able to see) are set to give anonymous access to contents. The two cases are:
- listing pages (ex: /pages)
- empty "viewers" field

When a user access to /pages with "frontend: [ 'viewers' ]" in permissions.yml as there is no database item for that listing page there is no "viewers" field that why the fallback.
When twigs are building this page they call getContent() or record() which would call getContent() which finally launch executeGetContentQueries().

executeGetContentQueries() receives $decoded array which contains all we need to know if this query or list of queries are to retrieve one or more item.

When executeGetContentQueries() tries to retrieve a list of items the SQL filter is applied which returns only database items according to user's roles including ownership.

This filter applies only if these 3 parameters are defined, if one is missing, no filter, no per content permissions:
- viewed page is in frontend 
- "frontend_permission_checks: true" in config.yml
- "frontend: [ 'viewers' ]" in permissions.yml
  So it is easily disable

But it is also quiet easily applied. Bolt simplify lot of things letting us doing complex stuffs as if we were good in creating web site. This is to simplify per content filtering.

This version moves added arguments at end of functions declaration. Thanks to GDmac.

### DIFF
--- a/src/Permissions.php
+++ b/src/Permissions.php
@@ -696,4 +696,5 @@ class Permissions
         }
 
         return $contentRoles;
+    }
 }


eduardomart:
I agree this would be real useful.

SahAssar:
@eduardomart Since this proposal would be using backend permissions for the frontend I don't think this would fly anymore... But it should be very possible to do something like this with for example the members extension and a custom controller for the frontend.

mathias213:
This was never accepted by Bolt dev team even if this was fully working.
The reason of refusal was, if my memory is correct, this would be too complex to manage for Bolt admins... This assertion seems to me considering Bolt admins as dumb admins (complexity exists everywhere but should not be a limitation as long as no simpler solution is proposed, which was not the case as this thread was unburied after almost two years).

The fact there is only one user database with different roles (backend & frontend) is not irrelevant for me: right now I'm working as Active Directory architect and AD is one user DB containing lot of different kind of users.
Of course managing that need admins to think about what they are doing but that's the same with two user DB.

In fact what I don't understand is for what reason having two user DB is an improvement? Security ? Please explain.

I'm now using CMSMS + Frontend users to get private frontend pages/areas. So two user DB. So I need to create two accounts when one is really needed.

Connecting to computer system with admin is not a good idea: using root account or "domain admins" account is not recommanded at all because these kind of account have access to the whole system, they can destroy the whole system because we are connected to that system.
In web site I don't see the same risks: the frontend is different from backend in many points and as far as I know there is no virus for website, so using a backend account on frontend should not be a security risk.

As said, I don't understand. So if anyone as an explanation better than "it is not a good idea", that would be great.

A last word: almost every company website (intranet) needs frontend users filtering. Most of micro-site would need that too (I mean building a home site in your bedroom to share some stuffs with your friends, I expect at some moment you would share something a little more private and here you need two web sites when one should be neeeded). Finally no frontend users seems to me aiming only middle size web sites.

GawainLynch:
> right now I'm working as Active Directory architect and AD is one user DB containing lot of different kind of users.

That is completely incorrect. 

As much as it is easy to pull apart AD, it is still susceptible to poor implementation design, such as exposing internal catalogues. It's been a while since I did an MCSA, but this is a key concept in passing the exam.

> I'm now using CMSMS

Which lacks pretty much any security, it simply implements obscurity mechanisms. Seriously, have a look at their internal code :wink: 

> As said, I don't understand. So if anyone as an explanation better than "it is not a good idea", that would be great.

Decades of established research and advancement in software science & engineering, e.g. 
  * Domain 
  * SOLID
  * Separation of Concerns
  * …

mathias213:
> > right now I'm working as Active Directory architect and AD is one user DB containing lot of different kind of users.

> That is completely incorrect.

AD, without ADAM/LDS, is not one LDAP tree hosting almost everything?

> As much as it is easy to pull apart AD, it is still susceptible to poor implementation design, such as exposing internal catalogues. It's been a while since I did an MCSA, but this is a key concept in passing the exam.

Except to tell us you have MCSA I don't see the point of that remark. I mean AD is complex and implies security. Complex + Security = You can do bull shit. No need of MCSA to understand that : )



 > > I'm now using CMSMS

> Which lacks pretty much any security, it simply implements obscurity mechanisms. Seriously, have a look at their internal code :wink: 

Here, again, I don't understand you. You punch CMSMS for what purpose? And which version do you accuse to be almost totally unsecure? It is neither clear nor relevant (not relevant not because I won't change it because some dude on the the net wrote that's bullshit, even you are right but because that's is not the subject)

> Decades of established research and advancement in software science & engineering, e.g.

> - Domain
- SOLID
- Separation of Concerns
- …

Here, I'll believe you. My knowledge about web site management is not good enough to argue the slightest.

Anyway to be back on the subject you avoided so much: there was something working which was refused because it would have been too complex for Bolt users or admins (I think I already wrote what I think about such assertion for your users).
Now I read the bad point is it was using same user database as backend users. To that I would like to ask one little thing: how difficult it would have been to add another table to host frontend users? How long it would have taken to replace usage of table containing backend users by the new one containing front end users?

Man, I wrote that patch and I take lot of time because I'm not a dev at all. But even for me these changes would not have needed so much time: Bolt was already containing almost all I would have needed.

Finally, in two words: Bolt don't want privacy in front end.

GawainLynch:
By the way, please take my response(s) with good humour and engagement … reading yours and mine to this point it would be easy to pick up otherwise … but a healthy debate is fun so lets go :smile: 

> AD, without ADAM/LDS, is not one LDAP tree hosting almost everything?

Your assertion was "and AD is one user DB containing" … I stand by my response and your refutation of your own.

> Except to tell us you have MCSA I don't see the point of that remark. I mean AD is complex and implies security. Complex + Security = You can do bull shit. No need of MCSA to understand that : )

Apart from nearly 20 years of work experience leading architecture & security teams, not much I guess. :wink: 

Actually my simple point has less to do with the usual Nerd Ego Stroke, and more with just trying to highlight what the vendor themselves document, including require answering in their industry based exams.

> Here, again, I don't understand you. You punch CMSMS for what purpose?

Whatever was latest as of November 2015. It took no time to compromise. There is so much technical debt in that codebase that it isn't economically viable to fix.

It isn't a punch, just an observation of fact.

> Man, I wrote that patch and I take lot of time because I'm not a dev at all.

You know what … and please note all jokes/fun/ribbing are aside at this point … but that annoys me. Not for a second because *you* did any thing wrong, but because I honestly think *we* (Bolt team) failed in communicating the "why" and "what for".

> Finally, in two words: Bolt don't want privacy in front end.

We do want privacy, we just know that tying that into the backend as it is currently is a bad idea. We have `bolt/members` that is leading the charge to bring that to a usable state for 3.x and is getting close to a stable release. It is just a lot of work, authentication systems are hard (again, why we want the separation) and we want to do a good job.

Finally, in two word: Thank you :+1:

mathias213:
First thank you no taking too bad my bad mood. Please accept my apologizes for that.
Nice to see some day Bolt would have private pages, I do hope the way it will be won't implies to move object from DB table to DB table (which was the case with permissions my dumb patch relied on) and I do hope access control would be applicable on objects (any objects) rather than on page (which lack granularity).
Perhaps one day I would try Bolt again.

GawainLynch:
> First thank you no taking too bad my bad mood.

All good, we had a good, and constructive, exchange … everyone wins :smile: 

> Nice to see some day Bolt would have private pages

We have them in production now, I just wouldn't call the code ready for the general public as there are still some edge cases to clean up, and most importantly I haven't had the time to write usable documentation.

> Perhaps one day I would try Bolt again.

We'll be waiting with open arms :+1:

Issue #1263, opened 2014-06-06
'View' Permission Not Obeyed in Front End


luxlogica:
I am not sure whether what I want to do is achievable with Bolt at the moment, but in trying to get where I want, it seems I came across a problem.

I wanted to setup a record with pages akin to an 'Account' page - that is: each user in the system has their own page, and a logged in user should be able to view only their own page. Admin users should be able to view everyone's. 

I setup a 'Memberships' contenttype, and gave it the following permissions:

```
contenttypes:
    memberships:
        view: [owner, admin]
        edit: [admin]
        create: [admin]
```

Other contenttype  permission settings are:

```
contenttype-all:
    edit: [ developer, admin, chief-editor ]
    create: [ developer, admin, chief-editor ]
    publish: [ developer, admin, chief-editor ]
    depublish: [ developer, admin, chief-editor ]
    delete: [ developer, admin ]
    change-ownership: [ developer, admin ]

contenttype-default:
    edit: [ editor ]
    create: [ editor ]
    change-ownership: [ owner ]
    view: [ anonymous ]
```

Using these settings, in the **front end** any user can still see everyone else's 'membership' page - that is, if I have 3 users, 'Peter', 'Paul' and 'Mary', logged in as any of these users I can directly access the membership pages of the others in the front end.

```
http://test.site/membership/peter -- accessible by all 3
http://test.site/membership/paul -- accessible by all 3
http://test.site/membership/mary -- accessible by all 3
```

Bolt's default record template even conveniently puts in a navigation bar, so the users can navigate to each other's records simply by accessing their own...

I thought this may be happening because of the 'view' permissions in 'contenttype-default' being set to 'anonymous' (it shouldn't, because the view permissions set in the contenttype itself should be overriding it), but just in case, I commented that line out, to test it:

```
    #view: [ anonymous ]
```

Even then, all users can still see and navigate to each other's 'membership' page in the front end. 

Am I doing something obviously wrong? Is this intended behaviour - or a bug?

PR #2297, opened 2014-12-08
Add Per Content Permissions with default set to ROLE_ANONYMOUS

mathias213:
This patch is to add permissions set at content level in addition to those set at content type level.

The idea to start that change came from there: https://github.com/bolt/bolt/issues/1263 where almost all needed tool to proceed were prepared. In that thread tobias2k warned about ACLs using terms "bloat" and "creepy" as ACLs are quickly complex.

To limit the complexity inside Bolt's code it was decided to use a user defined select field added to each content types you want to protect. This select will contain all groups of roles you want to be able to set up on these content type's contents.

A new role keyword is defined: "viewers". This role keyword is used in permissions.yml to protect some content types.
Ex: 
<pre>    pages:
        frontend: [ viewers ]</pre>

Then in contenttypes.yml the select is created:
Ex: this is an extract of "pages" content type declaration with addition of "viewers" select field
<pre>        template:
            type: templateselect
            filter: '*.twig'
        viewers:
            type: select
            default: anonymous
            values: [ anonymous, everyone, owner, 'admin, editor' ]
    taxonomy: [ chapters ]</pre>

Here each content of this content type would be grant-able to:
- role "anonymous": all users, a public content.
- role "everyone": only connected users
- role "owner": only owner could see the content in frontend
- roles "admin" or "editor": you would need to belong to admin OR editor to access that content.

So here was the way to extract from Bolt  a part of permissions complexity.

As most complexity is pushed on user side now the point - IMHO - would be to help users using this system. The code modification is meant to that purpose.
This modification applies only on frontend content - or at least is meant to apply only there - and it applies in two ways:
- the checkPermission() function in Permissions.php check also per content permissions when needed to add all roles contained in database field "viewers" in available roles array then these available roles are compared to current user's roles
- the executeGetContentQueries() in Storage.php is also modified to add a SQL filter when retrieving multiple contents. This filter avoid breaking limit parameter sent to requests.

So once the patch is added to the code user would need to:
- modify some content type in contenttypes.yml to add "viewers" select.
- modify this content type permissions in permissions.yml to add this internal "viewers" role to "frontend"
- set to "true" flag "frontend_permission_checks" in config.yml
And contents will be filtered.

At this point Bolt will complain the database must be updated, if user do it Bolt will add an empty "viewers" field to this content type table structure. To avoid the whole site becomes unavailable in frontend as no "viewers" roles are defined in our empty field the default behaviour is to consider the content as public and so granted to ROLE_ANONYMOUS in case "viewers" is empty.

How listing views work:

Considering the modified content type is still "pages" with "frontend: [ 'viewers' ]" in permissions.yml.

For now all fallback behaviour (I was able to see) are set to give anonymous access to contents. The two cases are:
- listing pages (ex: /pages)
- empty "viewers" field

When a user access to /pages with "frontend: [ 'viewers' ]" in permissions.yml as there is no database item for that listing page there is no "viewers" field that why the fallback.
When twigs are building this page they call getContent() or record() which would call getContent() which finally launch executeGetContentQueries().

executeGetContentQueries() receives $decoded array which contains all we need to know if this query or list of queries are to retrieve one or more item.

When executeGetContentQueries() tries to retrieve a list of items the SQL filter is applied which returns only database items according to user's roles including ownership.

According to Permissions::isProtected() function this filter applies only if these 3 parameters are defined, if one is missing, no filter, no per content permissions:
- viewed page is in frontend 
- "frontend_permission_checks: true" in config.yml
- "frontend: [ 'viewers' ]" in permissions.yml
So it is easily disabled.

But it is also quiet easily applied. Bolt simplify lot of things letting us doing complex stuffs as if we were good in creating web site. This is to simplify per content filtering.

This version moves added arguments at end of functions declaration. Thanks to GDmac.

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -39,7 +39,7 @@ class Frontend
                 $contenttypeslug = (string) $content;
                 $contentid = null;
             }
-            if (!$app['users']->isAllowed('frontend', $contenttypeslug, $contentid)) {
+            if (!$app['users']->isAllowed('frontend', $contenttypeslug, $contentid, $content)) {
                 $app->abort(403, 'Not allowed.');
             }
         }
--- a/src/Permissions.php
+++ b/src/Permissions.php
@@ -36,6 +36,14 @@ class Permissions
      */
     const ROLE_OWNER = 'owner';
 
+    /**
+     * A special role that is used to tag the viewers of a resource when some
+     * field is defined in database for some contenttype
+     * granting possibility to set percontent permissions in addtion
+     * of per contenttype permissions.
+     */
+    const ROLE_VIEWERS = 'viewers';
+
     private $app;
 
     // per-request permission cache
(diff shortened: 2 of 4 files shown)


bobdenotter:
I'm doing a bit of spring cleaning, and i'm going to close this one. Related, the `frontend: [ .. ]` will be removed, because it's deemed confusing, and complicated. I can only imagine how complex (mainly for the users) a "Per Content Permissions" strategy would be.

Using a separate contenttype, or a front-end authentication solution would make more sense, in the philosophy of Bolt. 

See #3087 for related discussion.

Thanks, though! :-)

PR #2383, opened 2014-12-31
Set default width/height of showImage() to '0'

### LINKED ISSUE #2372: calculate height according to aspect ratio if it is not supplied in a template image tag
when using template image tags like `thumbnail` or `showimage`, it would be super userful if the height was calculated automatically according to the aspect ratio if only the width is supplied.

right now you need to:
```twig
{% set info = imageinfo(record.thumbnail) %}
{% set ratio = info.height / info.width %}
{{ record.thumbnail|showimage(400, (400 * ratio)|round) }}
```
or else the height will be 100

rarila:
Resolves #2372 part 2


bobdenotter:
:+1:

PR #2388, opened 2015-01-03
Import macro instead of direct usage of _self in theme base-2013/14

### LINKED ISSUE #2328: Use of `_self` for twig macros should not be encouraged
I see in the cheatsheet and in the sample theme that comes pre-installed with bolt, you are using the `_self` object in twig templates to utilize a macro defined in its own file. (See https://github.com/bolt/bolt/blob/master/theme/base-2014/_sub_menu.twig#L10 for example).

From the Twig documentation (http://twig.sensiolabs.org/doc/tags/macro.html):
> When you define a macro in the template where you are going to use it, you might be tempted to call the macro directly via _self.input() instead of importing it; even if seems to work, this is just a side-effect of the current implementation and it won't work anymore in Twig 2.x.

It is definitely nice to be able to declare macros within the same file you use them, so perhaps it is worth the "un-pureness" of this approach (especially if you know that you will never be upgrading Twig to version 2.0). Just thought I'd bring it up in case you were unaware that this is not recommended by the Twig folks.

rarila:
Fixes #2328


bobdenotter:
Cool! Glad we could fix that relatively easy. :+1:

PR #2389, opened 2015-01-03
Fix omitting parentheses issue

rarila:
As criticised by @bobdenotter  in [#2378](https://github.com/bolt/bolt/pull/2378#discussion_r22375739)
I think that is the way to go, ok?


bobdenotter:
Sweet!

PR #2390, opened 2015-01-03
Twig cleanup

rarila:
Mostly adding or removing spaces – OCD


bobdenotter:
I don't see anything that could break with this! Merging in!

PR #2391, opened 2015-01-03
duplicate index error on new installs

winglian:
On new installs, it currently causes a duplicate index error.


GawainLynch:
Nice catch, thank you!

PR #2396, opened 2015-01-03
Use null defaults for dates

### LINKED ISSUE #2395: Password reset links failing on PostgreSQL
```
An exception occurred while executing 'UPDATE bolt_users SET password = ?, shadowpassword = ?, 
shadowtoken = ?, shadowvalidity = ? WHERE id = ?' with params ["$P$F <removed> ", "", "", 
"0000-00-00 00:00:00", 2]: 
SQLSTATE[22008]: Datetime field overflow: 7 ERROR: date/time field value out of range: 
"0000-00-00 00:00:00"
Code: 0
```

GawainLynch:
Fixes #2395 

### Changelog
* Fields that previously used `0000-00-00` and `1900-01-01` now default to using `NULL` instead

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -210,9 +210,9 @@ class Content implements \ArrayAccess
                 // @todo Try better date-parsing, instead of just setting it to
                 // 'now' (or 'the past' for datedepublish)
                 if ($key == 'datedepublish') {
-                    $value = date("1900-01-01 00:00:00");
+                    $value = null;
                 } else {
-                    $value = date("Y-m-d H:i:s");
+                    $value = date('Y-m-d H:i:s');
                 }
             }
         }
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -969,14 +969,14 @@ class Backend implements ControllerProviderInterface
 
         $duplicate = $app['request']->query->get('duplicate');
         if (!empty($duplicate)) {
-            $content->setValue('id', "");
-            $content->setValue('slug', "");
-            $content->setValue('datecreated', "");
-            $content->setValue('datepublish', "");
-            $content->setValue('datedepublish', "1900-01-01 00:00:00"); // Not all DB-engines can handle a date like '0000-00-00'
-            $content->setValue('datechanged', "");
-            $content->setValue('username', "");
-            $content->setValue('ownerid', "");
+            $content->setValue('id', '');
+            $content->setValue('slug', '');
+            $content->setValue('datecreated', '');
+            $content->setValue('datepublish', '');
+            $content->setValue('datedepublish', null);
+            $content->setValue('datechanged', '');
+            $content->setValue('username', '');
+            $content->setValue('ownerid', '');
             $app['session']->getFlashBag()->set('info', Trans::__('contenttypes.generic.duplicated-finalize', array('%contenttype%' => $contenttype['slug'])));
         }
 
--- a/src/Database/IntegrityChecker.php
+++ b/src/Database/IntegrityChecker.php
@@ -409,7 +409,7 @@ class IntegrityChecker
         $authtokenTable->addColumn('lastseen', 'datetime', array('notnull' => false, 'default' => null));
         $authtokenTable->addColumn('ip', 'string', array('length' => 32, 'default' => ''));
         $authtokenTable->addColumn('useragent', 'string', array('length' => 128, 'default' => ''));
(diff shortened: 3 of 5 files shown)


bobdenotter:
See! I _knew_ there was a reason why i originally took 1900 as default dates for everything. :-)

GawainLynch:
> However, is "0001-01-01" acceptable too in postgres

Honestly, I find it quite offensive that all major database engines don't measure the [Planck epoch](http://en.wikipedia.org/wiki/Planck_epoch).

OCD aside, `1900-01-01 00:00:00` is our defacto standard for date minimums, and this particular one is used to determine last logon (or something like that - it's late here)...  If we are going to change this, can we RFC and do it across the board in a standard fashion instead, as this is a breaking bug for PostgreSQL users in the mean time :sob:

rarila:
For me it would be much more logically to set such values to ``null``.

GawainLynch:
> For me it would be much more logically to set such values to ``null``.

Agreed, but let's:
1. Fix actual bug first
2. Decide/agree on standard way forward
3. Refactor
4. Drink #beer or go to gym

bobdenotter:
The thing with `0000-00-00 00:00:00` is, that it's parsed in PHP as "november 30, anno domini -1". Which is stupid.

 - `01-01` -> Jan 1st. 
 - `01-00` -> Dec 1st.
 - `00-00` -> Nov 30th. 

So, `0001-01-01` would be better at least. Or `null`, if MysQL, Postgres and SQlite all accept that.

rarila:
MySQL and Postgres should do fine and SQlite [also](https://www.sqlite.org/nulls.html)

GawainLynch:
So I should expand this to include all uses of  `0000-00-00 00:00:00` and `1900-01-01 00:00:00` to be converted to `0000-01-01 00:00:00`?

rixbeck:
Isn't possible to set date field's default value at schema definition?

rarila:
@rixbeck ``create table foo (foocol datetime null default null);``

rixbeck:
@rarila, got it. I meant with not null, valid date defaults.

rarila:
@rixbeck ``create table foo (foocol datetime not null default '0000-01-01 00:00:00');`` ?

rixbeck:
Let's say

rarila:
Still ``null`` is unambiguous in any way. If you have a column ``lastlogin`` and get back ``null`` everybody knows immediately without reading any docs there was no login sofar. Having some special dates makes you either wonder, how old that person is ;-) or what this f*#+ing value was again meaning no login.

rarila:
Btw, just read that on real databases (so not MySQL;-) processing columns with ``null`` values is even faster.

GawainLynch:
Yes, but can we agree we are going to change this across the board, **not just** for one obscure bug that only affects people who run PostgreSQL and have editors that forget their password all the time..

rarila:
@GawainLynch fully agreed

rossriley:
Ref: https://github.com/bolt/bolt/pull/2317

GawainLynch:
Moved all `1900-01-01` dates and fields to default nulls.  Tested:
  * New site creation
  * Pre-filled records publication
  * Normal publication
  * Timed publication
  * Timed depublication
  * Lockout/throttle for more than 5 failed password attempts
  * Password reset
  * No crying kittens

rarila:
A few questions: Am I right that these are mostly changes to the database scheme? Isn't there some handling code to be changed also or are these dates only written but not used or checked anywhere? How about old values already stored?

GawainLynch:
> Am I right that these are mostly changes to the database scheme? 

Both, these schema and the code that relies on it

> Isn't there some handling code to be changed also or are these dates only written but not used or checked anywhere? 

Yes, should be included

> How about old values already stored?

Work fine, I left the checks to still account for '1900-01-01' values

rarila:
Ah, cool! :clap:
I wonder if we should lay some foundation to be able to change values inside an extended integrity check, so that values like '1900-01-01' could be replaced in the future. I think we still support some old db structures or say the username instead of id in codebase.

GawainLynch:
Yeah, I want to address that too...  Just wasn't sure the right way to go about it.

bobdenotter:
> I want to address that too...

In this PR, or in a separate one? :-)

GawainLynch:
> In this PR, or in a separate one

Separate one please...  I might roll that into the auto-updater work (circa 2.2 now)

bobdenotter:
So, this is looking good to me. Merging in.

PR #2415, opened 2015-01-07
Fix reference of stub functions

CarsonF:


bobdenotter:
Travis failed, but it seems like that's due to a network error on travis' side. Merging in, thanks!

PR #2420, opened 2015-01-08
Change priorities of taxonomy templates

Pinpickle:
For everything except taxonomies, the template priority is something like:

    global config < theme config < specific type config

Taxonomies go

    global config < specific type config < theme config

This is unexpected behaviour, and it makes it impossible to specify a custom taxonomy template if you have a listing template specified in your theme (which is pretty common). On top of that, the function's code for choosing taxonomy was slightly inconsistent with all of the other template choosing functions.

Anyway, it was a simple fix. This PR changes the priorities to the expected behaviour.

### DIFF
--- a/src/TemplateChooser.php
+++ b/src/TemplateChooser.php
@@ -129,19 +129,20 @@ class TemplateChooser
      */
     public function taxonomy($taxonomyslug)
     {
-        // Set the template based on the (optional) setting in taxonomy.yml,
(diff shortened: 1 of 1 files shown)


bobdenotter:
Agreed! This makes more sense. Thank you!

GawainLynch:
#karma @Pinpickle

PR #2423, opened 2015-01-09
Also use XMLHttpRequest to detect AJAX requests in getWhichEnd()

GawainLynch:
c.f. [Symfony\Component\HttpFoundation\Request](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Request.php#L1668)

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -910,11 +910,10 @@ class Config
 
         // If the request URI is '/bolt' or '/async' (or starts with '/bolt/' etc.), assume backend or async.
         $mountpoint = '/' . ltrim($mountpoint, '/');
-        if (isset($_SERVER['HTTP_X_REQUESTED_WITH'] ) && $_SERVER['HTTP_X_REQUESTED_WITH'] === 'XMLHttpRequest')
-            || elseif ($scripturi === '/async' || strpos($scripturi, '/async/') === 0) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Travis says no. I think the relevant section should be like this: 

```
        // If the request URI is '/bolt' or '/async' (or starts with '/bolt/' etc.), assume backend or async.
        $mountpoint = '/' . ltrim($mountpoint, '/');
        if ( (isset($_SERVER['HTTP_X_REQUESTED_WITH']) && $_SERVER['HTTP_X_REQUESTED_WITH'] === 'XMLHttpRequest')
            || $scripturi === '/async' || strpos($scripturi, '/async/') === 0) {
            $end = 'async';
        } elseif ($scripturi === $mountpoint || strpos($scripturi, $mountpoint . '/') === 0) {
            $end = 'backend';
        } else {
            $end = 'frontend';
        }
```

GawainLynch:
>Travis says no

Oh some days I just feel so special...  :panda_face:

rossriley:
Just a note that referring to the $_SERVER global directly will break any tests or constructed requests. We should be able to refer to `$request->server->get("HTTP_X_REQUESTED_WITH")` from the request object and all should work the same.

bobdenotter:
@rossriley Often true, but this has given us trouble before. See #2243 for example. In this case it should be ok to use `$request->server->get`, which would allow us to skip the `issets()`, making it easier to read.

bobdenotter:
Whoop, merging in!

PR #2430, opened 2015-01-09
Refactor Twig setup

CarsonF:
The problem I had is twig was getting initialized to add globals and extensions before my customizations were added.

Twig shouldn't be initialized until it is needed, so others can extend it further.
I also moved the globals into the TwigExtension class, which also dries them up. Same with the SetcontentTokenParser.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -185,6 +185,13 @@ class Application extends Silex\Application
                 )
             )
         );
+        // Add the Bolt Twig Extension.
+        $this['twig'] = $this->share($this->extend('twig', function(\Twig_Environment $twig, $app) {
+            $twig->addExtension(new TwigExtension($app));
+            return $twig;
+        }));
+
+        $this->register(new Provider\SafeTwigServiceProvider());
 
         $this->register(new Provider\RenderServiceProvider());
         $this->register(new Provider\RenderServiceProvider(true));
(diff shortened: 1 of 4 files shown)


bobdenotter:
With the debug toolbar enabled, I'm getting errors on every page: 

```
Template "@WebProfiler/Profiler/info.html.twig" is not defined ().
```

Does this work correctly on your end?

screenshot 2015-01-10 10 27 41

bobdenotter:
After I reset my git-checkout, it works as it should! Merging in, thank you!

GawainLynch:
I am having this issue on *all* of my clones of master.  Can we revert please?

CarsonF:
I can try to debug this on Monday. The only thing I can think of is something is calling for twig before  this code is executed. That would be fine for the old code and not for the new code

GawainLynch:
No stress, it is just that we need to keep master always-on post-2.0.0 and this breaks stuff.

I am on limited time at the moment, so didn't have time to debug myself. :grin:

bobdenotter:
Weird.. I'm going to merge in the revert for now, so we can figure out properly what's going on.

GawainLynch:
@CarsonF I think you are on the right track. O e of the things I did notice on one install where I am unit testing the Composer refactor is that the CommandRunner stuff would get called on Dashboard... Something that most certainly shouldn't happen. But I was getting these errors on two other vanilla installs as well, so not limited to my branch.

Increasingly, as I type, I think the Extend loading is responsible. We had some ugly bugs early in 2.0-dev cycle to do with loading and i know we didn't get all cases down. 

If you have the chance to look deeper, it would be much appreciated!

CarsonF:
Ok I think I found [it](https://github.com/bolt/bolt/blob/master/src/Extensions.php#L292). Extensions are added before App::initAfterHandler is called. Extensions can have twig extensions which currently initializes twig.

@bobdenotter I also see [here](https://github.com/bolt/bolt/blob/master/src/Extensions.php#L253) a global variable is added to twig. The commit (b834110) message says "Allow for better detection of installed extensions, without breaking stuff." Can you explain where this global is used?

PR #2431, opened 2015-01-09
Url matcher updates

CarsonF:
The UrlMatcher was doing duplicate logic as Silex's RedirectableUrlMatcher. 
I just extended it which removed the need for a lot of the code.


bobdenotter:
Less code! \0/

Thanks again!

PR #2435, opened 2015-01-10
Fix "simpleredirect" function. Fixes #2408

### LINKED ISSUE #2408: Redirect does not work in Twig
If you try to redirect from Twig. You'll get the following error:
```
Bolt - Fatal error.
Class: Twig_Error_Runtime
Message: An exception has been thrown during the rendering of a template ("Redirecting to 'http://www.google.com/'.") in "entry.twig" at line 12.
Code: 0
```

It used to be a function `simpleredirect` now it's in `Bolt\Library`, so it might've been something to do with that.

See [Simpleforms Issue #2](https://github.com/jadwigo/SimpleForms/issues/2).

bobdenotter:


GawainLynch:
LGTM

PR #2440, opened 2015-01-11
JS/(CSS) reorganization

rarila:
- Added grunt-contrib-cssmin
- Relocated JS libs
- Built and usage of one unified/minifed ``lib.css``
- Built and usage of one unified/uglified ``lib.min.js``
- JsHinted and included ``bolt-extend.js`` into ``bolt.min.js``, also fixd a few bugs
- Extracted single plugins out of ``jquery.plugins.js`` for easier handling/updating/removing
- Tried to find out version numbers and if internal patches were done
- …

``ckeditor`` is not included in these changes sofar, could get a it complicated (plugins, …)


bobdenotter:
Amazeballs.

PR #2449, opened 2015-01-12
Translated Routing Setup to new keysystem

chilion:
I wanted to see how I should translate these things and how the process would work. Rarila did help me bigtime.


rarila:
Thanks, I really would love that someone brings that work started a while ago to an good end! I really appreciate it!

bobdenotter:
Merging in! Please keep them coming. ;-)

PR #2452, opened 2015-01-12
Refactor Twig setup v2

CarsonF:
#2430 - With bug fixes.


bobdenotter:
Ok, looks good to me. I've only tried it with a few extensions, though.. Will do some more testing, and see if there are extensions out there that try to manipulate `$this->app['twig']` in initialisation.

bobdenotter:
Merging in!

PR #2453, opened 2015-01-12
Refactor db initialization methods to separate methods. 

CarsonF:
Primarily so the database service can be extended before creation with the connection check.


bobdenotter:
Looks good to me! Thanks.

Issue #2457, opened 2015-01-13
Change Administrator name gives error


chilion:
When you change the administrator name you get a nice error. See also screenshot.
Refresh the page and everything works fine, but you shouldn't get the error ofc.

With debugging disabled:
![image](https://cloud.githubusercontent.com/assets/1157531/5719935/a1636fb0-9b1d-11e4-8ce6-8bb42a419332.png)

With debugging enabled:
![image](https://cloud.githubusercontent.com/assets/1157531/5719957/da4e18f2-9b1d-11e4-8bca-294440040e0d.png)

bobdenotter:
I think this is related to #2456

bobdenotter:
I've not been able to reproduce this. Doesn't seem to be related to #2456 after all. Any more info on how / when / why this happens?

chilion:
Everytime I change the account that I'm logged in with AND that has root privileges I get this error. Might be my server or something, you can login to mine if you want to, will send you credentials.

xiaohutai:
This happens when editing the current user (the one who's logged in). Then you need to re-login with the new credentials (username) and then everything is fine.

bobdenotter:
Ah, I see! Will fix!

xiaohutai:
Hm, you get the same message when creating the first user:
> You've been logged off, because you changed your own username. Please log on again.

PR #2467, opened 2015-01-14
Fix for #2457. Remove changed self message when creating first user.

xiaohutai:
#2457 - Removes the changed own username message when creating first user.


bobdenotter:
I bumped into this as well, but you already fixed it. \0/

xiaohutai:
Aww yiss \o/

PR #2470, opened 2015-01-14
Fix debug check in TwigExtension

CarsonF:
Debug flag shouldn't be checked from config, because config assigns it to app['debug'] which can be overridden by others.


bobdenotter:
Good catch. Merging in! ^_^

PR #2472, opened 2015-01-14
Separate table queries for MySQL and Postgres

### LINKED ISSUE #2471: Different table queries for MySQL and Postgres
Per discussion with @rossriley on IRC, line 2934 of Storage.php should be rewritten. As-is it attempts to use the same query for MySQL and Postgres, but using table_catalog in WHERE on MySQL on some shared hosts causes significant performance issues.

https://github.com/bolt/bolt/blob/61bb2f0a6eeb8f2116ec3e4388c7a7178040200d/src/Storage.php#L2934

dpfavand:
Fixes #2471 
I saw vast MySQL performance increases after implementing this change.

### DIFF
--- a/src/Storage.php
+++ b/src/Storage.php
@@ -2928,10 +2928,14 @@ class Storage
         if ($dboptions['driver'] == 'pdo_sqlite') {
             // For SQLite:
             $query = "SELECT count(*) FROM sqlite_master WHERE type='table' AND name='$name';";
-        } else {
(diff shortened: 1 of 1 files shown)


GawainLynch:
The query works for me on PostgreSQL and the PR doesn't seem to break my Bolt installs using it.

bobdenotter:
Yup, looks good. No apparent breakage here either. ^_^

PR #2473, opened 2015-01-14
setcontent first order is wrong

CarsonF:


bobdenotter:
As discussed on IRC.. Thanks!

Issue #2448, opened 2015-01-12
Filter for templateselect doesn't work


SlowFox089:
In order to let the filter field take effect for the templateselect in contenttypes.yml, the listTemplates-function in TwigExtensions should use the filter.

In my install I changed the listTemplate-function to:

        $finder = new Finder();
        if ($filter == "") {
           $filter='/^[a-zA-Z0-9]\V+\.twig$/';
        }
        $finder->files()
               ->in($this->app['paths']['themepath'])
               ->depth('== 0')
               ->name($filter)
               ->sortByName();

        $files = array();
        foreach ($finder as $file) {
            $name = $file->getFilename();
            $files[$name] = $name;
        }

        return $files;

GawainLynch:
Whoops... My bad.

bobdenotter:
> Whoops... My bad.

As in, you'll take this on?

bobdenotter:
See #2477

SlowFox089:
Updated my first post, because now I can use the suposed syntax in contenttypes.yml. Somehow I made a mistake the first time I tried. So now it should work with all options like mention in #2477.

bobdenotter:
#2477 also fixes the issue, right? Or are you still having issues?

GawainLynch:
Assuming fixed and closing.  Please re-open if problems are still there for you.

PR #2477, opened 2015-01-15
Makes filters work for templateselect fields.

xiaohutai:
Hm, can't find anything about `filter` in the docs but it's there by default in `contenttypes.yml` as `*.twig`. This is quite a useful feature to have. E.g. use `filter: invoice*.twig` to only select invoice templates. Not sure about its security implications though.

### DIFF
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -693,11 +693,16 @@ class TwigExtension extends \Twig_Extension
             return null;
         }
 
+        $name = '/^[a-zA-Z0-9]\V+\.twig$/';
+        if ($filter) {
+            $name = $filter;
+        }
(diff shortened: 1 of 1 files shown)


bobdenotter:
Is this different than #2448 ?

xiaohutai:
It's a bit different. #2448 (only) uses regexes.

According to the [Finder docs](http://symfony.com/doc/current/components/finder.html):
> The `name()` method accepts globs, strings, or regexes

So I suppose having all these options available would be the most flexible. Besides if `*.twig` is the default, you'd need this for backward compatibility.

bobdenotter:
Right! Topkek! ^_^

PR #2480, opened 2015-01-15
Extendable Config

CarsonF:
I'm open to talking about some of these changes.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -19,18 +19,23 @@ class Config
 {
     protected $paths;
 
-    private $app;
-    private $data;
-    private $defaultConfig = array();
-    private $reservedFieldNames = array(
+    protected $app;
+    protected $data;
+    protected $defaultConfig = array();
+    protected $reservedFieldNames = array(
         'id', 'slug', 'datecreated', 'datechanged', 'datepublish', 'datedepublish', 'ownerid', 'username', 'status', 'link'
     );
 
-    private $cachetimestamp;
+    protected $cachetimestamp;
 
+    /**
+     * Use {@see Config::getFields} instead.
+     * Will be made protected in Bolt 3.0.
+     * @var Field\Manager
+     */
     public $fields;
 
-    private $yamlParser = false;
+    protected $yamlParser = false;
 
     /**
      * @param Application $app
@@ -64,7 +69,7 @@ class Config
      * @param  string $path     The (optional) path to the YAML file
      * @return array
      */
-    private function parseConfigYaml($filename, $path = false)
+    protected function parseConfigYaml($filename, $path = false)
     {
         // Initialise parser
         if ($this->yamlParser === false) {
@@ -561,7 +566,7 @@ class Config
     /**
      * Assume sensible defaults for a number of options.
      */
-    private function setDefaults()
+    protected function setDefaults()
     {
         $this->defaultConfig = array(
             'database'                    => array('prefix' => 'bolt_'),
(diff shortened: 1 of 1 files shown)


GawainLynch:
Not sure I am seeing the valid use case here, can you possibly expand a bit?

rixbeck:
The question is was there any reason for @bobdenotter to declare them with private? I suppose there was. Generally I prefer the practice declaring protected members except if I have a good reason for privates. It helps to keep OOP's extendability and good for possible unit testing. I can't see drawbacks if we are eliminating privates as much as possible and necessary, although I'm not sure in which stage of development phase, in which version. Not sure also but maybe @rossriley is planning to implement a bigger bite of unit test on Bolt so these kind of changes may necessary sooner or later, who/he knows.

rixbeck:
...also a use case would be welcomed! :)

CarsonF:
I agree with @rixbeck. I always default to protected variables and functions, unless I have a good reason to make it private. Thus arguments should be made to make variables and functions private, not the other way around.

The use case I have at the moment is master/slave database connections, which are pulled from another location. I want to override getDBOptions but the data variable is private. I don't want to bypass all the logic that method is doing, but the data has to be there first to be worth while.

I don't want to be limited in the future because something is private. @GawainLynch what's your reasoning for not wanting them to be protected?

GawainLynch:
> what's your reasoning for not wanting them to be protected

You said you were open to discussion but hadn't provided any reference here. I hadn't stated an opinion one way or the other lacking that.

CarsonF:
Yeah, sorry I was being lazy :) 

On a similar note: It's better practice to return results instead of assigning them to a class variable, like setDefaults(). It makes it easier when testing and it can be extended without having to have access to the class variable. 

Also, I think the constructor should be broken out into another method. Currently, it violates single responsibility.

bobdenotter:
> The question is was there any reason for @bobdenotter to declare them with private? I suppose there was.

Mostly lazyness on my part. Whenever I don't want something `public`, i'll just make it `private`. I'll put more thought into this, in the future. ^_^

bobdenotter:
Merged manually, and I accidentally stole the credits! Whoops, sorry! 

https://github.com/bolt/bolt/commits/canary

GawainLynch:
> Merged manually, and I accidentally stole the credits! Whoops, sorry!

Shouldn't be a problem, this branch will be re-pulled into master when ready...  Don't delete is, @CarsonF :grinning:

CarsonF:
Are you merging feature branches into master after they are merged into
canary? I thought canary would just be merged into master

GawainLynch:
Yes, keep your feature branch until it lands in master please :-)

CarsonF:
Ok something isn't adding up here. I based my feature branch on canary. How
is that going to get merged into master without all of the changes below
mine in canary coming with it? Like Ross' tests.

GawainLynch:
You're, like me with the Composer stuff, dependant on Ross' tests merging down to master first.

CarsonF:
So what happens when Ross' changes have lots of bugs and yours are ready to
go? You HAVE to wait for him to go first?

You're, like me with the Composer stuff, dependant on Ross' tests merging
down to master first.

—
Reply to this email directly or view it on GitHub
<https://github.com/bolt/bolt/pull/2480#issuecomment-70655346>.

GawainLynch:
In my case, I need to modify tests.  So I guess I really am not dependant, as I am pushing that to @rossriley's testing branch anyway.  Either way, we figured that, on the overall, these are corner cases.  

We could design process perfection, but it would be impossible to use :grinning:

rarila:
I hope someone finds out how things are supposed to work in the not to far future and is able to explain it in a simple way to the dumb. I'm a bit [insert that animated panic Beaker gif here].

PR #2483, opened 2015-01-16
Changed visibility of the frontend permission check so derived classes can keep using this.

TrafeX:
I'm using my own controller, but I still want to use the frontend permission check.
Currently you're forced to copy the whole function to your own controller, this PR solves that by changing the visibility to protected.

I think the impact will be minimal, but I'm not sure.


bobdenotter:
beaker

bobdenotter:
I agree, don't think this will break anything. Thanks!

PR #2485, opened 2015-01-16
Refactor Config

CarsonF:
Most of this is breaking getConfig() into separate methods. 
I made resources a class variable since it is used all over the class.

I'm not sure if anything in here is considered a BC break. BC only counts for using the class right, not extending it? 
I ask because I changed some of the methods to now return their results instead of setting them to class variables.
This wouldn't affect users of the class only people who are extending it.


GawainLynch:
First comment is that I would consider this in need of time in `canary` that @rossriley is setting up.

bobdenotter:
Just a quick update: Once #2488 has landed, we'll merge this into the 'canary' branch. Soon! :-)

rixbeck:
:+1:

CarsonF:
Well I just rebased onto canary. I'm closing this PR and making a new one for canary branch

PR #2486, opened 2015-01-16
Disable CodeMirror whitespace formatting in favour of HTML Writer

### LINKED ISSUE #2481: CKEditor removes whitespace in <pre> tags
- Set `wysiwyg->codesnippet` to true so you can add code snippets.
- Make a contenttype with an html field.
- Add a new record, and add some code (with indents) to the html.
- Click the source button in the editor after you add the code.
- Watch in horror as your indents mysteriously disappear

This looks to be an issue with Bolt's specific implementation of CKEditor because the [demo page](http://sdk.ckeditor.com/samples/codesnippet.html) for Code Snippet doesn't have this issue.

Pinpickle:
Fixes #2481 

What I did was disable CodeMirror's whitespace formatting completely for a [different plugin](http://ckeditor.com/addon/htmlwriter). This fixes a lot of problems I've been having beyond the `<pre>` issue. Whitespace was randomly cropping up in areas which was mucking up with my formatting. CodeMirror's syntax highlighting is still used.

~~However, this breaks the icons in CKEditor. It looks like there has been an `!important` added to the icon style declarations to the per element `style=` are overridden. Not sure what to do about this, or what is the real root of the problem. It also looks like the icon sprites themselves have changed and this build of CKEditor has removed a bunch of plugins (despite the fact that I used the link in the current build config file). I feel like I've overridden a setting somewhere.~~

~~By broken, I mean that they're just using the wrong icon:
image~~

I think everything is OK now!


bobdenotter:
Hmm, codemirror does more than _just_ formatting the HTML. I think it's main purpose is to make the HTML legible, and color-coded and whatnot. 

Also, there's the issue that some people have brought up, where CKEditor inserts &nbsp; in places where it shouldn't. I think this issue is a bit more complex to fix properly than just removing codemirror. 

I'm leaning towards disabling _both_ `codemirror` and `codesnippet` by default, and document the caveat of enabling them both. I didn't know about `HTML writer`, and it seems like a good basic formatter.

Pinpickle:
Sorry @bobdenotter, I worded that all very poorly.

What I have done is removed CodeMirror's *whitespace formatting* completely. The syntax highlighting still works. HTML Writer, in my opinion, producers cleaner looking markup (not a line break for every new tag, mainly just block elements but it is completely configurable). This prevents the random insertion of whitespace that I was getting quite a lot with CodeMirror. Here is a screenshot of the source area for reference:

image

rarila:
Hmm, the formatting really looks nicer.

bobdenotter:
Ah, I completely missed that. Yes, this looks _much_ nicer! Thank you!! :-)

bobdenotter:
@Pinpickle Could you do an merge upstream? I'm getting a conflict in the generated file `app/view/js/bolt.min.js`.

Pinpickle:
Done! Minified files do get a bit sticky with merges.
If you get another conflict, just empty out the file and run `grunt` to regenerate it.

bobdenotter:
Thanks! Merging in!

PR #2488, opened 2015-01-18
Test improvements

rossriley:
OK, consider this the inaugural PR against canary.

Notes for using the test suite. There are still some problems in running the tests as an entire suite, runnin in process isolation makes it far too slow so don't be surprised initially to see a few errors and failures.

Running the tests at module level do all pass at this point, so for instance to test storage run:
`phpunit tests/Storage`
or for all extensions related tests:
`phpunit tests/Extensions`

This takes us to just over 60% test coverage so still plenty more to go, but most of the key areas are tested now.


bobdenotter:
As I've said before. This is epic! :crown:

PR #2494, opened 2015-01-18
Use Twig’s resolveTemplate instead of file_exists in Frontend Controller.

lavoiesl:
Instead of doing a manual search for a template prior to actually loading the template, use Twig’s built-in function.

Advantages include:

 1. Doing the same check as the real loader a couple lines below.
 2. Using Twig’s path loader.

This PR is needed to make https://github.com/wemakecustom/bolt-parent-theme work for base templates (not only includes).

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -112,7 +112,7 @@ class Frontend
             self::checkFrontendPermission($app, $record);
         }
 
-        return $app['render']->render($template);
+        return self::render($app, $template, 'homepage');
     }
 
     /**
@@ -157,19 +157,6 @@ class Frontend
         // Then, select which template to use, based on our 'cascading templates rules'
         $template = $app['templatechooser']->record($content);
 
-        // Fallback: If file is not OK, show an error page
-        $filename = $app['paths']['themepath'] . "/" . $template;
-        if (!file_exists($filename) || !is_readable($filename)) {
-            $error = sprintf(
-                "No template for '%s' defined. Tried to use '%s/%s'.",
-                $content->getTitle(),
-                basename($app['config']->get('general/theme')),
-                $template
-            );
-            $app['log']->setValue('templateerror', $error);
-            $app->abort(404, $error);
-        }
-
         // Setting the canonical path and the editlink.
         $app['canonicalpath'] = $content->link();
         $app['paths'] = $app['resources']->getPaths();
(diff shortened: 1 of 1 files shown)


bobdenotter:
I had never heard of this function. Looks good, though! I'll run a quick test tomorrow, and merge it in, unless somebody beats me to it! ^_^

lavoiesl:
I realized I only fixed one function. I redone the PR to always use (the new) `Controllers/Fronted::render`.  It has an added benefit of reducing quite noticeably the number of lines.

src/Controllers/Frontend.php
```diff
@@ -437,14 +385,31 @@ public static function template(Silex\Application $app, $template)
...
         }
 
-        $themePath    = realpath($app['paths']['themepath'] . '/');
-        $templatePath = realpath($app['paths']['themepath'] . '/' . $template);
+        return self::render($app, $template, $template);
+    }
 
-        // Verify that the resulting template path is located in the theme directory
-        if ($themePath !== substr($templatePath, 0, strlen($themePath))) {
-            throw new \Exception("Invalid template: $template");
```
lavoiesl: You may notice that this check is gone, I don’t believe it is necessary to limit to current theme, especially if we want to support parent themes or having extensions adding custom templates. It could be argued that a check could be added to at least be in the themes directory.

src/Controllers/Frontend.php
```diff
@@ -437,14 +385,31 @@ public static function template(Silex\Application $app, $template)
...
+     * @param  Silex\Application $app
+     * @param  string            $template   Ex: 'listing.twig'
+     * @param  string            $title      '%s' in "No template for '%s' defined."
+     * @return mixed                         Rendered template
+     */
+    private static function render(Silex\Application $app, $template, $title)
+    {
+        try {
+            return $app['twig']->render($template);
+        } catch (\Twig_Error_Loader $e) {
```
lavoiesl: `Twig_Error_Syntax` is not handled, it could be added in the future, but it would have to handle the case where the 404 page itself triggers an error (like in `_header.twig`)

bobdenotter:
This is lovely! Getting rid of a few `file_exists`, reducing duplicated code, having less code overall. Thank you, again!

lavoiesl:
Excellent, once this get released, my extension will work and I’ll be able to develop base theme(s): https://github.com/wemakecustom/bolt-parent-theme.

bobdenotter:
That looks great! :-)

We'll be releasing a 2.0.4 very soon, because of this: 523bdaa10a50b0fbf0f6cc78f0f664290107fa30

Issue #2495, opened 2015-01-19
Log in screen shows a horizontal scroll bar


gwbonline:
A little thing... The log in screen of Bolt shows a horizontal scroll bar. It doesn't bother me much (just a little), but it's not necessary, it has no function. I see it both in Firefox and IE.

chilion:
Chrome also does this. I do volunteer.

cdowdy:
There is actually more than one issue here just from a cursory glance:  

1) https://github.com/bolt/bolt/blob/master/app/view/sass/modules/_bolt.scss#L5-L10  
  * ```overflow: scroll-y``` should be ```overflow-y: scroll;``` [Mozilla Dev Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y#Syntax)  

2) the horizontal scroll bars would be the x axis so by using the right syntax this may fix the issue. Or a combination of if you want the y axis to still have scroll bars   
```css
body {
  color: $bodytextcolour;
  font-size: 13px;
  min-height: 100%;
  overflow: auto;
}

body.login {
  overflow-x: hidden;
}
```   
seeing as how the login page is responsive down to 300px scroll bars aren't really needed for horizontal or vertical axis. Overflow auto would remove the bars but you'd still be able to scroll.

bobdenotter:
Fixed by #2498

PR #2498, opened 2015-01-19
Remove horizontal scroll 

cdowdy:
This is a fix for #2495  

Updated ```overflow: scroll-y;``` to ```overflow-y: scroll```  
Moved ```body.login``` to ```_login.scss```

I screwed up committing this that's why there's a blank page.. I'll fix if its an issue.  The scroll happen because there is a -15px margin on the ```.row``` class. I didn't investigate far enough to tell if removing those margin would break any other functionality. If you remove that from the login page you don't have any horizontal scrolling


bobdenotter:
Hey Cory! Looks good to me. Thanks!

PR #2499, opened 2015-01-19
fixes nav links in admin dashboard when accessed over https

glasspelican:
when you use the nav links it attempts to load extensions and file-managment pages using http


bobdenotter:
Good catch! Thank you.

PR #2500, opened 2015-01-19
Refactor config

CarsonF:
Most of this is breaking getConfig() into separate methods. 
I made resources a class variable since it is used all over the class.

I'm not sure if anything in here is considered a BC break. BC only counts for using the class right, not extending it? 
I ask because I changed some of the methods to now return their results instead of setting them to class variables.
This wouldn't affect users of the class only people who are extending it.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -49,6 +49,7 @@ class Config
 
     protected function initialize()
     {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Yay! Merging in!

PR #2501, opened 2015-01-19
Extendable Config initialization

CarsonF:
Moved config service provider to its own method, so it can be overridden before it is initialized when initializing the session.

Edit: Ok I made the config calls from session init lazy, which would remove the need for two methods. But initializing the session in initConfig() is still maybe not the best. Thoughts?


GawainLynch:
Looks fine to me. Wait on @rossriley's next PR to land in canary first?

PR #2505, opened 2015-01-20
Correct method chain in ExtensionInstaller class

GawainLynch:


rossriley:
I can't remember the precise problems caused but I changed to getOperation on this commit here: https://github.com/bolt/bolt/commit/fa6e75d854c3fc3c9ba6e81f4a7c7abc88f49586#diff-04b9c44928ad02ce0f9466ec5211f666 

And I think it was to solve a potential for different types of events to be passed in. Not sure whether it's still an issue with current versions though.

GawainLynch:
I get b0rkage in it with current canary.

My immediate (and probably) incorrect thought is that we've changed the events that are called by interacting on a different code path... Composer internals and all :grin:

PR #2509, opened 2015-01-20
Moved Nut commands to a service provider

CarsonF:
Those who want to modify their Bolt install mostly likely need to create their own `nut` file which has their own bootstrap. 
The problem is they then have to add all the nut commands manually. 
What happens when these are removed/changed or others are added?

Moving these commands to a service provider fixes that.


GawainLynch:
Travis failure is testSaveSerializeErrors(), so nothing to see here.

Code looks good though.

PR #2512, opened 2015-01-21
Have Cache::clearCache() use $app['resources']->getPath('root') for the 'thumbs' directory

### LINKED ISSUE #2511: Clear cache does not clear thumbs folder
Scenario:
- Bolt installed with composer and outside the web root.
- Config option `thumbnails/save_files` set to `true`.

Clearing the cache from admin does not clear the thumbs folder.

I think the culprit is this line in the `Cache::clearCache()` method, that is not taking into account the real location of the thumbs folder under the web root:

        // Clear the thumbs folder..
        $path = dirname(dirname(dirname(__DIR__))) . "/thumbs";
        $this->clearCacheHelper($path, '', $result);

Cheers.

GawainLynch:
Fixes #2511

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -2,6 +2,7 @@
 
 namespace Bolt;
 
+use Bolt\Configuration\ResourceManager;
 use Doctrine\Common\Cache\FilesystemCache;
 use Symfony\Component\Filesystem\Filesystem;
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
Looks good to me!

rossriley:
I have a feeling this should use getPath('web') rather than root. Location of thumbs isn't necessarily in the root path.

bobdenotter:
Hmm, i think you might be right.

PR #2517, opened 2015-01-21
Fix doctrine/dbal at version 2.5.1, since it's the last to support PHP 5.3.3

bobdenotter:
See also: https://github.com/doctrine/doctrine2/pull/1255


GawainLynch:
R.I.P. PHP 5.3 :wine_glass:

PR #2521, opened 2015-01-21
Don't initialize the timefields if the date is '0000-00-00'. Fixes #2519

### LINKED ISSUE #2519: When updating an older Bolt, "empty" datedepublish should not trigger the 'date is in the past notification.
See screenshot: 
screenshot 2015-01-21 21 05 56

bobdenotter:


GawainLynch:
Shall do the bolt.min.js here, now

PR #2524, opened 2015-01-22
[RFC] Bolt logging/monolog update for 2.1

GawainLynch:
### Features
* New provider — LoggerServiceProvider — provides
  - $app['logger.system'] — System logging 
  - $app['logger.change'] — Record change logs
  - $app['logger.firebug'] — Log debug messages to Firebug
  - $app['logger.manager'] — Logging interface
* Monolog handlers:
  - SystemHandler
  - RecordChangeHandler
* Twig logging to Firebug with optional message:
  - {{ firebug( myvar, mymessage) }}


bobdenotter:
If we merge this in, we should add `monolog/monolog` as a dependency in our `composer.json`. Currently, it's included in Bolt, because it's a dependency of Codeception. If you run composer with `--no-dev`, Codeception is not installed, and monolog isn't either.

GawainLynch:
So I wasn't imagining it the other day when I thought I saw it evaporate!

GawainLynch:
Ugh, and a tests fail.

PR #2527, opened 2015-01-22
Feature/refactor config

CarsonF:
Canary already has a lot of these changes. But none of them are in master. Hopefully this one branch will make it there.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -62,7 +62,7 @@ class Application extends Silex\Application
     /**
      * Initialize the config and session providers.
      */
-    private function initConfig()
+    protected function initConfig()
     {
         $this->register(new Provider\ConfigServiceProvider());
         $this->register(
--- a/src/Application.php
+++ b/src/Application.php
@@ -67,18 +67,14 @@ class Application extends Silex\Application
 
     protected function initSession()
     {
-        $this->register(
-            new Silex\Provider\SessionServiceProvider(),
-            array(
-                'session.storage.options' => array(
-                    'name'            => 'bolt_session',
-                    'cookie_secure'   => function($app) {
-                        return $app['config']->get('general/cookies_https_only');
-                    },
-                    'cookie_httponly' => true
-                )
-            )
-        );
+        $this->register(new Silex\Provider\SessionServiceProvider());
+        $this['session.storage.options'] = function($app) {
+            return array(
+                'name'            => 'bolt_session',
+                'cookie_secure'   => $app['config']->get('general/cookies_https_only'),
+                'cookie_httponly' => true
+            );
+        };
 
         $this['session.storage.handler'] = $this->share(
             $this->extend(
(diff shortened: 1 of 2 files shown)


GawainLynch:
This branch is clear comparing to master and canary, you're still fine :grin:

CarsonF:
Umm what? ha

GawainLynch:
> Hopefully this one branch will make it there.

CarsonF:
This branch gets merged to canary, and then to master right?

GawainLynch:
That is currently the plan

GawainLynch:
OK, review time.  @bobdenotter and I talked about these changes and we have some feedback.

>-                $this->app['session']->getFlashBag()->set('error', $error);
>+                $this->flashError($error);

This varies from how Symfony documentation reads and makes it harder to come-up-to-speed on Bolt for Symfony/Silex developers...  Or anyone.  Would only mean we have to document it.

>-            $this->app['resources']->initializeConfig($this->data);
>+            $this->resources->initializeConfig($this->data);

Same with these.

Other than that, I think @bobdenotter might have a comment, but looks good.

CarsonF:
flashError method I can understand. I saw in one place that it was doing a couple checks to make sure session exists I thought it might be better to be abstracted.

> $this->app['resources']->initializeConfig($this->data);
> $this->resources->initializeConfig($this->data);

I want to push back a little on this. It makes it very hard to see which pieces are used in what places. If I pull up a class that just has $app passed into the constructor I have to search through the entire class to find out what it's using. If they were to be passed in to the app I would know immediately. 

Pimple is supposed to be a DI container. The dependencies are supposed to be injected into the constructor to make object creation easier, not to be a passed IN to the actual object. This is why there are actual classes and then, separately, the service providers that build up the object tree. That wouldn't be necessary if all the classes just took the DI container.

The other thing is (and I can't believe no one has complained about this before) is type hinting. In PHPStorm, and I'm sure any other IDE, you can say "find usages" of this method. Well that doesn't work if the method is called right from a silex service because it doesn't know the return type of the array get.

I know there are some common "services" used in a lot of places (config, session, etc.). An arguement could be make that, It would be a hassle to always pass those into objects. As I compromise, I would suggest that we create getters in the Application. `$app->config()->get('foo')` I would suggest those as traits, but not for now since we still support 5.3. Silex actually has several traits builtin that are for this purpose.

GawainLynch:
Relevant parts of conversation copied over from IRC:

> Bopp: What I'd like, is that we're sticking to silex idioms, because we happen to use silex.
> rossriley: only service providers should reference $app ideally…
> rossriley: although this is the age old DI versus Service locator argument
> gawainlynch: I just don't see what `$this->resources = $app['resources'];` buys us given above points
> Bopp: I really have no opinion whatsoever on DI vs service locator patterns
> Bopp: The main reason I like $this->app['resources'] over $this->resources is consistency.
> Bopp: "this->app['bla'] makes it clear to anyone reading the code what kind of construct is going on there.

GawainLynch:
Closing as we're moving to master! :tractor:

Issue #1351, opened 2014-07-04
PHP built-in server broken. (was: "Error "Template "editconfig.twig" is not defined")


h4cc:
I have installed bolt using these instructions `http://docs.bolt.cm/installation#option-3-the-developer-way-using-git-and-composer` and get the following error:
```
Error "Template "editconfig.twig" is not defined"
```
when accessing one of the urls

* `/bolt/file/edit/theme/index.twig`.
* `/bolt/file/edit/app/config/config.yml`
* `/bolt/file/edit/app/config/contenttypes.yml`
* `/bolt/file/edit/app/config/taxonomy.yml`
* `/bolt/file/edit/app/config/menu.yml`

To run the site i have been using this command and php version:

```
$ php -S localhost:8000 -t . index.php
PHP 5.5.3-1ubuntu2.3 Development Server started at Fri Jul  4 13:42:25 2014
Listening on http://localhost:8000
Document root is /home/julius/silpion/Speedpool/2014-07-04_Basis_CMS/bolt/bolt
Press Ctrl-C to quit.
```

bobdenotter:
Closing this, since it is a bug in PHP, not in Bolt.

PR #2529, opened 2015-01-22
Refactor Config::getWhichEnd() to work with PHP CLI Server and use Symfony Req Object

caseyamcl:
The `Config::getWhichEnd()` method relies some custom logic that uses the `$_SERVER` superglobal to determine which 'end' of the application is being used.  I propose instead using the Symfony `Request` object, since it does a better job of determining path information for various environments.

If the `$app['request']` hasn't been initialized yet, we can create a local copy of the Symfony `Request` object using `createFromGlobals()`.  This will ensure that the `getWhichEnd()` uses the same HTTP interpretation logic that the routing does.

This benefits of this pull request include:

- simplifies the code in the `getWhichEnd()` method, and
- fixes an issue where certain pages in the admin section throw "missing Twig template" errors; specifically pages where the URI ends with a file extension (e.g. `/bolt/file/edit/app/config/contenttypes.yml`).  This error occurs when using the PHP built-in webserver

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -7,6 +7,7 @@ use Bolt\Library as Lib;
 use Bolt\Helpers\Arr;
 use Bolt\Helpers\String;
 use Bolt\Translation\Translator as Trans;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Yaml;
 use Symfony\Component\Yaml\Parser;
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
I think this looks good! :+1:

GawainLynch:
Thanks @caseyamcl.  That whole function is about to get refactored, but this will get you going in master now.

GawainLynch:
@caseyamcl & @CarsonF, what versions of PHP are you testing on?

c.f. #1351

caseyamcl:
@GawainLynch - I tested on 5.4 and 5.5.

PR #2531, opened 2015-01-22
Added content to request attributes so after handlers can access it.

CarsonF:


GawainLynch:
@rossriley and I have a few concerns here.

> rossriley: hmm, is that a ref to the actual content object
> gawainlynch: I think that is his intent
> rossriley: i'm not sure about that then, I don't mind the idea, but the request is meant to model the http request
> gawainlynch: OK, so I am not losing my mind then :-)
> rossriley: so hinting at a contenttype and id would be ok but i think referencing the entire object goes against the spirit of a request object

CarsonF:
Hmm I don't disagree with those statements. Somehow I need to get access to the content object...without recreating it. Where do yall suggest is a better place? I know in pimple is an option...but I don't really like it.

GawainLynch:
Closing as we're moving to master! :tractor:

CarsonF:
I want to continue this conversation here for context. I can see why request isn't the right spot for this. I need it somewhere though...does anyone have thoughts? I guess in app?

Issue #2283, opened 2014-12-06
https://extensions.bolt.cm/ is unreachable.


peterboorsma:
It seems the 'install extensions' screen is b0rken in bolt-2.0.0-beta5pl4.

The message on /bolt/extend/ is: 
https://extensions.bolt.cm/ is unreachable.
Unable to install/update extensions!

The extension log screen and all buttons are inactive.

bobdenotter:
More bumps for @GawainLynch @rossriley

GawainLynch:
I saw something in the API today (can't remember exactly) about using downgraded SSL that I don't think the console app uses... Either way. I flagged it to look at when I get home next week.

peterboorsma:
No problem :)

GawainLynch:
OK, so a fix for this has landed in the `canary` branch (#2503) which will become 2.1.  I'll leave open for testing results from other and eventually @peterboorsma

bobdenotter:
Closing this one. Fixed by gawain for 2.1..

PR #2543, opened 2015-01-25
Extend's Composer Implementation for v2.1

GawainLynch:
"And...  In the blue corner, weighing in at 146 commits..."
daffey-boxing-small

So here is, a canary merge branch, branched fresh off my feature branch. A big thank you goes out to @rossriley for figuring so much of this out in the first place.

### New/Changed
* LOTS!
* Using the Composer API more directly, rather than proxying through console commands
* Each implemented `Composer\Command` has a corresponding `Bolt\Composer\Action` class that closely mimics how the `Composer\Command` classes individually interface `Composer\Installer`
* Actions currently implemented:
  - BoltExtendJson
  - CheckPackage
  - DumpAutoload
  - InstallPackage
  - RemovePackage
  - RequirePackage
  - SearchPackage
  - ShowPackage
  - UpdatePackage
* `Bolt\Composer\Action` classes are all declared `final`, for now. Our implementation of these classes is likely to change over time and as such, not valid as part of a 'public API'
* All jQuery `.get()` requests are now matched with a `.fail()` handler to report on Composer/Whoops errors and display feedback to end user
* Refactored `Bolt\Composer\CommandRunner` into `PackageManager` and `Factory` classes
* Most (eventually all) JSON calls use `Bolt\Composer\Action\` libraries
* Error handling all around, especially with a focus on AJAX requests
* Lots of work on a consistent AJAX response for available queries.  Composer itself is where this gets tricky, they really haven't gotten their heads/code into the idea that `Output\Console` is not how 100% of the library's user base want to (randomly) deal with things...
* Exceptions can now throw a BoltComposerException class
* Version button!  Everyone wants a version button these days, all the cool kids have one! 
  - Will display the installer dialogue for that package
  - Allow you to upgrade/downgrade on the spot
* cURL SSL/TLS certificate errors will now automatically downgrade the connection to HTTP 
  - See #2283
* Uses the full set of Composer JSON database files to compile package information, therefore we can list extensions that are installed as a dependency and manage those in the front end too
* JavaScript created HTML has mostly been moved to Twig generated Javascript variables
* UI strings are now translated in PHP and stored in a JavaScript object

**NOTE:** The old log has not been removed, just non-functional.  Will be fixed soon, but you will still get 500 errors on AJAX requests in your console log.

### Reviews/Tracking
* Feature tracking branch (PR against) `https://github.com/GawainLynch/bolt.git feature/composer-api`
* Feed back on unclear code would be very much appreciated
* PRs welcome
* Bolt's Composer installs have not been well tested

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -270,7 +270,6 @@ hash_strength: 10
 extensions:
   site: 'https://extensions.bolt.cm/'
   stability: stable
-  use_http: false
 
 # Enforce SSL: if set, all but the front-end pages will enforce an SSL
 # connection (and redirect to HTTPS if you attempt to visit the backend over
--- a/app/view/Gruntfile.js
+++ b/app/view/Gruntfile.js
@@ -214,6 +214,7 @@ module.exports = function(grunt) {
                     Stack: true,                // src/obj-stack.js
                     site: true,                 // src/extend.js/extend.twig
                     baseurl: true,              // src/extend.js/extend.twig
+                    rootpath: true,             // src/extend.js/extend.twig
                     // Bolt global functions
                     bindFileUpload: true,       // src/bindfileuploads.js
                     bindGeolocation: true,      // src/geolocation.js
--- a/app/view/lib/bolt/extend.js
+++ b/app/view/lib/bolt/extend.js
@@ -101,14 +101,16 @@ var BoltExtender = Object.extend(Object, {
                     
                     // Add an install button
                     target.append(bolt.data.extend.packages.install_new.subst({
-                        '%PACKAGE%': ext.name}));
+                        '%PACKAGE%': ext.name,
+                        '%VERSION%': ext.version}));
                 }
                 for (e in data.updates) {
                     ext = data.updates[e];
                     
                     // Add an update button
                     target.append(bolt.data.extend.packages.install_update.subst({
-                        '%PACKAGE%': ext.name}));
+                        '%PACKAGE%': ext.name,
+                        '%VERSION%': ext.version}));
                 }
                 active_console.hide();
                 controller.find('.update-list').show();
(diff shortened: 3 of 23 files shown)


mtotheikle:
@GawainLynch I have a system that we are building where we need to have plugins downloaded using composer. We were going to take a very similar approach to what was done in CommandRunner which this PR is removing. Do you mind providing some insight into what led to the decision of integration more closely with Composer?

GawainLynch:
Happy to. The big points were functionality, reliability, and error handling.

Using the Composer PHAR came with a lot of downsides. It uses the console directly, used several PHP functions that are blocked on a lot of hosts, and gave responses only in text strings… among others.  The responses,  despite what we tried in various tests branches, made error handling quite difficult overall.

Using the PHAR also required that `https://getcomposer.org/installer` would respond reliably. This is not always the case, in-fact it can be quite patchy at times and the Composer team are currently in the process of planing an infrastructure move for both `getcomposer.org` and `packagist.org` to cope with their growth, but in the meantime that has had the very real effect of causing our users to see *Bolt* the thing that was "not working properly".

Finally, the functionality.  CommandRunner is replaced with PackageManager, that while still being refined, should return a consistent result format to required commands, significantly better error handling and feedback the our AJAX callers, and allow us to extend functionality further.  

For examples of the desired uses one of the extended functionality.  In 2.2 we are aiming to extend the autoloader to extensions that have been installed in `extensions/local/` and there is an updater for Bolt core in the works as well, that can interface directly with PackageManager for our Composer libraries.  In fact, with a very simple tweak to PackageManager, I can already show *all* of Bolt core's installed packages and their dependencies in the Extend page in the same fashion as we do for extensions.

mtotheikle:
Awesome, thank you @GawainLynch. What you mentioned touched on some of the exceptions that I had so that's good to know.

PR #2545, opened 2015-01-25
Cleanup of composer.json

bobdenotter:
- Put `bolt/pathogen` in alphabetical order
- Use tagged version for `ircmaxell/random-lib`
- Use `dev-master` for `composer/composer`. We don't see that there will be an actual 1.0.0 release anytime soon, so we might as well use the latest.


GawainLynch:
Awesome-sauce!

PR #2563, opened 2015-01-25
regex the response rather than inspecting the flash message

rossriley:


GawainLynch:
Green is good!  Thanks mate.

PR #2564, opened 2015-01-26
More logging updates

GawainLynch:


rossriley:
Test failures are only minor, incorrect method call counts... we can revisit these soon.

rossriley:
Test failures are only minor, due to changing number of method calls. We can revisit these.

PR #2565, opened 2015-01-26
Display warnings in the back end for using development versions

GawainLynch:


bobdenotter:
This is butt-ugly and should be replaced with something that's slightly less annoying. It's perfect for now, because I sure will not forget it this way. ;-)

PR #2566, opened 2015-01-26
Reorganisation of datepicker/moment locale files

rarila:
- We now have both in meaningful folders
- All files are minified
- Files naming is unified to ``js/locale/<library>/xx_XX.min.js
- ``js/`` now only has generated files

What could be done to improve things further?

- All locale could be combined to one locale file that then includes both, datepicker and moment locale with fallback inserted. Would save one request.
- At the moment all available locale are added. Could be shaved down to a list aof supported locale. Otherwise it's just a file lying around that won't hurt and if someone adds translations without publishing here the correct lib locales are probably already there


GawainLynch:
Nothing to see here... ;-)

PR #2567, opened 2015-01-26
Initialize config and session separately

CarsonF:
This allows Config to be overridden before initialization


bobdenotter:
Looks good to me! :sparkles:

PR #2569, opened 2015-01-26
Config simpify parseConfigYaml method

CarsonF:
Depends on #2568


GawainLynch:
2568 merged...  and this is fine with me if it is fine with @bobdenotter and @rossriley :ship:

PR #2571, opened 2015-01-26
Refactor constructor into initialize method

CarsonF:
Generally it's better not to do logic in constructor.  This allows others to override initialize() to change caching/loading config behavior without having to recreate the constructor.


GawainLynch:
> Generally it's better not to do logic in constructor

Agreed.  Has my vote!

PR #2572, opened 2015-01-26
Config::setDefaults -> getDefaults

CarsonF:
Depends on #2571.

Generally, (IMO) it's better to return variables than set them in the class. It makes code easier to follow.
In this case, it allows others to modify defaults without having knowledge of where the defaults are set in the config.


GawainLynch:
I like, very much.

bobdenotter:
Yes, much nicer! Thx!

PR #2573, opened 2015-01-26
Config split parsing methods

CarsonF:
Here's the big one....I would look at each commit separately.


GawainLynch:
These are throwing weird test errors... @rossriley?

GawainLynch:
No kittens crying here :smile_cat:

rossriley:
That's more like kittens committing suicide...

GawainLynch:
> That's more like kittens committing suicide...

Yes, but they're happy!

CarsonF:
Yeah it's all the same logic just different variable names and indentation. Unfortunately, that doesn't diff well.

PR #2581, opened 2015-01-26
Using Doctrine Platform instead of our config values.

CarsonF:
Removed the calls to Config::getDBOptions. It is better to use the DBAL connection object.

- Using Platform name instead of driver name because that's really what we are checking for (this fixes pdo_mysql != mysqli).
- Getting database name from database connection


GawainLynch:
Elegant, good work!

CarsonF:
I only found it because of code completion :)

GawainLynch:
Ha! I usually find these things by reading code... I need new toys too!

PR #2583, opened 2015-01-26
Move post connect queries to an InitListener class

CarsonF:
This will execute the queries for every connection established, not just the first one. Also, it moves it out of Application.


GawainLynch:
:white_check_mark:

PR #2584, opened 2015-01-26
Update queries for slaves

CarsonF:
Replaced prepare() queries with executeQuery() since that supports read slaves. 
Replaced prepare() with query() on write queries because why do 3 method calls when you can do one?
Cleaned up tag queries in Async Controller as well.

Should some of those queries be on multiple lines? Like:
```sql
SELECT id 
FROM $tablename 
WHERE status = 'published' 
    and datedepublish <= :now 
    and datedepublish > '1900-01-01 00:00:01'
    and datechanged < datedepublish
```

### DIFF
--- a/src/Controllers/Async.php
+++ b/src/Controllers/Async.php
@@ -241,7 +241,10 @@ class Async implements ControllerProviderInterface
         $table .= 'taxonomy';
 
         $results = $app['db']->fetchAll(
-            "SELECT DISTINCT $table.slug from $table where taxonomytype = ? order by slug ASC",
+            "SELECT DISTINCT $table.slug
+            FROM $table
+            WHERE taxonomytype = ?
+            ORDER BY slug ASC",
             array($taxonomytype)
         );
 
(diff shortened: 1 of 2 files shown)


GawainLynch:
Database queries after my own heart! LGTM

CarsonF:
Multiline yes/no?

GawainLynch:
>Multiline yes/no?

Where it is long, absolutely!  [personal taste disclaimer]

CarsonF:
They are all long for me....

GawainLynch:
Historically, I've tended to use something like:

```
"SELECT slug, COUNT(slug) as count 
FROM $table 
WHERE taxonomytype = ? 
GROUP BY slug 
ORDER BY count DESC 
LIMIT ?"
```

But that can really piss some coders off

CarsonF:
That's what I was thinking. Figured I would ask because of that point

src/Controllers/Async.php
```diff
@@ -240,33 +240,26 @@ public function tags(Silex\Application $app, $taxonomytype)
...
         $table .= 'taxonomy';
 
-        $limit = $app['request']->get('limit', 20);
-
-        $query = sprintf(
-            'SELECT slug, COUNT(slug) as count from  %s where taxonomytype = ? GROUP BY  slug ORDER BY count DESC LIMIT %s',
-            $table,
-            intval($limit)
+        $results = $app['db']->fetchAll(
+            "SELECT slug, COUNT(slug) as count from $table where taxonomytype = ? GROUP BY slug ORDER BY count DESC LIMIT ?",
```
GawainLynch: s/from/FROM/
s/where/WHERE/

PR #2586, opened 2015-01-26
Async Controller clean up

CarsonF:


GawainLynch:
@CarsonF  is on :fire:

PR #2592, opened 2015-01-27
[WIP] Allow extensions to leverage getDecodedValue

### LINKED ISSUE #2591: Override getDecodedValue in extension
I'm trying to write an extension for the elusive list field at the moment, and I've quickly run into an issue. To decode JSON data stored in the database for ImageList and FileList fields, the `getDecodedValue` is called in [Content.php](https://github.com/bolt/bolt/blob/master/src/Content.php#L533).

However, it is using a switch-case statement so all of the different fieldtypes are hard coded in. If I'm writing an extension, my own implementation is not possible. This should be a fairly simple fix by adding `getDecodedValue` to [FieldInterface.php](https://github.com/bolt/bolt/blob/master/src/Field/FieldInterface.php), and making the implementation in [Base.php](https://github.com/bolt/bolt/blob/master/src/Field/Base.php) the function that is currently in Content.php.

I'll have a play and if I have success I will open a PR.

Pinpickle:
Fixes #2591 

I've done what I have described in the issue, but this is a breaking change for all current extensions that create new fieldtypes as I have added a new method to the FieldInterface.

I was wondering if there was any advice as to how I could make this change non-breaking? I don't PHP often so I don't know what is considered black magic - would taking the function out of the interface and then checking if the method exists using [`method_exists`](http://php.net/manual/en/function.method-exists.php) be something worth doing?

This also breaks the excerpt function for some reason, and all html tags have escaped brackets. Trying to get to the bottom of that one.

**Edit**: Part of me also wants to do something similar to the `setValues` function as I think that would allow for pre-saving hooks.

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -536,83 +536,14 @@ class Content implements \ArrayAccess
 
         if (isset($this->values[$name])) {
             $fieldtype = $this->fieldtype($name);
-            $fieldinfo = $this->fieldinfo($name);
-            $allowtwig = !empty($fieldinfo['allowtwig']);
-
-            switch ($fieldtype) {
-                case 'markdown':
-
-                    $value = $this->preParse($this->values[$name], $allowtwig);
-
-                    // Parse the field as Markdown, return HTML
-                    $value = \ParsedownExtra::instance()->text($value);
-
-                    // Sanitize/clean the HTML.
-                    $maid = new \Maid\Maid(
-                        array(
-                            'output-format' => 'html',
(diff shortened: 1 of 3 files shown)


bobdenotter:
I was wondering about this the other day: we need to be able to do this in extensions somehow. I'll look into this, when I have a bit of time! :-)

Pinpickle:
Maybe this should use events akin to `preSave`, just on individual field instances?

GawainLynch:
Let's merge, but remove the addition to the interface to prevent a BC break… and hopefully avoid an AD break too :clock1130: :beers:

bobdenotter:
Agreed. :-)

PR #2593, opened 2015-01-27
seems like `$this->app['end']` is not available here.

bobdenotter:


GawainLynch:
@bobdenotter to the rescue!

PR #2595, opened 2015-01-27
Feature/monolog

GawainLynch:
Summary:
* Queries now use QueryBuilder
* System log output can be user filtered (UI feedback gratefully accepted)


bobdenotter:
Looks good! :100:

PR #2597, opened 2015-01-27
Allow for 'keys: slug' in 'type: select' fields, to customize the used field that's actually stored in the DB.

bobdenotter:


GawainLynch:
This b0rks PHPUnit on PHP 5.6 on Travis and locally here... :-/

GawainLynch:
Actually, it isn't this!

PR #2598, opened 2015-01-27
Replace relative path regex with pathogen solution

CarsonF:
Y U REGEX?


bobdenotter:
Looks good to me! :100:

bobdenotter:
Merging in!

PR #2599, opened 2015-01-27
update gruntfile for magnfic popup

cdowdy:
I forgot to commit the updated gruntfile when I updated magnific popup to 1.0.0.  

Also removed version number from the directory to make future manual updates easier (if not using a package manager like Bower or NPM )


rarila:
@cdowdy I added them to make updating easier so that you can easily check what version is installed without reading readmes or sourcecode or where ever the version number could be hidden. I even compared sources with some libs where no version was given.

cdowdy:
ahhhh ok that makes sense... Ill revert back to version numbers then !

rarila:
Thx :-)

bobdenotter:
Merging in! :+1:

Issue #2140, opened 2014-11-09
Better Path Composition


rossriley:
At the moment there's quite a few places where we are using code like this... `$app['resources']->getPath('rootpath') . 'src'`

Since we now have the path library built into the ResourceManager I'd suggest that we bake this functionality into the class itself so that we standardise path management.

Suggesting the following syntax instead: `$app['resources']->getPath('rootpath/src')` or `$app['resources']->getPath('extensions/local')`

CarsonF:
Closed?

rossriley:
Closed

PR #2600, opened 2015-01-27
Feature - allow ResourceManager getPath to return a relative path based on a known base

rossriley:
As discussed in #2140

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -141,12 +141,37 @@ class ResourceManager
         if (array_key_exists($name . "path", $this->paths)) {
             return $this->paths[$name . "path"]->string();
         }
+        
+        if (strpos($name, '/') !== false) {
+            return $this->constructRelativePath($name);
+        }
 
         if (! array_key_exists($name, $this->paths)) {
             throw new \InvalidArgumentException("Requested path $name is not available", 1);
         }
 
         return $this->paths[$name];
+    }
+    
+    /**
+     * Takes a known path with relative additional atoms and returns a new path
+     * for instance constructRelativePath('public/images/example')
+     *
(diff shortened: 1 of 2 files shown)


src/Configuration/ResourceManager.php
```diff
@@ -148,6 +152,27 @@ public function getPath($name)
...
+    
+    /**
+     * Takes a known path with relative additional atoms and returns a new path
+     * for instance constructRelativePath('public/images/example')
+     *
+     * @return AbsoluteUnixPath Object
+     **/
+    public function constructRelativePath($name)
+    {
+        list($firstAtom) = explode('/', $name);
```
GawainLynch: Isn't this assuming *NIX?

rossriley: This is only an access interface, if you ask for getPath("cache/test") you will still get back an OS specific path, just we're using forward slashes as a divider, the same as we do in config->get('general/branding')

CarsonF:
I just did something similar, #2601. 
```php
$app['resources']->getPath('root', true)->joinAtoms('foo', 'bar');
```

Less code perhaps?

rossriley:
constructRelativePath refers to the fact that it is relative to the base that you pass in. The interface that you use is:
    $app['resources']->getPath('root/vendor');
    $app['resources']->getPath('public/files/demos');

It means that we can get rid of a lot of instances across the codebase where you see stuff like this:

    $app['resources']->getPath('rootpath') . '/src';

CarsonF:
Right, I was going for the same thing. 
Ironic that we both had this thought at the same time. 

Now that I look at this more, I like :+1: 

Can we still have optional argument from my PR so that other PathInterface methods can be used?

bobdenotter:
Looks good to me! :100:

PR #2601, opened 2015-01-27
Allow getPath to return PathInterfaces

CarsonF:
Paths are joined together with "/" hardcoded in several places. This is the first step to allow them to be joined properly.

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -136,9 +136,9 @@ class ResourceManager
 
         return $path;
     }
-
+    
     /**
-     * Gets a path either as a string or PathInterface.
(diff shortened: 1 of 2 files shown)


GawainLynch:
Nice mate!

rossriley:
I like the idea of this, and this works fine.. Just an alternative to think about in terms of keeping the API cleaner. Don't really mind if others disagree but I always prefer a method that returns an explicit type.

    public function getPath($path) {
        return $this->getPathObject($path)->string();
    }

    public function getPathObject($path) {
          // Logic identical to original getPath()
    }

GawainLynch:
Actually, yeah, I like your point @rossriley

PR #2603, opened 2015-01-27
Split Login Controller from Backend Controller

CarsonF:
Controller\Backend is about 2000 lines long. Way too big for my liking. 

There are two Login routes that do not use Backend::before middleware.
Grouping the login routes in their own controller collection allows all the 
Backend routes to define the before middleware once for the collection, instead of each route.

There are also other small clean ups. 
Like, 
```php
$ctl->get(...)->method('POST')
```
that doesn't make sense.


GawainLynch:
Nice, but can you clean up a few of the tests this breaks please...  Kitten are in danger! :cat:

CarsonF:
Looks better now. I need to move those tests to their own class though

bobdenotter:
Lookin' good! :+1:

rixbeck:
I like this direction!

PR #2605, opened 2015-01-28
Allow template directory

### LINKED ISSUE #2445: Template Directory
Currently all template paths are relative to the theme directory. I currently have a build system that puts all templates in `dist/templates/*.twig`. I've gotten used to typing 'dist/templates' in everything pertaining to templates, but it's not ideal. I think this will have more troubles if I am say, developing a public theme and this gets dropped into someone else's setup. Furthermore, this renders the template select field unusable as it can't detect any of my templates. I haven't had a need for it yet, but I imagine that when I do this will be really painful.

I am proposing a `template_directory` option in theme `config.yml` that sets the directory (in the theme directory) where the templates are stored. I think that the only internal stuff that would need to change is *some* references to `$this->app['paths']['themepath']` to `$this->app['paths']['templatepath']` where `$this->app['paths']['templatepath'] = $this->app['paths']['themepath'] + template_directory`.

Happy to do this myself, but I'm afraid I'll probably break something without knowing the ins and outs.

Pinpickle:
Fixes #2445

If you specify `template_directory` in your theme's `config.yml`, all related template activities will be relative to that directory (which is relative to the theme directory).

It adds `templatespath` to the paths object, and I *think* I managed to change all of the correct instances of `themepath`. This should be checked by someone who knows their stuff.

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -276,6 +276,13 @@ class ResourceManager
     public function postInitialize()
     {
         $this->setThemePath($this->app['config']->get("general"));
+
+        if (isset($this->app['config']->get('theme')['template_directory'])) {
+            $this->setPath('templatespath', $this->getPath('theme') . '/' . $this->app['config']->get('theme/template_directory'));
+        } else {
+            $this->setPath('templatespath', $this->getPath('theme'));
+        }
+
         $branding = ltrim($this->app['config']->get('general/branding/path') . '/', '/');
         $this->setUrl("bolt", $this->getUrl('root') . $branding);
         $this->app['config']->setCkPath();
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -276,7 +276,8 @@ class ResourceManager
     {
         $this->setThemePath($this->app['config']->get("general"));
 
-        if (isset($this->app['config']->get('theme')['template_directory'])) {
+        $theme = $this->app['config']->get('theme');
+        if (isset($theme['template_directory'])) {
             $this->setPath('templatespath', $this->getPath('theme') . '/' . $this->app['config']->get('theme/template_directory'));
         } else {
             $this->setPath('templatespath', $this->getPath('theme'));
--- a/src/Config.php
+++ b/src/Config.php
@@ -677,7 +677,7 @@ class Config
 
     public function getTwigPath()
     {
-        $themepath = $this->app['resources']->getPath("theme");
+        $themepath = $this->app['resources']->getPath("templatespath");
         $end = $this->getWhichEnd($this->get('general/branding/path'));
 
         if ($end == 'frontend' && file_exists($themepath)) {
--- a/src/Config.php
+++ b/src/Config.php
@@ -689,7 +689,8 @@ class Config
         // If the template path doesn't exist, flash error on the dashboard.
         if (!file_exists($themepath)) {
             $relativethemepath = basename($this->get('general/theme'));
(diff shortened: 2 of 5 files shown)


bobdenotter:
Cool! We're waiting for some other PR's to merge (finetuning, and general Travis weirdness), and then I'll look over this and merge it in, if it looks good!

Thanks, again! ^_^

Pinpickle:
I thought there were some changes afoot, with all of the failing builds and everything.

Would there be any objection to me removing the depth restriction in the template finder in [TwigExtension](https://github.com/Pinpickle/bolt/blob/issue-2445-template-directory/src/TwigExtension.php#L728)? It would make sense to store templates in subdirectories for logical division and filtering. E.g: people templates under templates/people and post templates under templates/posts and then in your config:

    filter: people/*

GawainLynch:
> I thought there were some changes afoot, with all of the failing builds and everything.

That was a misplaced boolean that had @rossriley and I crawling though the git repo forever! Good fun was had by all.

> Would there be any objection to me removing the depth restriction in the template finder [...]

I did want to bring this up with the reset of the team when I replaced the previous `fnmatch()` version of that function...  But it was late beta.  The call on that moves into UI space which I tend to not be too opinionated about as my brain is firmly "nerd" and not really in-line with how most people interact with the application (I personally like the idea though, but note my previous caveat).

GawainLynch:
P.S. We actively hang out on FreeNode in #boltcms if ever you want to catch us in person.

src/Configuration/ResourceManager.php
```diff
@@ -275,6 +275,13 @@ public function initialize()
     public function postInitialize()
     {
         $this->setThemePath($this->app['config']->get("general"));
+
+        if (isset($this->app['config']->get('theme')['template_directory'])) {
```
GawainLynch: This isn't valid for PHP 5.3, hence the Travis error. Try this:

```php
$theme = $this->app['config']->get('theme');
if (isset($theme['template_directory'])) {
```

Pinpickle:
Thanks for the feedback @GawainLynch, I've fixed up that issues and I've gone ahead and added the support for subdirectories. It's completely unobtrusive if you have no templates in subdirectories (they'll just appear as they always have).

image

GawainLynch:
[just going on picture]

Twig files that start with a '_' should not come up in this list, as they are considered include files (partials).

Pinpickle:
Huh, I was getting those before I made any changes, I think due to:

    filter: *.twig

In default contenttypes

Is it desired behaviour to always exclude partials? This can be done by
adding an exclusion name match into the finder, I believe.

Can't look into it now because laptop just died - I will when I get the
chance.

GawainLynch:
There might be a recent regression, but on my production boxes it definitely doesn't show partials.

The regex filter `'/^[a-zA-Z0-9]\V+\.twig$/'` should be covering that.

But yes, that is desired behaviour... While backend people might get the purpose of partials, ordinary editors likely won't.

Pinpickle:
I definitely still get it in `master`

image

This is because when you set a filter in the contenttypes (which is `*.twig` in default `contenttypes.yml`) it overrides the regex you've indicated there. I'll put in an exclusion so the partials are excluded no matter what.

Pinpickle:
There we go

image

GawainLynch:
> I definitely still get it in master

Yeah, I think there might have been an update recently that b0rked it...  No sresss, your fix works and that is the best part! :+1:

Pinpickle:
Awesome, happy to have helped!

bobdenotter:
MErging in! Thank you again!

PR #2609, opened 2015-01-28
"Fix" arrays in type:select fields.

bobdenotter:
In a `type: select` field, the current situation is that the YML parser does not distinguish between hashes and map typ arays. 
The reason that we would want to do this, is because if you set values to ['blue', 'green'], that is what you'd expect to see in the database, normally. Not '0' and '1', which is what would happen, if we didn't "correct" it here. 

I'm fully aware that this solution is conceptualy wrong. However before it was actually _broken_, it just wasn't very apparent. So, now we choose: "Wrong and fixed" or "correct and broken". ^_^

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -404,6 +404,17 @@ class Config
                 }
             }
 
+            // If field is a "Select" type, make sure the array is a "hash" (as opposed to a "map")
+            // For example: [ 'yes', 'no' ] => { 'yes': 'yes', 'no': 'no' }
+            // The reason that we do this, is because if you set values to ['blue', 'green'], that is
+            // what you'd expect to see in the database. Not '0' and '1', which is what would happen,
+            // if we didn't "correct" it here. 
+            // @see used hack: http://stackoverflow.com/questions/173400/how-to-check-if-php-array-is-associative-or-sequential
+            if ($field['type'] == 'select' && is_array($field['values']) && 
+                array_values($field['values']) === $field['values'] ) {
+                $field['values'] = array_combine($field['values'], $field['values']);
+            }
+            
             // If the field has a 'group', make sure it's added to the 'groups' array, so we can turn
             // them into tabs while rendering. This also makes sure that once you started with a group,
             // all others have a group too.


rarila:
My suggestion to make it bulletproof for the future:

Deprecate ``values``, introduce the following:
```
islist:
    type: select
    values_list: [red, green]
ishash:
    type: select
    values_hash: {0: red, 1: green}
```
First one saves ``red``, second one ``0``

bobdenotter:
@rarila: I fully agree with you from a developer's standpoint, however it would make things more complex for a novice user.

rarila:
Alternative …
```
islist:
    type: select_list
    values_list: [red, green]
ishash:
    type: select_hash
    values_hash: {0: red, 1: green}
```
… if one likes that better

rarila:
If think that can be easiliy explained in the docs, better than the caveats of the combined solution

rarila:
… Use hash when you want to store a value and display another one and list if you want to store the displayed value

GawainLynch:
What about:

```yaml
colours:
    type: select
    values: [red, green]
    prefer: key
```

rarila:
For me ``prefer`` is again like _fuzzy logic_. Either we like the key to be stored or not and so we should express what we want.

GawainLynch:
> For me prefer is again like fuzzy logic. Either we like the key to be stored or not and so we should express what we want.

Sure, but as @bobdenotter points out we want to make things easier for people how are not YAML experts and we have legacy to support going forward.

rarila:
For legacy ``value`` would stay as deprecated.

If you are a YAML expert then you should have no problems anyway, if not what would you do then? You would copy/pasta from docu or showcase example anyway and are done.

I think it's easier for a novice to follow a simple strict rule than to decide if using ``{}`` or  ``[]`` would do the trick.

And last but not least for us devs it would mean simplier, faster and easier to maintain code which in turn would help all. (There's not only this hack, don't forget [_select.twig](https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/fields/_select.twig), that probably could be cleaned up then also)

CarsonF:
```yaml
islist:
    type: select_list
    values: [red, green]
ishash:
    type: select_hash
    values: {0: red, 1: green}
```

I like this syntax. I feel like this could easily be documented. It keeps us from trying to figure out what type it is in code. 

I would even say we keep `select` instead of `select_list`. This maintains BC and nothing has to be deprecated.

GawainLynch:
Rest of this is RFC material.

tobias2k:
If you are a YAML expert, you'll notice that we use an unordered
key/value collection to encode an ordered key/value collection, and that
this only works because our YAML parser accidentally keeps ordering
intact when parsing hashes.

Interestingly, this same YAML parser is what's getting us into this mess
in the first place, because it uses PHP's "arrays" to represent both
hashes and lists, making it impossible to distinguish between
`{0:"foo",1:"bar"}` and `["foo","bar"]`.

However, switching to a YAML parser that is closer to the actual YAML
spec is kind of impossible, because AFAIK none exists, so the choice is
between "write our own YAML parser, break everyone's code, rewrite
everything in Bolt that touches YAML data structures, spend the
next few weeks fixing bugs and regressions", and "stfu and live with a
slightly broken solution that gets us a solution that works in 99% of
the cases, and cover up the brokenness with more documentation". Both
are painful, but the latter is more practical right now. (Not that I
wouldn't love writing a YAML parser, mind you).

On Wed, Jan 28, 2015 at 08:29:17AM -0800, rarila wrote:
> For legacy ``value`` would stay as deprecated.
> 
> If you are a YAML expert then you should have no problems anyway, if not what would you do then? You would copy/pasta from docu or showcase example anyway and are done.
> 
> I think it's easier for a novice to follow a simple strict rule than to decide if using ``{}`` or  ``[]`` would do the trick.
> 
> And last but not least for us devs it would mean simplier, faster and easier to maintain code which in turn would help all. (There's not only this hack, don't forget [_select.twig](https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/fields/_select.twig), that probably could be cleaned up then also)
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/bolt/bolt/pull/2609#issuecomment-71865587

-- 
Tobias Dammers - tobias@twokings.nl - 070-3457628 - www.twokings.nl
Maandag t/m donderdag van 9.00 tot 17.30
Voor dringende vragen, mail naar support@twokings.nl

Boorj:
i suppose there's an indirect but proof way to fix that:
1. parse contenttypes.yml by regexps
2. find `type: select` and following 'values :' (or above)
3. if there's `{` symbol after values - it's a hash.
4. if it's a hash, and yaml got straight array, split that to hash or at least assume in templates that it's a hash.

bobdenotter:
> 1. parse contenttypes.yml by regexps

This sounds like a _lot_ could go wrong. I'd be very wary to go down this route. Also, it disregards the other way to use hashes: 

```
  foo:
    bar: baz
```

And since basically _everything_ in the contenttypes.yml is structured like that, it would seem to me it would be brittle at best, and very likely introduce a lot of headaches.

Boorj:
i agree that it's a dirty way) but we can right in docs "want hashes - use {}" =) anyway...

SahAssar:
@Boorj We have an example for that here: https://docs.bolt.cm/howto/select-field-type

Not sure we want more in [defining-contenttypes](https://docs.bolt.cm/contenttypes-and-records#defining-contenttypes), that section is already pretty crowded... But if you want you can PR it and we'll see what the b*fl's say :)

PR #2619, opened 2015-01-29
better database seeding / resetting to improve test speed

rossriley:
This should save Travis a bit of money off their cooling bills.


CarsonF:
:beer: for you!!

PR #2624, opened 2015-01-30
Fix/improve provider tests

rossriley:
Should get tests/Provider up to 100% coverage.


GawainLynch:
:star: :star: :star: :star: :star:

PR #2626, opened 2015-01-30
exclude the tests directory from the code coverage reports

rossriley:
This should get rid of the tests directory from the code coverage report.


GawainLynch:
You mean we're not going to tests our tests...  How else do we achieve Boltception?

GawainLynch:
Maybe it is easy being green...

PR #2632, opened 2015-01-30
Fixes checks and creation of slugs when using multiple slugs on saving

StevendeVries:
When saving a contentype using multiple slugs, the slugs aren't set properly. Instead of using the string 'slug' it should use the $key of each slugfield, making sure each slugfield is checked and filled properly.


bobdenotter:
Good catch! Thank you! :+1:

PR #2633, opened 2015-01-30
Small clean up for Config::getWhichEnd

CarsonF:
Request::getPathInfo() always starts with a slash


bobdenotter:
Looks good to me!

PR #2637, opened 2015-01-30
Small Fix for Mobile Dashboard

dmm1:
On the Mobile Dashboard the titles of contenttypes overlap the "more" button.


rarila:
- Somehow your other PR apears here, too. Could it be you branched from that other branch instead from master? :-)
- Hmm, we try to get rid of the last inline styles, that are still in the codebase. Could you please add your patch to the stylesheet? I guess instead of outputting different markup it would be better to add a class to ``<h3>`` (or address it otherwise) and switch style with media query.

dmm1:
Sorry it was really just quick and dirty :smile: will do that today

rarila:
Hehe, no problem :-)

PR #2647, opened 2015-01-31
Scan file iteratively, instead of recursively. Fixes #2611.

### LINKED ISSUE #2611: YamlUpdater find nonexistent key fatal error
YamlUpdater::find on key that doesn't exist gives me:
PHP Fatal error:  Maximum function nesting level of '100' reached, aborting!

bobdenotter:
The idiot who wrote that in the first place apologises.


GawainLynch:
Nicely done!

PR #2654, opened 2015-01-31
fix mobile issues on Dashboard

dmm1:
- fix thumnail behavior
- fix Titles are overlapping
- fix button placement


bobdenotter:
Sweet! Thank you! :+1:

app/view/twig/dashboard/_recently_edited.twig
```diff
@@ -1,11 +1,11 @@
-<h3>
+<div style="contenttype-title">
```
CarsonF: This should be class, not style.

dmm1:
Uups will fix it thx

PR #2655, opened 2015-02-01
Have require failures report an error for extension loading

### LINKED ISSUE #2652: Local extensions not showing proper error reporting
I get a white screen when there is a php error in my extension. This makes writing code pretty difficult.
Looking at https://github.com/bolt/bolt/blob/master/src/Extensions.php#L171-L178 it just ignores any error in the autoload initialize() function.

GawainLynch:
Fixes #2652


bobdenotter:
Topkek! :100:

PR #2656, opened 2015-02-01
Remove xdebug.max_nesting_level from Travis tests

GawainLynch:
After #2647 Travis no longer requires `xdebug.max_nesting_level = 2000` to pass.  This also makes HHVM tests run again...  For what that is worth.


bobdenotter:
Merging in!

PR #2662, opened 2015-02-01
Edit field includes (geolocation/html)

rarila:
- Load include files only once (includes would have been multiple times for geolocation)
- Load ckeditor only on pages with a html field

Not the big art of software design, but pragmatically and working in the sense of our beloved k.i.s.s.


bobdenotter:
Not the cleanest solution ever, but I think i like it! ^_^

PR #2666, opened 2015-02-01
Move jQuery-gomap and CodeMirror plugins to view/js/…

rarila:
- Fixes bug in #2662 (CKEditor not included  for edit files/locale)
- Move jquery-gomap to ``view/js``
- Move CodeMirror language plugins to ``view/js``

@bobdenotter Finally all stuff from ``view/lib`` should be (if I haven't overseen something) serve from ``view/js``or ``view/css`` so that ``view/lib`` could be removed :scream_cat: from distribution file. The kittens will cry :crying_cat_face:


bobdenotter:
Whoooop!

PR #2667, opened 2015-02-01
Don't try to initilize CKEditor if it isn't loaded

rarila:


bobdenotter:
Green lights!! :traffic_light:

PR #2671, opened 2015-02-02
[RFC] Enabling Extension Dependencies from Packagist

rossriley:
As discussed here is an implementation that allows extensions (and their dependencies) to require packages from Packagist as well as the Bolt extensions repo.

To protect the core Bolt app we automatically mark as satisfied all of the core dependency graph, so should a package be part of this then the extension has to use the version installed and will be unable to replace it.

Potential Drawbacks to Consider
1. Speed: Parsing the packagist database is significantly slower than just our own repo, installing extensions will take a few seconds longer and possibly more as the dependency tree gets larger.

2. Clashes in future versions: For example, the scenario where a popular extension requires a package that in the future we want to include in core. We currently don't overwrite any existing composer requires, but we may have to consider that in the future.

3. Version nuances. There will be an increased onus on extension developers to ensure the extension works with the Bolt version of the package as this is upgraded. An extension that requires x/x: ~2.7 will be ignored when Bolt upgrades to 3.0 of x/x independently of the extension. 

### Possible Improvements:

1. We could make extension developers opt in to Packagist dependencies by specifying 
    "packagist": true

in the extra section of the extension composer.json.

The potential advantage of this is that we can do a scan of required extensions and leave Packagist disabled if none require it.

### DIFF
--- a/src/Composer/Action/BoltExtendJson.php
+++ b/src/Composer/Action/BoltExtendJson.php
@@ -103,7 +103,7 @@ final class BoltExtendJson
         );
         $json['provide']['bolt/bolt'] = $app['bolt_version'];             
         foreach($this->options['rootdependencies'] as $corePackage) {
-            $json['provide'][$corePackage] = "*";
+            $json['replace'][$corePackage] = "*";
         }
         $json['scripts'] = array(
             'post-package-install' => "Bolt\\Composer\\ExtensionInstaller::handle",
--- a/src/Composer/Action/BoltExtendJson.php
+++ b/src/Composer/Action/BoltExtendJson.php
@@ -25,6 +25,14 @@ final class BoltExtendJson
     {
         $this->options = $options;
     }
+    
+    /**
+     * @return $options  array
+     */
+    public function getOptions()
+    {
+        return $this->options;
+    }
 
     /**
      * Convenience function to generalise the library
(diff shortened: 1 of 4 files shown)


GawainLynch:
Carrying over from IRC, the conditional addition of Packagist is a one for my vote.  I just feel that forcing its inclusion might hurt those that don't need it (performance wise) as every load of Extend processes the JSON from the repositories listed enabled in `extensions/composer.json`

GawainLynch:
Thought it might be useful to do a comparison with `composer update` with and without Packagist enabled (cache is warm):

**Packagist Disabled (showing slowest run)**
```bash
$ time composer update
[..snip..]
real	0m3.411s
user	0m0.444s
sys	0m0.459s
```

**Packagist Enabled (showing fastest run)**
```bash
$ time composer update
[..snip...]
real	0m36.436s
user	0m18.588s
sys	0m0.853s
```

rossriley:
Yes, that's a fairly drastic difference then 12 times slower if I'm interpreting correctly. 

I did think of one other idea that might be a halfway point between this and turning off packagist. 
We do have the ability to require individual packagist packages as part of our own Satis repository.

We could either:
1. Parse the extension dependencies at submission time and add these to the Satis repo.
2. Allow a small number of requested packages to be whitelisted that extension developers can use, we could audit these to keep to a small number and thus keep the speed high.

With option 1 we probably end up getting slow pretty quickly too, option 2 allows us to keep to a minimal number of packages to parse, but obviously reduces the choice for developers.

GawainLynch:
Yes, you're interpreting correctly...  There were worse results obviously, but that was the smallest delta that I saw. Also, adding packagist.org adds one more point of failure to an increasingly complex web of requirements.

I think I like option 2 myself.  Though that in itself *could* create a little more work for us long term, or not, with keeping the whitelist updated/managed.

rossriley:
I guess we can always start with option 2 and then graduate from there if we think we need to. If we open it up completely to start with then there's not really any going back.

GawainLynch:
Agreed

PR #2677, opened 2015-02-02
Make ShowPackage::execute() optionally query the Bolt installs root composer.json

GawainLynch:
composer.json


bobdenotter:
No problems here!

PR #2680, opened 2015-02-02
Fix filebrowser route binding name

CarsonF:


bobdenotter:
Looks good. Not even going to wait until Travis finishes. Yolo.

PR #2683, opened 2015-02-02
Generating thumbs route

CarsonF:
Depends on https://github.com/bolt/bolt-thumbs/pull/16


bobdenotter:
Good to go!

rarila:
@bobdenotter It's unredable and simply a possible source of errors if you have to count %s and simply another unnecessary function call for no reason that's 2,5 times slower. :scream_cat:

rarila:
@CarsonF Thanks :+1:  :-)

PR #2684, opened 2015-02-03
Prevent apache headaches for new installs

jkazimir:
This fixes an issue with .htaccess for people using apache without mod_negotiation which has been mentioned on IRC a few times now.

Cheers!


GawainLynch:
Legendary @jkazimir!

PR #2685, opened 2015-02-03
[RFC] Indeterminate config values

CarsonF:
### Problem
Values can be based on other values, which can change. After changing a value, the dependent value(s) have to be re-set to account for this change. 
An example of this are the url paths for ckeditor (I think modifying them in the actual config file is broken right now....which defeats the purpose of having them in a config in the first place).

### Solution
Determine the value at the latest possible moment. 
I created an interface, `ResolvableInterface`, for place-holder-like objects. These objects hold the variables necessary to _resolve_ the actual value when the time comes.

When the value is accessed via `Config::get` the value is _resolved_ at that point (uncached). 
The values can also be _updated_ via `Config::set` or the actual yaml files when parsing.

Any value that is resolvable should be defined as such in `Config::getDefaults`.

----
I created two implentations, `Url` and `Path`, which reference values related to `ResourceManager::getPath/Url`. 
Although currently only `Url` is used, because the only use case for `Path` may break BC (thumbnail images).

----

There is a minor performance implication as values are _re-resolved_ for each call. Also, when pulling an array of values, iteration is needed to verify all values have been _resolved_.

Thoughts? Comments? This is awesome/mehh...over-engineered?

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -2,6 +2,7 @@
 
 namespace Bolt;
 
+use Bolt\Configuration\Type;
 use Bolt\Exception\LowlevelException;
 use Bolt\Library as Lib;
 use Bolt\Helpers\Arr;
@@ -127,7 +128,11 @@ class Config
             $part = & $part[$key];
         }
 
-        $part = $value;
+        if ($part instanceof Type\ResolvableInterface) {
+            $part->update($value);
+        } else {
+            $part = $value;
+        }
 
         return true;
     }
@@ -163,7 +168,9 @@ class Config
             $part = & $part[$key];
         }
 
-        if ($value !== null) {
+        if ($value instanceof Type\ResolvableInterface) {
+            return $value->resolve($this->app);
+        } elseif ($value !== null) {
             return $value;
         }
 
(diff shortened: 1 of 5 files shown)


CarsonF:
Ah yes. I still need to add serializing support. Tests fail due to config cache unserializing incorrectly.

src/Config.php
```diff
@@ -1045,4 +1044,34 @@ public function getCurrentTimestamp()
...
+            } else {
+                $merged[$key] = $value;
+            }
+        }
+
+        return $merged;
+    }
+
+    /**
+     * @deprecated Will be removed in Bolt 3.0
```
GawainLynch: Just kill it already :-)

CarsonF: Well we need this one to handle Resolvable objects and accept_file_types key. But all the other cases of Lib::mergeRecursiveDistinct can be replaced with array_replace_recursive.

CarsonF:
I'm closing this for now because these serializable problems are not going to go away easily. We can readdress this after TaxonomyType and ContentType classes are created.

GawainLynch:
I am so looking forward to when we branch `release/2.1`… :octocat:

CarsonF:
It will be :tada: :beer: time

PR #2691, opened 2015-02-03
Allow setting of regex pattern, replacement and case of uploaded file names

### LINKED ISSUE #1249: Bug: Upload files/images behaviour (re)naming and/or overwriting
- In some instances (from the uploaded files page) it seems that dashes are converted to underscores. 
- From the uploaded files page, if you upload a file with the same name, the file is then overwritten (which can be desired or non-desired)
- When uploading from a Image field, the file is not overwritten (e.g. number attached: ``image (1).jpg``)

I think this can be thought out a bit and imho handled by a file-manager (see also #1024 )

GawainLynch:
The maintainer of siriusphp/upload added the ability today to provide a sanitisation callback for the upload handler.  This PR adds config options to allow us to use this.

Might be worth thinking about deprecating `accept_file_types:` key and making it part of `upload:`.

Fixes #1249 and #2237

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -144,6 +144,15 @@ thumbnails:
   exif_orientation: true
 #  browser_cache_time: 2592000
 
+# Uploaded file handling
+# 
+# You can change the pattern match and replacement on uploaded files and if the 
+# resulting filename should be transformed to lower case.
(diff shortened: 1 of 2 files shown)


bobdenotter:
Merging in!

Issue #2628, opened 2015-01-30
[RFC] Version numbers in view/lib/ on folders


rarila:
After this came up by @cdowdy in #2599 and some refactoring I did in Gruntfile.js came up with the idea of removing the versions from the folders and instead adding a ``Versions.md``. in ``view/lib/`` that lists the version numbers of all libs. That would still make it easy to compare if there are new versions and make it easier – although that was only a search/replace – to update. Just edit the version number one time.

As additional sugar we could add links there to the source of the libs that would even more ease the process. Also we could add some info there how update and if there's something special to care of.

Anyone against that approach?

GawainLynch:
:+1:  from me.

Wondering if we shouldn't do the same with licences of said libraries.

bobdenotter:
Yes, please remove the version numbers from the folders! :100:

cdowdy:
this is also why I [mentioned earlier](https://github.com/bolt/bolt/issues/2585#issuecomment-71571454) using a package manager like [Bower](http://bower.io/) or the others.  

As a dev you'd just have to have bower installed. Seeing as how you already use grunt this shouldn't be a big issue. To update you'd just ```bower update``` then compile your scripts. All license info is still in the folder, version numbers in a json file or in your bower json file for all bower components and not on a folder or file name.   

I've started a branch few days ago [here](https://github.com/cdowdy/bolt/blob/convert-to-bower/app/view/bower.json) with bower if you'd like to possibly see how it may look.

rarila:
@cdowdy I have some other changes in the pipeline and will do them first. Then we could have a look at it.

But we would have to be careful and have to check, if there were no changes made to the code and then the risk is higher, that some changes may break things. And then there's stuff like CKEditor, where we use a customized version.

Do we need this extra file or could we just use [package.json](https://github.com/cdowdy/bolt/blob/convert-to-bower/app/view/package.json) as it is done with moment.js there?

cdowdy:
I did a quick (read extremely dirty ) dif against the files I placed into bower that I linked above. For instance I didn't include jquery hotkeys because that is a customized version. I also tried to include the current version used in Bolt and lock it to that version. Most of these in the bower json pull from their associated github   

Bower uses a bower.json to "hold" the dependencies. We could use the same package.json if you were to install those dependencies through NPM.  Instead of using ```bower install <package-name``` you'd just install like we currently do for a grunt plugin ```npm install <package-name> ```  . I don't know if many of these are available through npm though.     

[...]

PR #2697, opened 2015-02-03
converting front-end JS to bower

cdowdy:
related to: https://github.com/bolt/bolt/issues/2628  

created  
* ```.bowerrc``` file to tell where to install bower components.   
* ```bower.json``` for bower dependencies  
* updated ```lib.js``` with new bower included files  

I've deliberately only added a few files to keep possibly breakage at a mininmum. I diffed these and they weren't custom versions.  

```bootbox.js ```, ```jquery-cookie```, ```jquery.formatDateTime``` and tagcloud are all "locked" to their current version used in bolt inside bower.json.  

This will need to be thoroughly checked before commited :(. Is it more suited toward the canary branch?


bobdenotter:
Sweet! 

> app/view/lib/tmp/bower-assets.js

Should that file be committed to git?

cdowdy:
> Should that file be committed to git?

I have that there just to mirror the other files in ```lib```.  It doesn't need to be there since its included into lib.min.js. I committed it just in case anyone came across and wondered "where the hades is magnific coming from" since bower_components isn't committed.  

in short no it doesn't need to be committed really haha

rarila:
I would either put ignore the tmp folder or run a delete target on it after run. But I think the whole process can streamlined a bit anyway probably by minify everything and get a sourcemap for free without the need for a tmp file at all. Will eventually have a look into it after merge.

bobdenotter:
In that case.. Merging in, and we'll finetune as we go along. Thanks again, @cdowdy

PR #2700, opened 2015-02-04
Cleanup Routing Controller

CarsonF:
- No more statics
- I think config parsing is easier to follow
- Moved two methods from Storage to Routing (10d2296)


bobdenotter:
Sweet. :-)

PR #2701, opened 2015-02-04
flashbag messages

CarsonF:


bobdenotter:
Also sweet. We might have to take a good look if it turns out that this change will flood users with notices, though. But, we'll see that when we get there.

PR #2702, opened 2015-02-04
Translation Service Provider updates

CarsonF:


bobdenotter:
Whoop!

GawainLynch:
This breaks front end:

```
Whoops \ Exception \ ErrorException (E_WARNING) 
is_dir() expects parameter 1 to be a valid path, array given

src/Provider/TranslationServiceProvider.php::49
```

PR #2703, opened 2015-02-04
fixes issue using file manager over https

glasspelican:
attempts to redirect user to http when using some links over https


bobdenotter:
Good catch! Thanks!

PR #2705, opened 2015-02-04
Translation Service Provider updates #2

CarsonF:


GawainLynch:
The statics were one of the big concerns for @rossriley as well.

CarsonF:
Yeah I saw that on IRC. I'm not the biggest fan either but I wanted to make the logic publicly available. This was the easiest spot to do that. We could extend the translator and add a method for this but that was deeper than I wanted to go.

For the record, I just removed a bunch of statics in #2700. Surely this makes up for the two I've added :)

GawainLynch:
Yeah, while I did add a lot of statics via `Bolt\Library`, that was very much a temporary work around until we found a better solution to them.

I'll let @rossriley chime in though.

CarsonF:
Soooo. merge? :8ball:

GawainLynch:
This has a "no" from @rossriley last time he commented on IRC.

GawainLynch:
…only on the static from memory.

CarsonF:
Ok I'll change the static to non-static. Which means you'll have to create the object just to call that method, which is stupid.

GawainLynch:
Consensus reached! :+1:

PR #2707, opened 2015-02-04
Bugfix multiple flash messages

CarsonF:
This is caused by Config::checkConfig being called for ajax requests. Flash bag messages are only removed when the backend twig template pulls them. So every ajax request will add one error, which will build up until they are flushed with a backend twig response.

For example, here's a request list:
1) /bolt -> removes all errors (now 0)
2) /async/latestactivity -> +1 error (now 1)
2) /async/latestactivity -> +1 error (now 2)
3) /bolt -> +1 error (now 3). They are flushed here, so 3 errors are rendered and the list is reset.

I moved `Config::checkConfig` call to only on the `Backend` controller, since this is the only time they will be rendered.

Although the could be rendered on the Frontend as well, should I call be added there as well?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -358,10 +358,6 @@ class Application extends Silex\Application
             return $response;
         }
 
-        // Sanity checks for doubles in in contenttypes.
-        // unfortunately this has to be done here, because the 'translator' classes need to be initialised.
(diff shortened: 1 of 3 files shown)


bobdenotter:
Looks good to me! Thanks! :100:

PR #2708, opened 2015-02-04
R.I.P. Library::fixPath() 

GawainLynch:
Library::fixPath() had only three remaining users:
* `Library::loadSerialize()`
* `Library::saveSerialize()`
* `TwigExtension::menuHelper()`

Both `loadSerialize() and `saveSerialize()` are only called from `Bolt\Config` and the paths are constructed fine.

`TwigExtension::menuHelper()` was a bit of fun, it now handles menu `path:` keys a lot better (I hope).

The following permutations all work:
* `{contenttype}`
* `{contenttype}/`
* `/{contenttype}`
* `/{contenttype}/`
* `{contenttype}/{id|slug}`
* `/{contenttype}/{id|slug}`

Also if someone (mis)configures `path:` with any of the following, they will create working links, but get an error logged in the system log:
* `http://`
* `https://`
* `//`

**NOTE:** If none of these match, `menuHelper()` will just assume that whatever the string given in `path` is a relative site URI.

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -14,49 +14,6 @@ use Bolt\Translation\Translator;
 class Library
 {
     /**
-     * Cleans up/fixes a relative paths.
-     *
-     * As an example '/site/pivotx/../index.php' becomes '/site/index.php'.
-     * In addition (non-leading) double slashes are removed.
-     *
-     * @param  string $path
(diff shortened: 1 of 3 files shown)


bobdenotter:
Schweeeeet.

PR #2710, opened 2015-02-04
Use ISO/IEC 80000 format

GawainLynch:
Wrong is wrong.

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -65,11 +65,11 @@ class Library
     public static function formatFilesize($size)
     {
         if ($size > 1024 * 1024) {
-            return sprintf("%0.2f mb", ($size / 1024 / 1024));
+            return sprintf("%0.2f MiB", ($size / 1024 / 1024));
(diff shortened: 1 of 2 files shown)


bobdenotter:
Eew. I still have nightmares from when phpmyadmin suddenly started putting everything in "Mebibytes". "The maximum upload size is 8.192 KiB". Uhm, wait, so?

We could also fix this the other way round, and just divide by 1000, instead of 1024. ;-)

rarila:
Could we use feet, gallons, ounces, fahrenheit, jiffies and dog years, please ;-)

bobdenotter:
"This image is 0.0004 furlong by 0.000001 parsec"

GawainLynch:
To be honest, I was most objecting to case rather than base... However I figured tht either @rarila or @tobias2k wouldn't force me to up my ante...

bobdenotter:
Enough chitchat. merging in!

PR #2718, opened 2015-02-05
Implement slugify and kill one more old lib.php function

### LINKED ISSUE #1232: makeSlug() vs.slugify()
So we have ``makeSlug()`` and ``util::slugify()`` where the latter isn't used at all (as ``util::remove_accents()``

So, is ``util::slugify()`` somehow left over and could/should be removed or already there for future use?

Another thing: both doesn't seem to handle umlauts right/nice. I'd like to have the slug ``bratwuerste`` when I input ``Bratwürste``. Instead I get ``bratwurste``.

So the next question is, would changing the behavior of ``makeSlug()`` break existing installations? Some ``strtr()`` in front of that code would do the least. Also ``iconv("UTF-8", "ASCII//TRANSLIT", $slug)`` and ``Transliterator::transliterate`` would be worth a consideration.

GawainLynch:
Fixes #1232

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
         "bolt/pathogen": "~0.6",
         "bolt/thumbs": "~2.0",
         "brandonwamboldt/utilphp": "~1.0",
+        "cocur/slugify": "~1.0",
         "composer/composer": "dev-master",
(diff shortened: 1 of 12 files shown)


bobdenotter:
Merging in!

PR #2720, opened 2015-02-05
[RFCish] Symfony Filesystem service

GawainLynch:
We use `Symfony\Component\Filesystem\Filesystem` in several places, some of them are very early and have to stay that way, and there are more still where we could leverage it.

This implements a service provider and uses it where we currently create a new instance.  Basically this is an opportunity to expand on that with fractional overhead.

**Note for the OCDs:** `$app['symfony.filesystem']` chosen as a name due to `$app['filesystem']` being used as a project specific service provider class...  Suggestions welcome prior to merge...  or after, we're in alpha after all :camel:

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -352,6 +352,7 @@ class Application extends Silex\Application
             ->register(new Thumbs\ThumbnailProvider())
             ->register(new Provider\NutServiceProvider())
             ->register(new Provider\GuzzleServiceProvider())
+            ->register(new Provider\SymfonyFilesystemServiceProvider())
             ->register(new SlugifyServiceProvider());
 
         $this['paths'] = $this['resources']->getPaths();
--- a/src/Content.php
+++ b/src/Content.php
@@ -2,12 +2,11 @@
 
 namespace Bolt;
 
-use Silex;
-use Symfony\Component\Filesystem\Filesystem;
 use Bolt\Library as Lib;
 use Bolt\Helpers\String;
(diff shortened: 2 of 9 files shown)


CarsonF:
What do we have against flysystem?

bobdenotter:
> What do we have against flysystem?

We quite like Flysystem, but Flysystem's abstraction adds a (small) overhead, when there's nothing to abstract because we're just using the local filesystem. We use both Symfony's Filesystem as well as Frank's Flysystem, and I suspect it will stay that way.

CarsonF:
Yeah I guess some of the operations we are doing wouldn't make since on a remote filesystem.

The main thing is we are using an abstraction so a testing one can be swapped in easily :+1:

GawainLynch:
What @bobdenotter said...  Invoking all of Flysystem to test if a file is writeable...  :8ball:

CarsonF:
It's not that much extra, if the filesystem is already initialized

GawainLynch:
Creating a base class object of each and checking if a file exists:

| Library | Peak Memory |
|---------------|------------------|
| Filesystem | 145,224 bytes|
| Flysystem | 321,320 bytes (local adapter only) |

To be honest, the more I look at Flysystem's code, the more I am leaning towards @CarsonF's idea of using that though, with the exception of the calls that need to be made statically.

GawainLynch:
#2747 wins.

PR #2723, opened 2015-02-05
Refactor out last instance of "URLify".

bobdenotter:


bobdenotter:
This is giving errors in Travis. "Slugify" does not give the same results as "Urlify", so we can't use it as a drop-in replacement. At least, not in it's current form.

PR #2724, opened 2015-02-05
Dashboard UI

dmm1:
- new Panellayout
- fix quicklink margin
- adjust font-size
- thumnail styling & size changed from 54 to 80


app/view/sass/modules/_dashboard.scss
```diff
@@ -2,8 +2,12 @@
  * Dashboard specific stuff
  */
 
+h3 {
+    font-size:20px;
+    }
+
```
GawainLynch: Intended as un-classed?

dmm1: sorry missed that

bobdenotter:
Hi dmm1, 

I'm probably going to sound a little bit harsh, but please know that I mean this in the best way possible. I'm in no way trying to deter you in making contribution. In fact, I rather hope that once you get the know a bit more about how certain things have evolved, you'll make more contributions to help move forward on that course. :-) 

Anyhow, I think I see what you were going for with this PR, but we can't merge it in, in this form. For reference: here's side-by-side screenshots of your changes and the current version, for a smallish screen: 

screenshot 2015-02-07 08 22 22

screenshot 2015-02-07 08 22 28

- You've added rounded corners, and more boxes. When I made the initial layout for v2, there were rounded corners and boxes everywhere. I've asked a few designer friends/coworkers to help finetune, and they all were pretty firm about that "boxyness" needs to go, and rounded corners should _only_ be on the buttons. same applies to the 'frames' around the thumbnails. The developer in me had those everywhere, because they looked neat. Designer friends shot them down.

- We've made a difference in the thumb-sizes for the "dashboard" and "single contenttype overviews" to keep the former more compact and 'at-a-glance', while having a bit more information on the latter. See also the 2 vs. 3 lines of excerpt and the added "meta column" 

- I _do_ like the added whitespace in a bunch of places. We developers tend to cram as much stuff into one screen, but I know whitespace is very important. I do think that we should do that a bit more deliberate, gradually, and consistent. Adding 1 or 2 pixels of margin/pagging to each row in the table will already make a big difference i think. And it should also be done to the other overview pages, then.

dmm1:
@bobdenotter Thanks for you answere. This is the first time i am not disappointed that a PR is refused. Im not used to get such detailed, respectful and straight out response in OS Work.
After reading your answere i totally agree with your opinion. I'll try to give my best to follow these CI instructions on my next PR's :smile: .
Without to fawn this way of response is the way OS work should be done and makes Bolt (for me) unique.

PR #2728, opened 2015-02-06
Insight quickies

jkazimir:
Removed a few unused imports, and added newlines to the end of a couple of files.

Left the minified JS files with the missing newlines, as this would probably need to be fixed in how they're generated.

Cheers!


GawainLynch:
LGTM

PR #2731, opened 2015-02-06
Fix the "mysql was selected but is not supported" error notice.

bobdenotter:
Simply bumping the version will invalidate the cached config, ensuring the config files are reloaded which allows Bolt to pick up the correct values.


GawainLynch:
It's been a while... But I am calling #1555 on this puppy!

PR #2733, opened 2015-02-06
Fixes "Uncaught ReferenceError: CKEDITOR is not defined"

rarila:
Yet another one.

Sidenote: we should move ``jquery-watches`` to our own sourcetree and add it to init.js as it is 30% patched anyway and no updates are to expected


bobdenotter:
Thanks! :+1:

PR #2738, opened 2015-02-06
Flip array replace arguments

CarsonF:

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -430,12 +430,12 @@ class Config
             }
 
             // Make sure we have these keys
-            $field = array_replace($field, array(
+            $field = array_replace(array(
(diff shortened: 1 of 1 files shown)


bobdenotter:
water

GawainLynch:
Strangely `git blame` doesn't assign any blame to that one...

bobdenotter:
Gawain: That's probably because this error has been in there since _before_ we moved `app/src/` to `src/`.

CarsonF:
I'm to blame #2707

GawainLynch:
Nah mate, it was the gnomes!

CarsonF:
@bobdenotter's water gif is the real winner here!

PR #2739, opened 2015-02-07
Fix double autoloading of extension support files

rossriley:
This should solve the extension installer script bug where the post-install  script was throwing a fatal error.

The class is copied as is and only autoloaded via the PSR-0 namespace rather than via autoload_files.


GawainLynch:
Lovely, doing related stuff right now!

PR #2745, opened 2015-02-07
Fix/no more nbsp in CKeditor. Fixes #2660

### LINKED ISSUE #2660: &nbsp; in editor happening in 2.0.5
Bolt 2.0.5 apparently still has issues with nbsp stuff.

They appear when editing text in the WYSIWYG editor: backspacing, copying & pasting, entering, etc.
I'm not exactly sure which action mentioned above causes it..

I do know that the editors add their article like this (especially long texts):
Type stuff in a text editor. 
Paste it into the WYSIWYG editor
Use the xT button to discard all shitty code from editors
Save the article
...
(Another) editor opens the article, edits it (by backspacing, copying & pasting, entering, etc.)
Saves it once again.

bobdenotter:


GawainLynch:
One less script to run in my publishing workflow!

Issue #2264, opened 2014-12-03
Feature request: create/edit user without options like admin and root


gwbonline:
I want to grant the chief editor permission to create and edit users. Problem is that the chief editor also has the options 'root' and 'admin' to select for their roles. Isn't a simple solution possible to make 'root' and 'admin' invisible for other users than the root and admin accounts? Now it is a potential threat to give the chief editor permission to create and edit a user, because the chief editor can deactivate the root user or give someone else root access.

bobdenotter:
This has been fixed in 2.1.0. :-)

PR #2757, opened 2015-02-08
User Role-based Hierarchy.

MrAdamCook:
userEdit() function modified to check which roles the currentuser can add/edit to other users.
This prevents a 'developer' role from giving himself (or others) the 'root' role.

userAction() function modified to prevent the currentuser from enabling, disabling or deleting himself.

before() function modified to prevent the currentuser from editing, deleting, disabling or enabled users with roles higher than his/her own.
This stops the user deleting the 'root' access.

### DIFF
--- a/app/config/permissions.yml.dist
+++ b/app/config/permissions.yml.dist
@@ -79,6 +79,19 @@ global:
     # Likewise, 'logout' needs to be granted to 'everyone', otherwise people
     # cannot log out anymore.
     logout: [ everyone ]
+    
+    # User roled-based hierarchy settings. This will restrict users from being
+    # able to add, edit, enable, disable and delete users with a higher role
+    # than their own.
+    # For example:
+    # Users with a role of admin can only be added, edited, enabled, disabled and
+    # deleted by [ root, admin ] roles.
+    users:roles-hierarchy:root: [ root ]
+    users:roles-hierarchy:admin: [ root, admin ]
+    users:roles-hierarchy:developer: [ root, admin, developer ]
+    users:roles-hierarchy:chief-editor: [ root, admin, developer, chief-editor ]
+    users:roles-hierarchy:editor: [ root, admin, developer, chief-editor ]
+    users:roles-hierarchy:guest: [ root, admin, developer, chief-editor ]
 
 # For content type related actions, permissions can be set individually for
 # each content type. For this, we define three groups of permission sets.
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -121,6 +121,7 @@ class Backend implements ControllerProviderInterface
         $ctl->match('/users/edit/{id}', array($this, 'userEdit'))
             ->before(array($this, 'before'))
             ->assert('id', '\d*')
+            ->value('checkUserRoleHierarchy', 'true')
             ->method('GET|POST')
             ->bind('useredit');
 
@@ -140,6 +141,7 @@ class Backend implements ControllerProviderInterface
 
         $ctl->get('/user/{action}/{id}', array($this, 'userAction'))
             ->before(array($this, 'before'))
+            ->value('checkUserRoleHierarchy', 'true')
             ->method('POST')
             ->bind('useraction');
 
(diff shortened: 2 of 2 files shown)


MrAdamCook:
This also resolves open issue #2264.

bobdenotter:
Hi @MrAdamCook, This is neat! I'll try it out, and give feedback. 

First thing I noticed: Can you update permissions.yml.dist for this as well?

tobias2k:
Took a quick glance; Controllers/Backend.php has a regex blooper, line
1952 - I don't think you should use regular expressions here at all, but
instead throw the current route against a list of routes to match, or,
better yet, do things properly and set some request state in a before()
handler for the relevant routes. Either way, the regex that is currently
in place (/(user)\w?/i) doesn't look right at all. The parentheses don't
do anything at all, and neither does the \w?. The regex basically
amounts to stripos($route, 'user') !== false, and it will match all
sorts of undesirable things, such as 'Lusers', 'AbuserReport', etc.

And one more thing: preventing users from changing their own roles
doesn't add any security unless you also prevent users from creating
users. The current mindset is that "useredit amounts to full access",
and preventing users from modifying themselves doesn't change that - if
you can create new users, you can use that path to modify your own
account indirectly, and users will find out pretty quickly.

On Sun, Feb 08, 2015 at 04:58:40AM -0800, Adam Cook wrote:
> userEdit() function modified to check which roles the currentuser can
> add/edit to other users.  This prevents a 'developer' role from giving
> himself (or others) the 'root' role.
> 
> userAction() function modified to prevent the currentuser from
> enabling, disabling or deleting himself.
> 
> before() function modified to prevent the currentuser from editing,
> deleting, disabling or enabled users with roles higher than his/her
> own.  This stops the user deleting the 'root' access.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/bolt/bolt/pull/2757
> 
> -- Commit Summary --
> 
>   * User Role-based Hierarchy.
> 
> -- File Changes --
> 
>     M src/Controllers/Backend.php (29)
> 
> -- Patch Links --
> 
> https://github.com/bolt/bolt/pull/2757.patch
> https://github.com/bolt/bolt/pull/2757.diff
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/bolt/bolt/pull/2757

-- 
Tobias Dammers - tobias@twokings.nl - 070-3457628 - www.twokings.nl
Maandag t/m donderdag van 9.00 tot 17.30
Voor dringende vragen, mail naar support@twokings.nl

MrAdamCook:
That's fair enough - Still trying to pick up on all the possibilities with Bolt, only really picked it up yesterday for a client. So I'll improve the way it handles picking up the user pages and update.

Also I think I may have possibly confused you on the prevention of users from their own actions and editing. I've created an imgur album with some screenshots and descriptions which will hopefully make it clearer.
http://imgur.com/a/VwEMs

However, I have picked up on a new potential issue.
A user can edit his/herself, but can also remove their own / highest access role. This might not be a huge issue. But might be more ideal if their highest access role (or all of them) are locked/disabled fields. What do you think?

bobdenotter:
> A user can edit his/herself, but can also remove their own / highest access role. This might not be a huge issue. But might be more ideal if their highest access role (or all of them) are locked/disabled fields. What do you think?

This might've been solved with these PR's: #2764 and #2765.

MrAdamCook:
Looks like it has. I'll merge and confirm tonight.

MrAdamCook:
Not quite.
The logged in user can still remove their own 'admin' and 'developer' roles.

GawainLynch:
Just tested and something has caused a recent regression.

GDmac:
Bolt currently isn't a "full-fledged" community solution out of the box (forum, comments, accounts, etc). Another cms i work with has an option to let users delete their own account. There are some precautionary measures as: when an admin deletes an account, should comments and posts be assigned to another user? Also (more important) at least one super-admin account has to stay up.

GawainLynch:
Changes cherry-picked into master

PR #2764, opened 2015-02-08
Adjust the logic that checks if a user is enabled

### LINKED ISSUE #2763: After disabled a user, exception :/
Hello,
after disabled a user and reload his page -> exception ...

<pre>
An exception has been thrown during the rendering of a template ("You must be logged in to use this.") in "dashboard/_aside.twig" at line 23.</pre>

GawainLynch:
Fixes #2763

* Splits first user creation out of editUser into its own method
* Move test for a user being enabled into Backend::before() to catch as early as possible

### Changelog
- Fixed: Logged in user can no longer disable themselves
- Fixed: Disabling a logged in user with force a logout

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -1745,6 +1745,13 @@ class Backend implements ControllerProviderInterface
             return Lib::redirect('userfirst');
         }
 
+        // Confirm the user is enabled or bounce them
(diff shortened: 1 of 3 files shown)


bobdenotter:
Looks good to me.. Not sure if all bases are covered like this, but it's a definite improvement. :+1:

PR #2768, opened 2015-02-09
Check correct array element

GawainLynch:
Local & pending extensions won't show if there are no existing composer installed packages...  :see_no_evil:


bobdenotter:
Topkek! :100:

PR #2771, opened 2015-02-10
Fix Flysystem requiring write access

### LINKED ISSUE #2770: Flysystem has changed behaviour to not accept readonly filesystems
https://github.com/thephpleague/flysystem/commit/cfa08d0aa5ef861ed7614b2b8d9ca88465d150e0
https://github.com/thephpleague/flysystem/pull/358

```php
if (! is_writable($realRoot)) {
    throw new \LogicException('The root path '.$root.' is not writable.');
}
```

CarsonF:
Fixes #2770

- Subclassed Local adapter. Any write operations just return false.
- Fixed some CCP bugs
- Fixed some usages of pathogen with flysystem

### DIFF
--- a/src/Filesystem/Local.php
+++ b/src/Filesystem/Local.php
@@ -80,6 +80,9 @@ class Local extends LocalBase
     {
         $location = $this->applyPathPrefix($dirname);
 
+        // mkdir recursively creates directories.
+        // It's easier to ignore errors and check result
+        // than try to recursively check for file permissions
         if (!is_dir($location) && !@mkdir($location, 0777, true)) {
             return false;
         }
@@ -96,6 +99,4 @@ class Local extends LocalBase
 
         return parent::deleteDir($dirname);
     }
-
-
 }
--- a/src/Filesystem/Local.php
+++ b/src/Filesystem/Local.php
@@ -15,7 +15,7 @@ class Local extends LocalBase
 
     protected function ensureDirectory($root)
     {
-        if (is_dir($root) === false || !is_writable($root)) {
+        if (!is_dir($root) || !@mkdir($root, 0755, true)) {
             return false;
         }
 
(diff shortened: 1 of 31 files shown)


CarsonF:
Idk why those two tests are failing. They pass locally...

src/Filesystem/Local.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    protected function ensureDirectory($root)
+    {
+        if (is_dir($root) === false || !is_writable($root)) {
+            return false;
+        }
+
+        return realpath($root);
+    }
```
GawainLynch: Changing the behaviour here doesn't make sense.

GawainLynch: Commenting before coffee doesn't ether…  

What I meant was that if the parent directory **is** writeable, and the directory being *ensured* doesn't exist, we should create it.

src/Filesystem/Local.php
```diff
@@ -0,0 +1,102 @@
...
+
+        // mkdir recursively creates directories.
+        // It's easier to ignore errors and check result
+        // than try to recursively check for file permissions
+        if (!is_dir($location) && !@mkdir($location, 0777, true)) {
+            return false;
+        }
+
+        return array('path' => $dirname, 'type' => 'dir');
+    }
```
GawainLynch: I agree with @bobdenotter's reasoning for a safeguard of `777`, but I know there are lots of us that find `777` **totally** unacceptable for production systems.  Wondering if it isn't better to have a config option for this to keep the paranoid happy? :guardsman:

src/Filesystem/Local.php
CarsonF: Doh, you're right

GawainLynch: About the coffee… Tell me about it :eyes:

src/Filesystem/Local.php
CarsonF: What we determined was:
- directories always need executable permissions (and this is only for directories)
- umask modifies the 777 (usually to make it 755).

bobdenotter:
Merging in!

PR #2775, opened 2015-02-10
Only added flash errors on backend requests. Fixes #2774

### LINKED ISSUE #2774: Excessive error messages when template isn't found
bolt-theme-does-not-exist

The sheer amount of errors when a theme isn't found might be a bit excessive...

To recreate: set theme to something that isn't a theme in config.yml and save.

This is Bolt 2.1.0 - alpha2 running on Apache 2.4.7 with php 5.5.9-1.

CarsonF:
Fixes #2774

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -3,9 +3,9 @@
 namespace Bolt;
 
 use Bolt\Exception\LowlevelException;
-use Bolt\Library as Lib;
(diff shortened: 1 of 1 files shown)


bobdenotter:
Green lights too! :traffic_light:

PR #2780, opened 2015-02-10
Issue 2359 and 2424

### LINKED ISSUE #2359: bug in tab group, default group assignment is incorrect.
The incidence occurs when some of the fields do not have a defined group, rather than grouped into "ungrouped" appear in the last group of the iteration.

for example.
```yml
orders:
    name: orders
    singular_name: order
    fields:
        title:
            type: text
            class: large
        image:
            type: image
            group: media
        message:
            type: html
            height: 300px
```
In this case the first go to "Ungrouped", the second "Media" and also the third show in "Media" and the latter is the error, apparently the attribute data-tab is assigned to the last variable instead of "ungrouped" .

merry christmas (jo jo jo), and I wish that consumerism does not take over this party ^^

### LINKED ISSUE #2424: Can't add content type to default Meta group
Setting a field to group: meta creates a second meta tab.
v: 2.0.2

rarila:
Fixes #2359
Fixes #2424 

A bit hackish, but to make it clean would mean a bigger refactoring

### DIFF
--- a/app/view/twig/editcontent/_buttons.twig
+++ b/app/view/twig/editcontent/_buttons.twig
@@ -1,0 +1,57 @@
+<div class="form-group" style="margin-top: 40px;">
+    <div class="col-sm-12">
+
+        <div class="btn-group">
+            <button type="button" class="btn btn-primary" id="savecontinuebutton">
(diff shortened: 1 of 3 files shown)


rarila:
Hold it for now, I think there's something missing…

rarila:
Will be redone…

PR #2788, opened 2015-02-11
debug_print_backtrace() returns void and can't be passed to a variable

### LINKED ISSUE #2610: ResourceManager::getApp bad exception message
https://github.com/bolt/bolt/blob/master/src/Configuration/ResourceManager.php#L391

`debug_print_backtrace` returns void

@rossriley

GawainLynch:
Fixes #2610. Ugly, but does the job until this is refactored out.

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -425,7 +425,9 @@ class ResourceManager
     public static function getApp()
     {
         if (! static::$theApp) {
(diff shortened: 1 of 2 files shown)


bobdenotter:
I don't readily see how we can refactor this out, without losing functionality.. Are we sure `dump()` is already available when this is called? 

Anyways: :+1:

GawainLynch:
`dump()` maybe, maybe not.  "*Thankfully*" I used `debug_backtrace()` for this one.

bobdenotter:
Choo choo!! :steam_locomotive:

PR #2765, opened 2015-02-08
New users and the current users don't need to disable themselves

GawainLynch:
Fixes #2763 for stable branch

### Changelog
- Fixed: Logged in user can no longer disable themselves

PR #2791, opened 2015-02-11
Prevent logged in user from disabling themself through useractions.

MrAdamCook:
Fixes: #2763
Updates: #2764, #2765

* Prevents the logged in user from seeing user actions for their own user account within the admin/users page.
* Prevents the logged in user from submitting a post request method to /user/{action}/{id} if the id parameter is their own.

### Changelog
* Fixed: Logged in user can no longer enable, disable or delete themselves.

### DIFF
--- a/app/view/twig/users/_userlist-actions.twig
+++ b/app/view/twig/users/_userlist-actions.twig
@@ -7,39 +7,41 @@
         <i class="fa fa-info-sign"></i>
         <span class="caret"></span>
     </button>
-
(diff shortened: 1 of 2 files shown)


MrAdamCook:
#2764, #2765 only prevent the user from disabling themself with the userEdit page.
This expands it to the userActions in the dropdown list on the users page.

GawainLynch:
Changes cherry-picked into master

PR #2800, opened 2015-02-12
Fix Omnisearch to generate paths

CarsonF:
Also a couple other things.

Is everyone ok with `app->path()`? (3658157)

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -578,7 +578,7 @@ class Application extends Silex\Application
      *
      * @return string The generated path
      */
-    public function path($route, $parameters = array())
+    public function generatePath($route, $parameters = array())
     {
         return $this['url_generator']->generate($route, $parameters);
     }
--- a/src/Application.php
+++ b/src/Application.php
@@ -5,16 +5,16 @@ namespace Bolt;
 use Bolt\Exception\LowlevelException;
 use Bolt\Helpers\String;
 use Bolt\Library as Lib;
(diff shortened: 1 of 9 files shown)


src/Application.php
```diff
@@ -569,4 +566,20 @@ public function getVersion($long = true)
...
+     *
+     * @param string $route      The name of the route
+     * @param array  $parameters An array of parameters
+     *
+     * @return string The generated path
+     */
+    public function path($route, $parameters = array())
+    {
+        return $this['url_generator']->generate($route, $parameters);
+    }
```
GawainLynch: Like the idea, not sure if the name isn't too ambiguous though (per IRC discussion)

CarsonF: This is the same as what `UrlGeneratorTrait` and `Library` call. But yes I agree. 
I thought about changing it to `generatePath`

GawainLynch: That is actually pretty valid reasoning.

bobdenotter: Yes, I like `generatePath` too. :+1:

bobdenotter:
@CarsonF Don't forget to change the name to `generatePath()`;-)

GawainLynch:
Travis failures are due to timeouts on `http://loripsum.net/` again.

bobdenotter:
Choo Choo! :steam_locomotive: :train:  :train:  :train:  :train:

PR #2804, opened 2015-02-12
Fixed a bug with some utf8 characters becoming question marks after savi...

dwr:
...ng into database.

Fixes #2648

PR #2806, opened 2015-02-13
Add some comments to the fix introduced in #2804

bobdenotter:


GawainLynch:
It might be just me, but I am calling #1555 on this one!

Issue #2807, opened 2015-02-13
Create user page is showing "Edit user"


GawainLynch:
![create user page title](https://cloud.githubusercontent.com/assets/1427081/6184511/921dd356-b35a-11e4-87be-5dfcc683563c.png)

GawainLynch:
Fixed in #2808

PR #2808, opened 2015-02-13
Make Backend::userEdit() correctly select 'new' or 'edit' user actions

GawainLynch:
Fixes: #2807

Also fixes a Codeception failure that was missed.

Uncovered a second hidden error, that is also fixed in second commit.


bobdenotter:
Whoooop!

Issue #1129, opened 2014-04-18
Meta in body and  insertAfterMeta


truffo:
You can have meta tag in html body, for example : 

```
<body>
 ... 
 <meta itemprop="datePublished" content="2011-03-25">March 25, 2011
</body>
```

If you use insertAfterMeta method, this tag is matched and the data add in the body but not in the html header.

SahAssar:
I ran into the same issue when using microdata (#2786). Since bolt adds it's own meta tags in "after meta" that are not valid to use in body and since HTML5 allows for meta tags in body this is clearly an issue in my mind.

jadwigo:
pinging myself as this is related to an extension I'm writing

bobdenotter:
I've taken the liberty of assigning you to this issue, then. ;-)

jadwigo:
In that case...
My opinion is that 'aftermeta' should insert code after the last meta tag **in the head** of the html document.

jadwigo:
and maybe 'beforecss', 'aftercss', 'beforejs' and 'afterjs' should know about the head too

SahAssar:
I'd agree about beforecss and aftercss, but since js is often last in the body beforejs and afterjs should maybe be treated differently.

bobdenotter:
fixed in https://github.com/bolt/docs/pull/179

PR #2811, opened 2015-02-13
added extra snippet locations

jadwigo:
To make the situation in https://github.com/bolt/bolt/issues/1129 clearer I've added the following lcoations for snippets:

* beforeheadmeta
* afterheadmeta
* beforeheadcss
* afterheadcss
* beforeheadjs
* afterheadjs
* beforebodycss
* afterbodycss
* beforebodyjs
* afterbodyjs

The locations are essentially aliases for startofhead, endofhead, startofbody and endofbody because in daily usage those will be all you need.

In the theoretical case where the specific ordering must be even more controlled we'll have to switch to simpledom parsing or something similar.

(Docs will be updated after merge)

### DIFF
--- a/src/Extensions.php
+++ b/src/Extensions.php
@@ -583,6 +583,9 @@ class Extensions
             // then insert it into the HTML, somewhere.
             switch ($item['location']) {
                 case SnippetLocation::END_OF_HEAD:
+                case SnippetLocation::AFTER_HEAD_JS: // same as end of head because we cheat a little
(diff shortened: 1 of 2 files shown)


jadwigo:
the fun part is that this is forward and backward compatible because existing behaviour is not modified :)

GawainLynch:
Awesome :+1:

bobdenotter:
Topkek!

PR #2812, opened 2015-02-13
Return only a single related ID when required in Content::related()

### LINKED ISSUE #2469: Filter related content by id not working
Just updated a site from v1.6 to v2 and had an issue with fetching related content.

It seems the method Content::related has a parameter $filterid which is not used anymore.
This worked in v1.6.

GawainLynch:
Fixes #2469


bobdenotter:
:+1:

PR #2813, opened 2015-02-13
Base2014 theme cleanup

cdowdy:
I've updated Magnific Popup to match what bolt distributes in the backed ( to 1.0.0).  

I also moved the Magnific Popup css ```link``` tag from the footer to the head of the document. A link tag for css should be in the ```<head>``` so that it wont block painting/rendering of the page  (* more info in link below) . The page isn't very large for the base 2014 theme with ipsum content but if a person were just use the base20{14/15} theme and build a site from there in my personal opinion the base2014 theme should at least be a good web citizen towards web performance. This change doesn't break anything that I've tested in a few loads in FF/Chrome/IE and doesn't change much in load times but mileage may vary.   

* css locations/async loading of css  https://gist.github.com/scottjehl/87176715419617ae6994


bobdenotter:
Sweet. Thank you! :-)

PR #2815, opened 2015-02-14
Twig thumbnail, remove double encoding paths

GDmac:
url_generator already rawurlencodes paths, no need for lib::safe name as well


bobdenotter:
Thanks! :100:

PR #2825, opened 2015-02-15
Make X-Frame-Options configurable.

### LINKED ISSUE #2814: X-Frame-Options headers
For a reason, I had to remove these headers from a Bolt website and noticed that I can't:
https://github.com/bolt/bolt/blob/master/src/Application.php#L448-L449

I found no method to remove headers here neither: http://api.symfony.com/2.6/index.html

I also wonder if a CMS application is meant to send this kind of headers? Wouldn't be better to leave it to the server responsability?

bobdenotter:
Fixes #2814. Changelog line: 

- Change: The 'X-Frame-Options' header is now only sent for backend pages, and can be disabled in `config.yml` (See #2825)

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -270,7 +270,7 @@ session_use_storage_handler: true
 # '12' is better.
 hash_strength: 10
 
-# Bolt sets the `X-Frame-Options` and `-Frame-Options` to `SAMEORIGIN` by default, to prevent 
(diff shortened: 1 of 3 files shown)


GawainLynch:
For going forward, I am wondering if it is better to have:
```yaml
headers:
    x_frame_options: true
```
… in case there are other headers we may want to manipulate?

bobdenotter:
PR updated.

PR #2826, opened 2015-02-15
Fix up the nut cache command tests

rossriley:


GawainLynch:
Thanks mate :+1: . My bad for not cleaning that up.

PR #2833, opened 2015-02-15
Set timeout for http://loripsum.net/api and gracefully handle interface errors

GawainLynch:


bobdenotter:
Green lights! :traffic_light:

Issue #2627, opened 2015-01-30
file upload on mobile phones does not work


jadwigo:
The upload file, browse on server and select from stack buttons do not work on android of iphone at the moment. They are unclickable.

Some verzion information - bolt version 2.0.5 from the zip on php 5.5

![2015-01-30 09 20 03](https://cloud.githubusercontent.com/assets/1488214/5973513/ab83f9de-a869-11e4-9a6c-17fe8c6f8d3e.png)

dwr:
Hi, see PR, worked for me, but changes layout for mobile a bit this way:

<img src="http://i.imgur.com/lMEe16W.png" />

dwr:
Also there's a bug on mobile devices when image container is less than 200px width.

Suggested fix is adding max-width:200px; to imageholder and removing width and height attribute of image:

```twig
    <div class="col-xs-4">
        <div class="right-on-large">{# to make the preview align right on wide screens #}
            <label>{{ __('Preview') }}</label>
            <div class="content-preview imageholder" id="thumbnail-{{ key }}" style="max-width:200px;">
                {% if context.content.get(key) != '' %}
                    {% set preview_path = image.file|thumbnail(preview_w, preview_h, 'c') %}
                {% else %}
                    {% set preview_path = app.paths.app ~ 'view/img/default_empty_4x3.png' %}
                {% endif %}
                {# <img src="{{ preview_path }}" width="{{ preview_w }}" height="{{ preview_h }}" alt="Preview"> #}
                <img src="{{ preview_path }}" alt="Preview">
            </div>
        </div>
    </div>
```

dwr:
I submitted a PR with an idea on how to fix it without breaking other things, the solution is dirty imo using inline-styles to keep preview_w and preview_h variables working.

GawainLynch:
Fixed in #2846

PR #2846, opened 2015-02-17
Fix for #2627.

dwr:
It seems that preview block got over the buttons so it was impossible to click them (click was triggered on preview instead of buttons).

This will fix the layout without breaking anything.


bobdenotter:
Thank you! ^_^

PR #2848, opened 2015-02-17
Make loripsum mockable

rossriley:


GawainLynch:
:+1:

src/Provider/PrefillServiceProvider.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Silex\Application;
+use Silex\ServiceProviderInterface;
+use Bolt\Storage\Prefill;
+
+class PrefillServiceProvider implements ServiceProviderInterface
+{
+    public function register(Application $app)
+    {
+        $app['prefill'] = $app->share(
```
GawainLynch: I know this to some extent belongs in our discussion about the 2.2 refactor, but while adding a `Bolt\Storage` namespace, should we also namespace the services, i.e. `$app['storage.prefill']`?

CarsonF: I agree

rossriley:
Just hold off on merging this one for a bit.. need to add a test for Prefill.

src/Provider/PrefillServiceProvider.php
GawainLynch: Consensus on IRC was this is a 2.2-dev discussion. :nut_and_bolt:

rossriley:
Ok all good now.

GawainLynch:
:cake:

PR #2849, opened 2015-02-17
Bugfix/bolt Adding support for multiple slugs when creating uri through ajax

StevendeVries:
When using multiple slugs the backend previously depended on the 'slug' field. When using multiple slugs (or even a different named slug) you could create content with duplicate uri's.
With this fix we send the slugfield through AJAX/Async.php to the getUri (in Storage.php) function. Instead of using a hardcoded 'slug' in the query we can use slugfield. 'slug' is still being used as fallback and when an invalid slugfield is provided.

Please use grunt to recreate the minified bolt.min.js


bobdenotter:
Looks good to me! Thanks, @StevendeVries :+1:

Issue #2491, opened 2015-01-18
Self-repair after related-content mistake


evertalbers:
I screwed up a site in development because I indented relations one level too much in contenttypes.yml 

After removing the mistake from contenttypes, the error remained, so had to reinstall Bolt (no big deal, I had to update to 2.0.3 anyway) but I'd expect the corrected contenttypes.yml to be sufficient to get things up and running again.

Edit, some extra info: what seems to be odd with this particular installation is that I have to chmod each new twig-template file from 000 to 644 after uploading.

The error notification was the one below:
![php_fatal_error_bolt_core_-_2015-01-18_12 51 15](https://cloud.githubusercontent.com/assets/3974120/5792376/f906013c-9f15-11e4-97ae-2bb3618cd523.png)

bobdenotter:
This is a weird one.. I'll see if i can reproduce it!

GawainLynch:
Fixed in #2852

bobdenotter:
beat me to it, gawain. ;-)

GawainLynch:
We have each other's back covered. Team work for the win! :+1:

PR #2852, opened 2015-02-17
fix fatal when field type is unknown

jkazimir:
This should take care of #2491.

Cheers!


bobdenotter:
Pinging @xiaohutai ^_^

bobdenotter:
Awesome! Thanks, Jared. :+1:

PR #2854, opened 2015-02-18
Fix PHP builtin webserver serving static files

CarsonF:
That snippet assumed it was in the webroot. /app is not the webroot.


GawainLynch:
:traffic_light:

PR #2858, opened 2015-02-18
prevent search query from growing too fast (fixes #2850)

### LINKED ISSUE #2850: Search is broken for many records
Tested in Bolt 1.6.13 and Bolt 2.0.6, so it's been here quite a while. If you add lots of dummy records (e.g. 100 entries and 100 pages, using SQLite), and use the search function, you'll get an error. The search function is quite an old implementation, so we might want to re-write the search functionality anyway. If anyone with experience has good ideas?

error

jkazimir:
Quick fix to prevent the search query from growing too large when there are lots of records.

This isn't a permanent solution, but should hold things over until the 2.2 refactor/rewrite of this bit of code.

Cheers!

### DIFF
--- a/src/Storage.php
+++ b/src/Storage.php
@@ -669,13 +669,14 @@ class Storage
 
         // Build SQL query
         $select = sprintf(
-            'SELECT %s.id FROM %s LEFT JOIN %s ON %s.id = %s.content_id WHERE %s',
+            'SELECT %s.id FROM %s LEFT JOIN %s ON %s.id = %s.content_id WHERE %s GROUP BY %s.id',
             $table,
             $table,
             $taxonomytable,
             $table,
             $taxonomytable,
-            implode(' AND ', $where)
+            implode(' AND ', $where),
+            $table
         );
 
         // Run Query


bobdenotter:
Shouldn't we _also_ add something like `array_unique()`, like you suggested in the issue?

jkazimir:
That was just a 'quick hack' suggestion. I don't think it's needed anymore, as the `GROUP BY` prevents the duplicates at the source.

bobdenotter:
Pinging @xiaohutai to the thread. :bell:

GDmac:
+1 for the extra array unique, to put less strain on WHERE IN() many possible duplicates as well

xiaohutai:
I've checked the queries with @bobdenotter . The values are already unique by using `GROUP BY`, so there's no need to add another `array_unique`. Still there could be an issue with https://github.com/bolt/bolt/blob/master/src/Storage.php#L689, if you have over a few thousands matching results. In that case, we need to rewrite that part to `WHERE `id` IN (...)`.

PR #2861, opened 2015-02-19
Read options from package manager correctly

mtotheikle:


GawainLynch:
Thank you!

On request for modification though, you're actually better off adding this around line 38:
```php
        $options = $this->app['extend.manager']->getOptions();
```

mtotheikle:
Good call, change applied

GawainLynch:
Awesome, thank you! :+1:

PR #2864, opened 2015-02-19
Fixes #2863, use PHP `mail` transport when not SMTP is available

### LINKED ISSUE #2863: Email sending fails silently with SMTP
Hi there,

There's an issue with those comments:
https://github.com/bolt/bolt/blob/master/src/Application.php#L328
https://github.com/bolt/bolt/blob/master/app/config/config.yml.dist#L232

If you don't fill the SMTP parameters, emails won't be sent.
The Silex `SwiftmailerServiceProvider` uses SMTP by default (see https://github.com/silexphp/Silex-Providers/blob/master/SwiftmailerServiceProvider.php#L51), and does fall back to the PHP `mail` function that is provided by the `Swift_MailTransport` class: 
https://github.com/swiftmailer/swiftmailer/blob/master/lib/classes/Swift/MailTransport.php

I submitted a pull request that fixes this.

Let me know if you want me to change something in this PR.

nlemoine:


src/Application.php
```diff
@@ -329,6 +329,11 @@ public function initProviders()
         $this->register(new Silex\Provider\SwiftmailerServiceProvider());
         if ($this['config']->get('general/mailoptions')) {
             $this['swiftmailer.options'] = $this['config']->get('general/mailoptions');
+        } else {
+            $this['swiftmailer.use_spool'] = false;
```
GawainLynch: This turns of the in-memory spooler. Not sure that is a good "default" approach, or am I missing something? :grin:

GawainLynch:
Using PHP's mail() as a "default" overall is a bad idea.  The implementation differs on Windows and Linux, is problematic and not allowed on all hosts.

GawainLynch:
From the [Swiftmail documentation](http://swiftmailer.org/docs/sending.html#the-mail-transport)

> It's a last resort, and we say that with a passion!

src/Application.php
nlemoine: The transport won't be used if use_spool is set to true:
https://github.com/silexphp/Silex-Providers/blob/master/SwiftmailerServiceProvider.php#L37

nlemoine:
I answered your comment here: #2863

src/Application.php
GawainLynch: Ah well caught, I was looking directly as the Swiftmailer codebase.

bobdenotter:
Ok, there's more to this than meets the eye. 

For now, the first step.. Let's make it an explicit option like @GDmac suggested: 

```
mailoptions:
   mailservice: php 
```

We should also fix up the description, because this is no longer accurate, then: 

```
# by default, mail is sent using PHP's built-in mail function. In general, it's advised to use SMTP for sending mail
# instead. Uncomment the following lines to use an SMTP server with authentication.
# Please check http://silex.sensiolabs.
```

After 2.1 we'll add two new features to make this easier for people: 

- On the dashboard we'll show a warning, telling people to configure the mail-settings, if they didn't do so. 

- We'll make a small tool to send out e-mails, so users can easily test if Bolt is sending out e-mails OK.

GawainLynch:
Superseded by #2907

PR #2867, opened 2015-02-20
Fix login redirect loop

CarsonF:
Don't tell a user to login if they are requesting the login page.

For the record I don't like what Backend::before is turning into. 
We should be applying the the middleware to routes that need it, not adding it all routes and excluding certain parts of the them for certain routes. 
It just needs to be split up better. Partly my fault for not creating Login::before.


GawainLynch:
> Partly my fault for not creating Login::before.

Yeah, I did a "what the?" the other day when I was in Backend::before()… Just that we're in freeze now.

bobdenotter:
Merging in! Refactoring later. ;-)

PR #2872, opened 2015-02-20
Handling login retreats in after handler so that functionality isn't dependent on Library::redirect

CarsonF:
Prepping to deprecate Library::redirect and Library::path.

Note:
My thought for 2.2 is to have a base controller that will have shortcut methods like "redirectToRoute"
See [Symfony's Controller](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/Controller.php)

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\Exception\ConnectionException as DBALConnectionException;
 use RandomLib;
 use SecurityLib;
 use Silex;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\HttpException;
@@ -487,6 +488,10 @@ class Application extends Silex\Application
             }
         }
 
+        if ($response->isRedirect() && $response instanceof RedirectResponse) {
+            $this->handleLoginRetreatRedirects($request, $response);
+        }
+
         // Stop the 'stopwatch' for the profiler.
         $this['stopwatch']->stop('bolt.app.after');
     }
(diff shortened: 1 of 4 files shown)


GawainLynch:
Quoth @CarsonF 

> `<CarsonF>` Ummmmm. My thought is to do that along with a BaseController. Probably in 2.3

zomars:
Excuse my ignorance, but I was wondering if someone could explain to me how is the retreat supposed to work, every time I use a `{{ redirect('bolt/login') }}` from a front-end page I get redirected to the dashboard each time. I'm also planning on doing a PR to add a manual override with a query parameter like `/bolt/login?redirect=/path-I-want-to-be/redirected-to/after-login`.

CarsonF:
@zomars This has been moved to [RedirectListener](https://github.com/bolt/bolt/blob/release/3.2/src/EventListener/RedirectListener.php). It checks for any response that's a redirect to the login route. If it is, then we save the current route in the session. After successful login, the controller looks for this session value to redirect the user back to [here](https://github.com/bolt/bolt/blob/release/3.2/src/Controller/Backend/Authentication.php#L167).

I don't think you should be linking to the login route directly. Instead route to your `/path-I-want-to-be/redirected-to/after-login` path directly. Then in that controller check if logged in and redirect to login route if not. This is better because you need to check if you are logged in on that page anyways and the redirect back to your page is already handled for you.

zomars:
So in theory if in my homepage I have:
```twig
{% if not user %}
  {{ redirect('bolt/login') }}
{% endif %}
```
It should take me back to the homepage right? If that so it seems like it's not working.. Or I'm doing something wrong.. Probably the second one.. Thanks for the quick response BTW.

CarsonF:
The `{{ redirect() }}` twig function is deprecated and it doesn't work with the retreat logic. You need to redirect in PHP, not Twig.

zomars:
I see, so I should make a extension for it?

PR #2877, opened 2015-02-21
Logging prep for beta

GawainLynch:
If it hasn't broken now, we're good.  Shifting calls to PSR-3 to keep @CarsonF from complaining 

**Added bonus:** That damn array dump in the tests is gone!


CarsonF:
:+1:

GawainLynch:
Before:
before

After:
after

bobdenotter:
Whoop!

PR #2880, opened 2015-02-21
Display incoming relations in relationships tab

### LINKED ISSUE #2399: Incoming relationships don't appear in editor
I found a bug where incoming relationships (relationships defined in other records, but not in the contenttype of the record being edited) do not appear in the editing screen if the following two conditions are true:

1, There is a group (tab) set (in my case, set once under a title field)

    carriers:
        name: Carriers
        singular_name: Carrier
        fields:
            title:
                type: text
                class: large
                required: true
                group: content

2, There are no taxonomies set for that contenttype

Under these conditions, the DOM inspector shows that the DIV is created (I can show it by setting display:block) but it is not attached to a tab, and is set `display:none`.

If I add a taxonomy, the relationships DIV is shown under the taxonomy tab. If I remove the group, the relationships DIV shows up toward the end of the tab-less page.

rarila:
Fixes #2399


GawainLynch:
:sake:

Issue #2081, opened 2014-11-04
2.0 beta should relations be listed under the tab "taxonomy"?


ChangePlaces:
Should the tab for relations be called "Relations" / "Relationships" instead of taxonomy?

PR #2881, opened 2015-02-21
Split combined Relationships/Taxonomy tab

rarila:
As a testbed to see how it feels and being merged if it pleases …

- ``[Relationships]`` ``[Taxonomy]`` isn't much wider than ``[Relationships/Taxonomy]``
- It should wrap better on smaller screens
- Code easier to maintain
- Headlines can be removed in tab group mode as tab name functions as headline
- Fields can directly added to Relationship tab using ``relations`` group
- Probably easier to update to a system where you can add relations and taxonomy on whatever position you want because of decoupled logic

Related:  #2081

### Changelog:

Changed: Relationships are now edited/displayed in it's own tab (relations)


bobdenotter:
:-)

PR #2906, opened 2015-02-23
add translation for taxonomy link

joanlouis:


bobdenotter:
I'm not in favor of this. If you want the name of a taxonomy translated, just translate it in `taxonomy.yml`.

joanlouis:
and how for multi-lingual sites ?

bobdenotter:
You can only set one language at a time, so this will not solve that. Instead, you might find the "labels" extension useful: http://extensions.bolt.cm/view/318f0a8b-973e-11e4-866c-43d66b93a270

joanlouis:
ok.
 I use taxonomy.yml for now. but i don't understand why not use translation in templates.

bobdenotter:
Some of us want us to, and we might eventually. For now, i'm just being the evil dictator, preventing ourselves from losing focus on the "simple" part of the CMS. ^_^

joanlouis:
:+1:

PR #2907, opened 2015-02-23
Alternative solution to fix #2863

### LINKED ISSUE #2863: Email sending fails silently with SMTP

bobdenotter:
Do not merge yet!!!

This is a proposed alternative solution to #2863 / #2864. It makes the selection for either 'smtp' or 'mail()' explicit. It needs some more testing (on actual linux machines) to see if it works as intended. 

Comments/critiques welcome.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -240,15 +240,17 @@ wysiwyg:
 
 #mailoptions:
 #  transport: smtp
+#  spool: true
 #  host: localhost
-#  port: 25
 #  username: username
 #  password: password
+#  port: 25
 #  encryption: null
 #  auth_mode: null
 
 #mailoptions:
 #  transport: mail
+#  spool: false
 
 # Bolt allows some modifications to how 'strict' login sessions are. For every option
 # that is set to true, it becomes harder for a bad-willing person to spoof your login
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -229,16 +229,28 @@ wysiwyg:
     disableNativeSpellChecker: true # If set to 'true' it will stop browsers from underlining spelling mistakes
     # viewSourceRoles: [ 'admin', 'developer' ] # the roles which are allowed to use the [Source]-button.
 
-# by default, mail is sent using PHP's built-in mail function. In general, it's advised to use SMTP for sending mail
-# instead. Uncomment the following lines to use an SMTP server with authentication.
-# Please check http://silex.sensiolabs.org/doc/providers/swiftmailer.html for a full range of options
+# Use the 'mailoptions' setting to configure how Bolt sends email. Uncomment one
+# of the two blocks below, to tell Bolt to use 'smtp' or PHP's built-in `mail()`
+# function. Note that the latter might _seem_ easier, but it's been disabled by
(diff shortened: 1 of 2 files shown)


app/config/config.yml.dist
```diff
@@ -240,6 +247,9 @@ wysiwyg:
 #  encryption: null
 #  auth_mode: null
 
+#mailoptions:
+#  transport: mail
+
```
GawainLynch: My immediate feeling is one of unease at having this in two places, people could uncomment *both* and then not understand the reason that `mail()` isn't firing.

bobdenotter: If they uncomment both, the YML breaks, and an error is shown. If you combine them, i'm not sure it would be easier to grok for people. Neither is really very good.. Hmm..

GawainLynch: Hence my use of "*unease*". Either way, just flagging it :flags:

src/Application.php
```diff
@@ -325,10 +325,17 @@ public function initProviders()
         // Make sure we keep our current locale..
         $currentlocale = $this['locale'];
 
-        // Setup Swiftmailer, with optional SMTP settings. If no settings are provided in config.yml, mail() is used.
+        // Setup Swiftmailer, with the selected Mail Transport options: smtp or `mail()`.
         $this->register(new Silex\Provider\SwiftmailerServiceProvider());
-        if ($this['config']->get('general/mailoptions')) {
+        if ($this['config']->get('general/mailoptions/transport') == 'mail') {
+            // Use the 'mail' transport.
+            $this['swiftmailer.transport'] = \Swift_MailTransport::newInstance();
```
nlemoine: Hi @bobdenotter,

In order to make the 'mail' transport to work, you alos have to set `use_spool` to false. 
See: https://github.com/bolt/bolt/pull/2864#discussion_r25054879

bobdenotter: Thanks, Nicolas. I was wondering about that! :-)

GawainLynch: That is in master, not in current Silex.

bobdenotter: Reading this, I *think* it's nog bad to set this: http://silex.sensiolabs.org/doc/providers/swiftmailer.html#usage-in-commands

GawainLynch: Unless you *want* to use the PHP `mail()`er and you're sending multiple emails.

**Note:** Why someone would *chose* to use the default mailer is beyond me, but still :grinning:

bobdenotter:
PR Updated.

PR #2918, opened 2015-02-24
Remove unneeded check for `checkUserRoleHierarchy`. Fixes #2894

### LINKED ISSUE #2894: Root not allowed to edit records by other roots..
See screenshot, when trying to edit a 'page'.

screenshot 2015-02-22 14 50 19

Error is triggered here, in Backend.php, line 1809

```PHP
        } elseif (!$request->attributes->get('checkUserRoleHierarchy') && $user = $app['users']->getUser($id)) {
            if (!$app['permissions']->isAllowedToManipulate($user, $app['users']->getCurrentUser())) {
                $app['session']->getFlashBag()->add('error', Trans::__('3. You do not have the right privileges to view that page.'));

                return Lib::redirect('dashboard');
            }
         }
```

I don't think that particular bit of code should be there at all.. Checking whether a user is allowed to edit another user happens elsewehere. Shouldn't be in the 'after' handler at all.

bobdenotter:
The check is also done in `usersEdit` and `users`, so it’s not needed
here.

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -1805,13 +1805,7 @@ class Backend implements ControllerProviderInterface
             $app['session']->getFlashBag()->add('error', Trans::__('You do not have the right privileges to view that page.'));
 
             return Lib::redirect('dashboard');
-        } elseif (!$request->attributes->get('checkUserRoleHierarchy') && $user = $app['users']->getUser($id)) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
`$request->attributes->get('checkUserRoleHierarchy')` :gun:

PR #2923, opened 2015-02-24
Fixes #2904 (Tab groups with spaces in their name won't open)

### LINKED ISSUE #2904: Tab groups with spaces in their name won't open
tabs

rarila:
Tab id's are now slugified and prepended by tab numbers to prevent collisions when necessary.

### Changelog
* Fixes #2904


bobdenotter:
aw yiss

PR #2925, opened 2015-02-24
Fix permissions check

CarsonF:
Derped. checkPermission wants current user's roles not current user.

This wouldn't happen if we had actual classes and not arrays. 2.2 can't come soon enough :)


GawainLynch:
> This wouldn't happen if we had actual classes and not arrays. 2.2 can't come soon enough :)

But perfection never will. I think your derp makes that point :Well caught :+1:

Issue #2893, opened 2015-02-22
"Root" is not allowed to edit other "Root" users. 


bobdenotter:
See screenshot: 

![screenshot 2015-02-22 14 46 04](https://cloud.githubusercontent.com/assets/1833361/6318710/b1c0cf30-baa1-11e4-89a5-d635fb367e35.png)

in permissions.yml

```YML
    users:roles-hierarchy:admin: [ root, admin ]
    users:roles-hierarchy:developer: [ root, admin, developer ]
    users:roles-hierarchy:chief-editor: [ root, admin, developer, chief-editor ]
    users:roles-hierarchy:editor: [ root, admin, developer, chief-editor ]
    users:roles-hierarchy:guest: [ root, admin, developer, chief-editor ]
```

bobdenotter:
pinging @MrAdamCook

CarsonF:
This should be fixed in [Permissions](https://github.com/bolt/bolt/blob/master/src/Permissions.php#L184)

CarsonF:
If @bobdenotter is having issues with upgrading to 2.1 we are going to be facing a lot of "issues" from others as well. 

We need to figure out a better solution for these "unversioned" changes.

MrAdamCook:
Looks like you might be missing the following from your permissions file.
users:roles-hierarchy:root: [ root ]

I'll investigate further tonight after work.

CarsonF:
I found the problem.

CarsonF:
This problem should be fixed now. @bobdenotter can you confirm?

bobdenotter:
Fixed by #2931

Issue #2892, opened 2015-02-22
Getting a lot of "Configuration error: users:roles-hierarchy:everyone is not granted to any roles."


bobdenotter:
When editing a page.  See screenshot. 

It points to `src/Permissions.php` line 254. 

![screenshot 2015-02-22 14 25 28](https://cloud.githubusercontent.com/assets/1833361/6318666/acb5091e-ba9e-11e4-9663-cfc35475b837.png)

Pinging @MrAdamCook: Any idea what causes this?

CarsonF:
Did you update your permissions file?

GawainLynch:
What @CarsonF said. Since #2868 we don't send these to La-La Land any longer. Previously they would only be showing in a Codeception run, but I thought for now that annoying us with them would prompt discussions like… um… this one :ballot_box_with_check:

GawainLynch:
Tracking this one though #2893 as it is not  "problem" as such, just a redirection of logging, where #2893 is a problem that needs addressing.

CarsonF:
I'm still getting this error. I think it's still a bug

GawainLynch:
It's actually a "feature" that has been there forever, it is only that we redirected the logging from error_log(). :palm_tree:

CarsonF:
Well who said to move it from error_log, shame on them.

PR #2931, opened 2015-02-25
Fix role permissions

### LINKED ISSUE #2927: Editing a user's roles doesn't take into account roles the current user doesn't have permission to change
The edit user form takes the roles as an absolute list, instead of accounting for the other roles the user could have that the current user doesn't have permission to change.

Ex: With default setup, UserA has developer role and UserB has admin and developer roles. UserA can remove UserB's developer role, but in doing so it will remove UserB's admin role as well.

CarsonF:
Ok this should fix #2927 #2893 #2892.
Additionally:
- It removes the role "everyone" from user's roles list in Backend.
- All roles are shown in useredit form, but unmanipulatable ones are disabled. (this is accounts for #2927)

I did change the permissions config. I think the spot it is now is better suited for the future. Possibly if we want to break out _manipulate_ into _add_, _edit_, _enable/disable_, _delete_, etc.

### DIFF
--- a/app/config/permissions.yml.dist
+++ b/app/config/permissions.yml.dist
@@ -33,6 +33,21 @@ roles:
         description: This user can log on, but is not allowed to edit any content.
         label: Guest Editor
 
+# Role hierarchy settings
+roles-hierarchy:
+    # This will restrict users from being able to add, edit, enable, disable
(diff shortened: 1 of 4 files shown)


bobdenotter:
Looking good! :+1:

Issue #2455, opened 2015-01-12
[UX] Upload errors aren't shown to user


CarsonF:
I just had an error uploading a file. A RuntimeException was thrown since files folder wasn't writable. However, this exception left the UI looking like the file was about to upload, instead of showing an error.
![screen shot 2015-01-12 at 2 50 58 pm](https://cloud.githubusercontent.com/assets/932566/5710785/992ab4ec-9a6a-11e4-9ba3-0d37ce051bb3.png)

bobdenotter:
Yes, we should fix this. :-)

PR #2933, opened 2015-02-25
Upload UX improvements

rossriley:
Fix for #2455 

Checks for private visibility and show error message if path is inaccessible.
Adds a new check for `readonly` and shows normal listing but with upload form disabled.


bobdenotter:
:+1:

PR #2941, opened 2015-02-25
Solves another pagenr parameter decoding...

rixbeck:
... in case of textquery is something like '(table1,table2,table3)'. Parameter 'page_table1' may refer another paging content currently and makes collision.


bobdenotter:
Thank you, Rix!

PR #2950, opened 2015-02-26
Fixed warning and incorrect weighting when searching

tobias2k:
In the content search functionality, the result weighting code would
produce a warning message and incorrect behavior when non-scalar fields
(e.g. images, files, addresses...) were included in the search.

This fix first checks for arrays and implodes them as needed.


bobdenotter:
Whoop whoop!

PR #2952, opened 2015-02-26
Remove debug alert

rarila:
Thanks @GDmac


app/view/lib/bolt/obj-stack.js
```diff
@@ -17,7 +17,6 @@ var Stack = Backbone.Model.extend({
         // In the modal dialog, to navigate folders..
         $('#selectImageModal-stack').on('click', '.folder', function (e) {
             e.preventDefault();
-            alert('hoi');
```
bobdenotter: Lol! :wave:

rarila: And nobody noticed that for a long time. Is this as with TODOs. If that nobody notices for half a year the function can be removed? ;-)

GawainLynch: So much #1555 here it's not funny

GDmac:
What wonders me slightly more is, the alert never shows up, does it?
i couldn't find ``$('#selectImageModal-stack .modal-content')`` content in views anywhere? 
we only seem to have ``selectModal``.

PR #2953, opened 2015-02-26
Fix for #2947 (Adding image to wysiwyg html editor from stack fails.)

### LINKED ISSUE #2947: Adding image to wysiwyg html editor from stack fails.
If you enable images in your wysiwyg config and then attempt to insert images from the stack using that functionality it fails. What happens is you click on the image in the popup that appears and instead of that being put into the editor you end up looking at the full size version of the image in the tiny popup window. 

If you choose to navigate to the image via the filesystem view beneath the stack and click on one of the images presented there it works fine. i.e. the image is inserted into the image dialog correctly and appears in the preview. 

This is using a new bolt 2.0.6 install. I have verified the issue exists with a friend also using bolt on a different stack.

rarila:
Fixes #2947 as reported by @cooperaj 

### Changelog
+ Fixes #2947 (Adding image to wysiwyg html editor from stack fails.)

### DIFF
--- a/app/view/twig/files_ck/_files.twig
+++ b/app/view/twig/files_ck/_files.twig
@@ -1,6 +1,4 @@
-{% import '_macro/_macro.twig' as macro %}
-
-<table class="dashboardlisting"{{ macro.attr({_bind: ['ckfileselect']}) }}>
+<table class="dashboardlisting">
     <thead>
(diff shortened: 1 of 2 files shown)


bobdenotter:
Sweet! Merging in!

PR #2955, opened 2015-02-26
Fix redirect loop.

### LINKED ISSUE #2653: Removing CMS permissions from user causes infinite loop
I have 2 user-roles (`client`, `registered`) who I want to block from everything in the CMS (`/bolt/`) and I have 1 `admin` role that can edit all content.

```
global:
  about: [ everyone ] # view the 'About Bolt' page
  activitylog: [ admin ]
  clearcache: [ admin ]
  contentaction: [ admin ]
  dashboard: [ admin ]
  dbcheck: [ admin ]
  dbupdate: [ admin ]
  dbupdate_result: [ admin ]
  extensions: [  ]
  extensions:config: [  ]
  fileedit: [ admin ]
  files:config: [  ]
  files:theme: [  ]
  files:uploads: [ admin ]
  files: [ admin ]
  prefill: [  ]
  profile: [ everyone ] # edit own profile
  settings: [ admin ]
  translation: [  ]
  useraction: [ admin ] # enable/disable/delete
  useredit: [ admin ] # user settings
  users: [ admin ] # view user overview
  roles: [ admin ] # view the roles overview  
  maintenance-mode: [ everyone ] # view the frontend when in maintenance mode
  omnisearch: [ admin ]
  # The following permissions are particularly important: login and postLogin
  # determine who may see and use the login form. If you set them to anything
  # but 'anonymous', only users will be able to log in that are logged in
  # already, which is probably never what you want.
  login: [ anonymous ]
  postLogin: [ anonymous ]
  # Likewise, 'logout' needs to be granted to 'everyone', otherwise people
  # cannot log out anymore.
  logout: [ everyone ]

contenttype-all:
  edit: [ admin ]
  create: [ admin ]
  publish: [ admin ]
  depublish: [ admin ]
  delete: [ admin ]
  change-ownership: [ admin ]

contenttype-default:
  edit: [ admin ]
  create: [ admin ]
  change-ownership: [ admin ]
  view: [ admin ]
  frontend: [ admin ]
```

This seems to cause an infinite loop in the cms part after login with either the `client` or the `registered` user
How can I make the CMS part inaccessable for the 2 users but still let them redirect properly after login (to the last page they where on for example)?

CarsonF:
Fixes #2653

If the user doesn't have access to the backend, redirect them to the frontend.

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -138,6 +138,12 @@ class Library
     {
         $app = ResourceManager::getApp();
 
+        // If the user doesn't have access to the backend, redirect them to the frontend
+        if ($path === 'dashboard' && !$app['users']->isAllowed('dashboard')) {
+            $app['session']->getFlashBag()->clear();
+            $path = 'homepage';
+        }
+
         // Only set the 'retreat' when redirecting to 'login' but not FROM logout.
         if (($path == 'login') && ($app['request']->get('_route') !== 'logout')) {
 
--- a/tests/Controller/LoginTest.php
+++ b/tests/Controller/LoginTest.php
@@ -131,4 +131,26 @@ class LoginTest extends BoltUnitTest
 
         $app->run($request);
     }
+
+    public function testDashboardWithoutPermissionRedirectsToHomepage()
+    {
+        $app = $this->getApp();
+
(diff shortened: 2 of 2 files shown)


bobdenotter:
Sweet!

PR #2957, opened 2015-02-27
Fix for #2916 (Required validation not working 'html' field type)

### LINKED ISSUE #2916: Required validation not working 'html' field type
Using 2.0.6, none of my 'html' field types are being validated as 'required'. Neither 'introduction' or 'body' fields in config are being flagged as 'required'.

Docs state 'html' type can have 'required' attribute...https://docs.bolt.cm/contenttypes-and-records

Any help appreciated.

Here is content type config:

```yaml
recipes:
    name: Recipes
    singular_name: Recipe
    fields:
        title:
            type: text
            class: large
            group: content
            required: true
            pattern: ".{2,255}"
        slug:
            type: slug
            uses: title
        introduction:
            type: html
            height: 100px
            required: true
        body:
            type: html
            height: 300px
            required: true
        in_carousel:
            type: checkbox
            label: "Add to carousel"
     relations:
        categories:
          multiple: false
          order: title
          label: Select category to add
    record_template: recipe.twig
    listing_template: recipes.twig
    listing_records: 3
    default_status: draft
    sort: -datepublish
    searchable: true
    recordsperpage: 10  
```

rarila:
- Fixes #2916
- Translation moved into template (now translatable)
- Alertbox template moved into template
- Cleanup

Don't know how to style so that it looks like say an text field with error. At least we have a red border.


bobdenotter:
Whoop whoop!

PR #2962, opened 2015-02-27
Use $app->redirect() on valid first user creation to skip checks

### LINKED ISSUE #2960: Setting up "first user" has wrong redirect.
Steps to reproduce: 

1) Set up a new bolt somewhere: 

```
curl -O http://bolt.cm/distribution/bolt-git-2015-02-27-1792cd9.tar.gz
tar -xzf bolt-git-2015-02-27-1792cd9.tar.gz --strip-components=1
chmod -R 777 files/ app/database/ app/cache/ app/config/ theme/ extensions/
```

2) Open page in browser:

screen shot 2015-02-27 at 09 38 27

3) Fill in form and click "Create the first user" 

screen shot 2015-02-27 at 09 38 52

4) See empty front page of website, instead of login for the backend.

screen shot 2015-02-27 at 09 39 18

GawainLynch:
Fixes #2960

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -889,8 +889,8 @@ class Backend implements ControllerProviderInterface
         } else {
             // For existing items, we'll just keep the current owner.
             $contentowner = $app['users']->getUser($content['ownerid']);
-        } 
-        
+        }
+
         // Test write access for uploadable fields
         foreach ($contenttype['fields'] as $key=>&$values) {
             if (isset($values['upload'])) {
@@ -904,7 +904,7 @@ class Backend implements ControllerProviderInterface
                 $values['canUpload'] = true;
             }
         }
-        
+
 
         // Info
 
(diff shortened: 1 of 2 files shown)


GawainLynch:
First user before this patch:
Grumpy Cat

First user after this patch:
Quokka

jirimedved:
LOL ...

rarila:
Never gonna create a new first user in the future

CarsonF:
I don't like this. You're just using another method to avoid the logic. This method is changing in 2.2 so that you cant skip around it like this. We should be checking `hasUsers` instead.

GawainLynch:
Using another method? This is just using the same method further down the chain. But meh.

GawainLynch:
Oh, and who broke it in the first place………… :grinning:

GawainLynch:
> We should be checking hasUsers instead.

That won't work as the database table has been populated at that point. You've added that check to `redirect()` to stop a loop, we can direct  to dashboard or login at this point, if that would make you happier?

P.S. No comments before coffee, remember.

CarsonF:
Further down the chain. But we are kinda expecting the entire codebase to use `Lib::redirect` at this point. 

Yeah I was going to suggest just logging them in at that point too. It seems like a better UX anyways.

GawainLynch:
> But we are kinda expecting the entire codebase to use Lib::redirect at this point.

Sure, but you and I both have designs on reworking this approach in 2.2. 

Right now, your last set of changes causes unexpected and unwanted behaviour and we want to get a beta — which this was one of two reasons it got delayed today — and stable release out soon, rather than invest more wasted time in refactors that can't, by definition, go in after feature freeze.

bobdenotter:
Let's put this on the list of things we want to refactor for 2.2, 2.3, or whatever..

For now: :shipit:

Issue #2961, opened 2015-02-27
No fields are shown in editor..


bobdenotter:
See screenshot. No fields are shown: 

![screen shot 2015-02-27 at 10 06 39](https://cloud.githubusercontent.com/assets/1833361/6409797/5fafb5c4-be68-11e4-9aee-57b56c8380c7.png)

The database seems OK, and after "loripsumming" it, they show fine in the _frontend_.

PR #2963, opened 2015-02-27
Have ./run-functional-tests exit with the return code from the Codeception run

GawainLynch:
There was a problem introduced with #2957 that was shown in the Codeception results, but the script we run exits with it's own value.  This fix captures and exits the script with that return value so that Travis will fail.

This is — obviously — going to trigger a Travis failure itself until #2961 is fixed, but can we please merge it once we know that the PHPUnit tests pass to prevent anything further from sneaking in.


bobdenotter:
Whoop! here goes!

GawainLynch:
Never been so happy to have a Travis failure on my PRs :-)

PR #2980, opened 2015-02-28
Log DBAL failures for timed (de)publication of contenttype records

GawainLynch:


bobdenotter:
:100:

GawainLynch:
There is a slim chance this might show us something that is not quite right, at the moment. But my hope is that if this is the case, we might see something during beta.

bobdenotter:
I was thinking this seems pretty "edge casey" to me, but I figured you had a reason for doing this. ;-)

GawainLynch:
Yeah, that was pretty much what is in my head… But I try to stick with the Samuel L. Jackson rule of assumptions… ergo, don't swallow the error.

PR #2997, opened 2015-03-02
Client side validation including #2956 fix

### LINKED ISSUE #2956: float type buggy for clients with locale fr
the fr locale use coma for decimal separation using input type number cause bug because the value have coma and the browser want a dot.
maybe it is safe to not use locale for the form values in the backend?
or not use the buggy input type number for floats (all browsers are buggy)

rarila:
- Add client side validation (first for floats only)
- Distinct error messages
- Translatable error messages
- Fixes #2956
- Float field now not html5 number field anymore
- ``,`` and ``.`` allowed as decimal point for floats
- You now can click on the field part of the alert to jump to the field with the error
- Correct label names, not internal field names. (Only floats sofar)

### Changelog

+ Fixes #2956


bobdenotter:
emmastone-thumbsup1

rarila:
:+1::+1:

PR #2998, opened 2015-03-02
Update CKEditor field objects if they exist on AJAX content saves

### LINKED ISSUE #2994: When a content is updated on pre_save, UI is not refreshed wth new values
(tested on bolt 2.1 beta2)

Using defaut "entry" content-type and having a listener which sets entry's teaser to a custom value. 

When clicking "save entry", ajax payload indicates the newest value but form is not updated. A single refresh of page doesn't display the new value, I need to click on entry in the listing to see the new value.

(tested on FF 38 and Chrome 40, release channels on both)

GawainLynch:
Fixes #2994

Will need a JS rebuild, but @rarila has a PR with modified JS, so merge that and one of us can rebuild.

### DIFF
--- a/app/view/lib/bolt/init.js
+++ b/app/view/lib/bolt/init.js
@@ -158,7 +158,13 @@ var init = {
                                         $(":input[name='" + index + "[" + subindex + "]']").val(subitem);
                                     });
                                 } else {
-                                    $(":input[name='" + index + "']").val(item);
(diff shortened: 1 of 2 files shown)


rarila:
Looks good.
Waiting for #2997 then …:zzz:

bobdenotter:
:100:

PR #3004, opened 2015-03-03
Use QueryBuilder for Async::tags() and Async::populartags()

### LINKED ISSUE #2996: HTTP 500 error retrieving tags
Loading a page in edit mode with tags enabled (standard `entries` in this case) give the following error:
```
GET http://example.com/async/populartags/tags?limit=40 500 (Internal Server Error)
```

GawainLynch:
Fixes #2996

Ping @CarsonF — can you check with the slaves configuration please.

### DIFF
--- a/src/Controllers/Async.php
+++ b/src/Controllers/Async.php
@@ -277,37 +277,40 @@ class Async implements ControllerProviderInterface
 
     public function tags(Silex\Application $app, $taxonomytype)
     {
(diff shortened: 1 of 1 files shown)


CarsonF:
Ok just checked, Query Builder uses executeQuery() which supports slave connections. 

Unfortunately it doesn't support caching. Something we should think about for 2.2

bobdenotter:
cheering_minions

PR #3005, opened 2015-03-03
added config var replacement ability

digitalkaoz:
this PR adds the ability to use service container parameters in your configuration files.

imagine you have this:

```php
$app['debug'] = true
$app['mysql-host'] = "lolcathost"
```

you can then use this vars in your yaml files like this

```yaml
debug: %debug%
database:
    host: %mysql-host%
```

this is useful if you configure your application through environment variables (e.g. with this tool: https://github.com/ivoba/dotenv-service-provider)

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -182,7 +182,7 @@ class Config
      * @param mixed $value
      * @return mixed
      */
-    private function doReplacements($value)
+    protected function doReplacements($value)
     {
         if (!is_array($value) && ('%' !== substr($value,0,1) && '%' !== substr($value, -1, 1))) {
             return $value;
--- a/src/Config.php
+++ b/src/Config.php
@@ -170,10 +170,36 @@ class Config
         }
 
         if ($value !== null) {
-            return $value;
+            return $this->doReplacements($value);
         }
 
         return $default;
+    }
+
+    /**
+     * replaces placeholders in config values %foo% will be resolved to $app['foo'] from the container
+     *
+     * @param mixed $value
+     * @return mixed
+     */
+    private function doReplacements($value)
+    {
+        if (!is_array($value) && ('%' !== substr($value,0,1) && '%' !== substr($value, -1, 1))) {
+            return $value;
+        }
+
+        if (is_array($value)) {
+            foreach ($value as $k => $v) {
+                $value[$k] = $this->doReplacements($v);
+            }
+            return $value;
+        }
+
+        if (is_string($value)) {
+            return $this->app[substr($value, 1,strlen($value)-2)];
+        }
+
+        return $value;
     }
 
     /**


digitalkaoz:
in the long run i can imagine to use https://github.com/igorw/ConfigServiceProvider in the backend of the Bolt Config Class

CarsonF:
Looks cool. I would probably make that method protected. 

I definitely see the merit in allowing configuration values from other places in code without having to extend the Config class or overriding all values on initialization.

digitalkaoz:
ok its protected now

CarsonF:
Also, just so you know, we are in feature freeze right now. This won't be merged until we release 2.1 and branch it. 2.1 is in beta right now, so soon :)

CarsonF:
I'm looking at ConfigServiceProvider and I like the way it looks too. Currently though we are using Application as a service locator instead of dependency injection. If/when we do change more to DI, then ConfigServiceProvider would be a good tool to use. It would prevent us from having to inject Config everywhere.

digitalkaoz:
Damn would love to see it in 2.1 :/

Carson Full <notifications@github.com> schrieb am Di., 3. März 2015 21:49:

bobdenotter:
Looking at this, i think the impact of this is pretty minimal. If we're sure this won't cause any side-effects, we might want to sneak it in for 2.0. I think,

CarsonF:
I'm worried about accessing the $data var inside the class without the getter.

bobdenotter:
Right. Ok, we should wait until after the 2.1 release. 

The 2.1 beta is going pretty well, so we're expecting to release 2.1 final soon. Shouldn't be too long.

bobdenotter:
This should absolutely be documented somewhere on docs.bolt.cm. 

@digitalkaoz Could you write a small piece of documentation explaining how this works for users? Perhaps it could be a new page in the "HOWTO" section?

bobdenotter:
Pinging @digitalkaoz Do you think you could do a short write-up on how this works, for the docs? :-)

digitalkaoz:
Hey,

exactly as i wrote ;) 

```php
$app = new Application();

$app['debug'] = true;
$app['mysq-password'] = 's3cr3t';
```

i use https://github.com/ivoba/dotenv-service-provider for injecting env vars into the pimple container.

then you can reference these variables in you config files:

```yaml
debug: %debug%
mysql:
    password: %mysql-password%
```

bobdenotter:
Well, let's just assume I'm dumb, and don't know what dotenv is.. ;-)

What would be the benefit of putting those vars in your `app.php`, instead of `config.yml`? You'd just be putting it in a different place, that would still go into your VCS. 
In fact, putting it in `app.php`, would make it easier for it to get overwritten during an update of Bolt.  

If we wanted to make this easier for dumb end-users like me, wouldn't it be better if we incorporated your Silex Provider, and make this a core feature?

digitalkaoz:
well basically it enables a multistage environment. imagine this:

```php
//staging.credentials.php
$app['mysql-password'] = 's3cr3t'
```

```php
//production.credentials.php
$app['mysql-password'] = 'r3allys3cr3t'
```

in `app.php` we load either the one or the other. (maybe copied to `env.php` from a CI Server, and always include `env.php`)

but yeah you know it...the best solution would be a core feature that builds upon those 2 

https://github.com/ivoba/dotenv-service-provider in combination with https://github.com/igorw/ConfigServiceProvider 

(dotenv for multi-environment stuff, and config for transporting those into the config files)

rossriley:
Just a question on this PR, since the replacement happens inside `get()` does that mean that replacements happen for contenttypes too, in which case I think this would be nice for allowing custom Entities to be provided for contenttypes, in the same way that we currently support the `class: ` option for a contenttype.

That way entities can be created using a service rather than needing a hard link to the class (obviously we can still support both)

SahAssar:
Opened a forum thread for this and #4572 here: https://discuss.bolt.cm/d/138-dynamic-variables-in-config-files-continuation-of-4572-and-3005

GDmac:
From the forum thread:

IMO shouldn't those config variables be namespaced?
Does this proposal allow arbitrary $app[varname] access?
$app[] is the core container and i'ld suggest against access/output arbitrary $app[stuff]

bobdenotter:
It's not documented. Your guess is as good as mine. 😉

rossriley:
I would personally prefer that it wasn't namespaced and we just ship it as is.

Unless anyone can think of a security reason, we don't block off access to Twig templates and thus theme developers so I don't see why we need to protect people who can write to the config file too.

This way is more useful since it allows you to traverse any registered services.

I wouldn't mind making one small modification to it, and that is to allow dot notation within replacements.

eg: `%db.default_options.port%` or `%resources.path.theme%` 

This is a little bit more complex though, so we can leave it for a future PR.

GawainLynch:
OK, @rossriley has stepped up to write up docs for this if merged… welcome to Bolt v3 #3005 :grin:

GDmac:
Without having looked, a little question... how "dynamic" is it? 
e.g. Is the config file cached as a COMPILED version of the yml?
(e.g. app[var] updated, but config FILE not, is the config VALUE updated?)

PR #3010, opened 2015-03-04
Added base path to current url

### LINKED ISSUE #3008: paths.currenturl (and probably other paths) doesn't take into account subdirectories/base
Spotted this in the RSS Feed extension which uses `paths.current` and I noticed that the url, which should be `http://localhost/site/contenttype/feed/rss.xml` was just `http://localhost/contenttype/feed/rss.xml` (note the lack of `site` in the second url).

The culprit is the code here in [ResourceManager at 277-285](https://github.com/bolt/bolt/blob/f3f3eae52c9165864844d17c1a20c368044dc760/src/Configuration/ResourceManager.php#L277), not taking into account `$request->getBasePath()` as the paths a few lines above do.

CarsonF:
Fixes #3008

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -277,7 +277,7 @@ class ResourceManager
         $this->setRequest("protocol", $protocol);
         $hostname = $request->server->get('HTTP_HOST');
         $this->setRequest('hostname', $hostname);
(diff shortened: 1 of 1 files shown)


Pinpickle:
:heart:

bobdenotter:
Whoop!

PR #3013, opened 2015-03-04
Don't duplicate 'beta 3' in the profiler nub.

bobdenotter:
See screenshot: 

screen shot 2015-03-04 at 17 40 11


CarsonF:
Hmmm I swore the default parameter was false not true. Derppp

GawainLynch:
:+1:

PR #3014, opened 2015-03-04
adding extension and library requirements

dwolf555:
adding pdo and gd extensions as well as curl library to composer.json to make sure minimum requirements are met


CarsonF:
What in core requires GD and Curl?

bobdenotter:
Thumbs requires GD, and Guzzle requires Curl. Not sure if either of those packages has these dependencies in their own `composer.json`.

CarsonF:
Both do.

bobdenotter:
Wait, did you add that second sentence as a ninja-edit?

CarsonF:
Ahh. I'm going back to IRC!

bobdenotter:
Ok, @dwolf555, so gd and curl are already implicitly required but other dependancies. If you leave in just the `"ext-pdo": "*",`, we'll merge this in. ^_^

bobdenotter:
Whoop! Thanks!

PR #3015, opened 2015-03-04
Update about.twig

bobdenotter:


GawainLynch:
Less that 24 hours between #1555s… is that a bad thing?

Issue #1000, opened 2014-03-13
[RFE] Allow for an 'Editorial Comments' section in contenttype editors


GwendolenLynch:
Numerous sites have review of posts, it would be useful for editors to be able to log comments about the pre-published record.

If you're happy for this to go in @bobdenotter then by all means assign to me.

Issue #555, opened 2013-08-31
Normal editors can edit all  contenttypes


evertalbers:
Maybe this is still on the to-do list, but #283 suggests otherwise, so I thought I'd let you know: 
In an installation of Bolt 1.1.4, I created a user (editor level) that can only edit (3) contenttypes, but when I log in as that user I can create all (5) contenttypes.

PR #3017, opened 2015-03-04
Hey Joe, …

rarila:
… where you goin' with that gun in your hand?
Hey Joe, I said where you goin' with that gun in your hand? 
Alright. I'm goin down to shoot that **debug output command**.


GawainLynch:
Ugh, ok I am going to just mention #1000 as it is #555 less!

rarila:
Shoot her one more time again, baby! :gun:

PR #3023, opened 2015-03-05
Copy lib.css images to img/lib

rarila:
Images adressed from ``lib.css`` are copied to ``view/img/lib/`` and  ``lib.css`` is changed to use these.

(With a bit of luck) the ``view/lib`` now only has a source function and can completly removed from distribution


bobdenotter:
Only one way to find out! :-)

rarila:
If you're gonna die, die with your beers on. :beers:

PR #3031, opened 2015-03-05
Use the X-Forwarded for IP address when an appropriate one exists and th...

cooperaj:
...e trustedProxies config contains a valid IP.


CarsonF:
We use [this](https://gist.github.com/CarsonF/af2cf15344ae68e4c48d) at my work.

cooperaj:
Well thats a lot more complicated :P. As best as I can make out Silex only utilises X-Forwarded-For* in it's request object so this change doesn't stray from that.

GawainLynch:
@CarsonF I can already here one of our colleagues invoking Godwin at the statics… :smile_cat:

CarsonF:
@GawainLynch Ugh I know. I saw that as I was pasting. We really don't use statics much, I swear. These are just util methods that never change.

@cooperaj What does Silex only using X-Forwarded-For have to do with anything? It doesn't tell the client what headers to use. Also, XFF can be a comma separated list.

bobdenotter:
>  I can already here one of our […]

Where?

CarsonF:
@bobdenotter your not making since. Its right they're.

GawainLynch:
Will you two learn to grammar! :see_no_evil:

cooperaj:
Ok. Scratch this for now. Commit incoming.

cooperaj:
Commit updated with sanity now included.

src/Users.php
```diff
@@ -42,7 +42,7 @@ public function __construct(Silex\Application $app)
         $this->authtokentable = $prefix . "authtoken";
         $this->users = array();
         $this->session = $app['session'];
-        $this->remoteIP = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : "";
+        $this->remoteIP = $app['request']->getClientIp();
```
GawainLynch: I've committed this change before too, doesn't it break CLI @bobdenotter>

bobdenotter: The one we had troubles with specifically was with 'cookies' in the request object, I believe. But, In general, we should be really wary to change the few leftover `$_SERVER`'s, because in most cases there's a reason they're still there.

CarsonF: Yeah nut user commands. We at least need to try/catch `app['request']` here and then use *localhost* or the like.

bobdenotter:
Looks like Travis b0rked. :-/

GawainLynch:
Yeah, this is the CLI breakage that I created the last time too.

cooperaj:
I'll work through and try to figure out solutions to the cli issues. I do
think it worth pursuing though as the prior behaviour wasn't great.

cooperaj:
Think this is working now. All unit tests (bar an apparently pre-borken string comparison) pass. Moved the getting of a clientIP to a place that is always in scope for a genuine request.

CarsonF:
There's no sense to store it Application. It's one isset check either way, mind as well do it at the point you need it.

cooperaj:
An isset check on a service causes a getOffset check in Application which invokes the exception that Silex stores in 'request' spot which causes the issue we're seeing in the first place. I.e. stuff being used out of scope. 

The *only* way to get access to a Silex Request is when you're in-scope for it and that only happens in before and after filters and controllers. I beat my head against this for about 3 hours and discovered the Silex/Pimple/Symfony documentation is a nightmarish recursive rabbit hole. :P

CarsonF:
Yeah sorry, I could've given you a heads up about that. 

I don't like putting random variables in Application like a global storage, it makes me feel icky.

I understand not liking to catch/squelch exceptions when there's a lot of different things that could happen, but in this case we know exactly what's happening. 

If the framework didn't intend for this they should've not set it in the first place, then an isset check would work.

bobdenotter:
I've marked this for "version 2.2", seeing how close we are to 2.1 final.

cooperaj:
Ok. I guess you don't want to muddy application space too much. That said, it is what it's there for and there are also a number of "request.*" items already in there so I think it's namespaced sufficiently.

CarsonF:
@cooperaj I think the bigger problem I have is the random logic in Application. I just wish the logic was closer to the spot it is being used. 

What about moving it to `UsersServiceProvider`?

```php
public function boot(Application $app)
{
    $app->before(function ( Request $request, Application $app) {
        $app['request.client_ip'] = $request->getClientIp();
    });
}
```

cooperaj:
That sounds like a sane plan. I'd put it in the main application cause I'm not all that familiar with silex/symfony layout and I knew that location worked at least.

CarsonF:
No worries, thanks for working with me on this!

CarsonF:
:+1:  :100:

bobdenotter:
This issue has a pretty high comments::code ratio. ;-)

I think it's good to go now. Any objections against merging in?

GawainLynch:
:shipit:

bobdenotter:
Merging in! Thanks, @cooperaj

GawainLynch:
Ugh, OK. This is causing immediate log out in the backend.

GawainLynch:
@CarsonF this is actually causing a deeper problem to show, from IRC:
```
<cooperaj> your ip is 127.0.0.1 all the way through until you’re logged in when it turns to 192.168.*
<cooperaj> but then your ip doesn't match your cookie value
<cooperaj> so it logs you out
<cooperaj> without the extension it’s always 192.168.*
```
So if you have an extension that mounts a controller, like `bolt/members` you're going to have a bad day.

cooperaj:
Looks to be a race condition. The $app->before closure in UserServiceProvider is not called until *after* the initial call to $app['user'] in Gawains extension. This could be an issue with Silex not running ->before middleware until it exits back up the stack out of the current request. Sadly that means it's run too late to be of any use.

Issue #3028, opened 2015-03-05
Euhm. what? (`[WARNING 1549] failed to load external entity`)


bobdenotter:
![screen shot 2015-03-05 at 15 47 01](https://cloud.githubusercontent.com/assets/1833361/6506940/ebf54cfc-c34e-11e4-8aa6-bcfb7f4e9e69.png)

bobdenotter:
I've not been able to reproduce this anymore.

bobdenotter:
As pointed out by Ross, seems to be the same issue as #716

bobdenotter:
This has not shown up anymore. Closing.

PR #3032, opened 2015-03-05
Only add yml and xlf files

CarsonF:
Only add yml and xlf files


bobdenotter:
Let's see if this happens to fix #3028 as well. :-)

PR #3038, opened 2015-03-06
Difference between files and images

evertalbers:
Oh, "files" in plural is even better, but I don't know how to change the pull request.


bobdenotter:
:+1:

PR #3042, opened 2015-03-06
Expand the list of acceptable HTTP return code for PackageManager::ping() in line with RFC 2616

GawainLynch:


bobdenotter:
No `418`?

GawainLynch:
Oh… my… how on Earth did I **miss** that!

CarsonF:
Why not `$response->isSuccessful()`?

GawainLynch:
Yeah, good idea. I blame BoltCamp where that was originally written.

GawainLynch:
In case you're also thinking of Guzzle, [this is why](https://github.com/guzzle/guzzle/blob/v3.8.1/src/Guzzle/Http/Message/Response.php).

CarsonF:
> In case you're also thinking of Guzzle, this is why.

You mean `isSuccessful` doesn't allow these - 300, 301, 302, 307, 410?

GawainLynch:
```php
    public function isSuccessful()
    {
        return ($this->statusCode >= 200 && $this->statusCode < 300) || $this->statusCode == 304;
    }
```

CarsonF:
Why is this a problem? 

301 and 302 will never be returned because Guzzle follows redirects. 
When would 300, 307, 410 be returned? And why would that be a positive thing?

GawainLynch:
If you can explain to me why at the time we originally did this we were getting weird response codes I can give you the answer. Past that, what is your point?

GawainLynch:
Oh and yes, we were getting 302 directly back from Guzzle.

CarsonF:
Hmm that's weird you were getting 302, maybe redirects were turned off. My point is let's not reinvented the wheel here. 

How about `!$response->isError`? That would allow 3xx responses.

GawainLynch:
> How about `!$response->isError?` That would allow 3xx responses.

Sure, put it on your 2.2 TODO list.

CarsonF:
:disappointed:

GawainLynch:
https://twitter.com/erikaheidi/status/573161393362481152

CarsonF:
That doesn't apply. It's not a different wheel, it's duplicated code. 

Fine, I'll change it in 2.2.

GawainLynch:
> My point is let's not reinvented the wheel here.

...

> That doesn't apply. It's not a different wheel, it's duplicated code.

That has been there since 2.0-dev in an attempt to fix shit that made no sense as to why people were getting problems and it solved a fucking problem…

CarsonF:
Thanks for finding/fixing the problem. Sorry for being difficult :/

bobdenotter:
:+1:

PR #3044, opened 2015-03-06
Allow for protocol in `canonical` setting in config.yml. fixes #3040.

### LINKED ISSUE #3040: "Canonical" should allow the developer to override the protocol.
For example, the following in your config.yml, should work:

```yml
canonical: 'https://example.org'
```

However, it shows up as: 

```HTML
    <link rel="canonical" href="http://https://example.org/page/refert-tamen-quo-modo">	    
```

bobdenotter:


src/Configuration/ResourceManager.php
```diff
@@ -279,21 +279,34 @@ public function initializeRequest(Request $request = null)
...
+     * makes no sense in this context: Bolt always determine the path for a page / record.
      *
      * @param \Bolt\Application $app
      */
-    public function initializeApp(Application $app)
+    public function initializeApp(Application $app, Request $request = null)
     {
-        $canonical = $app['config']->get('general/canonical', "");
-        $this->setRequest("canonical", $canonical);
+        $canonical = parse_url($app['config']->get('general/canonical', ""));
```
CarsonF: I'm pretty sure parse_url will fail if you don't include `http(s)://`

src/Controllers/Frontend.php
```diff
@@ -187,7 +187,7 @@ public function record(Silex\Application $app, $contenttypeslug, $slug = '')
 
         // Setting the canonical path and the editlink.
         $paths = $app['resources']->getPaths();
-        $app['resources']->setUrl('canonicalurl', sprintf('%s://%s%s', $paths['protocol'], $paths['canonical'], $content->link()));
+        $app['resources']->setUrl('canonicalurl', sprintf('%s%s', $paths['canonical'], $content->link()));
```
CarsonF: wouldn't this be `"example.com/foo" . "/foo"`?

CarsonF:
Is there a point to have `initializeApp` and `initializeRequest` separate with this change? They are both doing the same type of logic.

src/Configuration/ResourceManager.php
bobdenotter: Nope, works just fine. `$canonical['scheme']` will just be unset.

src/Controllers/Frontend.php
bobdenotter: Nope, because the canonical always drops the 'path' part, because we know it'll give this kind of trouble if we keep it around.

bobdenotter:
> Is there a point to have initializeApp and initializeRequest separate with this change?

I debated that internally, but i just wanted to fix the bug, not do refactorings. Small as they might be.

CarsonF:
Fair point. I'll just be over here in my corner twitching

bobdenotter:
Ok, ok, i'll put those two functions together. ;-)

src/Configuration/ResourceManager.php
```diff
@@ -279,21 +279,34 @@ public function initializeRequest(Request $request = null)
...
+            $canonical['scheme'] = 'http';
+        }
+        if (empty($canonical['host'])) {
+            if ($request === null) {
+                $request = Request::createFromGlobals();
+            }
+            $canonical['host'] = $request->server->get('HTTP_HOST');
+        }
+
+        $this->setRequest("canonical", sprintf("%s://%s", $canonical['scheme'], $canonical['host']));
```
CarsonF: What about base paths?

bobdenotter:
@CarsonF I've tested them, and they work fine. Let's say i have this: 

```
canonical: 'https://example.org'
```

Notice the difference in these screenshots below, yet they both give the same (correct) canonical: 

screenshot 2015-03-06 20 36 59

screenshot 2015-03-06 20 37 19

GawainLynch:
:shipit:

PR #3047, opened 2015-03-07
Return the error from Twig rather than swallow it

### LINKED ISSUE #3045: Error in template throws confusing exception.
I just had this issue where I got the exception "No template for 'xxx' defined. Tried to use 'xxx'." when the template was actually present.
Turns out there was an include statement with a typo, so the included template couldn't be found.

Shouldn't it just show the correct error message?

It's because of the try catch block around the twig render function in Controllers/Frontend::render (see 8d59739a23e0d52c80ac057c9401edd22d484e89)

GawainLynch:
Fixes #3045


bobdenotter:
Whoop!

PR #3048, opened 2015-03-07
Use Response constants for HTTP return codes

GawainLynch:
Dibs on early merge when we branch, before everyone starts pulling controllers apart.

### DIFF
--- a/src/Controllers/Async.php
+++ b/src/Controllers/Async.php
@@ -247,10 +247,10 @@ class Async implements ControllerProviderInterface
 
         // don't allow viewing of anything but "readme.md" files.
         if (strtolower(basename($filename)) != 'readme.md') {
-            $app->abort(401, 'Not allowed');
+            $app->abort(Response::HTTP_UNAUTHORIZED, 'Not allowed');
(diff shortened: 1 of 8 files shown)


src/Controllers/Async.php
```diff
@@ -754,7 +754,7 @@ public function before(Request $request, Silex\Application $app)
 
         // If there's no active session, don't do anything.
         if (!$app['users']->isValidSession()) {
-            $app->abort(404, "You must be logged in to use this.");
+            $app->abort(Response::HTTP_NOT_FOUND, 'You must be logged in to use this.');
```
bobdenotter: Now that i look at this.. Is `404` the correct response code for this?

GawainLynch: One of the reasons I've changed these! It makes one look a bit closer :+1: 

But, yeah I'll update this one to HTTP_UNAUTHORIZED and yell if any others don't look right.

CarsonF:
:+1:

bobdenotter:
MERGEFEST!!

PR #3052, opened 2015-03-07
handle proper event types from composer

jkazimir:
This should prevent Insight, Gawain, Ross, Carson, and myself from twitching.

Cheers!

### DIFF
--- a/src/Composer/ExtensionInstaller.php
+++ b/src/Composer/ExtensionInstaller.php
@@ -1,19 +1,24 @@
 <?php
 namespace Bolt\Composer;
 
+use Composer\EventDispatcher\Event;
+use Composer\Installer\PackageEvent;
+use Composer\Script\Event as ScriptEvent;
+
 class ExtensionInstaller
 {
     /**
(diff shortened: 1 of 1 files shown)


GawainLynch:
@rossriley this looks good to me and [makes sense](https://getcomposer.org/doc/articles/scripts.md).

As @jkazimir points out:
```
<slick0> both package install and update are calling the same handler
<slick0> they're different events!
<slick0> one is Composer\Script\Event, which we handle properly
<slick0> the other is Composer\Installer\PackageEvent
<slick0> we need different handlers for both
```

Pushing to 2.2 and could you have a look to make sure this doesn't send you into Godwin mode?

jkazimir:
Just clarifying my IRC comments: Both were actually handled properly because of the try/catch, but at least by the docblock, the PackageEvent was unexpected, which threw off insight as well as a few of us in the past. The PR hints for the base class common to both events we handle (which happens to be any that composer could send to it), and has an early exit for events the handler doesn't care about.

bobdenotter:
MERGEFEST!!

PR #3053, opened 2015-03-07
Missing s/addCritical/critical/ on some logger.system calls

GawainLynch:
Nobody tell @CarsonF that I missed these… I'll never live it down :grinning:


CarsonF:
*twitching*

bobdenotter:
MERGEFEST!!

PR #3075, opened 2015-03-10
Account for base path in menu links.

### LINKED ISSUE #3068: [Regression] Item.link doesn't take into account subdirectory in menus
Eagerly installed 2.1 today and I ran into this pretty big one :(

~~See #3008 as it seems to be the exact same issue (but worse as its for a more commonly used path)~~

The title explains it all. Somewhere around [here](https://github.com/bolt/bolt/blob/master/src/TwigExtension.php#L1250) is the culprit code.

Guess it's back to 2.0.6 for now!

\cc @CarsonF

CarsonF:
Fixes #3068

### DIFF
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -1271,7 +1271,7 @@ class TwigExtension extends \Twig_Extension
                             $item['link'] = $content->link();
                         }
                     } else {
-                        $item['link'] = $this->app['request']->getBasePath() . $path;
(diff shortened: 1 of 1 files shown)


Pinpickle:
This guy is the routing master

CarsonF:
@Pinpickle I'm just committing the stuff you're pointing out :) More like stealing your credit hehe.

GawainLynch:
:shipit:

PR #3083, opened 2015-03-10
[2.1] Fix redirect for backend trailing slash redirect (/bolt -> /bolt/). 

### LINKED ISSUE #3082: Using the /{slug} override for pages breaks Bolt redirects
I don't think this problem was here before, I've certainly never had it. I know that there has been a lot of work done on how routing functions so I think that's where this crops up.

Basically, if you make a `/{slug}` route, this overrides the system URLs that are required for Bolt to function. Going to `/bolt` will show a 404 error because the page isn't found - and it will attempt to do a redirect to /bolt/ (regardless of subdirectory, but that's another issue). That's not terrible, but nowhere near as nice as the automatic redirect I was getting with previous installs.

I know an easy solution is to change it to `/{slug}/`, as that will automate the redirect to to `/bolt/`, but I can't do that for this particular project due to another bug with menus not having `current` set properly if the routes have trailing slashes.

CarsonF:
Should fix #3082.

I would rather figure out a better way to handle this, but this will work for now. 
If we are at /bolt we should just match Backend::dashboard.

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -161,7 +161,7 @@ class Frontend
             // There's one special edge-case we check for: if the request is for the backend, without trailing
             // slash and it is intercepted by custom routing, we forward the client to that location.
             if ($slug == trim($app['config']->get('general/branding/path'), '/')) {
-                Lib::simpleredirect($app['config']->get('general/branding/path') . '/');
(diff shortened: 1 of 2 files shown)


bobdenotter:
:+1:

PR #3085, opened 2015-03-10
Who likes trailing slashes anyways?

CarsonF:
This removes trailing slashes from `/bolt` and `/bolt/extend`

This is the 2.2 fix for #3082.


bobdenotter:
Also :+1:

PR #3086, opened 2015-03-10
Fixes broken extension installer

rossriley:
I'm proposing to revert this back to how it was pre Jan 20th. I've not been able to get any extension scripts to install with either the type-hinting passed and additionally the current package info:

    $event->getComposer()->getPackage();

seems to always return the root package, not the package of the just installed package. Has anyone who changed and tested this got any more details as to how the current code can run correctly?

### DIFF
--- a/src/Composer/ExtensionInstaller.php
+++ b/src/Composer/ExtensionInstaller.php
@@ -20,6 +20,8 @@ class ExtensionInstaller
                 $installedPackage = $operation->getPackage();
             } elseif (method_exists($operation, 'getTargetPackage')) {
                 $installedPackage = $operation->getTargetPackage();
+            } else {
+                return;
             }
         } catch (\Exception $e) {
             return;
--- a/src/Composer/ExtensionInstaller.php
+++ b/src/Composer/ExtensionInstaller.php
@@ -15,7 +15,12 @@ class ExtensionInstaller
     public static function handle($event)
     {
         try {
-            $installedPackage = $installedPackage = $event->getOperation()->getPackage();
+            $operation = $event->getOperation();
+            if (method_exists($operation, 'getPackage')) {
+                $installedPackage = $operation->getPackage();
+            } elseif (method_exists($operation, 'getTargetPackage')) {
+                $installedPackage = $operation->getTargetPackage();
+            }
         } catch (\Exception $e) {
             return;
         }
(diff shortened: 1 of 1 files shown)


jkazimir:
Nice catch on the `$event->getOperation()->getPackage();`! I hadn't noticed that had changed.

I'd love to see this fixed in a once-and-for-all kind of way, though. Notably, the missing type-hint has raised a few eyebrows in the past (mine included), leading to some "fix" attempts, followed by obvious breakage.

I see depending on the event type, getOperation() can return a few different things. It'll get an `InstallOperation`, which has `getPackage()`, but the `UpdateOperation` has `getInitialPackage()` or `getTargetPackage()`.

Have updates ever successfully worked with this in the past? Both post-package-install and post-package-update call to the handler from composer.json, but it doesn't look like update's would have done anything, as the `try` block contained the attempt to call `getPackage`, which couldn't have existed on the operation.

rossriley:
@jkazimir you were completely correct it was the difference in interfaces between the two. This definitely works. now.

src/Composer/ExtensionInstaller.php
```diff
@@ -12,13 +12,19 @@ class ExtensionInstaller
...
+    public static function handle($event)
     {
-        if (!($event instanceof ScriptEvent || $event instanceof PackageEvent)) {
+        try {
+            $operation = $event->getOperation();
+            if (method_exists($operation, 'getPackage')) {
+                $installedPackage = $operation->getPackage();
+            } elseif (method_exists($operation, 'getTargetPackage')) {
+                $installedPackage = $operation->getTargetPackage();
+            }
```
GawainLynch: Shouldn't we also:
```php
} else {
    return;
}
```
… to prevent anything else randomly b0rking, if we've missed something, or Composer changes.

GawainLynch:
Mea culpa on the change in 620353daf5fbf8cbab38ee030b4ba4d318e0c2da, @rossriley… I spent too much time looking at the labyrinth that is Composer and got my brain in a twist.

Just looked though Composer and this seems to be right. :shipit:

rossriley:
@GawainLynch yes you're right on that, the try catch got it before. I'm gonna leave the try catch in just in case it ever gets passed an event without a getOperation() method.

PR #3089, opened 2015-03-11
Add composer branch alias

CarsonF:
Add composer branch alias. This allows composer installers to require `2.1.*@dev` and always have the latest version of this branch


GawainLynch:
Good call mate!

PR #3093, opened 2015-03-11
Revert code to a prior version to make sure we don't break BC with exten...

cooperaj:
...sions in the 2.x cycle

I've left the user provider code in place since it's not hurting and is the only way to access request properties where the availability of a request object is suspect.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,7 +6,7 @@ Not yet released.
  - Added: Stop Finder from recursing common build folders and place a limit on the maximum depth it will recurse otherwise. (Thanks @Cooperaj, see #3069)
  - Fixed: Removing default taxonomylink route leads to exception (See #3070)
  - Fixed: Don't reset urls when adding base path. (See #3074)
- - Added: Use the X-Forwarded for IP address when an appropriate one exists and the trustedProxies config contains a valid IP. (See #3031)
+ - Added: Use the X-Forwarded for IP address when an appropriate one exists and the trustedProxies config contains a valid IP. (Thanks @Cooperaj, see #3031, #3093)
  
 Bolt 2.1.0
 ----------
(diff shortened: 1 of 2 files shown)


GawainLynch:
:shipit:

CarsonF:
`HTTP_X_FORWARDED_FOR` could be a comma separated list. Need to check that as well

cooperaj:
And this is why getClientIp() is about 40 lines long :package: 

I'll get that up later.

CarsonF:
Ok so you don't want to duplicate all that logic, I totally agree. Why not just create the request from globals? We already do that in other places, for better or worse.

GawainLynch:
Woudn't building the request form globals there sort this whole problem?

CarsonF:
Isn't that what I just said @GawainLynch? :P

cooperaj:
Building request? Is that the Request::buildFrom... method?

GawainLynch:
Request::BuildFromGlobals()

CarsonF:
`Request::createFromGlobals`

GawainLynch:
Ugh, type… can't… good…

cooperaj:
I'll give that a crack. It can't hurt.

GawainLynch:
> Isn't that what I just said @GawainLynch? :P

Yes, but you ended with a question mark…

CarsonF:
>Yes, but you ended with a question mark…

So did you :P

GawainLynch:
> So did you :P

Yes, as I meant a question, the underlying of which was, why wasn't this the first thing **you** came up with this morning :-P

CarsonF:
> Yes, as I meant a question, the underlying of which was, why wasn't this the first thing you came up with this morning :-P

I blame the gnomes, they stole my :coffee:

CarsonF:
This isn't ready.

GawainLynch:
Done and added so we can cherry-pick for 2.1

GawainLynch:
@CarsonF see my additional commit

PR #3103, opened 2015-03-11
Refactor menu from twig

cooperaj:
Pulled the menu code out of TwigExtension so that I can use it as part of a breadcrumbs extension.

It now loads as a ServiceProvider giving access to the menu array wherever you might need.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -365,6 +365,7 @@ class Application extends Silex\Application
             ->register(new Provider\TemplateChooserServiceProvider())
             ->register(new Provider\CronServiceProvider())
             ->register(new Provider\FilePermissionsServiceProvider())
+            ->register(new Provider\MenuServiceProvider())
             ->register(new Controllers\Upload())
             ->register(new Controllers\Extend())
             ->register(new Provider\FilesystemProvider())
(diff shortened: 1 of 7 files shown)


GDmac:
small question, this is front-end menu, yes? 
to distinguish it from other possible menus (backend, chinese restaurant), 
maybe have a little less generic name.

GawainLynch:
We had a quick chat about this from a namespace PoV on IRC just now… 

Feeling is roughly that the namespaced classes should probably be along the lines of `Bolt\Helper\Menu` and `Bolt\Helper\MenuBuilder`… We're starting to explode with directories in `src/`

@GDmac specifically what are you thinking mate?

Other than that, LGTM…

bobdenotter:
With Gawain's addition, i'm all :+1:.

rossriley:
I'll have number 81 with prawn crackers

GawainLynch:
prawn_cracker_with_black_pepper

cooperaj:
In order to keep the helper directory under control is it worth keeping
menubuilder under menu?

bobdenotter:
Looks good to me! Thanks, Adam!

PR #3106, opened 2015-03-11
Catch "№" to translate.

gander:


bobdenotter:
As discussed on IRC, I'm going to close this one. the `№` sign is a typographic symbol, and as such it shouldn't be translated. 

http://en.wikipedia.org/wiki/Numero_sign

Thanks, @gander

PR #3110, opened 2015-03-12
Update .gitattributes to use a union merge for the changelog.md file.

rossriley:
As suggested in this article (not the original idea but the update):

https://about.gitlab.com/2015/02/10/gitlab-reduced-merge-conflicts-by-90-percent-with-changelog-placeholders/


GawainLynch:
Adding this to #1555… just 'cause!

rarila:
Does that mean were back to the old modus operandi?

GawainLynch:
It means that either works now.

PR #3112, opened 2015-03-12
Replacing `&nbsp;` with single space, instead of nothing. Fixes #3111.

### LINKED ISSUE #3111: We're still not free of &nbsp;
#2745 fixed most of these problems, relating to #2660 but there's still at least one (regularly occuring) case that's hit me up quite a few times.

Between inline tags and text nodes (and just between inline tags, I think), CKEditor will put a `&nbsp;`. To test, just open it up, type (or copy + paste):

> Testing *testing* testing

And look at the source

    <p>Testing&nbsp;<em>testing</em>&nbsp;testing</p>

:cry:

bobdenotter:


GawainLynch:
Here you go @pinpickle, this should dry those :cry:

PR #3123, opened 2015-03-13
Upload UX Improvements

Pinpickle:
This was initially going to be one change but it turned into a quite a few, just making the whole uploading thing make more sense.

You are now warned when a file is too large to send to the server, and the files are then abandoned. In a file list, files that aren't too big are still uploading. This change hasn't propagated to the file editor, unfortunately.

In the backend file manager that isn't in a record, you can now upload multiple files.

When uploading to an imagelist or filelist, the currently uploading files are displayed underneath to notify you of what's still uploading.


src/Library.php
```diff
@@ -33,6 +33,25 @@ public static function formatFilesize($size)
...
+     * Convert a size string, such as 5M to bytes.
+     *
+     * @param string $size
+     *
+     * @return integer
+     */
+    public static function filesizeToBytes($size)
+    {
+        $unit = preg_replace('/[^bkmgtpezy]/i', '', $size);
+        $size = preg_replace('/[^0-9\.]/', '', $size);
```
GawainLynch: Yottabyte is covered… that's forward planning!

src/Filesystem/FilePermissions.php
```diff
@@ -118,4 +125,25 @@ public function getAllowedUploadExtensions()
...
+            $this->maxUploadSize = $size;
+        }
+
+        return $this->maxUploadSize;
+    }
+
+    public function getMaxUploadSizeNice()
+    {
+        return Lib::formatFilesize($this->getMaxUploadSize());
+    }
```
GawainLynch: PHPDoc blocks on these would be awesome!

GawainLynch:
Overall LGTM

rossriley:
Tested here and all looks good. 

Caveat being that in the future we will need to change the maxFileSize to take into account size rule settings on the upload handler, but lets merge this as is and get to that when we can.

GawainLynch:
Sorted the docblocks at merge time. Thanks @Pinpickle

PR #3127, opened 2015-03-13
New config setting additional_locales

sintemaa:
I found that datelocal was not actually using the locale that I had
configured. This seems to be caused by environment specific issues. A
way to solve it is by specifying additional locales. This change adds an
optional config parameter where additional locales can be configured. If
configured they are appended to the local array. Using this, adding
"Dutch_Netherlands" I could get Dutch dates (which did not work by
simply setting nl_NL as the locale)

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -17,14 +17,11 @@ theme: base-2014
 
 # The locale that'll be used by the application. If no locale is set the
 # fallback locale is 'en_GB'. For available options, see: http://docs.bolt.cm/locales
+# In some cases it may be needed to specify (non-standard) variations of the locale
+# to get everything to work as desired. This can be done as [nl_NL, Dutch_Netherlands]
+# When specifying multiple locale, make sure the first is a standard locale
 locale: en_GB
 #timezone: UTC
-
-# In some cases it may be needed to add additional locales to force correct handling
-# in all cases. Additional locales defined here will be appended to the locale array
-# that is used in the SetLocale call
-# additional_locales:
-#   - Dutch_Netherlands
 
 # Set maintenance mode true or false
 # While in maintenance mode, only users of level editor or higher can access the site.
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -20,6 +20,12 @@ theme: base-2014
 locale: en_GB
 #timezone: UTC
 
+# In some cases it may be needed to add additional locales to force correct handling
+# in all cases. Additional locales defined here will be appended to the locale array
+# that is used in the SetLocale call
+# additional_locales:
+#   - Dutch_Netherlands
+
 # Set maintenance mode true or false
 # While in maintenance mode, only users of level editor or higher can access the site.
 # All other visitors are presented with a notice that the site is currently offline.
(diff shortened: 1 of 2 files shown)


rarila:
That's one of the reasons ``set_locale()`` will be removed in the mid-term. It's one high on my TODO list.

Concerning this patch I wonder if it would have been better to simply allow the ``locale`` setting being an array ``['nl_NL', 'Dutch_Netherlands']``.

But on the other hand this patch/setting could be easily removed without side effects while an array handling code would have to stay until using an array is deprecated.

bobdenotter:
My vote would go to allowing this: 

    locale: [nl_NL, Dutch_Netherlands]

@sintemaa Could you change the PR to support that?

rarila:
@bobdenotter Although that adds code that can't be removed without breaking. config.yml?

sintemaa:
@bobdenotter I had a look at what you suggested. Implementing this in src/Application.php in initLocale is not a problem. However after I had done this I noticed that $app['locale'] is used on a few other places. I found a reference in 5 other files including TranslationServiceProvider.php, WhoopsServiceProvider.php, some of which aren't part of Bolt but vendor components like Silex. 

I expect these all expect a string (I was getting an error from TranslationServiceProvider.php on my first test). Changing all these for this corner case problem I am having seems excessive?

bobdenotter:
Ah, I see. Good catch, @sintemaa. 

So, what we _actually_ want is for $app['locale'] to stay the same, but still allow an array in config.. How about this? 

```php

    public function initLocale()
    {

        $configLocale = $this['config']->get('general/locale', Application::DEFAULT_LOCALE);
        if (!is_array($configLocale)) {
            $configLocale = array($configLocale);
        }

        // $app['locale'] should only be a single value.
        $this['locale'] = reset($configLocale);

        // Set The Timezone Based on the Config, fallback to UTC
        date_default_timezone_set(
            $this['config']->get('general/timezone') ?: 'UTC'
        );

        // for javascript datetime calculations, timezone offset. e.g. "+02:00"
        $this['timezone_offset'] = date('P');

        // Set default locale, for Bolt

        $locale = array();
        foreach ($configLocale as $key => $value) {
            $locale = array_merge($locale, array(
                $value . '.UTF-8',
                $value . '.utf8',
                $value,
                Application::DEFAULT_LOCALE . '.UTF-8',
                Application::DEFAULT_LOCALE . '.utf8',
                Application::DEFAULT_LOCALE,
                substr(Application::DEFAULT_LOCALE, 0, 2)
            ));
        }

        // dump(array_unique($locale));

        setlocale(LC_ALL, array_unique($locale));
```

I did some monkey-testing with `nl_NL`, `[ en_GB, nl_NL, Dutch_Netherlands ]`, and they all seem to work as expected.

sintemaa:
Yeah, something like that could work. Only in the above when you have multiple locales defined now you will add the Application::DEFAULT_LOCALE multiple times. Let me change my work I had already done to match the configLocale of you

bobdenotter:
> Application::DEFAULT_LOCALE multiple times

The `array_unique()` takes care of that. ;-)

sintemaa:
Ah yes, I missed that. I had solved that by taking the Application::DEFAULT_LOCALE out of the loop and assigning it after, but I guess array_unique is better as it also removes duplicates if the Application::DEFAULT_LOCALE is also in the config list.

Do you want me to update the pullrequest?

bobdenotter:
@sintemaa Yes please, and we'll merge it in! In config.yml.dist we can also mention that the locale can be an array, now. :-)

bobdenotter:
Looks good to me! Merging in. Thanks, @sintemaa !

PR #3129, opened 2015-03-14
Show notice on update of Bolt. Fixes #850

### LINKED ISSUE #850: Notify in case of succesfull bolt core update
After updating to 1.4.4 I did not receive some sort of feedback the bolt core was succesfully updated. The update process was easy and straighforward (thank you for that ;)  ) but it was so simple it made me question if it had actually succeeded.

It would be a nice addition if there was feedback on screen after a cms core update (and upgrade).

bobdenotter:


GawainLynch:
:see_no_evil:

PR #3130, opened 2015-03-14
Prevent extra spaces in excerpts. Fixes #1388.

### LINKED ISSUE #1388: 'Excerpt' sometimes introduces unwanted whitespaces
By design, the excerpt function adds a [space] after a closing tag. This is to prevent type from being glued together when strip_tags deleted the tags. However, sometimes this is not preferable. 

For instance: `<a href="bla">linklinklink</a>`. becomes 
`<double space>linklinklink<space>`. 

The double spaces are filtered out by the brower, but the last added space created an unwanted space before the period.

I don't know yet how this should be solved or if it occurs often, and it is not very urgent, but I wanted to file an issue anyway.

bobdenotter:


GawainLynch:
:shipit:

PR #3131, opened 2015-03-14
Migrate to PHPUnit 4.5 & Codeception 2

GawainLynch:
@rossriley we need to look at an alternative way to do BaseExtensionTest::testSetComposerConfiguration()

### Changelog
- Updated: Unit and acceptance tests are now run with PHPUnit 4.5 and Codeception 2.0.11


rossriley:
I can't see the test failure result from the run you mentioned...

What was failing?

GawainLynch:
What is commented out in 5790efe39ac1000a986752c5f349a4b2a21caa49 is the problem, at least here.

Namely that  PHPUnit complains to me that an array is not passed in, instead of throwing an \Exception.

PR #3132, opened 2015-03-14
Big refactoring of Bolts JS - part 1

rarila:
Goals & things already done:

- Refactoring into module pattern
- Remove variables and functions from global space (Having only Bolt and external globals in the end)
- Complete source documentation
- Create documentation with ``grunt jsdoc``. Ends up in ``/apps/src/docs/``
- Config now to be get with ``Bolt.conf('foo.bar')``
- Data now to be get with ``Bolt.data('foo.bar', {'%FOO%': 'bar'})``

Part 2 to be expected soon, stay tuned


bobdenotter:
Merging in! Whoop whoop!

PR #3133, opened 2015-03-14
Remove "frontend permission checks". Fixes #3087

### LINKED ISSUE #3087: [RFC] Let's remove the "frontend" permission.
This one: https://github.com/bolt/bolt/blob/master/app/config/permissions.yml.dist#L149

```
# - frontend: allows front-end access to records (note that this requires the
#             'frontend_permission_checks' global configuration setting, for
#             backwards-compatibility reasons as well as performance)
```

Why? Because:

- It's confusing. See #3079
- It's broken (or at least counter-intuitive).
- Imho it clashes with points 4 and 5 of our manifesto.

bobdenotter:


GawainLynch:
:fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire: :fire:

PR #3134, opened 2015-03-14
Add an option to delete a record, when editing it. Fixes #3115

### LINKED ISSUE #3115: There's no way to *delete* a record while you're editing it. Only from the overview page.
I feel there should be an option to delete a page while on the "edit record" screen.

bobdenotter:


GawainLynch:
:traffic_light:

PR #3138, opened 2015-03-15
Backporting WatchChanges to Bolt 2.1.x

bobdenotter:


bobdenotter:
This almost works, except for the ckedit fields, and I don't know why. 

If i add console.log, like this: 

```javascript
               if ($(this).data('watch') !== val) {
                    changes++;
                    console.log("change before: ", $(this).data('watch'));
                    console.log("change after : ", val);
                }
```

I get output like this, but only _after_ i save the page first, and then try to navigate away. 

screenshot 2015-03-15 11 38 08

@rarila Halp! :-)

bobdenotter:
It's because in 'before' the indentation is 'spaces' and after the indentation is 'tabs'. :-/

bobdenotter:
Should be good to go now!

GawainLynch:
Yay! This is driving me nuts at the moment on 2.1.x!

rarila:
The whitespace replacement should only be done on CKEditor fields, easily detectable with code like this snippet from validation (the ``hasClass()`` part should do):
```js
isCkeditor = field.nodeName === 'TEXTAREA' && $(field).hasClass('ckeditor');
```

bobdenotter:
We only use it for fingerprinting, so it doesn't really matter that it's applied to all fields, right? 

The only thing is now, that if you add a 'space' to a title, the warning doesn't pop up. I can live with that. :-)

rarila:
The OCD is weak with this one!

PR #3142, opened 2015-03-15
Make the sanitisation of markdown fields configurable. Fixes #2992

### LINKED ISSUE #2992: Output raw HTML in Markdown
I am using markdown as my text field, but want the ability to output raw HTML and have it not be stripped. For example, let's say I have a little tutorial going on, and then I want to output some inline SVG and some custom styles in a style tag, it seems that markdown completely ignores the inline svg code, and actually outputs the styles as text on the screen.

Is this something to do with Parsedown itself? Is there any way I can alter this?

In particular, I'm looking at this part of the Parsedown documentation, where they run some tests on "block level html":

http://parsedown.org/tests/block-level_html/

The tests on their site give me the desired results, but the way Bolt outputs it doesn't satisfy those tests.

bobdenotter:


src/Content.php
```diff
@@ -678,12 +678,18 @@ public function getDecodedValue($name)
                     // Parse the field as Markdown, return HTML
                     $value = \ParsedownExtra::instance()->text($value);
 
+                    $config = $this->app['config']->get('general/htmlcleaner');
+                    $allowed_tags = !empty($config['allowed_tags']) ? $config['allowed_tags'] :
```
GawainLynch: Might be safer to set the defaults for those in [`Config::getDefaults()`](https://github.com/bolt/bolt/blob/master/src/Config.php#L7350) in case someone updates one set, but forgets the other

bobdenotter: Nah, won't work. 

The way our stupid getdefaults works with regard to merging, prevents this.. For example, if i set in config:

```
htmlcleaner:
  allowed_tags: [ x, y, z] 
```

And the default is `a, b, c, d, e, f`, you would end up with `x, y, z, d, e, f`. Not what somebody would want. 

(this had me puzzled for a while, because i didn't understand why this was happening ;-) )

PR #3143, opened 2015-03-15
Test improvements

GawainLynch:
Follow up to #3131

**Codeception**
* Clean up class implemented
* Installs a local extension that catches Storage events and tests results
* Cleaned up some and added more tests
* Running against Nginx & FastCGI
* HHVM runs against HVVM FastCGI

**PHPUnit**
* HHVM errors reduced


bobdenotter:
Whooop!

Issue #2711, opened 2015-02-05
[RFC] Template specific fields


Pinpickle:
Sometimes pages are sufficiently different from each other that using the same fields for both of them will make one, or both of the pages, not entirely editable. Say your home page is full of different sections describing the various areas of the website, some pulling information from other records and some statically sitting there - while your about page has has two columns that each have very different content, followed by a large image. Your client wants to edit everything.

I think this was the concern illustrated in #2476 and some solutions were identified but they're either clunky (having 'resource' records that will require a lot of jumping around to edit a few things) or not possible to edit (choosing a template that has all of the information hard coded in). What we can do, which will clutter the sidebar, is create a separate contenttype for every different page type. I've done this where the only substantially different page was the home page and it worked okay.

But as far as real, intuitive solutions, I see two possible ones (and I think both of them should be implemented):

1. List fields (#2261). This will be the killer because it will allow people to add content "blocks" akin to the [matrix field in Craft CMS](https://buildwithcraft.com/features/matrix) (as well as all of the other advantages). Hopefully this is coming in Bolt 2.1 and with some time I'd like to help out with that as well.
2. Template specific fields. This issue is about this method.

Essentially, you outline fields in the theme's `config.yml` the same way you do in `contenttypes.yml`:

```yaml
template_fields:
 - template: homepage.twig
   fields:
      section_1_image:
          type: image
      section_1_text:
          type: html
      map_location:
          type: geolocation
```

When a record uses a particular template (based on the record name, contenttype name or template chooser field), these fields will be displayed on the next refresh of the record. They'd be JSON encoded into a column specifically for template fields on save. When accessing the page, they'd be de-encoded and accessed on the `record` object like anything else:

    {{ record.section_1_text }}

Or if there are any worries about conflicting names, this could easily go under its own namespace

    {{ record.template.section_1_text }}

And that's the idea! I can easily see this being considered as out of scope, but it's been floating around in my head for the past few weeks so I thought I'd at least get it written down.

bobdenotter:
I was thinking about maybe creating a 'blocks' field type, that would just be a wrapper for (Sir Trevor](http://madebymany.github.io/sir-trevor-js/).

Pinpickle:
Where has this been all my life? This is all we need for a list contenttype, though its really minimalistic design will stick out a bit in the interface.

Can we make everything look like this?

However, that's more of my first solution. The per-template contenttypes are for layouts that are more rigid and defined.

[...]

Issue #3033, opened 2015-03-05
Feature: Ability to choose which fields are visible/editable on which templates


dwolf555:
Useful for `pages` content type. Often creating a few pages for a site you'll have fields that are relevant to only certain templates. Some fields may not even display, depending on the template. I'd like to be able to choose which fields are visible based on template. This will prevent confusion to an end user who is editing content.

**Proposed solution in contenttypes.yml:**

```yaml

pages:
    name: Pages
    singular_name: Page
    fields:
        // i think putting template first would make more 
        // sense since it chooses which fields are visible
        template: 
            type: templateselect
            filter: '*.twig'
            group: content
        title:
            type: text
            class: large
            templates: ['home.twig']
        slug:
            type: slug
            uses: title
        image:
            type: image
        teaser:
            type: html
            height: 150px
            templates: ['*-suffix.twig', 'prefix-*.twig']
        body:
            type: html
            height: 300px
    taxonomy: [ chapters ]
    recordsperpage: 100


```

The default value for the `templates` property will be ['*.twig'], or all. This field will use the same pseudo-regex that the `filter` property uses on template fields. If a field is not relevant to a template as configured, we can simply add the 'hidden' class to the form-group and not display it. It won't require changes to the saving process.

**Other potential solution:**

Another way to do this would be to add a list of relevant fields at the top of the template, using annotations or something. In a case where the configuration gets really complicated, this may help to keep it organized. I don't like this solution as much because it's putting the content type configuration in a bunch of different places.

Pinpickle:
Similar issue: #2711 

I would opt for specifying the fields in the template because that information is specific to the template. If you switched to a different template or theme, there's a good chance that those fields won't be relevant anymore.

rossriley:
Closing this, handled via #3229

PR #3144, opened 2015-03-16
[WIP/RFC] True WYSIWYG Experience (ie. Live Editor)

Pinpickle:
I'll let the GIF do the talking on the feature (give it time to load)

So there it is, the ability to edit content in the theme as it will appear online. Please give it a go, suggest/make improvements. I think this could be something that will make Bolt really attractive to a wider audience.

## What's there

- The basic live editor UI in the Bolt app (you click on the button, and you see the site + a lot of other things disappear). It's an iframe of the preview page with a special flag set for the editable side of things
- If you add `data-bolt-field="fieldname"` in an element in your theme markup, it will become editable in the live editor, if it is of field type text or html
- HTML field types have CKEditor bound to them
- Text field types are just contenteditable with some extra stuff to make sure markup doesn't get in them to muck everything up
- Everytime the view is switched, the two views are synced up
- Global and per contenttype config for `liveeditor`
- All links inside the editing page are disabled to our best ability
- Navigation warning when navigating away from the Bolt app while liveeditor is running
- Form is validated before live editor can be opened
  - I'd rather that this wasn't necessary but I can't think of any reasonable way to get around that. Incorrectly filled forms will break the them so they'll break the preview
- A little bit of extensibility for your particular theme. If you set `data-bolt-noedit` on a tag, Bolt won't do anything to that element, but it will still look to it for some. There is also a JS variable: `boltIsEditing` set to true in the window scope.

## What's not there

- UX is a bit cobbled together from existing elements. If you're in mobile view, it's not immediately obvious how you close it as it's stuck behind the menu.
  - The uploading method becomes even more difficult here, as there is no direct access to the stack.
- Lots of fields not there, that could very easily be. However, there needs to be a very easy way for theme developers to hook into this as the other fields likely have a larger effect than just changing a string: For example, generating thumbnails based on an uploaded image
   - As all fields don't have a generic "set" function, it's also quite difficult to go on the way back. Even the ckeditor implementation is quite hackish. It would be great if we could expose a JS field object for every field, that has a generic set function - but it looks like the recent refactors are heading in that direction
- ~~I really think this will sit well with #2711.~~
- ~~Methods to prevent accidental navigation inside the iFrame~~
- Lots of other things that I have forgotten/haven't seen yet
- ~~Because changes are only synced back to the form when the view changes, you can make edits in the live editor and navigate away without any warning. Easy fix: Show a warning whenever the live editor is open~~
- Every time you open the editor, the embeded page has to reload. For basic fields being edited, this isn't necessary (the changes can just be pushed to the iframe). It would be good if it is detected whether a change needs a refresh (could be specified in contenttypes.yml), otherwise it just does a "hot" update.
- ~~Config options. This is on by default and not optional. This should be optional, and maybe even off by default at first. Should probably also be settable per content type.~~
- I have only tested this in Chrome but it should work in all major browsers.
  - I just tested in IE8. It should work, but I think it's erroring out for other reasons in the master branch.

## Should this be in an extension?

I know there will be a strong argument that the place for this kind of functionality is in an extension, but I disagree. Bolt has everything pretty much nailed from the perspective of the designer and the developer and the *tech-savvy* editor, but this needs to earn the respect of those that aren't tech savvy. Experiences like this, will really help them get over the barrier of an editor where all you see is forms. Especially if there are a lot of fields for a particular record.

With the big players like Squarespace making WYSIWYG actually feasible, these projects need to catch up to stay relevant.

## Security

Initially I tried sandboxing it, but I ran into problems as when you can't share the same origin, the iframe has an origin of `"null"`, making it fairly difficult to authenticate messages posted to the Bolt application. Beyond that, it also meant a fair few new scripts had to be inserted into the preview which complicated the development environment considerably.

So now, this is in an un-sandboxed iframe, which means all scripts inside the preview have full access to DOM/JS of the Bolt application while it is open.

This sounds like it could be a concern, but this capability has always been there as the main website and the Bolt application share the same origin - there are no restrictions anyway. Which means if someone has access to inserting JavaScript into a page, that can then open a new window in the Bolt application and will have access to wreck whatever it wants.

**TLDR;** there are no new security threats imposed by this implementation, to the best of my knowledge


bobdenotter:
Whoa, that looks like magic. :sparkles: 

I'm betting a _LOT_ of people would love to have this feature baked in.

bobdenotter:
It has happened a couple of times, that people told me that we should incorporate "Live editing" as a core feature in Bolt. They usually say something like "But, the editors will hardly even know they're using a CMS!". This is _exactly_ why I usually think it's a bad idea to do this. I feel that this messes with the 'mental model' of the CMS that an editor has. In turn, this will lead to confusion and misunderstanding of how things work, and how things are related. "I edited this page, and suddenly this other piece of text changed too. How? What?" 

That said, I think your implementation looks very promising, @Pinpickle. I like how it allows editors to use 'live editing', but it doesn't completely take the editor away from the concept of "I'm editing a website". I think the main difference is that the _workflow_ of this PR is just right. 

Like you mentioned: Probably a lot of things to work on still, before this is production ready, but i'm really impressed by how it looks already. :+1:

rixbeck:
Like it! Agree with @Bopp the workflow is which makes this cool and why it might be a right place in the core. Although in lunchtime I've been thinking about this a bit further.

I think we need to find right answers for a question like
_when or why the user needs live edit feature?_

I was trying to find possible answers and thought this may helps find how this function could be more useful in Bolt. 

First trivial possible answer for me is _when he/she would like to see the content in its "production" environment to find optimal typographic solution (letter size, pagination, disjunctions, etc)_.

This case the problem is we need to solve that the editing content can be on different pages (listings, side panels), on different page fragments, and better if user can see the result on these existing layouts. Do we able to provide a mechanism to let him chose the page context to edit on somehow? Or don't we want to target this? If so, what possible reasons could be why the editor needs this feature?

Pinpickle:
@bobdenotter, I had absolutely no idea that you were against the idea of live editing. I'm happy that I didn't enquire beforehand because I probably would have been put off the idea and wouldn't have (accidentally) come up with something that you like.

@rixbeck: this is a *really* important question to ask, as it should underpin every feature added to anything.

Firstly, I think that being able to edit page fragments on listing pages or change a resource record that creates the header/footer on any page would be a cool feature, but to me that's the kind of workflow that @bobdenotter is against. It hides the fact that you're in a CMS and it's confusing as to how everything relates. Beyond that, it's quite a lot harder to implement as you're not just editing one record, but many at the same time.

Here are the reasons that I think this implementation would be useful to users:
- Some users don't fully understand the whole "CMS" thing, that editing the text on a page that is completely disconnected from their own website, changes their website. For these users, I'd even say that something like this should be the default editing view (easily configurable with permissions). This way, they understand that what they're editing is their website, and may help introduce them to the concept of a CMS so that they can become more advanced in the future.
- I feel like for long body content, this is a nicer editing experience. This editor takes up the entirety of the viewport and will probably have nicer styling than CKEditor's default styling. True, CKEditor does have a fullscreen button, but I don't find it that useable for some reason.
- @rixbeck's points about seeing the site *as it really is*. Sometimes we'll do a lot of work before checking a preview, only to find some problem after all that work.
- Using classes in the editor makes more sense now, as the styles associated with them will now be viewable. As opposed to adding a class that doesn't do anything in the editor, but makes a big difference on the site.
- We can circumvent "form chaos". Having lots of fields for pages where there is a lot to edit can seem daunting and I have refrained from adding lots of fields in some sites for this very reason. If the editable fields are exactly where they should be in the site, it won't look like there are too many. This will also go towards solving issues like #3033

rixbeck:
@Pinpickle, mostly agree with your points moreover like the whole idea. One thing is sure, Bolt will be more handy with this feature.

rossriley:
Big +1 for this (and showed a few people who agreed it would be very popular with our end users)

bobdenotter:
> Some users don't fully understand the whole "CMS" thing, that editing the text on a page that is completely disconnected from their own website, changes their website. 

This is where I have a different opinion than you do. I'd say: "Suck it up, learn what a CMS is". Having an actual understanding of what you're doing, makes you a better editor. 

But, let's not get into discussion here. Let's focus on getting this functionality stable, in a basic form. For me, that means: 

 - Fix bugs, quirks
 - Keep the UI as is, basically
 - Just the 'record view', no other views yet. 

As with other things we've done: Let's get it working as-is and release it. After that, people will comment and request additional features and we can take it from there. :-)

Pinpickle:
Just added support for template specific fields #3229 
You use selectors for the form names when using `data-bolt-field`, eg:

```html
<div data-bolt-field="templatefields[section_1]">{{ record.templatefields.section_1 }}</div>
```

Intendit:
This is completely awesome. I wish I could use boltkarma on github so that I could give it to @Pinpickle multiple times right here.

pubgit:
Nice work!  One thing I would add is that the admin user should see a rather visible call out at the top of the page when in edit mode.  Some kind of message that lets the admin user know that they are in live edit mode and how to get back to the regular admin mode.  It might need to be a bit redundant to the left nav area.  I have "less advanced" users who might get lost/ frustrated if they accidentally stumble into live edit ( at first anyway).

nealmarques:
Nice work! One thing I would add is that the admin user should see a rather visible call out at the top of the page when in edit mode. Some kind of message that lets the admin user know that they are in live edit mode and how to get back to the regular admin mode. It might need to be a bit redundant to the left nav area. I have "less advanced" users who might get lost/ frustrated if they accidentally stumble into live edit ( at first anyway).

Pinpickle:
Updates!

@bobdenotter if you're happy, I think this is ready now.

What I've added/changed:
- Simplified UX to make closing the editor more obvious, removing every other link because live editor should be closed before navigation
- Added warning for navigation while live editor is open
- Disabled links inside the page being edited so navigation doesn't happen which would much up everything
- Added form validation (things were bugging if the form wasn't valid)
- Added global and per contenttype config for `liveeditor`
- Added a JS variable `window.boltIsEditing` so theme editors can detect if the page is being edited

@GawainLynch, I did a rebase and got rid of all of the compiled static assets per your request

bobdenotter:
Cool! As far as i'm concerned, we're merging this into master, and fix any residual issues as we go. ^_^

bobdenotter:
Whoooo! Merging in!! Choo-choo!

psren:
Super happy with this feature :+1:

eduardomart:
This doesn't seem to work in 3.x anymore...

SahAssar:
@eduardomart Does your theme have the required `data-bolt-field` attributes and is the liveeditor enabled in your config?

PR #3145, opened 2015-03-16
Big refactoring of Bolts JS - part 3

rarila:
This time only bugfixing the fileupload and select from server as some minor cleanup.


bobdenotter:
Looks good to me! :-)

PR #3150, opened 2015-03-17
Hotfix composer install scripts

GawainLynch:
Un-breaks current master `composer create-project bolt/composer-install` installs


bobdenotter:
Choo! choo!

PR #3152, opened 2015-03-17
Add missing jquery..

bobdenotter:
In our backend we have it compiled in our `lib.min.js`, but any extension that adds jquery breaks, without this file present.


GawainLynch:
Travis won't get this… Merging.

Issue #3147, opened 2015-03-16
How to adjust height of WYSIWYG fields in 2.1.1?


peterboorsma:
In earlier versions of Bolt one could alter the height of a WYSIWYG field in contenttypes.yml.
Like this:
        body:
            type: html
            height: 600px

But not in The latest Bolt versions. Has something changed in the markup? Or is it just b0rken?

rarila:
Looks like for now you have to do it this way: #1782

I will have a look into this and add setting it directly from the height attribute later hitting in master.

bobdenotter:
The method described in #1782 doesn't work (anymore). :-)

rarila:
Because it has to be:
```
options:
                ckeditor:
                    height: 123
```

bobdenotter:
I really can't get it to work. Are you sure?

rarila:
You have to also set the autoGrow values!

GDmac:
Or set ``autoGrow_onStartup: false``
(see html field description/example)
https://docs.bolt.cm/contenttypes-and-records

Alternatively, we could have the html field ``height`` setting be some  
kind of macro to automatically set options:height and autogrow:false

bobdenotter:
Closed by #3156

PR #3156, opened 2015-03-17
Setting height of CKEditor fields

rarila:
Setting height of CKEditor fields now works as adveritised:

```yaml
ctype:
    type: html
       height: 600
```

To fully solve this the height parameter should be extended:
```yaml
       # Fixed height:
       height: 600
```

```yaml
       # Minimum height, maximum height (autogrow)
       height: [300, 500]
```

```yaml
       # Minimum height, initial height, maximum height (autogrow)
       height: [300, 400, 500]
```

The last one probably makes no sense, as just using autogrow will do.

Fixes: #3147


bobdenotter:
Sweet! :-)

PR #3160, opened 2015-03-17
JS bugfixes (save button + goto publishing status)

rarila:
- Fix save button does not change state back enabled after save
- Fix not changing to meta tab when sidebar publishing stated is clicked


bobdenotter:
Awesome!! I noticed that first second one today, and was thinking i had to dive in, to see it it was just on my end. I will do this tomorrow morning, because then I can do a cherry-pick for release/2.1 with a clear head! :+1:

rarila:
One thing left: the button stays in focus state. Ideas welcome.

bobdenotter:
`$(this).blur();` ?

rarila:
Hmm…

rarila:
Question is, if displaying a rotating wheel wouldn't be better anyway

PR #3164, opened 2015-03-17
Composer custom backend listing

Pinpickle:
When in composer, the root app/view/twig isn't added to the Twig renderer, so you can't use custom listing templates (or, override any template if you want to).

This PR fixes that.

This is a fix for 2.1, I'll now work on a master fix


bobdenotter:
Looks good! Thanks, @Pinpickle

PR #3165, opened 2015-03-17
Composer custom listing master

Pinpickle:
Master version of #3164 

Should have done this the other way around but this works too!


bobdenotter:
Merging in this one too.. Just going to do a small ninja-edit: `"path"` -> `'path'`. ;-)

PR #3168, opened 2015-03-18
In addition to ed39846: 'attrib' needs to be set for `addCss` as well. 

bobdenotter:


GawainLynch:
The :koala: strikes again! Damn those Drop Bears [(Thylarctos plummetus)](http://australianmuseum.net.au/drop-bear) are deadly!

PR #3169, opened 2015-03-18
Properly support both git and Composer install paths for Codeception

GawainLynch:
This should (hopefully) now allow acceptance tests to run/pass on `bolt/composer-install` installs


bobdenotter:
back from lunch! :bread:

GawainLynch:
Cool, I can have mine and play with this will I do :+1:

PR #3179, opened 2015-03-18
Add composer install acceptance tests

GawainLynch:
This takes us to:
* Acceptance tests for git **and** Composer install types
* Acceptance tests for all three database platforms, for both install types

I'm going to get some :beers: now.


rossriley:
Awesome work, :beers: :beers: :beers:

PR #3183, opened 2015-03-19
Flush the cache if the Bolt version has changed

### LINKED ISSUE #3182: Invalid cache objects can break upgrade
As pointed out by Netbulae on IRC after updating to 2.1.3

GawainLynch:
Fixes #3182

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -18,6 +18,9 @@ class Cache extends FilesystemCache
      */
     const DEFAULT_MAX_AGE = 600;
 
+    /** @var Application */
+    private $app;
+
     /**
      * Default cache file extension.
      */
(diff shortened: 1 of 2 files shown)


bobdenotter:
Shouldn't we just tie this to the changes in #3129 ?

GawainLynch:
My thought that this case is very lowlevel, fast-path, and provides certain guarantees. Therefore special case. Happy to be over turned though.

GawainLynch:
Oh wait, I think I might have misunderstood your point… pinging you on IRC in 5… 4… 3…

bobdenotter:
Nah, I guess you're right.. This does allow us to remove the "… and flush the cache." from the notice that we set in #3129, though.

GawainLynch:
Yeah, I might append this to do exactly that then.

GawainLynch:
Updated! :tada:

bobdenotter:
Heh. I sort of meant only the "clear the cache" part. The notice itself is a 'feature', to reassure the user: "Hey, we noticed you've update! Good job, old chap. Ta ta!".. something like that. ;-)

GawainLynch:
> Heh. I sort of meant only the "clear the cache" part. The notice itself is a 'feature', to reassure the user: 

Derp derp derp derp… With added derp.

GawainLynch:
Can I also suggest we back port this to `release/2.1` as that is where it is hitting.

bobdenotter:
Merging in, and will cherry-pick into release/2.1

Issue #3176, opened 2015-03-18
Error when viewing entry


gwbonline:
Everytime some displays an entry on the web, Bolt wil list an error like this:

624 	Error 	storage 	contentlink 	/bericht/noardske-balk-in-de-balk-in-balk 	Requested specific query 'bericht', not found.

The entry is visible and readable like I want it to be. What does the error say?

Something else: I also get these:

608 	Info 	authentication 	useredit 	/bolt/users/edit/8 	Configuration error: systemlog is not granted to any roles. 	
607 	Info 	authentication 	useredit 	/bolt/users/edit/8 	Configuration error: changelog is not granted to any roles. 	

I have installed 2.1.1 (pl1).

UPDATE:

Ah... in the newest permissions.yml file there are two new lines. I added...

changelog: [ admin, developer, chief-editor ]
systemlog: [ admin, developer ]

...to my own file. I think I will no longer get the authentication error. Do I have to check and edit such files manually everytime there is a new Bolt release?

bobdenotter:
> Do I have to check and edit such files manually everytime there is a new Bolt release?

We strive to keep everything as 'backwards compatible" as possible. That said, it never hurts to take a look, to see what new things have been added :-)

PR #3185, opened 2015-03-19
Only log content not found errors if slug isn't numeric, since next it next checks by id.

CarsonF:
Only log content not found errors if slug isn't numeric, since next it next checks by id.

Fixes first part of #3176

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -120,7 +120,7 @@ class Frontend
         $slug = $app['slugify']->slugify($slug);
 
         // First, try to get it by slug.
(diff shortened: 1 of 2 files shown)


bobdenotter:
:+1:

gwbonline:
When is this implemented in Bolt? I have 2.1.4 installed and I still get errors like below when viewing an entry on the frontpage...

730 	Error 	storage 	contentlink 	/bericht/lutsmuur-moet-versterkt-worden 	Requested specific query 'bericht', not found. 

Can someone tell me what this error means?

GawainLynch:
Hmm… This PR made it into 2.1.4.

gwbonline:
I'll wait for a newer release. Maybe it is solved by then. It's not a big problem, I only want as few errors logged as possible. They do mean something I think.

PR #3186, opened 2015-03-19
Only log content not found errors if slug isn't numeric, since next it next checks by id. 2.2

CarsonF:
Only log content not found errors if slug isn't numeric, since next it next checks by id.


bobdenotter:
As discussed, it might seem a bit hackish at first, but if it works, it works! :-)

PR #3187, opened 2015-03-19
Fix Storage not using sort from contenttype

CarsonF:
- Use sort order from contenttype if none is explicitly specified
- Set limit to 1 if returning single.


bobdenotter:
Topkek!

PR #3189, opened 2015-03-19
Allow ||| queries to be more complex

Pinpickle:
I was trying to put date operations in a `setcontent ... where {'date ||| other_date': '>today ||| >today'} but it wasn't having it.

This fixes that (and most likely some other cases).


CarsonF:
Ha I just brought this up on IRC. :+1:

PR #3191, opened 2015-03-19
Replace string responses from Controllers with a new BoltResponse object

rossriley:
## Overview

This PR replaces the traditional response from $app['render'] effectively with a proxy to the eventual output. 

The new `BoltResponse` object mimics the default Symfony Response object but overloads the sendContent method to compile the response from the loaded template and context variables.

This is primarily going to be useful in testing, here's some of the things we can now test for:

### Test for the correct template

    $request = Request::create('/bolt/extend');
    $response = $app->handle($request);
    $this->assertEquals('extend/extend.twig', $response->getTemplate());

### Test that a particular context variable matches what is expected

    $request = Request::create('/bolt/extend');
    $response = $app->handle($request);
    $context = $response->getContext();
    $this->assertEquals('/my/path', $context['extensionsPath']);

### Test what is in the global twig context

    $request = Request::create('/');
    $response = $app->handle($request);
    $globals = $response->getGlobalContext();
    $this->assertTrue(count($globals['records']) > 1 );

## Other uses

Since the response is only evaluated at the point the app asks to send the response it opens up the potential for extensions or other event listeners to manipulate both the template and the context variables right up to the last minute before it is sent to the browser.

## Test coverage

screenshot 2015-03-19 22 31 52


CarsonF:
Ross you beat me to it! 

What's the purpose of passing renderer into object? Why not a template string?

I don't think content should be rendered in constructor. It would need to be rendered twice if the context was modified.

At some point it will need to be rendered to an HTML string though for those after middlewares expecting that. Like snippets.

CarsonF:
`setRenderer` parameter name doesn't match. Can that be typehinted too?

Oh and great job! :)

rossriley:
Primarily because I didn't want BoltResponse to know whether it was Twig or some other mechanism which may come into play if we want to deal with other types of response.

If we type hint it to a Twig_TemplateInterface we're stuck with only being a wrapper for Twig, which may not be a problem or we may want to handle more.

What do you think?

GDmac:
BoltResponse::create('hello, world',200) wont work, so this breaks the response object. 
Now i have to go look at the source for expected parameters and deduce what they do.

rossriley:
Hold fire on merging this, I'll make a few adjustments.

GawainLynch:
Merge and the koala gets it! :koala: :gun: :guardsman:

GawainLynch:
Drop bear says…

CarsonF:
Yeah a compiled flag was the same thing I was thinking :+1:

It would probably be good to have a getter `isCompiled`. I could see one banging their head against a wall trying to figure out why their changes to context haven't shown up. 

Or maybe reset the compiled flag when context is modified. Well no that wouldnt be good because that would reset HTML modifications made, like adding snippets. 

I was saying typehint Renderer not twig.

CarsonF:
What about twig globals? Those those be placed in the response object until compilation? That way we don't have to know about twig at all in our controllers

rossriley:
Yes, how it is at the moment, Renderer is a pre-compiled Twig template, so that's the only thing we could Typehint it to.

We could probably make our own wrapper interface to this though to allow us to Typehint to BoltRenderer or something like that, I'll think about that for next PR.

Also globals are available via the getGlobalContext() getter is that what you meant?

CarsonF:
@rossriley Take a look at those changes. Renderer was confusing with Render class. I see what you mean though about adding another layer in the future. This isn't public api so does that matter?

I would be really hesitant about using "Renderer" and "Render" as separate concepts since they are so close in name. Maybe Render's functionality could morph into what you see it being and Render::render could be moved to a helper function in the soon to come BaseController?

As far as global context, [here's](https://github.com/bolt/bolt/blob/master/src/Controllers/Frontend.php#L86) what I'm talking about. The controller has to access twig directly to add global variables. Twig could be completely decoupled if there was a setter for that in BoltResponse.

PR #3194, opened 2015-03-20
Fixed ParsedownExtra usage.

### LINKED ISSUE #3195: Markdown fields do not translate definition lists correctly
If I use a definition list in a markdown field it does not come through. The syntax for a definition list is
```
List item title
:  List item description
```

Fixed in #3194

cooperaj:
Bolt was calling ParsedownExtra in a way that meant it was getting a Parseodown instance and consequently the functionality that ParsedownExtra provides was not made available. Namely definition lists and footnotes.

I've also made sure that the correct definition list tags were included in the allowed tags. Since 'dh' is incorrect and what is needed is 'dt'

Fixes #3195

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -14,6 +14,7 @@ use Symfony\Component\Routing\Exception\RouteNotFoundException;
 class Content implements \ArrayAccess
 {
     protected $app;
+    protected $parsedown;
     public $id;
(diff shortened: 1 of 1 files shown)


bobdenotter:
Thanks, Adam! :+1:

PR #3202, opened 2015-03-20
add error messages to the session flash if an upload fails

rossriley:
At the moment any upload validation errors go by without being reported to the user.

This only affects uploads from the /bolt/files upload form, ajax uploads via content editing do report the correct error.


GawainLynch:
Look… over there… :cat:

PR #3204, opened 2015-03-20
Watching CKEditor changes on watchChanges() done OCD

rarila:
Let's use CKEditors built in functions for this.


GawainLynch:
This would be why I can't rung grunt watch right now… grabbing to debug something.

PR #3207, opened 2015-03-21
Fix select from server not working anymore when modal content was reloaded

rarila:


GawainLynch:
Confirmed! The fix is good! Repeat! The fix… is… good!

rarila:
Roger, delete branch now. Over!

PR #3218, opened 2015-03-21
Handle canonical config settings that don't include a scheme

### LINKED ISSUE #3201: "Duplicate" pages with differing canonicals
I'm attempting to create a basic home page, accessible from the site root: http://example.com/

1. I created the page via the admin panel with the slug `page/home`
2. I then set the homepage property in `config.yml` to `page/home`

My homepage is configured properly, however it is accessible via the site root as well as `/page/home` and `/page/1` each with their own canonical value:

http://example.com/
`<link rel="canonical" href="http://example.com/">`

http://example.com/page/home
`<link rel="canonical" href="http://example.com/page/home">`

http://example.com/page/1
`<link rel="canonical" href="http://example.com/page/1">`

I would expect all of these canonical tags to have the same href value. It is possible that a site with pages like this could be penalised in search results for perceived duplicate content.

Extra info:
* This is a clean install (v2.1.3) done via composer with a separate public dir (public/)
* This issue persists (and is exacerbated by) routing changes like https://docs.bolt.cm/templates-routes#contenttype-overrides which will add another URL to the same page with it's own canonical value (`/home`)

GawainLynch:
Fixes #3201

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -277,7 +277,11 @@ class ResourceManager
         // This is where we set the canonical. Note: The protocol (scheme) defaults to 'http',
         // and the path is discarded, as it makes no sense in this context: Bolt always
         // determines the path for a page / record. This is not the canonical's job.
-        $canonical = parse_url($app['config']->get('general/canonical', ""));
+        $canonical = $app['config']->get('general/canonical', '');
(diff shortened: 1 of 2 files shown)


GawainLynch:
Thanks mate, it's late, and going all day and I looked and looked and was convinced it could go but left it for safety… that and Carson Detection would pick up on it :grin:

bobdenotter:
Whoop!! :-)

PR #3222, opened 2015-03-22
Improve `shyphenate()`: Only add breaks to long words, instead of everywhere. Fixes #3221

### LINKED ISSUE #3221: "Shyphenate" naively adds '&shy;' between every two character. It should do this only when needed.
.. otherwise, this is what you get:

ebrc9fw

bobdenotter:


GawainLynch:
Travis and I both agree :+1:

PR #3224, opened 2015-03-22
Normalising taxonomy before comparison for deleting old ones

silentworks:


bobdenotter:
Looks good to me! Thanks, Andrew! :-)

PR #3225, opened 2015-03-22
This fixes the *bug* that our markdown fields look too plain. Very serious BUG! Fixes #2522

### LINKED ISSUE #2522: Spice up our "Markdown" field.
Currently, our Markdown field is really just a `<textarea>`. We should add something like this: 

http://www.codingdrama.com/bootstrap-markdown/

bobdenotter:


bobdenotter:
After this is merged, We need to rebuild CSS/JS.

bobdenotter:
What it looks like: http://recordit.co/XfE4vHTNbN

PR #3229, opened 2015-03-23
Template specific fields

Pinpickle:
PR for #2711 

Use Case: You've designed an fairly complex homepage, but you want it to be really easy to edit everything. Your current setup for pages (with just one HTML field) doesn't fit into this. You either have to create a new contenttype, leverage resource contenttypes or just admit defeat and roll the changes yourself as the designer - instead of the editor. You refrain from making other pages unique and beautiful, because you know that edibility is very important.

A lot of the time, we put fields in contenttypes that don't represent valuable data or information pertaining to that contenttype, but information that is relevant to how it is going to be represented (in the content). It confuses things, really, and introduces problems like I've outlined above.

This PR allows you to define fields for a template, like you would an ordinary contenttype in the theme's `config.yml`, linked to a particular template. For example:

```yaml
template_fields:
    extrafields.twig:
        section_1:
            type: text
            label: Section 1
        section_2:
            type: text
            label: Section 2
        image:
            type: image
```

Then, if you set that template in a record that has a template selector, you'll get a friendly notice that this will make some changes:

image

After saving and refreshing, the new fields appear in the "Template" tab.

image

You can access these fields in a template as `{{ record.templatefields.field }}`

```html
<article>{{ record.templatefields.section_1 }}</article>
<article>{{ record.templatefields.section_2 }}</article>
<article>
    <a href="{{ image(record.templatefields.image) }}">
        <img src="{{ thumbnail(record.templatefields.image, 640, 480) }}">
    </a>
</article>
```

image

And it all just works.

I was going to wait until the new storage layer landed, but I'm impatient. I will happily rewrite all of this functionality for that.

It works by making an extra content object that contains all of the template fields. There is an extra set of "light" contenttypes created in the theme's config that get set to the `templatefields` of a content object. There is now a `isRootType` for contenttypes to determine if they're a "real" one, or created on the fly to serve a purpose such as this.

## Needs to be implemented/addressed

Tests! I can't get PHPUnit or Codeception to successfully run on master at the moment. I don't know if that's my fault or if there's something else going on right now. So I haven't written any tests for this behaviour. I have not yet tested manually, that every field type works.

The data stored in these fields is very fragile. By this, I mean, if you change to a different template that has different template fields, the old fields will be lost due to the nature of records not allowing fields outside of their contenttype. I do make a warning display if this is going to happen when you change templates:

image

It's very crude, all it does is check if a template has fields associated with it, it doesn't compare the actual fields in it.

I think there are better ways this could be handled:

- One could leave the data in the DB so if the template is ever returned to, the fields are restored.
- Better yet, the fields that aren't active, could be displayed as readonly fields so the user could reassign content as necessary (and then mark it for deletion).
- The application could "guess" where to reassign fields to: put HTML fields from one template into HTML fields of another (and similar with other field types)
- We could adopt a common naming convention so that when you switch from one template with fields to another, there is minimal loss (there is still loss, however).

## Bugfixes/refactors inbound

I think at some stage during the JS refactor, server select modals stopped working if they weren't named a certain way - the name `image` still worked, so it was pretty hard to pick up on. This fixes that bug.

I've changed the way fields handle rendering. Previously, they would base all of their rendering off a specific `key` variable. This worked well when there weren't arrays involved. I've split this up into three variables now:

- `key`: Is the ID of the field in the form - `templatefields-image`
- `name`: Is the name of the field in the form - `templatefields[image]`
- `contentkey`: Is the key that we can access this value from on the content object - `image`

Any older custom fields will continue to work, provided they aren't in any template specific fields.

I think this is an important step to making fields more useable, as now they can be placed in any form context fairly easily.


Pinpickle:
Looks to me like these failed tests are genuine! I'll work on fixing those

bobdenotter:
Once again, this looks awesome! I'll check it out soon!

GawainLynch:
What @bobdenotter said, @Pinpickle !

Just to get you started, the first acceptance test is failing with:

```
General error: 1 table bolt_resources has no column named templatefields
```

Pinpickle:
@GawainLynch I think I've fixed that bug but I'm still getting acceptance failures (but they're in master as well, so I think it's something to do with my setup). Travis will have the final say.

GawainLynch:
@Pinpickle I am not seeing that failure on `master`, however that is somewhat irrelevant as this break is a valid "flag" that something and the test simply needs updating.

Have a look [here](https://github.com/bolt/bolt/blob/master/tests/codeception/acceptance/100-BackendAdmin/BackendAdminCest.php#L381) and change:
```php
$I->see('title, slug, body, status, ownerid, datepublish, datedepublish, datecreated', 'td');
```

to

```php
$I->see('title, slug, body, status, templatefields, ownerid, datepublish, datedepublish, datecreated', 'td');
```

…and the *acceptance* tests should pass. The unit test on the other hand are still b0rking and *might* need adjustment too. I'll try to look at it for you later today (UTC).

GawainLynch:
I *may* have just done a background ninja run of PHPUnit… Looks like the database file we bundle would need to be updated.

@rossriley would you be able to note what's needed.

Pinpickle:
@GawainLynch I'm not sure that those are the cause of the error.

The changelog test is failing, but I'm seeing an empty changelog (on master and on this branch), so I think that's where the failure is coming from.

The other failures pertaining to DB updates are coming from this error here:

```
<b>Deprecated</b>:  Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. in <b>Unknown</b> on line <b>0</b>
```

This is getting sent at some point during the DB update test, which is also sending content before the application has sent headers. Resulting in the following error:

```
Fatal error:  Uncaught exception 'RuntimeException' with message 'Failed to start the session because headers have already been sent by "" at line 0.' in 
bolt/vendor/symfony/http-foundation/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php:144
Stack trace:
#0 bolt/vendor/symfony/http-foundation/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php(262): Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage-&gt;start()
#1 bolt/vendor/symfony/http-foundation/Symfony/Component/HttpFoundation/Session/Session.php(81): Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage-&gt;getBag('attributes')
#2 bolt/src/Application.php(219): Symfony\Component\HttpFoundation\Session\Session-&gt;has('user')
#3 bolt/src/Application.php(105): Bolt\Application-&gt;initProfiler()
#4 bolt/app/bootstrap.php(82): Bolt\Application-&gt;in in <b>bolt/vendor/symfony/http-foundation/Symfony/Component/HttpFoundation/Session/Storage/NativeSessionStorage.php on line 144
```

This causes the DB update to fail, which means a resource contenttype isn't added so all of the other resource contenttype tests fail because there is no resource contenttype database.

I don't get this issue when adding them manually.

These are both issues I'm getting in master, so maybe its something to do with my setup?

GawainLynch:
@Pinpickle I am 100% sure that the Codeception test is failing for the reason I stated. I pulled your branch, copy/pasted the change I gave above and the test passes.

Pinpickle:
@GawainLynch definitely something to do with my setup then, because I have 6 failing tests and making those changes doesn't make a scrap of difference.

That, and my changelog just doesn't work. Will do some troubleshooting, but in the meantime I'll make your changes and push up!

GawainLynch:
Weird… I am trying to think why your local set up would be failing and without seeing the failures it is all guess work.

One thing that might help you, when you run `./tests/scripts/run-codeception-tests` and there is failures on matching output, it will dump the HTML that it received to `tests/codeception/_output/` (just make sure you either remove old ones first of run `codeception clean` first.

Pinpickle:
@GawainLynch yeah I checked the output file. The comment before my last one outlines what errors are being thrown. Specifically, its the [last part of the DB update test](https://github.com/bolt/bolt/blob/master/tests/codeception/acceptance/100-BackendAdmin/BackendAdminCest.php#L197) that's failing (which is supposed to update the database with a resource contenttype) so all subsequent resource contenttype related tests also fail as there is no table for them.

GawainLynch:
Cool, the acceptance tests are passing on Travis now… just the PHPUnit unit tests to fix!

src/Content.php
```diff
@@ -127,79 +132,96 @@ public function getValues($json = false)
...
-        foreach ($contenttype['fields'] as $field => $property) {
-            switch ($property['type']) {
-
-                // Set the slug, while we're at it
-                case 'slug':
-                    if (!empty($property['uses']) && empty($this->values[$field])) {
-                        $uses = '';
-                        foreach ($property['uses'] as $usesField) {
-                            $uses .= $this->values[$usesField] . ' ';
+        if (is_array($contentttype)) {
```
GawainLynch: 3 t's here… test fail one

src/Content.php
```diff
@@ -238,7 +262,7 @@ public function setValues(array $values)
         );
         // Check if the values need to be unserialized, and pre-processed.
         foreach ($this->values as $key => $value) {
-            if (in_array($this->fieldtype($key), $serializedFieldTypes)) {
+            if ((in_array($this->fieldtype($key), $serializedFieldTypes)) || ($key == 'templatefields')) {
```
GawainLynch: Isn't 'templatefields' is a serializedFieldTypes value… add it to the array above, no?

GawainLynch:
Awesome work!

dwolf555:
Love it! :D

src/Content.php
```diff
@@ -332,6 +371,30 @@ public function setValue($key, $value)
...
+                        $unserdata = false;
+                    }
+                } else {
+                    $unserdata = $value;
+                }
+
+                if ($unserdata !== false) {
+                    $templateContent = new Content($this->app, $this->getTemplateFieldsContentType(), array(), false);
+                    $value = $templateContent;
+                    $this->populateTemplateFieldsContenttype($value);
```
GawainLynch: @Pinpickle can you remember what this was supposed to be for/doing? 

The function itself never got committed… very obviously the :koala: snuck in and messed with things :wink:

Pinpickle: I'm not too sure...

From memory I *think* that this was from one of my older implementations and was superseded by the whole `$isRootType` thing. I haven't seen this throwing any errors yet... This should have come up by now, shouldn't it? Especially as there are tests in place for template fields.

I think it would be safe to remove it.

PR #3239, opened 2015-03-23
[WIP] Allow 'name' in contenttype to override translation.

### LINKED ISSUE #3234: We should _not_ translate content.
Like, ever. See screenshot: 

screen shot 2015-03-23 at 15 50 39

I can set the `name` of the `entries` contenttype to whatever I want, but it doesn't change on the site, because it stubbornly keeps using the translation for the word 'entries'. Bolt shouldn't try to do this.

bobdenotter:
Resolves #3234

### DIFF
--- a/app/view/twig/nav/_secondary-content.twig
+++ b/app/view/twig/nav/_secondary-content.twig
@@ -33,7 +33,6 @@
                 ]) %}
             {% endfor %}
 
-
             {% set active = (page_nav == 'Content/*' and context.contenttype.slug|default == slug) %}
 
             {{ nav.submenu(contenttype.icon_many|default(''), label_plural, sub, active, true) }}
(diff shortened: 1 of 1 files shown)


bobdenotter:
Let's say we have a basic contenttype that looks like this: 

```yml

pages:
    name: "Þe olde pages"
    singular_name: "olde page"
    slug: pages
    singular_slug: page
    fields: 
        ... 

entries:
    name: Entries
    singular_name: Entry
    fields:
        ...
```

Before, this looked like this, the 'name' was overruled by the translation. 

screenshot 2015-03-23 20 24 00


With this change, it looks like this: 

screenshot 2015-03-23 20 20 22

If you set `locale: de_DE`, you'll notice that the ones that _aren't_ explicitly set will still use the translation (see 'Einträge').

screenshot 2015-03-23 20 20 49


This is a work in progress. To me this seems like a good change to make. If we're agreed, I'll make this same change in a few other places, before we can merge it in.

GawainLynch:
Bolt currently doesn't handle this properly… this fix addresses that adequately… what we should do long term might be another story, but for me :+1: 

@rarila does this work for you as an intermediates solution?

bobdenotter:
Closing this one.

PR #3243, opened 2015-03-24
Alternative to #3239

rarila:
Draft for an alternative to #3239.


bobdenotter:
Sorry, Rarila, but I really don't like the removal of 'name' and 'singular_name'. It would mean we'd have to change a _lot_ of existing websites or they'd break. 
Also, I don't like having to edit a separate language file. I just want to overwrite a label, and to me it makes most sense to do that in `contenttypes.yml`. :-/

bobdenotter:
@rarila: This can be closed, because we'll do it properly, soon. Right?

PR #3245, opened 2015-03-24
Fix typo/copy paste oopsie in homepage test

timcooper:
This is a fix for the fix (#3214) for #3201 

https://github.com/bolt/bolt/commit/9ed06a0f7aba358dbd7646db93f8a4dbfb87ffa9#diff-41eef01f213bbe09b595d7594b6f199dR956

Was checking `[singular_slug] / [id]` twice instead of `||`ing with `[slug]`


GawainLynch:
Anyone else glad that this wasn't the :koala: ?

PR #3247, opened 2015-03-25
Washers for Bolt's Nuts

### LINKED ISSUE #3172: Cannot write to log from nut command
Hi, Bolt team!! Latest Bolt 2.1.3 here.

When I try to write to the system log (with the new Monolog implementation) from a nut command, nothing gets writen. Debugging  'Logger\Handler\SystemHandler::write()'  I can see that the following code is eating an exception with message **"Accessed request service outside of request scope. Try moving that call to a before handler or controller."**

```php
        try {
            $this->app['db']->insert(
                $this->tablename,
                array(
                    'level'      => $record['level'],
                    'date'       => $record['datetime']->format('Y-m-d H:i:s'),
                    'message'    => $record['message'],
                    'ownerid'    => isset($user['id']) ? $user['id'] : '',
                    'requesturi' => $this->app['request']->getRequestUri(),
                    'route'      => $this->app['request']->get('_route'),
                    'ip'         => $this->app['request']->getClientIp(),
                    'context'    => isset($record['context']['event']) ? $record['context']['event'] : '',
                    'source'     => $source
                )
            );
        } catch (\Exception $e) {
            // Nothing.
        }
```
Of course I can't move the call to a before handler or controller because is being called from a nut command. 

I assume the problem is the three calls to `$this->app['request']`, that do not make sense from a CLI call.

GawainLynch:
* Fixes #3172
* Set $app['request'] in BaseCommand to enable logging and getContent() functionality in Nut commands
* Add an audit log wrapper function to BaseCommand
* Logs (by default) system modifying events in the system log
* Adds a 'nut' context to the system log UI
* PHPDoc love

### DIFF
--- a/app/view/twig/activity/systemlog.twig
+++ b/app/view/twig/activity/systemlog.twig
@@ -40,6 +40,7 @@
                         <option value="deprecated">deprecated</option>
                         <option value="exception">exception</option>
                         <option value="news">news</option>
+                        <option value="nut">nut</option>
                         <option value="security">security</option>
                         <option value="template">template</option>
                         <option value="translation">translation</option>
--- a/src/Logger/Handler/SystemHandler.php
+++ b/src/Logger/Handler/SystemHandler.php
@@ -99,8 +99,8 @@ class SystemHandler extends AbstractProcessingHandler
                     'message'    => $record['message'],
                     'ownerid'    => isset($user['id']) ? $user['id'] : '',
                     'requesturi' => $this->app['request']->getRequestUri(),
-                    'route'      => $this->app['request']->get('_route'),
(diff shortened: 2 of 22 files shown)


bobdenotter:
Sweet! :+1:

magabriel:
You guys rock! Ability to call getContent() from commands is great for us
CLI-addicts.

GawainLynch:
Yeah, it's been on my TODO forever… You might also then like what's [coming with this](https://github.com/bolt/bolt/compare/bolt:master...GawainLynch:feature/nut-database-record-import-export?expand=1)

magabriel:
Great!

PR #3251, opened 2015-03-25
Codeception speed tweaks

GawainLynch:
* Only login Codeception users once per user class run
  * Local acceptance tests dropped from an average of 103 seconds to ~55 seconds per run
* Turn up Twig's anger ratio on acceptance tests (strict_variables: true)


bobdenotter:
Travis taking his sweet time, today.

GawainLynch:
Yeah, still problems with packagist.org… Nils and Jordi are working on it though.

PR #3263, opened 2015-03-26
Reset dirty check after updating ckeditor field. Fixes #3256

### LINKED ISSUE #3256: HTML fields are behaving weirdly in master
Their internal textarea doesn't look like its getting updated, so the content isn't getting saved. To reproduce this, make a new record of a contenttype that has an HTML wysiwyg field. Save it without changing the HTML field. On the refreshed page, update the HTML field and then try to save again. The HTML field will clear and your changes won't be saved. If it doesn't clear, refresh the page and you'll see that your changes haven't persisted

It also seems like scrollbars in them have stopped working.

image

As you can see, there is more content in that ckeditor, but i cannot scroll (there's about a paragraph underneath it.

Ping @rarila

timcooper:
This fixes the remaining issue reported by @EJanuszewski on #3256 where templatefields of `type: html` would cause the page navigation confirmation to display despite having saved.

### DIFF
--- a/app/src/js/modules/editcontent.js
+++ b/app/src/js/modules/editcontent.js
@@ -385,6 +385,7 @@
             if (this.type === 'textarea' && $(this).hasClass('ckeditor')) {
                 if (ckeditor.instances[this.id].checkDirty()) {
                     ckeditor.instances[this.id].updateElement();
+                    ckeditor.instances[this.id].resetDirty();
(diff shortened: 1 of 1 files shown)


GawainLynch:
Travis won't pick up on this either way, grabbing and I'll rebuild the JS right after.

PR #3264, opened 2015-03-26
Since the username is not unique, we shouldn't use it as array key.

TrafeX:
We didn't saw users that have the same username. This proved to be the cause.

### DIFF
--- a/src/Users.php
+++ b/src/Users.php
@@ -771,7 +771,7 @@ class Users
                 $tempusers = $this->db->fetchAll($query);
 
                 foreach ($tempusers as $user) {
-                    $key = $user['username'];
+                    $key = $user['id'];
                     $this->users[$key] = $user;
                     $this->users[$key]['password'] = "**dontchange**";
 


bobdenotter:
I'm curious how yo got duplicate names in the DB, though.. Inserting them straight into the table, perhaps? ;-)

TrafeX:
@bobdenotter No, we have a registration form, validate/handle it and call the `saveUser` method: https://github.com/bolt/bolt/blob/master/src/Users.php#L94 :)

bobdenotter:
Merging this in will break the option to use getUser with a username: 

    {{ dump(app.users.getUser('admin') }} 

It's probably other than you had in mind, but the proper way to fix this would be to add a check to `saveUser` to prevent duplicate filenames.

TrafeX:
@bobdenotter Good catch! I guess we can close this PR. I've created an issue to solve the underlying problem: https://github.com/bolt/bolt/issues/3287

PR #3273, opened 2015-03-28
Start of PHP 7 compatibility fixes

GawainLynch:
The class name `String` is reserved in PHP 7. 
* Rename `Bolt\Helpers\String` to `Bolt\Helpers\Str`
* Update tests for `Bolt\Helpers\String` to `Bolt\Helpers\Str` rename
* Add a deprecated wrapper class for `Bolt\Helpers\String`


GawainLynch:
Thanks to the work of @remicollet on the PHP 7 packages, this is what we get with these fixes applied:
screenshot from 2015-03-28 10 09 07

bobdenotter:
whoohoo!

PR #3274, opened 2015-03-28
Change CKEditor Skin

Pinpickle:
The monochromatic gradients of Moono give me shudders and make me think of all kinds of bad web applications. I have received similar sentiments from some of my clients. HTML fields also stick out like sore thumbs when editing content.

image

While I think it may be beneficial to eventually roll out or own Bolt skin for CKE, I had a look at those that are currently available. ~~[Minimalist](http://ckeditor.com/addon/minimalist)~~ [Flat](http://ckeditor.com/addon/flat), I felt, was the best fit, but it didn't reskin the dialog boxes. [Bootstrapcke](http://ckeditor.com/addon/bootstrapck) did a good job of that, and made the dialog boxes look more like Bolt modals. However the main editor UI still left something to be desired.

So I took the editor of one, the dialog boxes of another and shoved them together to make Boltcke. I've also added one style edit (in our sass files because modifying those skin files is complicated) to add a border around the whole field.

Now the experience is a lot nicer, I think.

### Editor

image

### Dialog Box

image


bobdenotter:
That's a lot of changed files.. 

Also, isn't our CK version pulled in from somewhere and built by Grunt? I'm sure @rarila knows more about this. :-)

Pinpickle:
Nah, ckeditor is made from a local copy of the build as far as I know.

However, it looks like I've wrecked some of the config and locales stuff so
I'll work on fixing that (hence hundreds of files changing)

Pinpickle:
@bobdenotter I've fixed all of the above issues and changed the editor theme to Flat. Minimalist didn't actually specify a background which I know will lead to issues with the live editor!

GawainLynch:
Process request. Can you leave out rebuilt `bolt.min.js` and similar files as if someone else gets a change in ahead of you, then your PR will end up in a conflict state… and :crying_cat_face: 

No need to worry about it on this one though.

Pinpickle:
@GawainLynch, I had wondered about that one.

I'll leave it out in the future.

bobdenotter:
Looks good to me! Merging in! Thanks, @Pinpickle

PR #3276, opened 2015-03-28
Match sorted array keys with assertRegExp() where values are equal as it is an unpredictable outcome on equal elements.

### LINKED ISSUE #3275: PHP7/HHVM array sorting changes
This came up on php-internals:
> Sorting in php master in this specific case is now different on PHP 7 than on 5.3-5.6 (and actually different than PHP 7 at the beginning of the month). The sorting order is now the same as HHVM.

Which matches a unit test fail that we're seeing on HHVM, and with #3273 on PHP7, where we are doing:
```php
$input = [
    ['title' => 'Gamma', 'id' => 10, 'date' => '2014-01-19'],
    ['title' => 'Alpha', 'id' => 10, 'date' => '2014-02-20'],
    ['title' => 'Beta',  'id' => 8,  'date' => '2014-01-10'],
    ['title' => 'Delta', 'id' => 6,  'date' => '2014-01-19']
];
$result = array_values($twig->order($input, 'id', 'date'));
```
…and expecting `$result[2]['title']` to be 'Gamma', however in PHP7 and HHVM we're returning 'Alpha'.

So to quote some guy called [Rasmus Lerdorf](http://marc.info/?l=php-internals&m=142230002429001&w=2):

> Yeah, this is a known change and the documentation has had a note about
not relying on the ordering of equal elements. Although the note in the
manual is probably a bit too geeky. It should be more explicit and talk
about the effect rather than the cause:

>    Note: Like most PHP sorting functions, sort() uses an
          implementation of » Quicksort. The pivot is chosen in the
          middle of the partition resulting in an optimal time for
          already sorted arrays. **This is however an implementation
          detail you shouldn't rely on.**

This has the potential to eclipse #1555 in terms of impact… I want PHP 7, damn it!

GawainLynch:
Fixes #3275
Depends on #3273


bobdenotter:
Also whooohoo!

PR #3282, opened 2015-03-30
Add tests for templatefields

Pinpickle:
@GawainLynch for some reason when saving a page with title `Contact Page`, it is getting saved with `CONTACT PAGE`. Does this have something to do with the save hooks?

Regardless, this tests for working templatefields


GawainLynch:
Yes, there is a local extension installed/removed during the Codeception test runs that has [PRE/POST_SAVE hooks](https://github.com/bolt/bolt/blob/master/tests/codeception/_data/extensions/local/bolt/tester-events/src/StorageEventTests.php)

Pinpickle:
@GawainLynch, just found the code for that! Sneaky little `strtoupper`

Everything is right with the world.

PR #3284, opened 2015-03-31
Fix strange step behaviour when no min is set for integer fields

### LINKED ISSUE #3242: Without min set, step value in integer/float fields is unpredictable
The min value, by default, is set to the minimum value that the DB will allow (-2147483647). If a `step` value is set for the number fields, it will put that in the `input` element as well. The browser uses increments from the `min` value of `step` to verify if the value is allowed. So if I set `step` to 1000, the allowed values I get are undesirable.

image

Setting the `min` value mitigates this issue, but that solution isn't particularly obvious. Default functionality should be to round the min DB value up to the nearest multiple of `step`.

Pinpickle:
Fixes #3242


bobdenotter:
Looks good!

PR #3285, opened 2015-03-31
Restore BoltResponse __toString with nicer error handling

rossriley:
This should be a decent compromise between the two problems.

The __toString method is restored so extensions that depend on it will continue to function identically.

Whilst PHP disallows exceptions being thrown in the __toString method it seems to be fine with triggering normal errors so, we can catch the original exception and trigger an error with the original exception message.


bobdenotter:
Thanks! :+1: 

I'll cherry-pick this for 2.1.x as well!

PR #3286, opened 2015-03-31
Use 'executeQuery'. Fixes publishing for items with 'Timed Publish'. Fixes #3279

### LINKED ISSUE #3279: SQL Error trying to publish timed publications
Timed publication of records for Array failed: An exception occurred while executing 'UPDATE bolt_en_articles SET status = 'published', datechanged = :now WHERE status = 'timed' and datepublish < :now': SQLSTATE[HY000]: General error: mode must be an integer

bobdenotter:


CarsonF:
Stupid stupid. Who even decided query was better to use anyways.

CarsonF:
Actually it may be because master slave stuff. executeQuery doesn't detect the query it's running to know to switch to master connection

bobdenotter:
@CarsonF Sure about that? We can't just use timestamps for datetime columns, right? 

Anyhow, using this, it won't trigger for me: 

```
        $now = time();

        try {
            // Check if there are any records that need publishing.
            $stmt = $this->app['db']->executeQuery(
                "SELECT id FROM $tablename WHERE status = 'timed' and datepublish < $now"
            );

            // If there's a result, we need to set these to 'publish'.
            if ($stmt->fetch() !== false) {
                $this->app['db']->query("UPDATE $tablename SET status = 'published', datechanged = $now WHERE status = 'timed' and datepublish < $now");
            }
```

CarsonF:
Almost positive db stores them as unix timestamps, which are ints. 

Also, ints work for me and that timestamp string throws an exception. Pretty sure the timestamp string is converted by dbal.

CarsonF:
Haha waitttttt. *facepalm*. We don't even need the variable. sql has `now()`.

bobdenotter:
Ok, the following works, except for the offset you mentioned.. :-(

try {
            // Check if there are any records that need publishing.
            $stmt = $this->app['db']->executeQuery(
                "SELECT id FROM $tablename WHERE status = 'timed' and datepublish < CURRENT_TIMESTAMP()"
            );

            // If there's a result, we need to set these to 'publish'.
            if ($stmt->fetch() !== false) {
                dump($stmt);
                $this->app['db']->query("UPDATE $tablename SET status = 'published', datechanged = CURRENT_TIMESTAMP() WHERE status = 'timed' and datepublish < CURRENT_TIMESTAMP()");
            }
        }

bobdenotter:
PR Fixed!

bobdenotter:
@CarsonF 

> Sorry, "everywhere" meaning in this method. aka these 3 times.

If we only do that here, we'll break the current behaviour.. if we did it properly, we'd also have to make sure we fix the _other_ locations where `datepublish` is modified. Mostly in 'editcontent' i assume.

GawainLynch:
:traffic_light:

PR #3299, opened 2015-04-01
Add extra check for view permission for quicklinks

StevendeVries:
When allowing a user to create a contenttype but without view permission (perhaps create it with a custom extension). An exemple is a resource contenttype. I've added a check to make sure the user is able view the contenttype.

Another check that might be handy for the future is a check on show_on_dashboard: false


bobdenotter:
Green lights! Travis gave an error for this one at first, but that was just travis being flakey. 

Thanks, @StevendeVries

PR #3300, opened 2015-04-02
Enable live editing for textarea fields

timcooper:
I've attempted to ensure the features match the equivalent backend field, the main goals being allowing newlines (for cases like running the value through the nl2br twig filter) and getting rid of random cruft (`&nbsp;`s) added by the browser.

Browser default behaviour on enter is to wrap an HTML element (div, p etc.), I've changed this to insert `<br>`s whilst editing which are then converted into newline characters when the live editor is closed.

Ping for Cap'n Liveedit, @Pinpickle, in case he needs to weigh-in.

### DIFF
--- a/app/src/js/modules/liveeditor.js
+++ b/app/src/js/modules/liveeditor.js
@@ -193,13 +193,14 @@
             var fieldType = field.closest('[data-fieldtype]').data('fieldtype');
             var fieldValue = '';
 
-            if (fieldType === 'text' || fieldType === 'textarea') {
-                fieldValue = liveEditor.cleanText($(this), fieldType);
-            } else {
-                if (_.has(ckeditor.instances, fieldName)) {
-                    ckeditor.instances[fieldName].setData($(this).html());
-                }
+            if (fieldType === 'html') {
                 fieldValue = $(this).html();
+
+                if (_.has(ckeditor.instances, fieldName)) {
+                    ckeditor.instances[fieldName].setData(fieldValue);
+                }
+            }else{
+                fieldValue = liveEditor.cleanText($(this), fieldType);
             }
             field.val(fieldValue);
         });
--- a/app/src/js/modules/liveeditor.js
+++ b/app/src/js/modules/liveeditor.js
@@ -195,7 +195,7 @@
             if (fieldType === 'text') {
                 field.val($(this).text());
             } else if (fieldType === 'textarea') {
-                field.val($(this).html().replace(/&nbsp;/g, ' ').replace(/<br.*?>/g, '\n'));
+                field.val($(this).html().replace(/&nbsp;/g, ' ').replace(/\s?<br.*?>\s?/g, '\n'));
             } else {
                 if (_.has(ckeditor.instances, fieldName)) {
                     ckeditor.instances[fieldName].setData($(this).html());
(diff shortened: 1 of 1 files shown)


app/src/js/modules/liveeditor.js
```diff
@@ -122,11 +128,22 @@
...
+                                if(e.which == 13) {
+                                    e.preventDefault();
+                                    doc.execCommand('insertHTML', false, '<br><br>');
+                                }
+                            });
+                        } else {
+                            $(this).on('keypress', function (e) {
+                                return e.which != 13;
+                            }).on('focus blur', function (e) {
+                                $(this).html($(this).text());
```
Pinpickle: This should probably be executed for textarea types as well because it undoes HTML other html changes incurred by the browser. For example, hitting `ctrl/cmd + B` adds `<strong>` tags around the selected text.

app/src/js/modules/liveeditor.js
```diff
@@ -177,6 +194,8 @@
 
             if (fieldType === 'text') {
                 field.val($(this).text());
+            } else if (fieldType === 'textarea') {
+                field.val($(this).html().replace(/&nbsp;/g, ' ').replace(/\s?<br.*?>\s?/g, '\n'));
```
Pinpickle: I'm thinking we should d this for ordinary text field types as well. Maybe combine this functionality with the `onblur` stuff I mentioned in my other comment so that the sanitisation isn't happening in different places.

This is a better places for sanitisation to happen because people may close the editor without first unfocusing a field that they're editing.

app/src/js/modules/liveeditor.js
timcooper: The reason I did it differently was to ensure parity with the non-liveedit counterpart. In a textarea field you can have newlines and they will persist to storage. This allows things like outputting the field with `{{ content.textarea | nl2br }}` using `.text()` would mean you can have newlines when editing content via the panel and then they would disappear if you used live edit.

I wasn't aware of the "extra" contenteditable commands, browser vendors are terribly inconvenient. Seems like there are plans for a `contenteditable="plaintext"` functionality but nothing we could actually use yet.

Pinpickle: Yeah I completely agree with the handling of <br> characters being different, just concerned about other HTML sneaking in!

It's pretty shocking that there's no simple way to do plaintext at present, given that `contenteditable` has been around since IE5.

bobdenotter:
@timcooper @Pinpickle I see that there's room for improvement, but as it stands, I think this is good to go. Should we merge this in and do further work in consecutive PR's, or do you want to tweak this PR first?

bobdenotter:
@timcooper *bump* ;-)

timcooper:
Okay, I've added some extra handling for textarea field types which will get rid of anything added in via REALLY USEFUL browser features we can't control like <kbd>CTRL</kbd>+<kbd>B</kbd> bolding text in contenteditables.

I also did some general refactoring around the value juggling stuff to make it clearer what's going on.

timcooper:
The failing tests seem to be a connectivity issue? The file modified by this PR isn't covered by tests.

GawainLynch:
@timcooper looks to be Travis being derpy… I've restarted the failed ones nonetheless.

bobdenotter:
Why was this still open? :-)

Merging in. Thanks, @timcooper

PR #3305, opened 2015-04-03
Guzzle 5 — take 2

GawainLynch:
This is a follow up to #3280.

> Codeception pulls in Guzzle 5, so in preparation for us dropping PHP 5.3 support later in the year, this allows PHP 5.4.0+ to use Guzzle 5 and PHP 5.3 stays on Guzzle 3.

@rossriley I'll need some follow up help to get BackendTest::testPrefill() back online.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
     "license" : "MIT",
     "require" : {
         "bolt/dumper" : "~2.0",
+        "bolt/package-wrapper": "~1.0|~2.0",
         "bolt/pathogen" : "~0.6",
         "bolt/thumbs" : "~2.0",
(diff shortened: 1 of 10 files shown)


GawainLynch:
Oh, should also note that what got this over the line was `"bolt/package-wrapper": "~1.0|~2.0"` in composer.json.

Basically `bolt/package-wrapper` it just a wrapper package where:
  *  version 1.0 requires Guzzle 3.9
  *  version 2.0 requires Guzzle 5.2

Now to nag someone on #composer-dev :grinning:

bobdenotter:
Neat!

PR #3307, opened 2015-04-03
Frontend requests should not set cookies. Remove them, to allow Varnish to do a better job of caching the request.

bobdenotter:


src/Controllers/Frontend.php
```diff
@@ -70,6 +70,32 @@ public function before(Request $request, Application $app)
...
+    public function after(Request $request, Response $response, Application $app)
+    {
+        // Frontend requests should not set cookies. Remove them, to allow Varnish
+        // (or reverse-proxies) to do a better job of caching the request.
+        if (!headers_sent()) {
+            $headersList = headers_list();
+            foreach($headersList as $header) {
+                if (strpos($header, "Set-Cookie: bolt_session=") === 0) {
+                    header_remove("Set-Cookie");
+                }
```
GawainLynch: Would this be simpler:
```php
$response->headers->clearCookie('bolt_session');
```

bobdenotter: It would, if i had known about it. :-)

GawainLynch: It aslo does the [headers_sent()](https://github.com/symfony/HttpFoundation/blob/2.6/Response.php#L332)

bobdenotter:
PR updated.

bobdenotter:
Fixed for 2.3 by #3564

PR #3310, opened 2015-04-04
Slugs generation fixed

rarila:
Looks like BUGS Bunny was hopping around in the slugs code a long time ago (at least v1.6). :rabbit:

I won't go into detail as it was kind a Matroska of bugs but now the autogenerating of slugs with ``uses`` should work again. Works now also with multiple slugs and with all kinds of select fields.

Some other enhancements are part of a PR to come.


bobdenotter:
Choo Choo!! Merging in!

PR #3312, opened 2015-04-04
[WIP] Make editable fields 'magic'.

bobdenotter:
I'd like to get rid of having to add `data-bolt-field="{{ key }}"` to the fields you want to have editable. 

So far, this works, except for one thing: The floating toolbars don't show up anymore. 

@pinpickle: Do you have any idea if this is caused by the wrapping element being different? Can't say the CKeditor documentation was a joy to dig through, so I couldn't find much..

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -520,6 +520,24 @@ class Content implements \ArrayAccess
         $this->setValues($values);
     }
 
+
+    public function wrapValues()
+    {
+        $fields = $this->contenttype['fields'];
+        $allowedEditable = array('text', 'textarea', 'html');
+
+        foreach ($this->values as $key => $value) {
+            $fieldtype = $fields[$key]['type'];
+            if (in_array($fieldtype, $allowedEditable)) {
+                $this->values[$key] = sprintf(
+                        '<span class="editable" data-bolt-field="%s">%s</span>',
+                        $key,
+                        $value
+                    );
+            }
+        }
+    }
+
     protected function getTemplateFieldsContentType() {
         if (is_array($this->contenttype)) {
             if ($templateFieldsConfig = $this->app['config']->get('theme/templatefields')) {
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -183,6 +183,7 @@ class Frontend
             $app['extensions']->insertSnippet(SnippetLocation::BEFORE_HEAD_JS, '<script>window.boltIsEditing = true;</script>');
             $app['extensions']->addJavascript($jsFile, array('late' => false, 'priority' => 1));
             $app['extensions']->addCss($cssFile, false, 5);
+            $content->wrapValues();
         }
 
         // Then, select which template to use, based on our 'cascading templates rules'
--- a/theme/base-2014/record.twig
+++ b/theme/base-2014/record.twig
@@ -7,10 +7,11 @@
 
         <article>
 
-            <h2 data-bolt-field="title">{{ record.title }}</h2>
+            <h2>{{ record.title }}</h2>
+
 
             {% if record.subtitle %}
-                <h3 data-bolt-field="subtitle">{{ record.subtitle }}</h3>
+                <h3>{{ record.subtitle }}</h3>
             {% endif %}
 
             {% for key,value in record.values if key not in ['id', 'slug', 'datecreated', 'datechanged', 'datepublish', 'datedepublish', 'username', 'status', 'title', 'subtitle', 'ownerid', 'templatefields'] %}
@@ -54,7 +55,7 @@
 
                 {% elseif record.fieldtype(key) in ['html', 'markdown', 'textarea'] %}
 
-                    <div data-bolt-field="{{ key }}">{{ attribute(record, key) }}</div>
+                    {{ attribute(record, key) }}
 
                 {% elseif record.fieldtype(key) == "select" %}
 


Pinpickle:
Yes, to my understanding CKEditor is "smart" about its inline editing where:

- `div`s will allow you to edit full HTML inside the element (to the extend of what is permitted by config)
- Other block elements such as `p` and `h1` will only allow you to add inline elements (`span`, `strong`, `em`)
- When editing inline elements (such as the `span` that you've got there), you can not add any HTML

I'm not sure that what you're trying to do is reliably achievable, though. I'll express a few concerns about your implementation:

- You're altering the value of the field, which means if the theme does anything *other* than just outputting it (such as spliting it up into different words through its spaces), unexpected things will happen in live editor (such as `<span` being one of the words)
- You're wrapping things in `span` tags which is invalid HTML - you can't have block level elements (`p`) inside inline elements
- You're changing the markup of the page which is very intentionally put together by the theme designer. This could mess up with styling in all sorts of ways
- We don't necessarily know if the designer wants this particular part to be editable in live editor, for whatever reason - this should be an "opt-in" in my opinion.

The only way, I think, to implement this reliably, is do some manipulation on the output HTML. Where there is a field output in the template, and it is the **only** content in a tag apart from whitespace (so `<div> {{ record.body }} </div>` will pass but `<div> {{ record.body }} <div class="date">A Date</div></div>` will not, add the attribute to the element that contains it.

I imagine that would be a lot of work, but it isn't impossible. It would also need to keep the `data-bolt-noedit` attributes in tact of a designer opts out of the live editor for that field.

Still, I think this isn't really necessary - adding those attributes isn't a lot of effort and I think that the designer needs to know how the live editor is getting the info it gets. That being said, magic is pretty cool.

bobdenotter:
Hmm, I see your point. 

I guess what I wanted to achieve is that people are just trying out bolt, they click the "Live Edit" button because "Hey, fancy!", and then nothing happens, because they (or the theme developer) didn't add any editable tags to the theme. This attempt is a bit too lazy, I think.

> The only way, I think, to implement this reliably, is do some manipulation on the output HTML. 

That would be the _best_ way, but i'm not sure that's do-able. Apart from the tricky code to actually make this work, we'd need to build a DOM tree from the HTML to manipulate it. Because we all know what happens when we use [regular expressions](http://stackoverflow.com/a/1732454/21285) for that. 
Meanwhile, the average theme will not be valid XML, so this will have side-effects on the theme. Also not what we want. 

> I imagine that would be a lot of work, but it isn't impossible. It would also need to keep the data-bolt-noedit attributes in tact of a designer opts out of the live editor for that field.

Ideally, this would be controlled in the `contenttypes.yml`, i think.

Pinpickle:
> I guess what I wanted to achieve is that people are just trying out bolt, they click the "Live Edit" button because "Hey, fancy!", and then nothing happens, because they (or the theme developer) didn't add any editable tags to the theme. This attempt is a bit too lazy, I think.

Ah, I think I've thought of a potential solution because you're right - this is far from ideal.

Here's my idea:
- (Somehow) detect if a record has no `data-bolt-field` attributes when editing it
- When editing the record, if you're not a developer or root (these will be default permissions), don't show the live editor button if this record has no editible fields
- If you are a developer or root, have the "Live Editor" button, but clicking on it brings up a modal that explains that it needs to be set up, with a link to the documentation explaining how to do that.

> Because we all know what happens when we use regular expressions for that. 

It's been too long since the last time I read that. I used regex for HTML on**ce and ę̶̦͔̫̤̟̓͂̀̓̊̿v̷̧͉̗̭̳̙̾̉̏̾̿̆e̸̪̯̖̹̽̈́͊͜͜͝͝͝r̵̟̣̼͖͉̞̅̿͊̐̈́̈́yţ̶̱̰͎̳̼͑̓͋̓̽͝h̷̨̫̗͖͖͈̑̆̓͋̀̆i̶͕̙͚̻̬̠͊̽̈́͐̈́͝ṋ̵̨̨̛̞̺̲̉̊̎̉̕**g̸̡̥̯͖̹̫̉͂͑̅̍̅ ̴̧̨͚̠̭͚͑̈́̀́͗͝w̴̙͍̻̪̻͉̌̆̏͌̑͌ä̷͖̗͚̣̯̦̈̀̒̚̕s̸̳͚̥̺̬̈́͆̓̽̇͜͠ ̶̥̝̠͇̝͕͐̈̅̇͒̓ḟ̷͍̰͔͕̲͈̈́̀̌͐̕i̷̱̩̟̙̞̤͋̊͌̿̂͘*nę̶̧̗̫̰͉͒̂̈́̀͑̒.*

bobdenotter:
That sounds like a good idea to me! :-)

I'd just like to squeeze in the option to give the user a glimpse of what it might do. I'll think about that.

GawainLynch:
Additional input from the :koala: 

We need something! I just tried on a non-enabled template and spent the next 5 minutes trying to figure out why I couldn't edit any fields… 

Granted I need more :coffee: and should have known better as I am a project contributor… but expecting an average consumer of Bolt — in any of the 3 categories — to understand why that button does nothing, is just going to cause pain for all.

GawainLynch:
Ping @bobdenotter and @Pinpickle. Do we have progress on this?  Beta really, really, soon :wink:

Pinpickle:
@GawainLynch, nothing from my end, apologies! Did we reach a consensus on the best way to fix this, at least for the time being?

GawainLynch:
No consensus past what's here… I was just hoping to motivate all parties to get that discussion happening. Please feel free to expand/highlight any ideas though, they're most welcome.

Also, Bob is away this weekend, but should be back tomorrow.

Pinpickle:
Sounds like a plan.

My current feeling is: magic would be great but needs a lot of thought and effort to make it work properly. For the time being, we just need a way of notifying the theme developer what's going on.

So I propose we go with that suggestion I made before, except without the detection of editable fields (unless someone can think of a resource friendly way to do that). Essentially, all themes, by default have live editing in "limbo mode", and while that is the case, developers can see the live editor button but if the live editor loads, and then there are no editable fields, a warning comes up - in this situation, standard editors can't see the button. At this stage, they can set `liveedit` to true or false in their theme's config.yml. Setting it to false will get rid of the error for good, and setting it to true will enable it for everyone, and stop showing warnings (even if there are no editable fields on the page).

GawainLynch:
I am wondering if we don't set `liveedit: false` as the default and advertise the feature for 2.2 with instructions/docs as to how to enable it completely… with your warnings built-in.

Then for 2.3 we can try to tackle it more completely and change to a true default… Let's see what @bobdenotter has to say tomorrow(ish).

Pinpickle:
I wasn't suggesting false as a default, but some sort of middle ground: true for developers but false for everyone else. That way we won't get frustrated editors with a live editor that doesn't work.

The easier option is probably as you suggested, make it true for everyone, and then fix it up for a 2.2.x release or 2.3 - just so we can get a release out sooner.

But yes, let's wait for the @bobdenotter and his say.

Issue #3303, opened 2015-04-02
Related content only accesible if published


miguelavaqrod:
Hi,

I know that using bolt's backend, you can not assign a related content to a contenttype if it is not published, but, using extensions you can.
Problem arrises when you try to get that kind of related content by means of extensions.
Using the `.related(...)` method, if related content is, for example, `"held"`, you are not able to get it.

`.related(...)` does not have any parameter to bypass status.
However, for example, `getContent(...)` from storage class has it...

Just with a small modification, we can get the best of both worlds...

In Content.php, locate the 'related' function:

```
public function related($filtercontenttype = null, $filterid = null)
```

and replace it with:

```
public function related($filtercontenttype = null, $filterid = null, $status = null)
```

(`null` as default guarantees that it does not break anything...)

A few lines below there is this sentence:

```
$where = array('id' => implode(' || ', $ids));
```

Jut add, below, these ones:

```
if(!empty($status)){
   $where['status'] = $status;
}
```

Now you can use mycontent.related('myrelated', null, '!'), so it bypass the status of the related content.
Obviously, you can filter using a specific status.

Sorry guys, for posting that information this way and do not to generate a pull request or similar.
That's because I do not master github yet and I do not want to break things... so.. if somebody thinks it's interesting and generates that pull request for me.. I promise a beer as reward!.

Thanks.
miguelav

bobdenotter:
Hey, 

The idea is that by default only _published_ content is fetched. Would it not be simpler to just set the ones you're making the 'related' to as 'published' as well? 

Your solution would allow you to get them, but you still couldn't link to them. If you just need the information on that page, why not make the related items just 'published'? 

(i took the liberty of adding some formatting to your issue, for readability ;-) )

miguelavaqrod:
Hi!

Maybe I did not explain myself so well... ;-)

Imagine that situation (a real one, I am working on, so we can have an e-commerce extension):

By means of some "forms" (frontend-based), you get information about an order.
Those forms are processed by an extension, and generates an entry in a predefined contenttype (contenttype 'orders', for example).
However, as that order is not paid, I set it as held.
Once the customer pays (by means of a payment gateway, for example) my extension handles that and set that order to 'published' (so we get a 'green' dot ;-).

However, in a control panel (frontend-based) I designed, so customers can access and check their orders, you need to list all orders (held and published).
That's why I made that change to core code.

Sometimes, you need to show not published records in front-end.. extensions scope is infinite ;-)

[...]

PR #3319, opened 2015-04-05
Fix path for non-ajaxy file-upload.

bobdenotter:
As suggested by @miguelavaqrod in #3303

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -472,9 +472,11 @@ class Content implements \ArrayAccess
                     continue; // Skip 'empty' uploads.
                 }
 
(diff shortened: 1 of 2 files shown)


miguelavaqrod:
Dealing with the same topic... it seems that in that function, variables like:

... $file['name'][0] ...
... $file['error'][0] ...
... $file['tmp_name'][0]...

should not have the leading '[0]'...

And it's a pity that that function only creates new folders based on year-month, instead of respecting what contenttype field definition has in the upload tag (I will try to change that...)

PR #3320, opened 2015-04-05
Allow filtering in 'record.related' on other fields too. Fixes #3303

### LINKED ISSUE #3303: Related content only accesible if published

bobdenotter:


GawainLynch:
I think this might be related to…

PR #3323, opened 2015-04-05
Make backend submenu items mainmenu items if...

Intendit:
Make backend submenu items mainmenu items if they are the only submenu
item.

Not the cleanest solution, or very DRY, but it works and is the best I
can do with modifying half the menu files.

### DIFF
--- a/app/view/twig/nav/_macros.twig
+++ b/app/view/twig/nav/_macros.twig
@@ -29,12 +29,12 @@
 
     {# Only display the 'root' option, if any of the subitems are allowed to be shown. #}
     {% set allowedany = false %}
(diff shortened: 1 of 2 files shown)


app/view/twig/nav/_macros.twig
```diff
@@ -29,9 +29,13 @@
 
     {# Only display the 'root' option, if any of the subitems are allowed to be shown. #}
     {% set allowedany = false %}
+    {% set allowedammount = 0 %}
```
bobdenotter: "Amount" is with one 'm'. ;-)

bobdenotter:
Neat! Thanks! :+1:

PR #3327, opened 2015-04-06
Sub-namespace our Twig classes

GawainLynch:


bobdenotter:
Sure. But why? :-)

GawainLynch:
Oh that's easy… 
1. It fits with the refactoring of stuff, standardisation wise
2. There is a rumour that there might be some Twig stuff on the horizon and it keeps the namespaces obvious.

PR #3334, opened 2015-04-07
Only setting default timezone if config provides it

### LINKED ISSUE #3302: Default timezone to ini setting if not in config

CarsonF:
Fixes #3302


bobdenotter:
Perhaps not ideal, but better than it was! 

close_enough

PR #3342, opened 2015-04-08
Allow explicit setting of a Contenttype's table name suffix

GawainLynch:
…I may have also accidentally done some type hints and PHPDoc.


bobdenotter:
;-)

PR #3343, opened 2015-04-08
Scrutinizer config

GawainLynch:
Apparently we don't have our PHP source files in `app/src/` any longer… who went and did that?

…this should be fun!


bobdenotter:
Yolo!

PR #3344, opened 2015-04-08
slug and geolocation fields refactored

rarila:
One step towards repeatable fields (started now with these two field types).

- No more accessing with id's, instead relative accessing.
- Container and slot (geolocation for now).
- Changed way to bind things to make initialisation on the fly possible

Other improvements:

- goMap library is gone. We now access Google maps lib directly saving us at least 11kB to load (I assume even the code to handle gm is shorter that the code to handle goMap).
- Slug has a new better fitting outfit.
- Geolocation html code streamlined. Now resiszing correctly.
- Lots of other small improvements


bobdenotter:
"Gonemap"!! :clap:

PR #3348, opened 2015-04-08
Feature/bolt login with email

StevendeVries:
At this moment a non unique username is used to login. With this PR I want to add support to login using an email or even replace it with the use of an emailadres.
* [x] Increase the length of the emailfield in the database
* [x] Force the database to have a unique email
* [x] Replace getUser with more sensible functions and thus deprecating it
* [x] Rewrite the loginprocedure to use e-mail (forms and methods)
* [x] Add support for e-mail to reset password
* [x] Replace usages of getUser() with the appropriate new functions
* [x] Add unittest
* [x] Add acceptancetests
* [x] Test everything functional
* [x] Fix other tests which brake on unique users

Note: A note on backwardscompatability, forcing a unique email/username might break if someone already has multiple user accounts with the same e-mail/username. Eventhough they should already be unique.


bobdenotter:
I have never seen this before.. Neat..

screenshot 2015-04-08 22 44 30

GawainLynch:
> I have never seen this (2 of 6) before.. Neat..

GitHUb updates for handling Travis (and Co)…

Speaking of… the test failures are what becomes an out-of-date database that is used by PHPUnit. @rossriley, this makes me nervous to even ask… but should we be calling `nut update:database` on each instantiation of a new `bolt.db`?

GawainLynch:
We're going to need to remove f670745fd0d4a35c56367fbd22455ad3dccb1791 as that will create a mess.

Can you do the following (assuming you have bolt/bolt defined as the remote named 'upstream':
```
$ git checkout master
$ git pull upstream master
$ git checkout feature/bolt-login-with-email
$ git reset --hard 8ff936a
$ git rebase master
```

It will stop at a relevant commit that has a conflict.  Edit the file and fix the conflict then:

```
$ git add path/to/file.php
$ git commit --amend
$ git rebase --continue
```

If that all works, then you can just
```
$ git push --force
``` 

If things go badly you can start over with:
```
$ git rebase --abort
```

bobdenotter:
Sweet! Thanks for the good work, @StevendeVries :+1:

PR #3349, opened 2015-04-08
add selective handlers for looking up user and group id

rossriley:
This more than likely will have a few holes on systems without posix functions, but it should be better than breaking completely.


bobdenotter:
Sweet! :-)

PR #3364, opened 2015-04-11
Scrutinizer review clean ups

GawainLynch:
This could be a bit of a mission… and help from Club OCD would be good.


bobdenotter:
:+1:

PR #3369, opened 2015-04-12
Set the SystemHandler default log level to Logger::INFO

GawainLynch:
Probably should be cherry-picked into release/2.1 too


bobdenotter:
Whooop!

PR #3380, opened 2015-04-15
Fixed action intiialisation for stack modals

timcooper:
This fixes an issue where you would need to open the modal, close it, then open it again before anything worked.

Using "shown" was trying to bind events directly to elements that didn't exist yet, we need to do it after the contents have loaded.

### DIFF
--- a/app/src/js/modules/stack.js
+++ b/app/src/js/modules/stack.js
@@ -40,7 +40,7 @@
         });
 
         $(elements).on(
-            'shown.bs.modal',
+            'loaded.bs.modal',
             function (e) {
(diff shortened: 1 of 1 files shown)


timcooper:
Also this seems deprecated? https://github.com/bolt/bolt/blob/master/app/src/js/modules/stack.js#L31

I can't find anywhere this would be used, that ID only seems to be referred to in JS (and not in a way that would add it anywhere).

bobdenotter:
Thanks! I think the Stack needs some work soon. The UX of the whole thing has room for improvement. :-)

PR #3381, opened 2015-04-15
[RFC] Delete multiple items more efficiently

timcooper:
This PR adds the ability to batch multiple items together for deletion in a single request.

This is achieved by accepting a comma-delimited list of IDs as a URL and function parameter:
`/bolt/content/deletecontent/pages/1,2,3`
`$controller->deleteContent($app, 'pages', '1,2,3')`

Currently, on an overview page if you select `n` items and "Delete selected" it will trigger `2n` HTTP requests, one to delete and one to "refresh" per item. 

This is quite inefficient and could possibly cause browser/server pain depending on the amount of items you're trying to delete.

Batching these deletes together, this process will only ever trigger 2 requests regardless of the number of items being deleted.

Current functionality will not change - making a request with a single ID will simply be treated as a single length array of IDs.

### DIFF
--- a/app/src/js/init.js
+++ b/app/src/js/init.js
@@ -200,18 +200,21 @@ var init = {
                 bootbox.confirm(notice, function (confirmed) {
                     $('.alert').alert();
                     if (confirmed === true) {
-                        $.each(aItems, function (index, id) {
-                            // Delete request
-                            $.ajax({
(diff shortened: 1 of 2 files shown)


bobdenotter:
Nice work! Thank you, @timcooper ! :+1:

Issue #3297, opened 2015-04-01
Adding a "Styles" combo to HTML field's CKeditor


JeromeBreizhtorm:
I'd like to achieve that : 
http://docs.ckeditor.com/#!/guide/dev_styles

but faces a few problems/questions :

First i understand that i have to add a styleSet to my ckeditor field configuration : 
```
body:
            type: html
            height: 300px
            options:
                ckeditor:
                    stylesSet: 'page_styles:path/to/some/styles.js'
```

**Where do i put this JS file ?**
 * in my theme : why not even if configuration shouldn't rely on themes. And when the app is running in a subdir (like mine), you can't rely on an absolute path (unless there is some kind of PATH var i can use in conf files ??). So i can get it to work with something like (which is ugly) : 
  ```
  stylesSet: 'page_styles:../../../../theme/mytheme/js/ckeditor/ckeditor_styles.js'
  ```
 * in app/view/js, so it is readable : but it will be overwritten in the next Bolt update.

Anyway, once it's done, the Styles Combo Plugin should be activated for my styles to be loaded and displayed in CKeditor, as it's not part of the package for the moment.

**How do i load a plugin into Bolt CKEditor ?**
I tried using CKEDITOR.plugins.addExternal and CKEDITOR.config.extraPlugins in the previous path/to/some/styles.js (and the CKeditor object seams fine in the output), but still can't see anything in the editor...

Thanks

JeromeBreizhtorm:
I didn't manage to do what i wanted and had to deal with the basic "Format" styles, binding default styles to those i wanted in the final content. Using CKEditor's *contentsCSS* property will help the end user understand what he's doing :)

Anyway, unless i missed something in activating external CKeditor plugins (which is highly probable) , it might be a great addition to offer a bit more flexibility on those HTML fields configuration.

Intendit:
+1 I'd love to see this capability. Perhaps this should be handled by an extension instead though since it is technically adding code to the backend, which is strictly extension territory IIRC.

rarila:
Stay tuned…

bobdenotter:
This is fixed, right? If not, feel free to re-open.

PR #3384, opened 2015-04-16
Include plugins "Styles Combo" and "Stylesheet Parser" in CKEditor

rarila:
- Include plugins ``Styles Combo`` and ``Stylesheet Parser`` in our CKEditor build.
- Remove moono skin from our repository
- Move boltcke skin to a folder where it won't overwritten on update and make grunt handle it

First part solution to #3297. Next part will be easier integration of CKE config files and perhaps some standardlocation. I'm convinced that it only make sense to configure CKE with CKE way of doing things and we can't build a config wrapper around to make things easy for the not techsavy user. Also configuration of CKE could get big so neither ``config.yml`` nor ``contenttypes.yml`` is the best place to do such configuration, especially as these settings are transfered/parsed on every request and an external config-cke.js could be cached.

Perhaps some well commented standard config file is the way to go. But these hings need some deep thought to make them right. So I hope this helps ppl who know CKE to do things right until a better  system is established.

PS:
There's an undocumented switch ``styles: true`` in ``config.yml`` that works like the other toolbar switches. :warning: But be warned, this one probably vanishes.


bobdenotter:
Sweet! :+1:

PR #3386, opened 2015-04-16
Move the autoloader for testing classes into autoload-dev

GawainLynch:
At best a tiny performance optimisation, at the most obvious… CDO.


bobdenotter:
:8ball: Magic 8-ball says: #1555

PR #3387, opened 2015-04-16
Fix to use title and alt text on image field

Shyim:
This PR adds the ability to use on image field Alt text and title on same time


Shyim:
I have added a rooturl scheme fix, because on a https connection is build the root url with http.

bobdenotter:
Sweet! Thank you, @Shyim! :+1:

PR #3391, opened 2015-04-17
Allow editing of empty files.

SahAssar:
Do not set required html tag on editfile textareas to allow editing of
empty files.


bobdenotter:
:+1: Thanks, @SahAssar

PR #3392, opened 2015-04-17
Consolidate user lookup functions

GawainLynch:
I was looking at an error on password reset where we were calling getUserByEmail() but that wasn't implemented.  Looking further the getUser*() function it seems we can just consolidate them.

@StevendeVries this affects you guys, the core functionality remains unchanged, jut dropping those new functions… can you chime in if you see any issues.

### DIFF
--- a/app/view/twig/activity/changelog.twig
+++ b/app/view/twig/activity/changelog.twig
@@ -57,7 +57,7 @@
                             {{ entry.comment|default('') }}
                         </td>
                         <td>
-                            {% set user = getuserbyid(entry.ownerid) %}
+                            {% set user = getuser(entry.ownerid) %}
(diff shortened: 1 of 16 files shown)


StevendeVries:
@GawainLynch Ah I see we've added that function in our extended Users, you are free to either implement it or to drop it:
```php
/**
     * Get a User by its email
     *
     * @param string $email
     * @return string[]
     */
    public function getUserByEmail($email)
    {
        $users = $this->getUsers();

        foreach ($users as $user) {
            if ($user['email'] == $email) {
                return $user;
            }
        }

        return false;
    }
```

GawainLynch:
Ah cool, this just consolidates that too rather than having 3 lookup functions.

PR #3393, opened 2015-04-17
PSR-2 and styling clean up

GawainLynch:


rarila:
May the OCD be with you!

PR #3397, opened 2015-04-17
Always have a fallback for a timezone when it isn't set in either php.ini or config.yml

### LINKED ISSUE #3377: No timezone in config.yml and php.ini leads to RuntimeException
After upgrading from 2.1.3 to 2.1.6 i got errors from nut commands : 

> Warning: date(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in /myproject/src/Application.php on line 300

> Fatal error: Uncaught exception 'RuntimeException' with message 'Failed to start the session because headers have already been sent by "/myproject/src/Application.php" at line 300

I fixed it by adding a timezone in my config.yml (i had no timezone set in my php.ini either).

Maybe it should default to UTC when no timezone is set.

see https://github.com/bolt/bolt/commit/a4f3e47590ff6c9b38ddc5029e7cc7002017815c

GawainLynch:
Follow up to #3333 & #3334 

Fixes #3377

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -305,16 +305,14 @@ class Application extends Silex\Application
         $this['locale'] = reset($configLocale);
 
         // Set the default timezone if provided in the Config
-        if ($tz = $this['config']->get('general/timezone')) {
-            date_default_timezone_set($tz);
-        }
+        date_default_timezone_set($this['config']->get('general/timezone') ?: ini_get('date.timezone') ?: 'UTC');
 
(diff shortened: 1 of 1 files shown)


CarsonF:
:+1:

bobdenotter:
:clock1:

PR #3404, opened 2015-04-19
Less statics for Database\IntegrityChecker

GawainLynch:
The statics were introduced in 2a5c9d28700ce0eb2e8925426e7f5c522b511d1e for #692. 

The dependency loop that those commits worked around were, in effect, created by overloading the constructor needlessly.

This one is for you, @rossriley :grinning:

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -73,7 +73,8 @@ class Application extends Silex\Application
 
     protected function initConfig()
     {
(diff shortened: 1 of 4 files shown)


bobdenotter:
Green lights, choo choo!:train: :train: :train: :train: :train: :train:

PR #3407, opened 2015-04-19
Further clean up

GawainLynch:
**Notable code change**:
In Users I have just replaced the super globals. This needs attention when we branch for v3. But we're calling Request::createFromGlobals() to get the data out of scope, so let's leverage that in the meantime.

### DIFF
--- a/src/Users.php
+++ b/src/Users.php
@@ -35,11 +35,17 @@ class Users
     /** @var integer */
     private $hashStrength;
 
-    /** @var string */
-    private $remoteIP;
-
     /** @var boolean */
     private $validsession;
+
+    /** @var string */
+    private $remoteIP;
+    /** @var string */
+    private $userAgent;
+    /** @var string */
+    private $hostName;
+    /** @var string */
+    private $authToken;
 
     /**
      * @param \Silex\Application $app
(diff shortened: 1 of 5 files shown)


src/Users.php
```diff
@@ -259,10 +280,10 @@ private function getAuthToken($name = '', $salt = '')
             $seed .= '-' . $this->remoteIP;
         }
         if ($this->app['config']->get('general/cookies_use_browseragent')) {
-            $seed .= '-' . $_SERVER['HTTP_USER_AGENT'];
+            $seed .= '-' . $this->userAgent;
         }
         if ($this->app['config']->get('general/cookies_use_httphost')) {
-            $seed .= '-' . (isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : $_SERVER['SERVER_NAME']);
+            $seed .= '-' . $this->hostName;
```
bobdenotter: The last time we did this, we had to revert it, I seem to recall. Are you sure this won't give any weird issues?

GawainLynch: I think the last time(s) that we did, we tried using `$app['request']` and that's where we ended up out-of-scope.  

I was able to reproduce it in my testing today, and using the data from `Request::createGlobals()` caused the behaviour to be correct(ed), which is logical as it just builds a correct data from the super globals. The downside, as documented by Fabien, is that you miss out on modifying the class data on construct, but that isn't relevant to our need here.

The whole thing needs to move out to middleware, but there are BC implications that we need to handle until we branch v3 and go nuts on this stuff :grin: 

I am happy to drop the patch though. Alternatively, it is confined to a single cset and can easily be reverted in git.

bobdenotter: No, no, no need for dropping anything.. Just wanted to check that it wasn't an oversight. ^_^

bobdenotter:
Whoop whoop!

Issue #3376, opened 2015-04-14
Cannot set "from" email address


svivian:
I've set up the `mailoptions` on my site, but I didn't see any method to specify the email address and name from which emails are sent.

Currently the emails appear to be sent from "Bolt" with my own email address. In Gmail it's listed as "Bolt via mail133-25.atl131.mandrillapp.com". Potentially this could cause problems with spam filters because the email is not actually sent from that address, and some servers may not even send the email without the correct From address.

After a quick look at the code it seems that "Bolt" is hard-coded, for example in `src/Users.php` on line 616:

    ->setFrom(array($user['email'] => 'Bolt'))

At the very least, by default it could use the `sitename` from config. But preferably have options in the mail section e.g.

    mailoptions:
      fromName: My Site
      fromEmail: mysite@example.com

bobdenotter:
Fixed by @shyim in #3409!

PR #3409, opened 2015-04-19
Ability to set a Email Sender Mail in config.yml

Shyim:
Makes the ability to set the sender Mail and sender name in the config.yml

@issue: https://github.com/bolt/bolt/issues/3376


bobdenotter:
Thanks! :+1:

PR #3411, opened 2015-04-20
Move the ExtensionsInfoService to use the Guzzle provider

### LINKED ISSUE #3410: file_get_contents should not be used to fetch remote content
Many webhosts/administrators won't enable allow_url_fopen. It would be significantly better to use cURL instead.

Most noticeable in the Extras with Extensions.

GawainLynch:
Fixes #3410


bobdenotter:
Aaaight. :+1:

PR #3416, opened 2015-04-20
Refactoring getUser

Shyim:
Hey,
i have refactored the getUser function to make it effective when a single user will be getted

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -1211,9 +1211,15 @@ class Backend implements ControllerProviderInterface
         if (!empty($id)) {
             $user = $app['users']->getUser($id);
 
-            // Verify the current user has access to edit this user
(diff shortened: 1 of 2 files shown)


bobdenotter:
It seems your changes caused a few regressions: https://travis-ci.org/bolt/bolt/jobs/59294637#L3460

Could you look into this? :-)

Shyim:
i have no idea why the phpunit fails.
i have diffed the output of the function before the change and after its the same output on every test.

GawainLynch:
I'll have a look for you, I am pretty sure I know what needs to be done it is just fiddly and will probably need to wait until the weekend… someone bug me on Sunday if this isn't fixed :grinning:

GawainLynch:
Pulled locally and tests updated.

PR #3420, opened 2015-04-21
More Contenttype table name fun

GawainLynch:
This is a follow up to #3342… because… :koala:


bobdenotter:
Some :leaves: :leaves: for the :koala:

Issue #3422, opened 2015-04-21
[RFC] IIS installation


hyperTwitch:
Currently there is no easy way to install Bolt on an IIS based setup.

bobdenotter:
Most of us have not much experience with IIS, so we're not in a good position to write docs or provide good configuration settings for it. If you do, you're more than welcome to help us out in this area! :-)

bobdenotter:
Closing, because PR was merged!

GawainLynch:
@hyperTwitch I've submitted #5304 to remove the file from the stable v3 release branch as it no worky-worky in v3 and is creating a support issue we don't have the expertise to handle.

What would be awesome is, if you have the personal bandwidth, move this to a page in [docs](https://github.com/bolt/docs).

PR #3423, opened 2015-04-21
Add IIS web.config

hyperTwitch:
For Issue #3422

Add in IIS URL Rewrite module drop in replacement for Apache .htaccess
file. URL rewriting is handled, as is the woff2 file definition, but I
had to disable the cache access and yml,db,twig,md deny rules as they
interfered with the CMS backend.

WARNING: In order to have this work on IIS you must have the IIS URL
Rewrite module. You will also need to alter the permissions on the files
folder and the PHP extension php_fileinfo will need to be enabled.
Instructions to follow in the documentation repository.

### DIFF
--- a/web.config
+++ b/web.config
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <system.webServer>
+		<defaultDocument>
+            <files>
+				<clear />
+                <add value="index.php" />
+                <add value="index.html" />
+                <add value="index.htm" />
+            </files>
+        </defaultDocument>
+        <rewrite>
+            <rules>
+                <clear />
+                <rule name="Deny cache access" enabled="false" patternSyntax="Wildcard" stopProcessing="true">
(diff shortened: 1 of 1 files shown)


bobdenotter:
Well, that was quick! ;-)

web.config
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <system.webServer>
+		<defaultDocument>
+            <files>
+				<clear />
```
bobdenotter: small nitpick: Seems like the indentation is a bit off, here. :-)

hyperTwitch:
I don't have the time to put together the proper documentation for the site, but when I do I'll update that as well. It will be covering the steps required in the warning. If someone could verify that everything works as it should that'd be appreciated as I've not had much experience with Bolt itself.

bobdenotter:
Thank you, @hyperTwitch :+1:

Issue #3425, opened 2015-04-21
Weird stuff with bolt_session cookies.


bobdenotter:
A short while ago, I had an issue where Bolt would _always_ set a `bolt_session` session cookie. Also on pages in the frontend, and also when you're not logged on. 

I did not want this, because setting a cookie on "static" pages in the frontend will prevent Varnish or Cloudflare from doing their jobs: Every page requests/visitor gets their own cookie.

See here: #3309 / https://github.com/bolt/bolt/pull/3309/files

This code _should_ prevent the `bolt_session` cookie from being set, but _only_ in the frontend. Note:

- "Not setting a cookie" by removing the header that is about to set it, is _not_ the same as _removing_ a previously set cookie. 
- This shouldn't interfere with _other_ cookies being set.
- If you're _not_ logged in, you _don't_ need a `bolt_session` cookie by default. 
- Setting it happens during initialisation of `Application.php`, in `initSession()`. (line 80)

For me, for the project i needed it to be gone, it works as it should. But, there's been more than one report of people seeing stupid shit happening, for it to be a coincidence. We should fix this. Somehow.

GawainLynch:
![cookies](https://cloud.githubusercontent.com/assets/1427081/7260783/a0a6c9a2-e864-11e4-98d4-74f5cc40a968.jpg)

GawainLynch:
Related to #3413

Pinpickle:
I had a good read through this recent discussion and it led me to the conclusion that using `$app['session']` is not good for extensions and sessions in the front end, correct?

I was wondering if there are any docs on how an extension might have its own sessions for simple things such as form flashes. I looked through the Symfony docs and through your code for clientlogin, @GawainLynch, but I couldn't work it out. I think this would be a common usecase, and without docs warning against it, people will probably go immediately to `$app['session']`.

GawainLynch:
Well, frontend sessions/cookies were broken for a short while and fixed in #3427. As it stands for the moment, default behaviour will be OK to use `$app['session']` but you'd be right that it's questionable to rely on it.  What is implemented in ClientLogin is simply a workaround for the behaviour introduced, but is no longer needed.

@CarsonF and I are looking at this for the Bolt responses in general in 2.3… watch this space :wink:

Pinpickle:
Sounds like 2.3 is going to make extensions very happy in lots of ways! I will eagerly watch all of the spaces.

GawainLynch:
[...]

PR #3427, opened 2015-04-21
Make Application::unsetSessionCookie() optional and BC friendly

### LINKED ISSUE #3413: Bolt cookie in frontend
Hi!

I noticed that now, in frontend, bolt session cookie is not set.
Maybe it can be ok for most cases, but I feel that you are not taking into account the use cases that involve extensions.
There are, and there will be, few extensions that make use of that cookie to store relevant info needed to show in frontend (carts, frontend areas that will be only visible to logged in users, and so on...).

Better than just removing it, I think that it should be a choice developers have to select (maybe just in the main config file).

For now, I had to modify Application.php file by hand to avoid that behaviour.

What do you think about?

Thanks...

GawainLynch:
**NOTE:** This is ugly, and a workaround for all involved parties until we nail what is causing this to not react correctly to Symfony Response handling

Related to #3425
Fixes #3413

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -485,6 +485,8 @@ class Application extends Silex\Application
      *
      * Note, we don't use $request->clearCookie (logs out a logged-on user) or
      * $request->removeCookie (doesn't prevent the header from being sent).
+     *
+     * @see https://github.com/bolt/bolt/issues/3425
      */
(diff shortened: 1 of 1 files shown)


bobdenotter:
Mergin' in!

PR #3428, opened 2015-04-21
Change slug label when contenttype is viewless

Pinpickle:
Currently the slug says "permalink" regardless of whether the contenttype is viewless or not. When there is no page for a contenttype, this can be misleading.

This just adds a simple check for a viewless contenttype and changes 'Permalink' to 'Unique Alias' if that check is true.


bobdenotter:
Sweet! Thanks! :+1:

PR #3430, opened 2015-04-22
YAML repeated nodes

GawainLynch:
Technically this isn't limited to Contenttypes, but that is where it makes the most sense and add the greatest amount of value… Calling #1555  on the PHP code change!

Full credit for the idea goes to @rossriley, resulting from [this Gist](https://gist.github.com/rossriley/0a2dc17108ac9f2f13b3)

@bobdenotter if you agree that this is worthy of Bolt, I will follow up with a PR on bolt/docs.

:koala: goes and eats :leaves:

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -182,8 +182,64 @@ showcases:
     icon_many: "fa:cubes"
     icon_one: "fa:cube"
 
+# Using YAML repeated nodes
+#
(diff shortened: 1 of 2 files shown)


GawainLynch:
@bobdenotter if you agree that this is worthy of Bolt, I will follow up with a PR on bolt/docs.

bobdenotter:
Cool!

PR #3432, opened 2015-04-23
add optional filter to select field with contenttype values

rossriley:
Tiny little feature that allows you to specify an optional filter when the values for a select field are pulled from a contenttype.. eg:

    relatednews:
        type: select
        multiple: true
        values: pages/title
        filter: {categories:news}
        autocomplete: true


bobdenotter:
Amazeballs! 

I've added an issue to add some docs for this: https://github.com/bolt/docs/issues/192

PR #3433, opened 2015-04-23
Secondary menu refactoring (JS) + fixes for #2329 and #2347

### LINKED ISSUE #2329: Menu in mobile view has glitch on desktop
To reproduce (ive tried in chromium and firefox on linux):

menu_glitch1

Use a small width on the browser, open the configuration menu by pressing the arrow.
When you scroll to the menu items below that were hidden before scrolling they are white on white.

menu_glitch

### LINKED ISSUE #2347: Bolt on mobile devices
I want to use Bolt on my tablet. For me this is almost impossible. I can only click the main menu items. The submenus become visible, but I can't click the links in it because the main link is loaded.

rarila:
:tada: Merge this PR and you get 2 bugfixes for free! :tada: 

- ``obj-sidebar.js`` and ``obj-navpopups.js`` moved to the new ``modules/secmenu.js``

- Yet two backbone using files less.

- Removed three ``data-actions`` (expand/collapse/toogle). These actions are now simply bound to the buttons on secmenu initialisation

- Fixed #2329 (Menu in mobile view has glitch on desktop). Problem was wrong calculation of height that got visible when the content area was smaller than the menu.

- Fixed #2347 (Bolt on mobile devices). We now open the submenu on click. If the submenu was already opened (hover) we change to the overview page. This way we have almost the same behavior as before. Only the very fast clickers (300ms), clicking before the popover is opened, have to click twice now. I think we can live with that as now double click to open the overview is also supported. We get popover menu support on touch devices with that change and you even can go to the overview there by tapping twice on the menuitem.

@Pinpickle I have removed three lines from ``liveeditor.js`` where the autoheight adjusting is switched off/on. Doesn't seem to be neccessary (the navigation is made unvisible in the stylesheet) and the 3s timeouts won't hurt performance much. Right?


Pinpickle:
That code was left over from another look of the Live Editor, it can go!

Who could argue with some good refactoring?

rarila:
@Pinpickle Ah, ok. Just wanted to be sure :)

> Who could argue with some good refactoring?

Travis can ;-)

bobdenotter:
Huzzah! :beer: :beers:

rarila:
:beer: :beers: Yeah, that fits!Today is __National Beer Day__ in Germany!

PR #3439, opened 2015-04-23
Fixes #3435 by disabling browser XSS protection for file editing 

### LINKED ISSUE #3435: XSS Auditor error when saving file with <form action="">
When editing a twig file in the bolt backend with  `<form action="">` in it, I get the following error:

>The XSS Auditor refused to execute a script in '**path to the file (left this out)**' because its source code was found within the request. The auditor was enabled as the server sent neither an 'X-XSS-Protection' nor 'Content-Security-Policy' header.

The action attribute is what is causing the error, it only happens in files that contain an action attribute. The error only occurs in webkit browsers.

timcooper:
The linked issue #3435 describes the problem well, only thing that might need changing is the implementation.

I went with `after` middleware in the routing section, but it could be done in the method itself (slightly differently due to it returning a redirect response) if that makes more sense/fits better with bolt idioms.


timcooper:
Just noticed a similar bug has been handled before in the preview feature at https://github.com/bolt/bolt/blob/master/src/Controllers/Frontend.php#L194 by @bobdenotter 

Still doesn't quite clear up whether this would be the preferred method of handling it since the routing is handled differently there than in the Backend controller.

bobdenotter:
I think i'd prefer if we did this one way, and did it consistently. I think i'd prefer it, if it were in the body of the method itself, to keep the list of controller definitions a bit more clean, but i _do_ feel that we shouldn't use `header()`, but rather `$response->headers->set('X-XSS-Protection', '0');` like you did.. 

If you can make these two instances more conststen with eachother, i'll gladly merge it in :-)

bobdenotter:
As discussed on IRC: 

 - timcooper: so leave the PR as is? 18:45
 - CarsonF: Yep
 - CarsonF: Was Bopp saying to change the other case to use that as well?
 - timcooper: well he was thinking we could make it consistently set it on the Response in both cases
 - Bopp: i would like them to be the same, if possible
 - CarsonF: Bopp: Does that need to happen on 2.2 or can it wait for 2.3?
 - timcooper: but that's a no go, so to make it consistent for now both would need to be header()s in the method
 - Bopp: yes, let's merge this PR in as is, and fix it in 2.3
 - CarsonF: Cool

PR #3442, opened 2015-04-24
remove the strict typehint and change doc param to match

### LINKED ISSUE #3441: Error when using PRE_DELETE or POST_DELETE StorageEvents
When using the PRE_DELETE or POST_DELETE StorageEvents the following error is produced.

PHP message: PHP Catchable fatal error:  Argument 1 passed to Bolt\Events\StorageEvent::__construct() must be an instance of Bolt\Content, array given, called in /vagrant/vendor/bolt/bolt/src/Storage.php on line 442 and defined in /vagrant/vendor/bolt/bolt/src/Events/StorageEvent.php on line 57

We narrowed it down to the line 421 (inside method "deleteContent") in Storage.php which calls the method findContent() which returns an array. This array is later used to create a new StorageEvent(). The constructor does however expect an Content object and not an array.

This problem is not encountered on the PRE_SAVE and POST_SAVE StorageEvents

rossriley:
fixes #3441

### DIFF
--- a/src/Events/StorageEvent.php
+++ b/src/Events/StorageEvent.php
@@ -51,10 +51,10 @@ class StorageEvent extends GenericEvent
     /**
      * Instantiate generic Storage Event.
      *
-     * @param Bolt\Content $subject   A Content object that is being saved or deleted
-     * @param array        $arguments Arguments to store in the event.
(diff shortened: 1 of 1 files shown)


bobdenotter:
beaker

Issue #3338, opened 2015-04-07
Enhancement: templateselect to use template names instead of file names


JeromeBreizhtorm:
The dropdown for a templateselect field is populated with file names. It would be more user-friendly to use template names.

Chopping the file extension, replacing underscores and ucfirst-ing the filename could be a start, even if it would be even better to have an association name/filename somewhere, like this in contenttypes.yml for example :

```
template:
    type: templateselect
    templates:
        greattemplate:
            name:'A great template'
            filename:'great-template.twig'
        othertemplate:
            name:'Anoter template'
            filename:'bad_naming.twig'
```

Pinpickle:
The theme config would be the best place for defining template names, I think. This would do well to be paired with the templatefields definition.

JeromeBreizhtorm:
Of course ... would be way better in the theme config. +1

GawainLynch:
Addressed in #3447

PR #3447, opened 2015-04-25
Enhancement to define templates for the template chooser in backend

Shyim:
Enhancement to define templates for the template chooser in backend

@issue: #3338

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -218,27 +218,40 @@ class RecordHandler
             return null;
         }
 
-        if ($filter) {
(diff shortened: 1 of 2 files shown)


index.php
```diff
@@ -40,6 +40,12 @@
     }
 }
 
+if (!file_exists('vendor/autoload.php')) {
+    echo '<h2>Error</h2>';
+    echo 'Please execute "composer install" from the command line to install the required dependencies for Bolt.';
+    return false;
+}
+
```
GawainLynch: Not always valid. If someone has installed via an archive and for some reason that file doesn't exist they would get a confusing error.

bobdenotter:
Green lights!! Merging in!

PR #3474, opened 2015-05-02
Fix in Backend controller translation route to correctly handle an array of locales

sintemaa:


sintemaa:
Hmm I fumbled with my branches and accidentally also included an update to the nl_NL translations in this pull request

GawainLynch:
No stress, magic wand being waved here…

PR #3483, opened 2015-05-03
Fixed some broken special chars caused by earlier update & added some mo...

sintemaa:
...re translations

### DIFF
--- a/app/resources/translations/nl_NL/messages.nl_NL.yml
+++ b/app/resources/translations/nl_NL/messages.nl_NL.yml
@@ -1,20 +1,20 @@
-# app/resources/translations/nl_NL/messages.nl_NL.yml – generated on 2014-12-06 09:23:09 UTC
+# app/resources/translations/nl_NL/messages.nl_NL.yml – generated on 2015-05-03 09:07:43 UTC
 
 # Warning: Translations are in the process of being moved to a new keyword-based translation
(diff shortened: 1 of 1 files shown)


sintemaa:
**@GawainLynch** looks like your instructions worked.

However I must admit I don't fully understand what all those commands did, and so which were meant to clean up the mess I made and which I should repeat for future contributions.

If I got it right the approach should be:

1. Make sure my `master` is in line with Bolt's `master`:
    ```
    git checkout master
    git reset --hard upstream/master
    ```

2. Create a new branch
    ```
    git checkout -b branch-name-here master
    ```

3. Do my development on this branch

4. Push my branch to my fork @ Github
    ```
    git push --set-upstream origin hotfix/nl-translations
    ```

5. Make a pull request on the website

Did i get that right?

GawainLynch:
More or less.

Generally, you should now start with something like this:
```
git checkout master
git pull upstream master
git push
```
That will (obviously) checkout `master` and pull Bolt's `master` into your local `master` then push that to your GitHub `master` (`origin master`).

The `git fetch` and `git reset` combination are to firstly "fetch" but not merge ("pull" does both for you) from the target you tell it, and the "reset" if, well, a reset button if every you get your master branch out of order.  Be careful with it, as if you don't know what you're doing you can hurt yourself.

For the most part, if you remember to never commit to master, most of the time you'll never need to worry.

PR #3494, opened 2015-05-06
Added protip for testing email settings.

bobdenotter:


GawainLynch:
Can… happy… to commit… but the original idea for that route was a work-around until we implemented something in the admin area that allowed it to be done more simply… thoughts?

GawainLynch:
My worry being BC

bobdenotter:
When we get around to doing that, we remove it again from the .dist file, imho. 

This trick was _very_ convenient when helping krands debug his issues yesterday (forms wouldn't email, and he didn't know if it was forms, or mail itself that was broken)

CarsonF:
Yeah I worry about BC too. Once the user does first install that comment is in their config forever, even after major version updates.

bobdenotter:
I'm open to better suggestions. ^_^

bobdenotter:
closing this one for now.

PR #3496, opened 2015-05-07
Fix pagination for searching via getContent() or {% setcontent %}.

xiaohutai:
This allows pagination for using search in `Storage->getContent` calls (in Extensions) as well as `{% setcontent ... %}` calls in Twig. I have no idea why a fallback limit of 2000 was picked, but I've kept everything as it was just in case.

There also seems to be redundant params, e.g. `$decoded['parameters']` and `$parameters`. Not sure if there might be any duplicate functionalities, if you compare with the global front-end search in https://github.com/bolt/bolt/blob/master/src/Controllers/Frontend.php#L317

### DIFF
--- a/src/Storage.php
+++ b/src/Storage.php
@@ -1637,15 +1637,13 @@ class Storage
         $offset = 0;
 
         // set correct limit and offset if these are set.
-        if ($limit) {
-            if( isset($decoded['parameters']['limit']) ) {
-                $limit = $decoded['parameters']['limit'];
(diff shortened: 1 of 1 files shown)


xiaohutai:
Wan't paying attention. Fixed the `if` statements.

bobdenotter:
Looks good to me! :-)

bobdenotter:
Merging in! Thanks!

PR #3502, opened 2015-05-09
Because of the `\n` we want double quotes here.

bobdenotter:


GawainLynch:
:koala: is **not** here…

PR #3508, opened 2015-05-11
Changed column type from 'string' to 'text' on templatefields

pedronofuentes:
Template fields allows you to configure extra fields in a template that they can be any field type supported by Bolt. They are stored serialized in a `string` column in the content type table.

It caused that if you configure a field that needs more storage space (for example a `html` field) and you add more content that is supported by the `string` column Bolt can't save.

Changing the column type to `text`solves it as it's the maximum storage type of Bolt's field types.


pedronofuentes:
I saw the failed tests. I'm going to fix it.

bobdenotter:
Cool. Thanks in advance! :-)

GawainLynch:
Easy one for you…
```
cd /path/bot/bolt/
git checkout change-column-type-for-templatefields
cp tests/phpunit/unit/resources/db/bolt.db app/database/
./app/nut database:udpate
cp app/database/bolt.db tests/phpunit/unit/resources/db/
```

pedronofuentes:
Oh! Thank your very much :)

I was looking for the code that generated the database and I didn't realize about the predefined database in tests.

GawainLynch:
We're green… :traffic_light:

PR #3519, opened 2015-05-13
Handle Flysystem exception correctly when a file is not found.

ggioffreda:
Hi,
just a pretty minor fix. I think that the expected behaviour of the backend in case the "extensions" directory does not exist should be to catch the FileNotFoundException exception and show the error "The folder '%s' could not be found, or is not readable." instead of returning a 500 Internal Server Error.

Regards,
Giovanni


GawainLynch:
Looks good to me, thanks @ggioffreda

PR #3526, opened 2015-05-14
Make sure `$unserdata` is an array, and not merely `false`

bobdenotter:
See http://stackoverflow.com/questions/30242572/bolt-getting-blank-screen-after-upgrading-to-2-2beta


GawainLynch:
Hmm…

PR #3529, opened 2015-05-15
[Tests] Adjust cache flushing priority

GawainLynch:


rossriley:
certified #1555 free.

PR #3532, opened 2015-05-16
Translation for .hu additional conf description

rixbeck:


app/config/config.yml.dist
```diff
@@ -234,7 +234,10 @@ wysiwyg:
         allowNbsp: false     # If set to 'false', the editor will strip out `&nbsp;` characters. If set to 'true', it will allow them. ¯\_(ツ)_/¯
         # viewSourceRoles: [ 'admin', 'developer' ] # the roles which are allowed to use the [Source]-button.
 
-# Global option to enable/disable the live editor
+# Global option to enable/disable the live editor.
+# Liveeditor feature requires to include bounding editable html containers in your template
+# the 'data-bolt-field' attribute and a value with the name of the corresponding field.
+# Example: <h3 data-bolt-field="subtitle">{{ record.subtitle }}</h3>
```
GawainLynch: I am unsure about this. @bobdenotter and @Pinpickle are talking about making this dynamic post 2.2.0 and my concern is that this will end up in new install config files and we can't remove it.  This might confuse people.

Thoughts?

bobdenotter: I agree with gawain, we shouldn't include this in the config, where it might be "stuck" long after it's removed. :-)

rixbeck:
Oh, didn't knew that. I've been just looking for how it should be used when I found this option.

rixbeck:
Can you cherry-pick c800227 or should I post a clean PR with just xlation?

GawainLynch:
Happy to cherry-pick.

rixbeck:
alt text

GawainLynch:
Merged into master as commit cd09e404fc22cb2355934bb4b9b4eb41320af4e5

Issue #3452, opened 2015-04-25
[RFC] Choose the admin title of a contenttype


davidgenetic:
I've bumped into a situation where the generated content's title in the administration area was not fitting for some reason. Instead of forcing people to use reserved field names like 'title', 'name', 'caption' or 'subject', it would be nice we could choose the field or format ourselves.

I've prepared a little commit (9b87931ff458b5209c5d59db4c717e6b9826bf82) where I've made the following possible:
```yaml
pages:
    singular_name: page
    fields:
        title:
            type: text
        subtitle:
            type: text
    title_format: "{{ record.title }} - {{ record.subtitle }}"
```
I can make a pull request if you think this is ok.

bobdenotter:
Hi @davidgenetic. Sounds like a good suggestion! Please do!

GawainLynch:
We'll call that "RFC accepted", please PR when ready.

bobdenotter:
Hi @davidgenetic, 

We implemented this a bit differently than you suggested. Hope that's OK with you. see #3635 for details. 

The reasons I wanted to do it differently were: 

- The other solution takes 'excerpt' into account as well. If you use a field as 'title', you probably don't want to see it in the excerpt again. If we left it there, whatever the 'title' was, would show up twice that way. 
- Parsing it as Twig is a bit of a performance impact. So, we'd rather not do that. 

Thanks for the PR, though. Otherwise this still wouldn't be implemented, I guess. ;-)

davidgenetic:
Hi @bobdenotter,
That is AOK with me. Thx!

PR #3540, opened 2015-05-19
choose format for a contenttype's admin title

davidgenetic:
PR for #3452

### DIFF
--- a/app/view/twig/_macro/_macro.twig
+++ b/app/view/twig/_macro/_macro.twig
@@ -15,11 +15,14 @@
 {% macro contentlink(contenttype, content) %}
     {% import _self as crosslinks %}
 
+    {% set title = content.getTitle() %}
+    {#
     {% if content.title is defined %}
         {% set title = content.title %}
     {% else %}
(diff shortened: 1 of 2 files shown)


bobdenotter:
Nice. Shouldn't a sensible default be set for this, somewhere?

jmschelcher:
Very nice, I had the same idea yesterday (#3588) but your way is simplier and more powerful !
Thanks.

app/view/twig/_macro/_macro.twig
```diff
@@ -15,11 +15,14 @@
...
+    {% set title = content.getTitle() %}
+    {#
     {% if content.title is defined %}
         {% set title = content.title %}
     {% else %}
         {% set title = content.name %}
     {% endif %}
+    #}
     {{ crosslinks.contentlink_by_id(contenttype, title, content.id) }}
 {% endmacro %}
```
CarsonF: You can just use `content.title`, twig will call `getTitle()`. No need to even set a variable.

PR #3541, opened 2015-05-19
Make comment line lengths consistent, and break up text to be more readable.

GawainLynch:
Totally calling #1555 on this too.


bobdenotter:
:+1:

PR #3544, opened 2015-05-20
Make addCss() and addJavascript() function declaration and handling consistent

GawainLynch:
There are several reasons that this got overlooked:
* Gnomes
* :koala:'s not having the right :leaves: 
* Coding sober

Depends on #3528

### DIFF
--- a/src/BaseExtension.php
+++ b/src/BaseExtension.php
@@ -517,19 +517,33 @@ abstract class BaseExtension implements ExtensionInterface
     /**
      * Add a CSS file to the rendered HTML.
      *
-     * @param string  $filename File name to add to href=""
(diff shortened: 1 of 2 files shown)


bobdenotter:
Looks like some other commits snuck into this one.

GawainLynch:
Yes, it depends on the 2.3 bits... I tagged it as such, in my defence :grin:

GawainLynch:
Just thinking about it, if you're happy it's low enough risk, I can just rebase against master, amend the PHP comment and it can go into 2.2... Easy. 

I just thought it was late in the game for 2.2, but it would give us consistency of this two functions.

bobdenotter:
Mergin' in!

PR #3549, opened 2015-05-21
Just a typo in german translation

sbani:
In german it's "der" not "de". 
Just look at the other "no-proper-type"


rarila:
While at it: ``Bitte contenttypes.yml bearbeiten dies korrigieren`` doesn't sound that correct either ;-)

sbani:
Yeah sounds strange. I will read the whole file later today

sbani:
Updated the commit with new fixes

rarila:
That's probably because I converted them from "du" to "sie" on the fly a while ago. Someone (can't remember his complete name at the moment) changed from "sie" to the even more formal "Sie". Also ok with me. I decided to use my time to code instead of translating as there are enough German non core devs around now. :)

sbani:
Thanks.

I know my boss will task our working student to translate it. So there will be a translation from an expert.
I think we release next week. We won't have to wait that long :)

rarila:
Perhaps we should add a ``de-de-du``, a  ``de-de-sie``and a  ``de-de-bigsie`` translation ;-)

sbani:
Good idea. I will task her to do that when she's working on it. But we have to wait for release next week minimum.

rarila:
You noticed the ``;-)`` ?

sbani:
:) nope haha just thought: "Ok... as long as she's doing it and not me" :)

PR #3554, opened 2015-05-22
Add a try-exception when parsing `allowtwig` fields.

xiaohutai:
If you make a Twig typo, such as `{{ record.title }]`, then your front-end explodes. I believe it may be better to show the unparsed content. Though this behaviour does need to be documented.

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -908,7 +908,11 @@ class Content implements \ArrayAccess
         if ($allowtwig && preg_match('/[{][{%#]/', $snippet)) {
             $snippet = html_entity_decode($snippet, ENT_QUOTES, 'UTF-8');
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
Thing is, if we merge this in, it could be an 'information disclosure' issue. `{{ dostuff('mypassword') }]` would show the password in the frontend. It's not a very _likely_ scenario, because that code is pretty dumb, but it _is_ a concern nonetheless. 

As @GawainLynch pointed out: "Let's return `$e->getMessage()` instead.". That way it _won't_ break, but instead it will display a readable error message.

bobdenotter:
Merging in! 

(and I'll cherrypick this for 2.2)

PR #3556, opened 2015-05-23
Require Silex WebProfiler ^1.0.6

GawainLynch:
…would you like some derp with that derp, sir?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "rossriley/flysystem53" : "~1.1",
         "siriusphp/upload" : "~1.2",
         "silex/silex" : "~1.2",
-        "silex/web-profiler" : "1.0.4",
+        "silex/web-profiler" : "^1.0.6",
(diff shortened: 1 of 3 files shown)


GawainLynch:
Putting a hold on this, as it requires this [PR to be merged](https://github.com/silexphp/Silex-WebProfiler/pull/60) upstream and a version tag.

GawainLynch:
I've updated this to use a temporary, patched, WebProfilerServiceProvider until upstream patches and releases their currently broken version.

This will get it in and working for now, but we won't go to release (unless Fabien takes a couple of months) with this patched class.

Oh… and, why are we using Scrutinizer again?

bobdenotter:
Merging in.. We should just not forget to revert this when the change is done upstream.

GawainLynch:
You really think @CarsonF would let me forget? :stuck_out_tongue:

PR #3560, opened 2015-05-23
Add reset password command

tvlooy:
This adds a command and functionality to reset a user's password from the commandline.

### DIFF
--- a/src/AccessControl/Authentication.php
+++ b/src/AccessControl/Authentication.php
@@ -338,6 +338,41 @@ class Authentication
     }
 
     /**
+     * Set a random password for user / reset password. Accepts email or username
+     *
+     * @param string $username
+     *
+     * @return string|boolean new password or FALSE when no match for username
+     */
+    public function setRandomPassword($username)
+    {
+        $password = false;
+        $user = $this->app['users']->getUser($username);
+
+        if (!empty($user)) {
+            $password = $this->app['randomgenerator']->generateString(12);
+
+            $hasher = new PasswordHash($this->hashStrength, true);
+            $hashedpassword = $hasher->HashPassword($password);
+
+            $update = array(
+                'password'       => $hashedpassword,
+                'shadowpassword' => '',
+                'shadowtoken'    => '',
+                'shadowvalidity' => null
+            );
+            $this->app['db']->update($this->getTableName('users'), $update, array('id' => $user['id']));
+
+            $this->app['logger.system']->error(
+                "Password for user \"{$user['username']}\" was reset via Nut.",
+                array('event' => 'authentication')
+            );
+        }
+
+        return $password;
+    }
+
+    /**
      * Handle a password reset confirmation
      *
      * @param string $token
--- a/src/Nut/UserResetPass.php
+++ b/src/Nut/UserResetPass.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Bolt\Nut;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Nut command to reset a user password
+ */
+class UserResetPass extends BaseCommand
+{
+    /**
+     * @see \Symfony\Component\Console\Command\Command::configure()
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('user:reset-pass')
+            ->setDescription('Reset a user password.')
+            ->addArgument('username', InputArgument::REQUIRED,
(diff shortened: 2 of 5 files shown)


GawainLynch:
Cool. A couple of points though.

1. This is more intended for open review, but I am wondering if this command shouldn't have two additional features, firstly a confirmation prompt and secondly a `--no-interaction` option for scripted use.
1. Can we go for "password" over "pass" for consistency
1. The authentication functions are being moved to `src/AccessControl/Authentication.php` when @CarsonF and my [branch](https://github.com/bolt/bolt/compare/master...CarsonF:feature/controller-refactor) lands later tonight

With all of the above, I am more than happy to give whatever help you might need :grin:

tvlooy:
I wouldn't add a confirmation. People who use the cli should know what they are doing. Eg: if you use something like sqlite or rm -rf, it also doesn't ask for confirmation.

Changed everything to "password".

I have a question, do we log the resets to logger.system or not?

GawainLynch:
> I wouldn't add a confirmation. People who use the cli should know what they are doing.

Should but don't always. It adds a safety net in the case where you meant to reset your own password and while thinking what beer to buy Gawain typed:

```
./app/nut user:reset-password gawain
```

CarsonF:
Its not like it locks you out. It tells you what the new password is.

GawainLynch:
And if the account isn't yours? Are you going to email me a plain text version of my new password that you unexpectedly changed? Your local security auditor will show you the exit door.

Destructive operations should always come with confirmation, and alternatively an override for the confirmation, e.g. As per the stated counter point, try either `rm -r *` or `rm -rf /`, the former asks for confirmation, the latter overrides that confirmation, except the it adds a second confirmation to help you double check your apparent lack of sanity.

tvlooy:
You could say sorry I accidentally reset your password, you will need to reset your account from the login page (forgot password function). That e-mail contains a plain text password too by the way.

CarsonF:
Ok good point Gawain. It could ask for  confirmation normally. And not for
the no interaction option.

tvlooy:
actually rm -rf / doesn't ask permission, it warns you. And rm -rf /* removes everything. Just tried it in a vagrant box :-)

So I would for example ask for confirmation when you are about to delete your last admin user. But not for password reset.

Just checked the Symfony FOSUserBundle and it also doesn't ask confirmation for the fos:user:change-password command.

GawainLynch:
> That e-mail contains a plain text password too by the way.

It does, and there is zero guarantee that email is properly set up... But the debate gets academic at that point.

What is not academic is that destructive operations **must** have a default sanity check, and, optionally, an override.

tvlooy:
If you require this, I'll add it. But if you ask my opinion, I don't see this as destructive.

tvlooy:
Also we are now talking about rm but to compare the right commands, passwd never asks for a confirmation either.

bobdenotter:
I'd also be in favor of a confirmation, and the `--no-interaction` option.

GawainLynch:
> Also we are now talking about rm but to compare the right commands, passwd never asks for a confirmation either.

Yes, and unless you are the root user, you can only change your own password.

We are not trying to be UNIX, we cater very specifically to 3 different groups, usually most are lacking a great deal of technical skill as well. Our [manifesto](https://docs.bolt.cm/manifesto) sets it out.

GawainLynch:
Oh and the missed answer to your question about logging, yes please!

Something along the lines of:
```php
$this->app['logger.system']->error('Somebody reset the password for Tom via Nut.', array('event' => 'authentication'));
```

tvlooy:
ok, so we all made our point and I think I'm the only one against it. So, I'll add the option this evening, no problem :+1:

GawainLynch:
OK, we're merged… Send `setRandomPassword()` to `src/AccessControl/Authentication.php` and we should be good to go.

You'll either need to rebase, or start a fresh branch depending on how good your git-fu is… either way, call out if you need help.

tvlooy:
Done! Also added some docs https://github.com/bolt/docs/pull/210

GawainLynch:
:+1:

stefantalen:
Has this been implemented yet? Probably not, because I can't use the command in v2.2.11:

```
$ php app/nut user:reset-password
[InvalidArgumentException]                     
  Command "user:reset-password" is not defined.  
  Did you mean this?                             
      user:add
```

tvlooy:
it's in git (merged) but not in v2.2.11, not sure why ...

GawainLynch:
Basically it's a bit of work to backport and nobody has stepped up… It, of course, will be in 2.3.

PR #3561, opened 2015-05-24
Request cache fix

aheart:
Render.php checks if the user is logged in incorrectly on line 173.
$this->app['users']->getCurrentUsername() returns null for users that are not logged instead of an empty string.


GawainLynch:
Nicely caught, thank you!

aheart:
Glad I could help, good luck!

Issue #3516, opened 2015-05-12
HTML template fields are not visually updated after live edit


Pinpickle:
But it looks like their inner text area is. Tested this in beta 3.

I think this is because the code [here](https://github.com/bolt/bolt/blob/master/app/src/js/modules/liveeditor.js#L197-L199) checks if there is a ckeditor instance with the field name - but ckeditor stores instances by ID. When there are no template fields, this doesn't make a difference, but template fields have different IDs to their name.

bobdenotter:
Pinging @rarila Do you know how to fix this?

rarila:
Probably … if someone can point me to some HTML template fields documentation, so that I can find out what they are for and how they are used :-)

GawainLynch:
https://docs.bolt.cm/building-templates#template-specific-fields

Pinpickle:
@rarila the fix is really simple. The code [here](https://github.com/bolt/bolt/blob/master/app/src/js/modules/liveeditor.js#L197-L199) (I've fixed the link now) simply needs to be changed from 

```
if (_.has(ckeditor.instances, fieldName)) {
    ckeditor.instances[fieldName].setData(fieldValue);
}
```

To

```js
var fieldId = field.attr('id');
if (_.has(ckeditor.instances, fieldId)) {
    ckeditor.instances[fieldId].setData(fieldValue);
}
```

Untested, but should work as long as every field has an ID.

rarila:
@Pinpickle That will work for now but with repeating fields all inputs will lose their ids anyway (I have removed them from ``geolocation`` already)

Otherwise to make ``<textarea name="fieldname[]" …`` work with cke there has to changed quite some code later anyway, if it is possible at all. But I like to avoid numbered ids if possible…

Pinpickle:
@rarila ah I didn't know that was going to be the case. So it would be Bolt's job to keep a `ckeditor.instances` like array that contains references to the ckeditors based on the field names?

On another note, if you're dropping IDs for the fields, how are we going to provide the `for` attribute for the labels to maintain easy accessibility for screen readers and the like?

bobdenotter:
Ok, so should we do the fix as proposed by @Pinpickle for 2.2, and sort it oult for repeating fields later? Or do you have a beter suggestion, @rarila ?

rarila:
@bobdenotter I would say fix it for now and lets see what 2.3 gets us.

rarila:
[...]

PR #3562, opened 2015-05-24
Change section in templatefields example to html

Pinpickle:
As HTML is one of the most problematic fields when it comes to template fields, it makes sense that we have it here for testing.

Helps #3516 

Ping @rarila


bobdenotter:
Merging in.

PR #3565, opened 2015-05-25
Remove guzzle 3

GawainLynch:


src/Application.php
```diff
@@ -34,9 +34,6 @@ public function __construct(array $values = array())
         $values['bolt_name'] = 'alpha 1';
         $values['bolt_released'] = false; // `true` for stable releases, `false` for alpha, beta and RC.
 
-        /** @internal Parameter to track a deprecated PHP version */
-        $values['deprecated.php'] = version_compare(PHP_VERSION, '5.4.0', '<');
```
bobdenotter: Shouldn't we keep this around, only bump it to 5.5.x, or whatever Bolt 3 will be?

GawainLynch: We could… it was originally intended just for this…

bobdenotter:
Merging in!

PR #3572, opened 2015-05-25
HTML template fields updated after live edit. Fixes #3516

### LINKED ISSUE #3516: HTML template fields are not visually updated after live edit

bobdenotter:
Thanks, @Pinpickle

### DIFF
--- a/app/src/js/modules/liveeditor.js
+++ b/app/src/js/modules/liveeditor.js
@@ -194,8 +194,9 @@
             if (fieldType === 'html') {
                 fieldValue = $(this).html();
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
Link to #3516 for the lazy.

PR #3574, opened 2015-05-25
On "Save and return to overview", save and return to overview.

### LINKED ISSUE #2735: Issue 'Save and go to overview' is still there in 2.0.6 RC2
I just installed 2.0.6 RC2 and I noticed that the issue with saving entries is still there. This prevents us from going live with our website.

When I want to save an entry and immediately go to the overview of entries, Bolt stays in the edit window. When I use this option again, Bolt brings me to the overview page. I always have to do this twice to make it work. This does happen with both buttons (right top corner and left bottom) and also in different browsers.

GawainLynch:
Fixes #2735

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ before_script:
   - git clone https://github.com/bolt/composer-install.git .tmp
   # - cd .tmp
   # Add requires
-  - composer require --working-dir=.tmp/ --no-update "bolt/bolt:dev-master" "phpunit/phpunit:~4.5" "codeception/codeception:*" lstrojny/phpunit-function-mocker
+  - composer require --working-dir=.tmp/ --no-update "bolt/bolt:2.2.*@dev" "phpunit/phpunit:~4.5" "codeception/codeception:*" lstrojny/phpunit-function-mocker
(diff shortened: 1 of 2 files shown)


bobdenotter:
This breaks the current behaviour. If you have more than one page of entries, go to any page but the first, in the overview. Or filter on a taxonomy or keyword.. Then, edit that entry and click 'save and return', and you'll get returned to the overview.

GawainLynch:
Figs are climbing the wall…

GawainLynch:
OK… so updated version calls abort() on editContent() when `editreferrer` is set. Local testing has this returning to the specified page with query intact, e.g. pager or filter.

The added commit is required as the :koala: was too busy eating :leaves: when `release/2.2` was branched to remember that the Composer test requires us to stay on the `release/2.2` branch and not the new heavily divergent master.

bobdenotter:
Whoop! Merging in!

GawainLynch:
I know one user that's going to be quite happy :+1:

gwbonline:
That's me!

GawainLynch:
:grin:

Issue #2129, opened 2014-11-07
Better Extension access for getContent()


rossriley:
There's a slightly unintuitive situation at the moment where it's impossible to call $app['storage']->getContent() in the initialize() block of an extension. 

It seems that there's an unnecessary call to $this->app['request'] in the organizeQueryParameters() method, and this throws an exception if you call storage->getContent() before the $app->run() has taken place.

Ideally we should make sure that the Storage class works independently from the http request, but if not then we could perhaps formalise a workaround for use by extension devs.

For example @GawainLynch has a workaround in use here: https://github.com/GawainLynch/bolt-extension-opengraph/blob/master/Extension.php

bobdenotter:
Should go hand in hand with #691

GDmac:
> becomes an academic debate
Initialise is called on EVERY request, including front-end, nothing academic about that.
I think it gets time to look into the dispatcher functionality #1914 to call out for action and hooks

GawainLynch:
> nothing academic about that.

If you read what I wrote in the preceding sentence, you'll see it is at the point that I defined it.

GDmac:
Ok, sorry. I also like ross' suggestion: "make sure that the Storage class works independently from the http request". This effort could go hand in hand with modularisation of user and permission classes, fields, etc.

GawainLynch:
This close-worth now @rossriley?

rossriley:
yes

Issue #687, opened 2013-11-12
Refactor Storage.php to separate concerns.


bobdenotter:
Perhaps use an ORM? Should we do this only for Storage.php? Or should we abstract out all database access?

dominikzogg:
@bobdenotter it's not ready atm, but i am working on a model generator for doctrine orm. It's not meant for initial build only, its mean't to rebuild the entities on each configuration change. There are much to less tests, and a yaml to php configuration converter is missing, but this could be a possibility to use doctrine2 orm with nearly the same configuration as now.

https://github.com/saxulum/saxulum-model-generator

dominikzogg:
@bobdenotter I refactored a few things, added some tests, doc still missing https://github.com/saxulum/saxulum-entity-generator

PR #3578, opened 2015-05-26
Storage refactor

rossriley:
Ok, lets do this.

Phase One of the new storage engine, for an overview of the philosophy there's much more in the new docs: https://docs.bolt.cm/storage/introduction

### :warning: The Usual Caveats

May contain: :koala: :koala: 

Whilst hopefully stable, in that the new system shouldn't interfere or break any existing code, it's probably not wise to build anything too strictly to this API since the next phase involves refactoring Bolt internals to take advantage of the new API, as we do so it's very likely that both the API and the implementation may change.

There's also quite a few field specific formatting / cleaning code that still has to be implemented on the new field types, so it's not 100% ready to replace the old storage class as yet.

### DIFF
--- a/src/Entity/Content.php
+++ b/src/Entity/Content.php
@@ -1,14 +1,13 @@
 <?php
 namespace Bolt\Entity;
 
-use Bolt\Entity\Entity;
 
 /**
- * Entity for User.
+ * Entity for Content.
  */
 class Content extends Entity
 {
-    
+    protected $_contenttype;
     protected $id;
     protected $datecreated;
     protected $datechanged;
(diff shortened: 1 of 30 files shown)


bobdenotter:
awwweesssooommee

GawainLynch:
Do we consider this as closing #687 & #2129?

…I'm off to chase that damn :koala:

rixbeck:
Cooool!

src/Entity/Content.php
```diff
@@ -0,0 +1,45 @@
...
+    protected $id;
+    protected $datecreated;
+    protected $datechanged;
+    protected $datepublish = null;
+    protected $datedepublish = null;
+    
+    public function getDatecreated()
+    {
+        if (!$this->datecreated) {
+            return new \DateTime();
```
GDmac: The new DateTime() is not stored?

PR #3580, opened 2015-05-26
Symfony Forms session handling

GawainLynch:
This completes the explicit session handling issues.

Where a session isn't started, and a CSRF is still required, we will adjust the cookie path that `Session` uses to be specific route for the page…

:koala: power!

### DIFF
--- a/src/EventListener/FormListener.php
+++ b/src/EventListener/FormListener.php
@@ -54,21 +54,23 @@ class FormListener implements EventSubscriberInterface
      */
     public function onFormPreSetData(FormEvent $event)
     {
-        if (!$this->session->isStarted()) {
-            // Enable route specific sesion cookies, generally speaking for front end
-            $storage = new NativeSessionStorage(
-                [
-                    'name'            => $this->tokenName,
-                    'cookie_path'     => $this->request->getRequestUri(),
-                    'cookie_domain'   => $this->config->get('general/cookies_domain'),
-                    'cookie_secure'   => $this->config->get('general/enforce_ssl'),
-                    'cookie_httponly' => true,
-                ],
-                $this->handler
-            );
-
-            $this->session = new Session($storage);
+        if ($this->session->isStarted()) {
+            return;
         }
+
+        // Enable route specific sesion cookies, generally speaking for front end
+        $storage = new NativeSessionStorage(
+            [
+                'name'            => $this->tokenName,
+                'cookie_path'     => $this->request->getRequestUri(),
+                'cookie_domain'   => $this->config->get('general/cookies_domain'),
+                'cookie_secure'   => $this->config->get('general/enforce_ssl'),
+                'cookie_httponly' => true,
+            ],
+            $this->handler
+        );
+
+        $this->session = new Session($storage);
     }
 
     /**
--- a/src/EventListener/FormListener.php
+++ b/src/EventListener/FormListener.php
@@ -1,11 +1,14 @@
 <?php
 namespace Bolt\EventListener;
 
-use Silex\Application;
+use Bolt\Config;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\NativeFileSessionHandler;
 use Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Form/FormEventTypeExtension.php
```diff
@@ -0,0 +1,64 @@
+<?php
+namespace Bolt\Form;
```
CarsonF: Can this go under a Type namespace? `Bolt\Form\Type`

CarsonF:
You have to redo the entire FormSP?

src/Form/FormEventTypeExtension.php
GawainLynch: Yeah, easy

GawainLynch:
> You have to redo the entire FormSP?

At the end of the day, not really. There was just more and more that I had to override I thought it made sense to put that one in our court… much like we did with Whoops. That said, I can move the overrides back into Application if you want.

src/EventListener/FormListener.php
```diff
@@ -0,0 +1,83 @@
+<?php
+namespace Bolt\EventListener;
```
CarsonF: I know EventSubscribers have nothing *specifically* to do with Kernel events, but currently everything in this namespace does. Maybe this would be less confusing in `Bolt\Form\EventListeners`?

CarsonF:
> At the end of the day, not really. There was just more and more that I had to override I thought it made sense to put that one in our court… much like we did with Whoops. That said, I can move the overrides back into Application if you want.

Bleh no, definitely not back to Application. I had a thought thought though the other day. Instead of extending the original SP (like we do with Twig), we just check for service and call it. 
```php
public function register(Application $app) 
{
  if (isset($app['form.factory'])) {
    $app->register(new SymfonyFormServiceProvider());
  }
  ...
}

File: src/EventListener/FormListener.php [hunk 3]
```diff
@@ -0,0 +1,83 @@
...
+            $storage = new NativeSessionStorage(
+                [
+                    'name'            => $this->tokenName,
+                    'cookie_path'     => $this->request->getRequestUri(),
+                    'cookie_domain'   => $this->config->get('general/cookies_domain'),
+                    'cookie_secure'   => $this->config->get('general/enforce_ssl'),
+                    'cookie_httponly' => true,
+                ],
+                $this->handler
+            );
```
CarsonF: It would be better to pull this from DI

src/EventListener/FormListener.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Handle the form event at the beginning of the Form::setData() method.
+     *
+     * @param FormEvent $event
+     */
+    public function onFormPreSetData(FormEvent $event)
+    {
+        if (!$this->session->isStarted()) {
```
CarsonF: Can we return early here?

src/EventListener/FormListener.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Constructor function.
+     *
+     * @param Session                       $session
+     * @param Request                       $request
+     * @param Config                        $config
+     * @param NativeFileSessionHandler|null $handler
+     * @param string                        $tokenName
+     */
+    public function __construct(SessionInterface $session, Request $request, Config $config, $handler, $tokenName)
```
CarsonF: It would be better to pass in `RequestStack` instead of `Request`

CarsonF:
It just seems like there's a lot of code in SP that we don't need to retype. Whoops was different because we *were* changing the entire SP. It doesn't look like that's the case here.

src/EventListener/FormListener.php
GawainLynch: It was, but that involved passing in $app

src/EventListener/FormListener.php
GawainLynch: Ugh… what was I thinking… on to it!

src/EventListener/FormListener.php
GawainLynch: Not sure I agree… It's a namespace that can/will grow… unless we make the namespace `Bolt\KernelEventListener` then it's make more sense.

src/EventListener/FormListener.php
```diff
@@ -0,0 +1,83 @@
...
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\Session\Storage\Handler\NativeFileSessionHandler;
+use Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage;
+
+/**
+ * Symfony Forms listener.
```
CarsonF: I'm lost as to what this class is doing. If session isn't started, it's creating a separate one? Where does `$this->session` get used?

src/EventListener/FormListener.php
CarsonF: Ugh yeah....I guess that would be valid use case. Lemme think about it.

CarsonF: Can the whole session object just be passed in from DI? That would remove the need for Config and Request/RequestStack.

src/EventListener/FormListener.php
CarsonF: Maybe sub-namespaces then. I just see it getting cluttered very fast. What about `Bolt\EventListener\Kernel` and `Bolt\EventListener\Form`? Hmm...but that's the same as having `Bolt\Providers` which I don't like either. I would rather the logic be grouped together by topic.

src/EventListener/FormListener.php
GawainLynch: Done

src/EventListener/FormListener.php
GawainLynch: Yeah, that's one we differ on to some extent.

CarsonF: Ah ok. Well what do you think about Kernel and Form subnamespaces under EventListener?

src/EventListener/FormListener.php
GawainLynch: OK, so the problem is this. If you're going to build a form using the DI:
```php
$this->app['form.factory']->createNamedBuilder($formname, $type, $data, $options)
    ->getForm();
```

That's all fine. The problem is that when the form is built, it accesses the current session, which by default sets the session cookie path to `/`, which applies it over the whole site…

So you need to pass a new storage provider to session, and that needs to have the `cookie_path` set to the relative URL you want the session cookie to apply to…

Now the fun part, Symfony Forms requires you to use `$builder->addEventSubscriber()` as unlike kernel events, they're not "on" by default.

GawainLynch:
Closing this as we have a better solution in another branch now.

PR #3582, opened 2015-05-27
Session cookie handling at logon/logoff

GawainLynch:
Currently a session is started on loading of the login page, and this sets a session cookie, and that cookie isn't cleared at logoff.

This addresses this limitation meaning that a system user, once logged out, will get full caching from Varish & Co.

### DIFF
--- a/app/view/twig/_sub/_messages.twig
+++ b/app/view/twig/_sub/_messages.twig
@@ -8,7 +8,7 @@
 
 {% macro messages(key, class) %}
     {% set shown = [] %}
-    {% for msg in app.session.flashBag.get(key) %}
+    {% for msg in app['logger.flash'].get(key) %}
         {% if msg not in shown %}
             <div class="alert alert-{{ class|default(key) }}">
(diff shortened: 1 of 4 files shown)


CarsonF:
Shouldn't the cookie be cleared with the `Session`?

GawainLynch:
> Shouldn't the cookie be cleared with the Session?

You really should step back… right… now :tongue:

CarsonF:
Shouldn't sessions be :bomb: :boom:

GawainLynch:
Yeah, It's queued figuring it would encourage you to, you know, fix it…

CarsonF:
It's the codeception tests that scare me

GawainLynch:
Yeah, actually that is what stopped me and made me thing this was dependent on our other work… but still valid as a separate (rebased) PR… they are separate issues, but related to how we're not currently doing this well.

It can (obviously) just be worked around with a query parameter, but that made me want to get :leaves: out for the :koala:

GawainLynch:
Moving this into the new session branch

PR #3585, opened 2015-05-27
Performance improvement for registering nut commands

CarsonF:
The whole point of `NutServiceProvider::addCommand` was to make it easier to register nut commands **without** invoking/creating every command.

Currently if I have even a _single_ extension that adds a _single_ command, **ALL 21** (20 core + 1 extension) commands will be created for _every_ web request even though they aren't used.

Pimple's extend and share syntax is very verbose (fixed in v3.0) and this shortcut was to help with that.

Even though the method was static it still used the app object passed in; not a static variable (like `Library` or `ResourceManager`). It had no negative effect to unit tests.

----

Perhaps the purposed change could be a compromise?

### DIFF
--- a/src/BaseExtension.php
+++ b/src/BaseExtension.php
@@ -682,9 +682,6 @@ abstract class BaseExtension implements ExtensionInterface
      */
     public function addConsoleCommand(Command $command)
     {
-        $this->app['nut.commands'] = array_merge(
-            $this->app['nut.commands'],
-            [$command]
-        );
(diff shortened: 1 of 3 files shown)


GawainLynch:
:shipit: says "Nuts are tasty"

:+1:

GawainLynch:
Yoda Squirrel has reviewed our Nut…

bobdenotter:
:shipit:, indeed!

PR #3595, opened 2015-05-29
Replace deprecated symfony/locale with intl. Remove loading intl stubs

CarsonF:
`symfony/locale` component is deprecated and just calls to `symfony/intl`, requiring that component instead.

Stub files are autoloaded [here](https://github.com/symfony/Intl/blob/master/composer.json#L38-L39).


CarsonF:
This should probably be backported too

GawainLynch:
:koala: says :leaves: are tasty

PR #3602, opened 2015-05-30
System Configuration Checks

GawainLynch:
This implements an extensible suite of system tests.

The UI is pretty basic at the moment and I would very much appreciate help from a UI/JavaScript competent contributor to carry this forward.

The route, when hit, calls and `{{ dump() }}`s the decoded responses from the async routes so:
Ssytem Check UI

The email checker, if it passes, will send an email to the current user's registered email address like so:
Email check message

In case it's not obvious, there is no menu for this right now, but you can access the route at `/bolt/checks`.


bobdenotter:
Whoa, pretty cool stuff!

bobdenotter:
Whoop! Merging in!

PR #3633, opened 2015-06-05
[2.2] Extend SSL/TLS Handling 

### LINKED ISSUE #3575: BUG: Warning from PackageManager.php
Warning: Creating default object from empty value in src/Composer/PackageManager.php on line 556

When using 2.1.8 I get the above warning when checking for extensions. This has been untested in the latest version.

### LINKED ISSUE #3628: Extensions HTTPS fallback not working
When running bolt on a server without valid CA certs for cURL, the extensions page fails to load extensions. The following error is currently produced:

Testing connection to extension server failed: cURL error 60: SSL certificate problem, verify that the CA cert is OK. Details: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed
https://extensions.bolt.cm/ is unreachable.

The check should fallback to http so that hosts without a valid cert can still use extensions.

GawainLynch:
Fixes #3575
Fixes #3628

### DIFF
--- a/src/Composer/PackageManager.php
+++ b/src/Composer/PackageManager.php
@@ -17,6 +17,7 @@ use Guzzle\Http\Exception\CurlException as CurlException;
 use Guzzle\Http\Exception\RequestException as V3RequestException;
 use GuzzleHttp\Exception\RequestException;
 use Silex\Application;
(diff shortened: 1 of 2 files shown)


bobdenotter:
Merging in!

PR #3634, opened 2015-06-05
Two more overrides for symfony components that got bumped to v2.7.0

bobdenotter:


GawainLynch:
Passes locally… sending in!

PR #3640, opened 2015-06-06
Extension Composer events

GawainLynch:
* Change the namespace(s)
* Events now run for CLI installs/uninstalls
* No more copying events class to `extensions/`


bobdenotter:
Whoop, whoop! merging in!

PR #3644, opened 2015-06-06
Fixes the overlay of the Symfony Icon in our debug toolbar.

bobdenotter:
Symfony 2.7 switched from a bitmap to an SVG. Fix the CSS accordingly.


GawainLynch:
Yay! Wondered what that was!

PR #3645, opened 2015-06-06
remove extraneous entity value

rossriley:


GawainLynch:
This is somewhat more dangerous than, but close to, #1555… Nah, check that, #1555 was more dangerous!

PR #3650, opened 2015-06-07
Update composer dependencies

GawainLynch:
Commit messages explain pretty much everything. 

DBAL was locked to 2.5.1 in #2517 to maintain PHP 5.3.3 support… which is kind of invalid now :grin:


bobdenotter:
:+1:

PR #3651, opened 2015-06-07
Moar cleaning

GawainLynch:
* Added a tiny bug fix for the upload controller


src/Application.php
```diff
@@ -268,10 +268,7 @@ public function initProfiler()
 
     public function initLocale()
     {
-        $configLocale = $this['config']->get('general/locale', Application::DEFAULT_LOCALE);
-        if (!is_array($configLocale)) {
-            $configLocale = [$configLocale];
-        }
+        $configLocale = (array) $this['config']->get('general/locale', Application::DEFAULT_LOCALE);
```
bobdenotter: Does this work always? i did not know that..

GawainLynch: `'foo'` will be cast to `['foo']`
`['foo']` will stay `['foo']`
`1` will be cast to `[1]`
`true` will be cast to `[true]`
…and so on

When I benchmarked it for 10,000,000 iterations the difference in performance was significant. :grin:

bobdenotter: I use this a lot.. Faster, and three lines less of code. Awesome. :-)

bobdenotter:
Merging in!

PR #3660, opened 2015-06-10
Fixes bug in populateItemFromRecord

### LINKED ISSUE #3655: contenttype/slug as path in menu.yml -> menuitem isn't populated with data from record
Hey everyone,

i ran into following problem: I use routings for single record overrides (as described in https://docs.bolt.cm/templates-routes)

For example (routing.yml):
```
mypagebinding:
    path: /some/path
    defaults:
        _controller: Bolt\Controllers\Frontend::record
        contenttypeslug: page
        slug: mypage
    recordslug:     page/mypage
```
Now i would expect that if i set the 'path' attribute for some menu item to 'page/mypage', that the link in the resulting record has the value, that i specified in my routing (i.e. '/some/path'). I think this would be more consistent, as all generated links for a content-type behave like this (e.g. in a listing of pages).

sbani:
**Partly**(!) fixes #3655 
Thanks to @hartigdan!
And sry that I said you are wrong.

### DIFF
--- a/src/Helpers/MenuBuilder.php
+++ b/src/Helpers/MenuBuilder.php
@@ -177,11 +177,11 @@ class MenuBuilder
 
         if ($content) {
             if (empty($item['label'])) {
-                $item['label'] = !empty($content->values['title']) ?: '';
+                $item['label'] = !empty($content->values['title']) ? $content->values['title'] : '';
             }
 
             if (empty($item['title'])) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
Awesome @sbani :+1:

bobdenotter:
Other part of the fix is in #3661

bobdenotter:
Also merged into 2.2. Thanks, @sbani

PR #3635, opened 2015-06-05
Implement title_format

bobdenotter:
A slightly more robust implementation of the idea put forth in #3540. This also takes into account filtering it out from the 'excerpt' snippet, and it also allows you to concatenate fields: 

```
    title_format: [ firstname, lastname ]
```

If merged in, I'll port it to master / 2.3 

**EDIT:** Closes #3540

PR #3662, opened 2015-06-10
`content.TitleColumnName()` is an array now. Additional fix to #3635

bobdenotter:


CarsonF:
Why is TitleColumnName an array now?

PR #3666, opened 2015-06-11
Improve MenuBuilderTest to check $item[link]

sbani:
- If $content is empty link comes from $item
- If $content is set link comes from content

This difference has been added


GawainLynch:
More awesome :+1:

PR #3671, opened 2015-06-12
cache translation from yaml to cachedir

GDmac:


src/Application.php
```diff
@@ -300,6 +300,8 @@ public function initLocale()
             ['locale_fallbacks' => [Application::DEFAULT_LOCALE]]
         );
 
+        $this['translator.cache_dir'] = $this['resources']->getPath('cache/trans');
```
CarsonF: This could just be added to the array above, next to `locale_fallbacks`

GDmac:
The unit tests fail, even on my local computer. Some tests seem to mess with app/cache during tests.

Response test testString
Fatal error - cannot rename tmp/somefile -> cache/trans/somefile - permission denied

during tests the cache dir gets mangled, and the trans dir looses X (execute) bits ???
when i &lt;exclude>LibraryTest.php&lt;/exclude> the test runs without fatal errors.
However i cannot reproduce when i only run librarytest and responsetest isolated.
Nasty to debug.

Please fix the testsuite, IMO setting app['translator.cache_dir'] should not break fatally.

GDmac:
Extra: There also seems to mess something with the environment. - Translator cache should try to write to the cache dir and then rename the file. It is using ``tempnam()`` for this. Tempnam() falls back to the system temp dir if it cannot generate a temporary file(name). All tests seem to do this correctly (var_dumping dirname, filename, tempname), but just before the error, there seems to be something wrong with the dir

```
PHP Fatal error:  An exception has been thrown during the rendering of a template ("Cannot rename "/tmp/catalogue.en_GB.86a89c0c402748e28181a27efaaffba1dc2c8014.phpkZhUhG" to "/home/travis/build/bolt/bolt/app/cache/trans/catalogue.en_GB.86a89c0c402748e28181a27efaaffba1dc2c8014.php".")
```
https://travis-ci.org/bolt/bolt/jobs/66530694

GawainLynch:
The memory use is PHPUnit… it gets worse when you turn on code coverage. Your 486-DX2 is going to have a bad day fur shure!

The funky bit of this puzzle is I set the translation caching directory to `/home/gawain/tmp/trans` and the tests pass. The bigger worry for me is that something is trying to write to `/tmp` as this will break, badly, for anyone who's provider/admin has enforced `open_basedir`

GawainLynch:
So, the problem in the end was the `Library::saveSerialize()` was doing a somewhat random `umask()` on completion and that was breaking things in weird ways, c.f. https://github.com/bolt/bolt/commit/bb9706d3769c9eb9c1d9c864008234836b65855e

PR #3696, opened 2015-06-15
add hydrating find commands to the base repository

rossriley:
This adds two nice little helper methods that hydrate the results fetched from a QueryBuilder object.

Bonus points for the first person to use.. `$repo->findOneWith($theUniverse);`


GawainLynch:
:koala: eats his :leaves: and says :+1:

PR #3697, opened 2015-06-15
Add array access to base Entity

rossriley:
Allows hydrated entities to have property access via `$entity['val']` as well as `$entity->val`


CarsonF:
So `offsetExists` has different logic than `has`?

rossriley:
Yes, has in an internal method and it handles the dynamic fields, that is the ones that aren't setup as properties on the entity.

offsetExists is a check to see if the accessor returns a value and covers both properties and dynamic fields, it's also there for the compatibility with ArrayAccess.

PR #3699, opened 2015-06-16
Fix force logic on checkLocalAutoLoader

timcooper:
`$force` defaults to false and currently isn't ever set otherwise, the conditional in there meant it will always his that return and never check the autoloaders.


GawainLynch:
Travis… Some days… Some days… Merging.

PR #3702, opened 2015-06-17
Fix paste issue with fileuploader

timcooper:
Client reported this one, they were trying to copy from Excel into input fields (title etc.) but on paste fileuploader was kicking in thinking they were trying to upload something and popping dialogs up.

There is a similar fix in place already over here: https://github.com/bolt/bolt/blob/release/2.2/app/src/js/upload-files.js#L188

### DIFF
--- a/app/src/js/bind-fileupload.js
+++ b/app/src/js/bind-fileupload.js
@@ -11,6 +11,7 @@ function bindFileUpload(key) {
         .fileupload({
             dataType: 'json',
             dropZone: $('#dropzone-' + key),
+            pasteZone: null,
             done: function (e, data) {
                 $.each(data.result, function (index, file) {
                     var filename, message;


timcooper:
I think this is the right approach for a 2.2 hotfix, however I would suggest for 2.3/master that the underlying fileuploader lib is updated instead. In the latest version this option defaults to null.

GawainLynch:
> I would suggest for 2.3/master that the underlying fileuploader lib is updated instead

:+1: I've wanted this for ages myself as I have a couple of things I have to work around myself.

rarila:
I thought I already did here in my overhauling of src/app/ I have in the works here. But it seems it's still the old one. I probably add that to that PR as it seems we can move it from lib to bower now.

PR #3706, opened 2015-06-17
[2.2] Local extensions autoloader updates & tests

GawainLynch:
Same fixes as #3705


GawainLynch:
Standard 2.2.x bug in Codeception… everything else passing.

PR #3708, opened 2015-06-18
Allow custom news sources

SahAssar:
Compatible with bolt/jsonaccess

**Edit:** related to https://github.com/bolt/docs/pull/223

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -235,10 +235,18 @@ accept_file_types: [ twig, html, js, css, scss, gif, jpg, jpeg, png, ico, zip, t
 #
 # Keep the path simple: lowercase only, no extra slashes or other special
 # characters.
+#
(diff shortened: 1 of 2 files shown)


bobdenotter:
Have you tested this to see what happens in the case of various failures: 

- If the server doesn't respond?
- If the server returns a 404 or 5xx response?
- It the JSON is malformed?

SahAssar:
In case of a 404 an error is logged: "Error occurred during newsfeed fetch"
In case of invalid JSON an error is logged: "Invalid JSON feed returned"
In case of a connection timeout an error is logged: "Error occurred during newsfeed fetch"

All of these scenarios seem to respond the same as what would happen if news.bolt.cm had a similar issue, I tested them with my modification and without it, using hosts to simulate the same behavior with news.bolt.cm.

SahAssar:
Closing to reopen on a proper branch.

PR #3710, opened 2015-06-18
[2.2][Travis] Update compser require versions to cope with changes in Comp…

GawainLynch:
…oser


GawainLynch:
:koala: eats :leaves: and :see_no_evil: nothing…

PR #3716, opened 2015-06-20
Feature/storage improvements

rossriley:
A few minor refactors to improve runtime modification of the storage components.

1. Remove the indexby parameter for QueryBuilder, this adds unnecessary complexity for Bolt which will stipulate that collections are indexed by id only
2. Move repositories to a new namespace `Bolt\Storage\Repository`
3. Adjust the `createQueryBuilder` method to allow empty aliases, in SQL terms this allows both `SELECT u.* FROM user u` as standard along with `SELECT * FROM user` if an empty alias is supplied.
4. Remove hardcoding of `new ContentRepository` from `$app['storage']` instead this is setup in the Storage Service Provider as a factory, which importantly means that users can swap out for any repository class they like.


GawainLynch:
Merged locally to ninja some whitespace and that should've closed this… but yeah :-/

PR #3722, opened 2015-06-20
Fix incorrect namespace in mapping

rossriley:


bobdenotter:
That was quick :-)

GawainLynch:
It never happened… :wink:

PR #3723, opened 2015-06-20
Allow selection of homepage template based on content

### LINKED ISSUE #3721: Templateselect is ignored when used on the homepage
## Scenario:
page/1 is the homepage record, and has a templateselect field. By default this field is ignored when visiting the homepage and it uses homepage_template as defined by config.yml or the theme instead. I want the client to be able to change the template used (without editing configs), and therefore want it to use the template selected in the templateselect field.

## The issue:
When selecting the homepage template bolt only checks general/homepage_template and theme/homepage_template before it falls back to index.twig.

## Proposal:
Instead of falling back to index.twig as a last resort, use the template defined by the record (first checking templateselect and then using record_template). This would allow for better use of template specific fields for homepages.

## Temporary workaround:
Using routing.yml one can change the homepage routing to:

    homepage:
        path: '/'
        defaults:
            _controller: 'Bolt\Controllers\Frontend::record'
            contenttypeslug: 'page'
            id: '1'

SahAssar:
Fixes #3721. Only used if there is no homepage template selected.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -853,7 +853,6 @@ class Config
             ],
             'sitename'                    => 'Default Bolt site',
             'homepage'                    => 'page/*',
-            'homepage_template'           => 'index.twig',
             'locale'                      => \Bolt\Application::DEFAULT_LOCALE,
             'recordsperpage'              => 10,
             'recordsperdashboardwidget'   => 5,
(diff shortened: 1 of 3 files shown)


bobdenotter:
Thanks! Will merge manually into `release/2.2`.

PR #3737, opened 2015-06-24
fix fullpath file hashing

jkazimir:
This fixes an asset hashing bug that made itself known as nice big flashy warnings in the live editor while running under the php server.  Turns out the `substr` wasn't doing too much there.

Looks like the koala slipped on some wet :leaves:!

Cheers!


GawainLynch:
:koala: :gun:

PR #3739, opened 2015-06-24
Update Integrity Manager

ntomka:
The database platform's method getCreateTableSQL creates only indexes default, however foreign keys also can be added by the documentation: http://doctrine-dbal.readthedocs.org/en/latest/reference/schema-representation.html
We just have to add the second parameter to getCreateTableSQL call.

This also affects stable branches. I can make PRs to them/it as well if you accept this.

Source and documentation for AbstractPlatform for quick reference: http://www.doctrine-project.org/api/dbal/2.4/source-class-Doctrine.DBAL.Platforms.AbstractPlatform.html#1010

### DIFF
--- a/src/Storage/Database/Schema/Manager.php
+++ b/src/Storage/Database/Schema/Manager.php
@@ -235,7 +235,7 @@ class Manager
 
                 /** @var $platform AbstractPlatform */
                 $platform = $this->app['db']->getDatabasePlatform();
-                $queries = $platform->getCreateTableSQL($table);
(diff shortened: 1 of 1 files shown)


GawainLynch:
I haven't like seen you all day bro!

I think this fixes something I am seeing in another branch, but LGTM

ntomka:
@GawainLynch Yep,it's been a long time since I was here. :D

bobdenotter:
:+1:

GawainLynch:
Thanks mate, just took a while do double check and catch @rossriley when we weren't otherwise distracted with the storage layer refactor. :+1:

ntomka:
:beers:

PR #3747, opened 2015-06-27
cleanup imports

jkazimir:
Some :leaves: for the :koala:.

Cheers!

### DIFF
--- a/src/Storage/Database/Schema/CheckResponse.php
+++ b/src/Storage/Database/Schema/CheckResponse.php
@@ -1,8 +1,6 @@
 <?php
 namespace Bolt\Storage\Database\Schema;
 
-use Doctrine\DBAL\Schema\Column;
-use Doctrine\DBAL\Schema\Table;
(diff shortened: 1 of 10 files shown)


src/Storage/Database/Schema/CheckResponse.php
```diff
@@ -1,8 +1,6 @@
 <?php
 namespace Bolt\Storage\Database\Schema;
 
-use Doctrine\DBAL\Schema\Column;
-use Doctrine\DBAL\Schema\Table;
```
GawainLynch: Doesn't this one break PHPDoc?

jkazimir: Nope. The only doc references are either TableDiff, or FQCNs.

GawainLynch: **Note to self:** `grep '@param Table' | wc -l` also counts `TableDiff`

PR #3748, opened 2015-06-27
Update .travis.yml

silentworks:
Speed up TravisCI test process.


GawainLynch:
Nope, can't do. We set up web servers and databases to test all of our install configurations.

PR #3750, opened 2015-06-28
Updated grunt modules, libraries and … bye, bye IE8!

rarila:
Most notably changes:
- All node modules updated to latest version.
- All bower modules updated to latest version.
   - Notably here that we now use latest **Bootstrap 3.3.4**.
   - Notably also the switch to **jQuery 2.1.4**.
- Some dependencies moved from node to bower so that only tools are now handled by node.
- Special patches for IE8 removed :tada:, **minimum IE** is now **9**.
   - No two stylesheet versions anymore.
- Switch to **postcss** for css postprocessing.
   - We now use **autoprefixer** with  ``"last 2 versions, > 5%, IE >= 9"`` browser setting.
   - We pack same CSS media query rules into one media query rule.
   - We now use ``csswring`` now minifiy instead of ``cssmin``.
- ``grunt-sass`` replaced by home brew version because of problems upstream.
- All sourcemaps now land in ``view/maps`` (if generated).

I recommend deleting folders ``bower_components`` and ``node_modules`` before updating as AFAIK now unused packages aren't removed automatically.


GawainLynch:
I am getting a scroll bar on the bottom of the screen with this PR:
screenshot from 2015-06-29 06-05-40

GawainLynch:
Footstrap fix coming in a later PR…

PR #3752, opened 2015-06-29
Fix a couple of storage bugs which were causing issues on Postgres

rossriley:


GawainLynch:
Tested locally… Screw Travis.

PR #3761, opened 2015-07-03
Helper/Str.php non-strict makeSafe() can has Uppercase

GDmac:
Don't forget to ``composer update``
Makes use of the new cocur/slugify 1.2 feature to preserve uppercase chars.
Also added Tests.


GawainLynch:
Bonus points for the added tests! Good stuff mate!

PR #3764, opened 2015-07-03
local extensions readme and config

jadwigo:
this was itching me - it should probably be DRY but this is better than nothing

it adds the json vars for the version, readme, and configfile on local extensions


src/Composer/PackageManager.php
```diff
@@ -269,10 +269,13 @@ public function getAllPackages()
                 $packages['local'][] = [
                     'name'     => $json['name'],
                     'title'    => $ext->getName(),
+                    'version'  => 'local',
```
GawainLynch: Ha! I never thought of something so obvious! :+1:

jadwigo:
I can't even program ..

PR #3767, opened 2015-07-03
Replace Storage hardcoded instantiation with container reference

fabschurt:
Hi folks,

I have a project requirement which has me redefine the `storage` service in `Bolt\Application`, for filtering results globally based on some criteria.

The fact that a `Bolt\Storage` standalone object is instantiated here bypasses the service in the container, so in my opinion it had to be fixed.

Is there a particular reason why the `storage` service wasn't used here, especially when Twig can get access to the container via `$context['app']`?

### DIFF
--- a/src/Twig/SetcontentNode.php
+++ b/src/Twig/SetcontentNode.php
@@ -20,7 +20,7 @@ class SetcontentNode extends \Twig_Node
 
         $compiler
             ->addDebugInfo($this)
-            ->write('$template_storage = new Bolt\Storage($context[\'app\']);' . "\n")
+            ->write('$template_storage = $context[\'app\'][\'storage\'];' . "\n")
             ->write('$context[\'' . $this->getAttribute('name') . '\'] = ')
             ->write('$template_storage->getContent(')
             ->subcompile($this->getAttribute('contenttype'))


GawainLynch:
> Is there a particular reason why the storage service wasn't used here

Well, I think the simple answer [is in here](https://github.com/bolt/bolt/tree/v0.8.4/app/src/Bolt)… this is from some of the very early Bolt code before we had service providers… and it currently Just Works&trade;… So I would guess that, for the most part, it hasn't been looked as since 2012 :grin:

fabschurt:
I must admit that I was betting on something like that :sweat_smile: I'm glad it's that way, because I was worried that there actually was a legitimate reason for not using the container here. Until this is merged and released, I'll use some trickery to make it work.

GawainLynch:
>  Until this is merged and released, I'll use some trickery to make it work.

FYI: The v2.3 release is going to be a little while in the making, we're doing some pretty cool refactoring work and giving Bolt a whole lot of love along the way. Potentially of great interest to you is that one of the big features in 2.3 is going to be a massive overhaul of the storage layer.

But you sound like you've got a workaround strategy in hand.

fabschurt:
Many thanks for the information, indeed I believe I've seen some ORM stuff going on in the `master` branch, and I think it'll be really cool! Is `Bolt\Storage` planned to be removed (or have its API modified) as the main data-fetching gateway or is it gonna stay around? In any case I'll just stick to my dirty workaround for now, hoping to use a cleaner solution in the future.

GawainLynch:
> Is `Bolt\Storage` planned to be removed (or have its API modified) as the main data-fetching gateway or is it gonna stay around

`Bolt\Storage` will remain until 3.0-dev for backward compatibility… But over the course of 2.x we'll move more and more data to object handling. There has been a large amount of core tables done already, and @rossriley is working on some of the `getContent()` parts at the moment (once he finished some more unit tests for the existing functionality), but how much makes it to 2.3 is hard to say right now as there is a lot to do.

fabschurt:
Okay, thanks for the thorough explanation :smiley:

As a side note, this is my first time using Bolt, and it's for a professional project. The project is almost done and I can say I don't regret my choice one bit! I was searching for a replacement for WordPress, built on solid foundations and with modern standards, and with Bolt I can say I found it.

Cheers from France :fr:

PR #3774, opened 2015-07-10
Added canonical setting test for Frontend

sbani:
Checks if the canonical in relation to config and url schema. Added this because the bahaviour could change.


bobdenotter:
Sweet, thanks @sbani!

PR #3785, opened 2015-07-11
Local extension autoloader enhancements

GawainLynch:
* **[General]**  If vendor/autoload.php is missing, include LowlevelException.php manually.
* **[General]** Added a very early shutdown function to pre-catch errors that occur before the application object is built. Hopefully less *White Screens of Death*

* Added a nut command to force the rebuild `./app/nut extensions:autoloader`
* When a fatal error occurs in an extension that is a result of a missing class, we will try to rebuild the autoloader and then either redirect to the intended page, or display the full error:
screenshot from 2015-07-11 15-59-41

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -50,15 +50,18 @@ return call_user_func(
             );
         }
 
-        // Register a PHP shutdown function to catch fatal error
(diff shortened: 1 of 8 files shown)


bobdenotter:
:+1:

sbani:
:+1:

PR #3786, opened 2015-07-11
[2.2] Local extension autoloader enhancements

GawainLynch:
* **[General]** Added a very early shutdown function to pre-catch errors that occur before the application object is built. Hopefully less *White Screens of Death*

* Added a nut command to force the rebuild `./app/nut extensions:autoloader`
* When a fatal error occurs in an extension that is a result of a missing class, we will try to rebuild the autoloader and then either redirect to the intended page, or display the full error:
screenshot from 2015-07-11 15-59-41

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -50,17 +50,18 @@ return call_user_func(
             );
         }
 
-        // Register a PHP shutdown function to catch fatal error
-        register_shutdown_function(array('\Bolt\Exception\LowlevelException', 'catchFatalErrors'));
+        // Register a PHP shutdown function to catch early fatal errors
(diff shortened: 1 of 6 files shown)


GawainLynch:
@StevendeVries and @sbani this should address both of the problems you were seeing during the week

bobdenotter:
Cool stuff!

GawainLynch:
The usual Codeception bug… Otherwise GTG from my end.

bobdenotter:
Cool. merging in!

StevendeVries:
Hats of to you guys, a real pro-active approach! Karma + ☺

sbani:
nice! I like!

PR #3789, opened 2015-07-12
Don't flush output buffers for tests

GawainLynch:
This restores our unit tests to a full 100% pass with a green light!


bobdenotter:
Whoop!

PR #3792, opened 2015-07-13
Send "New Bolt site" e-mail upon first user creation only

fabschurt:
Currently in `2.2`, the **New Bolt website** e-mail is sent to any new user that gets created in the backend. This message should only be sent to the root account e-mail address upon first user creation, as it's just a confirmation that the `mailer` service is working correctly, which normal users don't care about, and don't really have to know about.

What's more is that the Twig template used for generating the message content doesn't use translation, and I don't really feel like explaining to my French-speaking clients why they're receiving unintelligible e-mails in English when I create their account in the backend :sweat_smile:

I'm pull requesting this against `release/2.2` as I consider this behavior a bug.

As far as I can tell, the problem is solved in `master`.

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -1422,28 +1422,30 @@ class Backend implements ControllerProviderInterface
             } else {
                 $app['logger.system']->info(Trans::__('page.edit-users.log.user-added', array('%user%' => $user['displayname'])), array('event' => 'security'));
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
It is that damn :koala: again!

fabschurt:
I knew that :koala: couldn't be trusted!

PR #3793, opened 2015-07-13
Add custom sidebar groups

### LINKED ISSUE #3783: [RFC] - Sidebar Customization
The primary objective of this would allow ‘custom’ categories on the left sidebar which group content inside each (vs. forcing that specific content type as the only option). Need to ensure roles are also respected for the display of these ‘custom categories’. 

Team currently plans on building it out as an extension.

SahAssar:
Fixes #3783 as @rossriley suggested.

### DIFF
--- a/app/view/twig/nav/_macros.twig
+++ b/app/view/twig/nav/_macros.twig
@@ -19,7 +19,7 @@
  # @param bool    wide
  # @param array   subitems
  #}
-{% macro submenu(icon, label, popoveritems, active, wide, subitems) %}
+{% macro submenu(icon, label, popoveritems, active, wide, subitems, force_submenu) %}
 
     {% from _self import label, icon %}
 
@@ -43,7 +43,7 @@
 
     {# Show stuff! #}
     {% if allowedany %}
-        {% if allowedamount == 1 %}
+        {% if allowedamount == 1 and not force_submenu|default(false) %}
             {% set item = allowedsingle %}
             <li>
                 <a href="{{ item.link }}">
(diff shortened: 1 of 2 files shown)


rarila:
Hmmm, the thing I don't like here is [``__(submenuName)``](https://github.com/SahAssar/bolt/commit/6478ad4000667cdab30ca7d89e025c2f5175882d#diff-815e7b1eee5d0c2c68c1e3d6ad859c40R64).

It should be in a more structured way using keybased translation from the beginning as it can't be changed back as other translations as it's userdefined. Had not much time to think about what would be the best place to put but it should look more like ``__('contenttypes.menu.anothermenu')``. I also don't know if this would be better located in ``contenttypes.xx_XX.yml`` instead of ``messages.xx_XX.yml``. Difficult to say.

SahAssar:
You're right, that is a bit sloppy. How about skipping the translation until we have a good place/solution for these kinds of translations? I always hated `contenttypes.xx_XX.yml` since it overrides changes in contenttypes.yml.

I'm not even sure @rossriley even intended for it to be in his example... :flushed:

bobdenotter:
I think this one needs some work. If I set _one_ contenttype to `show_in_menu: false`, it's moved to the bottom, but still shown, without a menu. Like 'Showcases' in this screenshot: 

screen shot 2015-07-22 at 13 30 59

Once I change two contenttypes to have `show_in_menu: false`, it's OK: 

screen shot 2015-07-22 at 13 31 22

However, if i then add a `show_in_menu: Pompidom` on a third contenttype, it breaks again. No expanding menu, and the Icon is changed to a generic one. 

screen shot 2015-07-22 at 13 32 57

Intendit:
Fixed per @bobdenotter suggestions

bobdenotter:
Works like a charm!

PR #3777, opened 2015-07-10
Big backend ui refactoring, part 1

rarila:
First of all, sorry for that not really exemplary  PR :-)

What was done sofar:
- Fixing of (allmost) all boostrap errors (slug has one remaining)
- Fix of scrollbar-problem from #3750
- New and (hopefully) lovely checkbox
- Fieldsets are now generated in a new way
- All fieldsets can now be easily styles separately
- Main labels styled (grayblue for now, just to see them)
- New (smaller) infopopup: :information_source: instead of [Info]
- Cleanup of fieldset html here and there, e.g. video field
- Removal of some inline styles
- Moving of some templates and translations from fieldsets to global (making them unique)
- Some filestructure reorganisation
- Fixing of some minor bugs

Still a long way to go (removing remaining inline styles, inline javascript, …), but I thought it's time to push it before it get's even longer.

Known minor errors (just noticed and hopefully fixed soon):
- Left margin of Relationships/taxonomy wrong
- Inline labels should also be bold.

PR #3799, opened 2015-07-13
Big backend ui refactoring, part 2

rarila:
- Many bugs mentioned in #3777 fixed
- Taxonomy, Relationships and Meta now handled as fieldsets
- Direct rendering of fields without looping (You now can add fields to very groups, not just meta)
- Direct loading of templates without looping
- Light horizontal line between fieldsets
- Smarter output of headlines in tab mode

Again still a long way, but coming nearer …


bobdenotter:
Good stuff! :+1:

Issue #3553, opened 2015-05-22
Load user translation files to extend core translations


sbani:
If you use bolt/composer-install, the translation files are placed in vendor dir, but should be placed in /app/ where the configs are placed.
An even better way is to let the user overwrite the current translations with his files in his app dir.

Else, the an composer update will overwrite the translation files

rarila:
Having those in config/ is on my list since at least 2.0. The place where they are now should as dist files then. Problem is there's much more code that needs some love than spare time :-(

rossriley:
There's a fairly simple workaround for this while we're waiting for native support, if you have a service provider in your app, (or don't mind adding to the bootstrap file some similar code you can do this..)

    public function boot(Application $app)
    {
        $path = $app['resources']->getPath('root/app/resources/translations/en_GB');
        $app['translator']->addResource('yml', $path.'/overrides.en.yml', 'en_GB', 'messages');
    }

Then you can override anything in the messages.yml file and it will take precedence over the default bolt translations.

Something like this, for instance in `app/resources/translations/en_GB/overrides.yml`

     page:
        dashboard:
            title: "Control Centre"

bobdenotter:
> "Translation files are placed wrong"

Just stating for the record.. Because you would like them in a different location, that doesn't mean it's currently _wrong_.

fabschurt:
Nice trick @rossriley, thanks! So far, I was overriding `ResourceManager::getPath()` in a subclass, which was far from ideal:

```php
public function getPath($name)
{
    if (preg_match('#^app/resources/translations/[a-z]{2}(?:_[A-Z]{2})?$#', $name, $matches)) {
        return parent::getPath("root/{$matches[0]}");
    }

    return parent::getPath($name);
}
```

Anyway, the problem here is not that translation files are placed _wrong_ or anything, it's just that Bolt by default doesn't load other translation resources than the core resources. :+1: for the upcoming native support, even though the workaround is clean and easy.

fabschurt:
Would #3800 do the trick?

SahAssar:
Closing since this was fixed in #3800, no?

If that is not the case please feel free to reopen! :)

sbani:
Thanks @SahAssar and @fabschurt. Very busy these times I'm sorry!

PR #3800, opened 2015-07-14
Make translations overridable

fabschurt:
See #3553 for the initial discussion.

I've added a test case to validate this new functionality, and also 2 other test cases covering the existing behavior of `TranslationServiceProvider`.

Let me know what you think about this.

### DIFF
--- a/src/Provider/TranslationServiceProvider.php
+++ b/src/Provider/TranslationServiceProvider.php
@@ -46,23 +46,34 @@ class TranslationServiceProvider implements ServiceProviderInterface
      */
     public static function addResources(Application $app, $locale)
     {
-        // Directory to look for translation file(s)
-        $transDir = $app['resources']->getPath('app/resources/translations/' . $locale);
+        // Directories to look for translation file(s)
+        $transDirs = array_unique([
+            $app['resources']->getPath("app/resources/translations/{$locale}"),
+            $app['resources']->getPath("root/app/resources/translations/{$locale}"),
+        ]);
 
-        if (is_dir($transDir)) {
+        $needsSecondPass = true;
+
+        foreach ($transDirs as $transDir) {
+            if (!is_dir($transDir) || !is_readable($transDir)) {
+                continue;
+            }
             $iterator = new \DirectoryIterator($transDir);
             /**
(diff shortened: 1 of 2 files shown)


rossriley:
Looks good to me :+1:

fabschurt:
Scrutinizer seems angry, I'ma check if I can do something about it.

GawainLynch:
Scrutinizer is just getting upset about the cyclomatic complexity of `addResources()` due to the internal loop.

fabschurt:
Ah, right, so… I don't think there's actually something we can do about that, as 2-level iteration is required here AFAIK.

rossriley:
Yes ignore scrutinizer, it just needs some :beers: and to chill out more.

rarila:
Better let it continue complaining and drink the :beers: yourself :)

fabschurt:
Yeah it kind of reminds me of a grumpy old man, never satisfied about anything :older_man:

fabschurt:
(Just force-pushed a modification to a DocBlock description.)

GawainLynch:
For the sake of the academic conversation… the best approach is generally to break the inner loops out into methods as that makes them easier to examine in a debug context.

With this, that's largely just :cow:

fabschurt:
Yeah I guess the cyclomatic complexity could've been reduced by using `array_filter()` and `array_walk()` for example; it'll be something for me to keep in mind in the future. The use of the `foreach`/`if`/`continue`/`break` idiom is kind of a widespread bad habit in the PHP ecosystem, but the language doesn't exactly make the use of internal iterators very developer-friendly, as opposed to Ruby for example.

GawainLynch:
Hashmaps for the win! The performance gain is quite noticeable for **hot paths**… For every thing else readability of code is often a bigger concern. 

Personally, I like to split logic out and keep functions small as dump()ing a value returned is much easier than dozens of iterations… But that is just my OCD and most certainly not a project requirement.

fabschurt:
I feel you mate, don't even get me started on _my_ OCDs :smile:

rarila:
I more and more get the feeling that Bolt is a collecting basin for OCD affected people.

Issue #3787, opened 2015-07-11
Record editing lookups failing


GawainLynch:
Fresh master install with a pre-filled database:
![relationships](https://cloud.githubusercontent.com/assets/1427081/8634635/b2a9ad18-27f7-11e5-8588-dea7d94a0ba5.png)
![selectentries](https://cloud.githubusercontent.com/assets/1427081/8634636/b2c0e42e-27f7-11e5-96b7-1f0699bd01d3.png)

bobdenotter:
Yup, looks broken. ^_^

bobdenotter:
The correct results are in the rendered HTML.. I think it's the select/autocomplete that's having issues. 


![screenshot 2015-07-12 10 56 15](https://cloud.githubusercontent.com/assets/1833361/8637174/d4753fe2-2884-11e5-8713-0d7b7146523e.png)

GawainLynch:
Yeah, that'd make sense as the dropdown is sized to a larger content set

bobdenotter:
No idea why it's not working yet, but i'm going to dig in.

GawainLynch:
Fixed during the closing scene of *Refactor Wars Episode VI: Return of the Koala* #4189

PR #3806, opened 2015-07-15
Big backend ui refactoring, part 3

rarila:
Here we go again …

- Bugfix: datetime fields work again
- Inline styles eliminated
- Inline Javascript moved to the js part
- Logic moved from twig to php
- Many cleanups
- …

Now all select2 javscript is in the appropriate place so that the next step can be making all select2 based components work again. (see #3787)


GawainLynch:
Tests passing locally… Sending in.

PR #3811, opened 2015-07-16
Minor fixes for database export command

rossriley:
1. Fix typo on this 
2. Fix difference between how contenttype array was constructed… for single vs multiple contenttypes


GawainLynch:
:koala: says thanks :grin:

PR #3816, opened 2015-07-16
[2.2] Trigger database update notifications for changed field names

### LINKED ISSUE #3808: [Regression] Changing names of fields does not trigger DB update
When I change the name of a field in `contenttypes.yml`, I am not prompted for a DB update. I have tried clearing the cache and deleting the DB, recreating it with the old schema and changing the field name again doesn't work either.

So then when I try to save, I get an error as the column doesn't exist.

Tested in 2.2.4

GawainLynch:
Fixes #3808


bobdenotter:
Aaaight!

PR #3842, opened 2015-07-20
Optimisation for backend by caching requests for users.

bobdenotter:
As pointed out by @rossriley: No need to fetch users every single time. This PR fetches them once, and returns from memory if possible. This yields a significant improvement in queries / time taken. 

Before: 

screenshot 2015-07-20 20 11 16

After: 

screenshot 2015-07-20 20 25 29

### DIFF
--- a/src/AccessControl/Permissions.php
+++ b/src/AccessControl/Permissions.php
@@ -638,7 +638,7 @@ class Permissions
                     // Either fetch $content, or use the one we passed along.
                     $isContent = $content instanceof \Bolt\Content;
                     if (!$isContent) {
-                        $content = $thiss->app['storage']->getContent("$contenttype/$contentid", ['hydrate' => false]);
+                        $content = $this->app['storage']->getContent("$contenttype/$contentid", ['hydrate' => false]);
                     }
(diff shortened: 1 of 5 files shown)


bobdenotter:
Would you look at that! 

screenshot 2015-07-20 22 10 57

rossriley:
Can we have 2.2 version too... this is awesome.

bobdenotter:
@rossriley Surely. I'll backport this to 2.2.x after it has landed.

GawainLynch:
Just a few issues I a looking at :grin:

GawainLynch:
@bobdenotter acceptance tests were broken due to the permissions changes. I have modified the flow a bit and it passes locally… So back over to you for approval I guess.

**Note:** With your original changes, we already know if we have a `\Bolt\Content` object and `isAllowedSingle()` only cares about the owner ID at that point anyway, which we will have if either an object of array has been passed into `isAllowed()` in the first place… so we're not needed to re-request the object at that stage… Query count unchanged.

bobdenotter:
Hooray for teamwork! :tada:

bobdenotter:
Booom!

PR #3851, opened 2015-07-22
Avoid a missing array key from displaying a warning

### LINKED ISSUE #3712: PHP warning showing when making a search
## Bug summary

As a logged-in or logged-out user, when I make a search via the upper-right search form of the default theme (bringing me, for example, to the following URL: `http://localhost:8080/search?search=my_search_query`), I have this PHP warning message popping at the top of the HTML source code:

`Warning: Invalid argument supplied for foreach() in /Users/my_name/Sites/my_bolt_project/vendor/bolt/bolt/src/Content.php on line 1535`

(The message was anonymised.)

## Specifics

**Bolt version:** 2.2.3
**PHP version:** 5.5.20 (OS X)
**Web server:** PHP built-in server
**Browser**: Google Chrome 43.0.2357.124 (64-bit)
**OS**: OS X 10.10.3

I have `display_errors = On` in `php.ini`.

Fresh Bolt install (via [composer-install](https://github.com/bolt/composer-install)) with default theme (`base-2014`) activated.

I have tweaked `contenttypes.yml`, here's its content:

```yaml
pages:
  name: Pages
  singular_name: Page
  default_status: draft
  icon_one: fa:file-text-o
  icon_many: fa:file-text-o
  fields:
    title:
      type: text
      group: content
      required: true
    slug:
      type: slug
      group: content
    content:
      type: html
      group: content
      height: 300px
```

## Steps to reproduce

Edit the `contenttypes.yml` file as mentioned above and try to make a search with any search query.

## Expected result

The search results without the error message showing.

## Actual result

The search results with the error message showing. See the following screenshot:

screenshot

## Suspected issue

In `src/Content.php`, line [1535](https://github.com/bolt/bolt/blob/c65c0262002cd1d0baac1e7404890daffc2d677c/src/Content.php#L1535), there's a `foreach` statement which iterates over `$config['uses']`, but `$config['uses']` doesn't exist. I think that's because I didn't define a `uses` option for the `slug` field of the `pages` content type in `contenttypes.yml`.

Adding a check on the existence of `$config['uses']` solves the problem, like so:

```php
if ($config['type'] == 'slug' && !empty($config['uses'])) {
    foreach ($config['uses'] as $ptrField) {
        /*...*/
    }
}
```

I can create a PR to fix the issue, but I first wanted to check with you folks if that's actually normal that `$config['uses']` doesn't exist at this point in the source code. I mean, even in the absence of the `uses` option in `contenttypes.yml`, the slug is still generated, based on the value of the `title` field of the content type if it exists, or a random slug is generated otherwise.

So, is the addition of the `empty()` check enough? Or shall we put more work into it so that `record.contenttype.fields.field_of_slug_type.uses` is always defined, even if it was omitted in `contenttypes.yml`?

fabschurt:
This fixes #3712.

Now `$config['uses']` is checked for existence _and_ explicitly converted to an array in case it's originally a single string.

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -1524,8 +1524,8 @@ class Content implements \ArrayAccess
         }
 
         foreach ($this->contenttype['fields'] as $config) {
-            if ($config['type'] == 'slug') {
-                foreach ($config['uses'] as $ptrField) {
+            if ($config['type'] == 'slug' && isset($config['uses'])) {
+                foreach ((array) $config['uses'] as $ptrField) {
                     if (isset($fields[$ptrField])) {
                         $fields[$ptrField] = 100;
                     }


bobdenotter:
Looks good to me. 

Disregard Scrutinizer, acquire merge!

PR #3855, opened 2015-07-23
Handle "add" parameter for url generator a new way

CarsonF:
`Lib::path` and `Lib::redirect` both have an _add_ parameter which takes a string and appends it onto the end of the generated url. An example:
```php
$add = 'foo=bar#derp'; // I'm just an unstructured blob :(
$path = Lib::path('dashboard', [], $add);
echo $path; // /bolt?foo=bar#derp
```
As this class is deprecated (or trying to be :gun:), we need a new way to handle this functionality. 

Symfony's UrlGenerator will already take extra parameters (not needed in path) and add them to the query string of the generated url. This only leaves the _fragment_ part of the url functionality missing. 

I created a wrapper for the UrlGenerator that will take a `#` key from the parameters and append it. This also keeps the same interface/object, so no need for a "helper" method. 
New way:
```php
$params = ['foo' => 'bar', '#' => 'derp']; // I'm a nice readable array!
$path = $app['url_generator']->generate('dashboard', $params);
echo $path; // /bolt?foo=bar#derp
```

Functionality applied to:
-  `editcontent` route which uses a fragment/anchor for _return to_
- MenuBuilder - updated to maintain BC for _add_ key. New functionality works out of the :package:
- There might be other usages the :koala: hid?... beuller?

Who hates naming? :raising_hand: Let me know if you don't like how something is named, and I can talk you out of it...err..we can talk about it.

### DIFF
--- a/src/Helpers/MenuBuilder.php
+++ b/src/Helpers/MenuBuilder.php
@@ -3,6 +3,7 @@
 namespace Bolt\Helpers;
 
 use Bolt\Translation\Translator as Trans;
+use GuzzleHttp\Url;
 use Silex\Application;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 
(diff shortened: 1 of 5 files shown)


GawainLynch:
@rossriley I think we need to reject this, as the submitter **actually** added units tests for new code… That's not how we roll, right? :wink:

CarsonF:
Brought out all the bells and whistles for this one :smile: :tada:

rarila:
I wonder if the key shouldn't be simply be ``#``
```
$params = ['foo' => 'bar', 'fragment' => 'derp', '#' => 'koala'];
```
To get ``/bolt?foo=bar&fragment=derp#koala``

- you can see at first sight that this is a special parameter
- it's almost wysiwyg
- it's more likely you have real parameter namend ``fragment`` them one named ``#``

CarsonF:
@rarila I almost went with that, but I was afraid I was being too terse

rarila:
@CarsonF k.i.s.s.

rarila:
Not to forget: ask 100 people how that # part of the url is named. I believe the majority answers "don't know" and probably "anchor".

CarsonF:
Works for me!

rarila:
:+1:

GawainLynch:
:shipit:

bobdenotter:
Looks good to me!

PR #3857, opened 2015-07-23
Update usages of "fragment" key to "#"

CarsonF:
Who's a derp? :koala:


GawainLynch:
Someone didn't write an acceptance test… and we won't talk about the unit tests :stuck_out_tongue:

PR #3866, opened 2015-07-25
[2.2] Handle custom Content classes and clean up the class slightly

### LINKED ISSUE #3864: Record Change Handler
I've hit a bug when saving a conttenttype without a "title" field.

RecordChangeHandler.php
The problem is that this file is not looking at the class attribute on the conttenttypes, I have created my own class that extends Content.php to change the output of getTitle.

I believe the fix is to re-write the following 
```
 protected function write(array $record)
...
 if ($record['context']['new']) {
	            $content = new Content($this->app, $record['context']['contenttype'], $record['context']['new']);
	        } else {
	            $content = new Content($this->app, $record['context']['contenttype'], $record['context']['old']);	        
}
```

to the following

```
$contenttype = $record['context']['contenttype'];
        if ($record['context']['new']) {
            $values = $record['context']['new'];
        } else {
            $values = $record['context']['old'];
        }

        // Make sure $contenttype is an array, and not just the slug.
        if (!is_array($contenttype)) {
            $contenttype = $this->app['storage']->getContentType($contenttype);
        }
        // If the contenttype has a 'class' specified, and the class exists,
        // Initialize the content as an object of that class.
        if (!empty($contenttype['class']) && class_exists($contenttype['class'])) {
            $content = new $contenttype['class']($this->app, $contenttype, $values);
            // Check if the class actually extends \Bolt\Content.
            if (!($content instanceof Content)) {
                throw new \Exception($contenttype['class'] . ' does not extend \\Bolt\\Content.');
            }
        } else {
            $content = new Content($this->app, $contenttype, $values);
        }
```

GawainLynch:
Fixes #3864

### DIFF
--- a/src/Logger/Handler/RecordChangeHandler.php
+++ b/src/Logger/Handler/RecordChangeHandler.php
@@ -74,18 +74,7 @@ class RecordChangeHandler extends AbstractProcessingHandler
         $context = $record['context'];
 
         // Check for a valid call
-        if (!in_array($context['action'], $this->allowed)) {
-            throw new \Exception("Invalid action '{$context['action']}' specified for changelog (must be one of [ " . implode(', ', $this->allowed) . " ])");
-        }
-        if (empty($context['old']) && empty($context['new'])) {
-            throw new \Exception("Tried to log something that cannot be: both old and new content are empty");
-        }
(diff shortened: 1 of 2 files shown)


bobdenotter:
Also merging in!

PR #3870, opened 2015-07-25
Unit test parameter handling

GawainLynch:
This (should) make the unit test environment fully configurable for configuration files, theme and Sqlite database.

The biggest win here is for extension authors that want to unit test extensions with customised configuration (routes/contenttypes/etc) and Sqlite databases. Of course if you don't want to use Sqlite, that will be overridden in the passed in `config.yml` file.

There is an update to the docs [pending in a branch here](https://github.com/bolt/docs/compare/bolt:master...GawainLynch:update-phpunit-docs?expand=1) for once this is merged and back-ported into `release/2.2`, including the work in and #3850.

Special thanks go out to the :koala: who was busy eating :leaves:  and left me alone long enough to finish this today!

Ping @slahav… this should be exactly what you all need. :grin:

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,10 +17,45 @@
     <listeners>
         <listener file="tests/phpunit/BoltListener.php" class="Bolt\Tests\BoltListener">
            <arguments>
+              <!-- Configuration files. Can be either .yml or .yml.dist files -->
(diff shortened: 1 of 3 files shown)


slahav:
Thank You Gawain.
We'll start using it right away.
This is great!

Best Regards,

Sivan Lahav

PR #3872, opened 2015-07-25
Add foreign key checks to IntegrityChecker

ntomka:
This PR is related to #3745. That change added foreign keys only on table creation leaving alters out. So I added few more lines to handle new or modified foreign keys after creation.

I don't know for sure if this is needed in master branch.

### DIFF
--- a/src/Database/IntegrityChecker.php
+++ b/src/Database/IntegrityChecker.php
@@ -184,7 +184,6 @@ class IntegrityChecker
                         $msg = 'Table `' . $table->getName() . '` is not the correct schema: ';
                         $msgParts = array();
 
-                        // No check on foreign keys yet because we don't use them
                         /** @var $col Column */
                         foreach ($diff->addedColumns as $col) {
                             $msgParts[] = 'missing column `' . $col->getName() . '`';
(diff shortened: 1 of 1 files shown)


GawainLynch:
> I don't know for sure if this is needed in master branch.

Potentially… Thoughts @rossriley ?

rossriley:
yes, let's keep the master/2.2 the same with IntegrityChecker. I personally haven't touched this particular method yet so if we have a master version too we can get them both merged in.

ntomka:
@rossriley If I use an extension to add a table it is created with correct foreign keys, but if I change the definition I get an exception [here](https://github.com/bolt/bolt/blob/master/src/Storage/Database/Schema/Manager.php#L310) as `$this->app['schema.tables']` contains only Bolt's base tables. If I surround this statement with a try block to pass the thrown exception everything works as expected, the change is committed to database.

So it seems to me my change in this PR is not needed in master, but it is possible that I'm doing something completely wrong. :)

bobdenotter:
Is this one 'good to go'?

GawainLynch:
@bobdenotter I think so, I was going to marry it with a forward… wait… did I just say "marry"… 

Seriously… well I tried first go… but I wanted to combine it with a forward port to master so it didn't get lost. My apologies, @ntomka

ntomka:
I can't get your point @GawainLynch. I'm married and I'm proud of it! :gem: 

As I stated above changes in master maybe already resolved the issue in this PR, so it is only affects 2.2 branch (and faced another problem in master when changing FK definitions, but that's another story).

You should try a sample in both master and 2.2, like this in an extension:

```PHP
$this->app['integritychecker']->registerExtensionTable(function (Schema $schema) {
    // Define table
    $table = $schema->createTable($this->tableName);

    // Add primary column
    $table->addColumn('id', 'integer', array('autoincrement' => true));
    $table->setPrimaryKey(array('id'));

    // Add working columns
    $table->addColumn('userid', 'integer');

    // Index column(s)
    $table->addIndex(array('userid'));

    // Constraints
    $table->addForeignKeyConstraint(
        $this->app['config']->get('general/database/prefix', 'bolt_') . 'users'
        , array('userid')
        , array('id')
        , array("onUpdate" => "CASCADE", "onDelete" => "CASCADE")
    );

    return $table;
});
```

Try to change the FK constraint. I know in master this looks a bit different, but how you define FKs is the same.

GawainLynch:
> I can't get your point @GawainLynch. I'm married and I'm proud of it!

Yes, and your wedding photos are wonderful… They put a smile on Erin and my faces :+1:

GawainLynch:
On the DB thing… it's just been a time constraint on my part… RL work versus Bolt core versus having written too many damn extensions :frowning: 

Help wanted :flags:

PR #3873, opened 2015-07-25
Update config.yml.dist

dadaxr:
'hostname' was confusing : it does not match the available key ( 'host' and 'port' ) specified in the online documentation.


bobdenotter:
Good catch! Thank you, @dadaxr

PR #3879, opened 2015-07-27
[2.2] Storage publish/depublish events

### LINKED ISSUE #3862: Send storage event in publishTimedRecords
Shouldn't bolt send a Storage event in publishTimedRecords and depublishExpiredRecords ? https://github.com/bolt/bolt/blob/master/src/Storage.php#L1077

For example if we want to make live indexing in elasticsearch, it's not possible when items are timed published.

GawainLynch:
Fixes #3862


bobdenotter:
Whoop!

PR #3881, opened 2015-07-28
Fixing spacing for thumbnails cell, if no thumbnail is present.

bobdenotter:


rarila:
Ok, the stuff I'm trying hard to refactor out at the moment by using correct markup and css: inline styles :rage:

GawainLynch:
A new update to Chrome was just pushed by Google too, we might not need this now… Needs verification.

bobdenotter:
> A new update to Chrome was just pushed by Google too, we might not need this now… Needs verification.

It fixed it indeed! I will remove those inline styles! :-)

PR #3884, opened 2015-07-28
Break here if $contentId is `` or `0` instead of only on strictly `null`

### LINKED ISSUE #3883: Weird memory leak / loop in "new content"
Encountered this on one site, so i'm not sure if it's a but in Bolt, or something quirky with the site just yet. Thought i'd log it just in case. 

Most pages on that site hover around 25mb memory usage: 

screen shot 2015-07-28 at 17 33 20

But, when clicking "new entry", it breaks: 

screen shot 2015-07-28 at 17 33 37

bobdenotter:
Fixes #3883 

If we try to get permissions for `pages/` (for a new record, without existing id), don't get stuck in a loop where we retrieve all records for that contenttype.

### DIFF
--- a/src/Permissions.php
+++ b/src/Permissions.php
@@ -621,7 +621,7 @@ class Permissions
                 }
 
                 // Handle special case for owner.
-                if ($contentId === null) {
+                if (empty($contentId)) {
                     break;
                 }
 


bobdenotter:
if this one is merged, I'll cherry-pick for master!

GawainLynch:
Ha! It wasn't the :koala: this time!

PR #3891, opened 2015-07-29
Rewrite of getContent to use new storage layer

rossriley:
This is the first phase, in that it contains no deletions of the current functionality but exposes the new services via a `QueryServiceProvider` that can be called via `$app['query']->getContent(...........)`

If all goes to plan we do one of two things, leave $app['storage']->getContent() as a deprecated call that forwards to the query provider, or leave both running concurrently and gradually refactor internals to point to the new system.

### Highlights

I'll write more details up in the documentation but this effectively turns the old black box of calling `getContent()` into a composable parsing and handling library that allows more potential to extend or modify on the fly. 

This is especially useful for the new Field types that will need the ability to affect the building of database queries.

#### 1: The ContentQueryParser

This service splits up every Bolt query into commands, directives and parameters. For example the following query:

    {% setcontent featured = '(events,entries)', {'featured':1, 'datepublish': '< last month'} limit 5 %}

This class parses the contenttypes, in this case `events` and `entries`, recognises `limit` as a directive to later modify the query, and then recognises the parameters that will be used to filter the query.

Another type of query:

    {% setcontent latest = 'pages/latest/5' %}

will also be parsed into a contenttype `pages` and operation `latest` and a directive, in this case a limit of 5.

The important part is that the handlers for each of these can be swapped out from the defaults, or alternatively you can quickly add your own handler to support custom query handlers. For instance the following query:

    {% setcontent featured = 'pages/featured/5' %}

Would be possible by registering a `featured` operation and a corresponding handler for it.

#### 2: The QueryParameterParser

This class handles all of the smart parameter tests used when fetching content, eg:

    {'ownerid':'!1', 'id': '1 ||2 || 3', 'datepublish': '>now', 'username || email': 'fred', 'title':'%welcome%' }

This parser turns all of those into SQL expressions but again it exposes all its matchers and handlers to the application via `$app['query.parser.handler']`

So all the above defaults will work out of the box, but adding new matchers is straightforward, for example the test for this functionality adds an additional syntax for like queries that allows the following to work: `'filename': '~example'` identically to the standard `'filename': '%example%'`

Adding this syntax is as simple as adding a new matcher to the stack:

    $app['query.parser.handler']->addValueMatcher('\~(\w+)', ['value' => '%$1%', 'operator' => 'like']);

#### 3: Field delegation

The final important part of this is that each field handler is passed the QueryBuilder object prior to any query execution. This for example allows the date fields to replace values like the now in: `'datepublish': '> now'` by passing them through strtotime first. 

It also allows taxonomy and relation fields to modify the query so that they can be filtered on. It will also be important for new field types in the works, like Repeating Fields and Composite Fields.

## Still to do / obligatory warning

For the next couple of weeks this is primarily going to be for testing / evaluation only. I wouldn't point all your code to this new method yet. 

1. Searching / weighting will be handled via a separate PR.
2. The results of the query will return the new `Entity\Content` classes rather than the old `Content` classes, these aren't feature complete as yet.
3. Automatic filtering of unpublished isn't happening yet
4. Paging integration will be handled via a separate PR

### DIFF
--- a/src/Provider/QueryServiceProvider.php
+++ b/src/Provider/QueryServiceProvider.php
@@ -3,7 +3,10 @@
 namespace Bolt\Provider;
 
 use Bolt\Storage\NamingStrategy;
+use Bolt\Storage\Query\ContentQueryParser;
 use Bolt\Storage\Query\Query;
+use Bolt\Storage\Query\QueryParameterParser;
+use Bolt\Storage\Query\SelectQuery;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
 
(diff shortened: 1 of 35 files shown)


GawainLynch:
:+1:

bobdenotter:
bubbles

GawainLynch:
:shipit:

PR #3892, opened 2015-07-29
Deprecated class aliases

GawainLynch:
This moved the aliases to an autoloaded file that we can use for deprecated stuff like this.

Logic being that this give the BC coverage to all install types.


bobdenotter:
:+1:

PR #3897, opened 2015-07-30
Add .jshintrc

rarila:
- Moved the jshint config into ``js/.jshintrc``
- Added ``.jshintrc`` for grunt files
- Removed inline jshint directives for globals
- Added ``jshint:grunt`` grunt target
- Fixed two small issues

Now IDEs and editors can make use of ``jshint`` with Bolts settings.


GawainLynch:
More :de: awesome… :+1:

PR #3900, opened 2015-07-30
Sessions Rewrite

CarsonF:
_I want CSRF protection!!_
_I want caching!!_

# Context
_Wait what's the problem again?_

- Cache proxies (like Varnish) do not like when cookies are used, since they make the request unique to the individual
- Forms are protected from [CSRF](https://en.wikipedia.org/wiki/Cross-site_request_forgery) with sessions
- Sessions are identified using cookies

_Go on..._

Any time a user visits a page with a form on it a CSRF token is set in _the_ session. This starts the session and sends the cookie header. A) The session cookie is set to the root path `/`, so that any page of the application can access it. This renders caching useless. B) The path is restricted to the current request. This would fix caching on other pages, but render the session useless for the rest of the application. `goto start;`

_Why can't the cookie be removed if the session data isn't needed?_

In short, because it's not possible with the built-in session.

# Solution

We aren't going to use the built-in session anymore. _Stick it to the man!_ This is completely transparent, since the application is written against Symfony's `SessionStorageInterface`.

So what does this allow us to do?

## Session Storage
- Multiple Sessions. Since there are no [superglobals](http://replygif.net/i/120.gif)/[singletons](http://replygif.net/i/882.gif), this is easy to wire up.
- Works inside of the Symfony HTTP lifecycle
	- Individual sessions for sequential requests, inside the same php process, are not a problem (Unit tests for example).
	- Session cookie is added to the response object (via a [listener](https://github.com/CarsonF/bolt/blob/feature/csrf-sessions/src/Session/SessionListener.php)). This allows other middlewares to change or delete the cookie before it is sent to the client.
- Session components can be swapped out in user-land code.
	- [**Generator**](https://github.com/CarsonF/bolt/tree/feature/csrf-sessions/src/Session/Generator) - Responsible for generating session IDs.
	- [**Serializer**](https://github.com/CarsonF/bolt/tree/feature/csrf-sessions/src/Session/Serializer) - Responsible for (un)serializing session data.
	- **Handler** - (Using [built-in interface](http://php.net/manual/en/class.sessionhandlerinterface.php)) Responsible for persisting session data.
- DI all the things! Everything is hooked up with DI making it easy to configure at runtime.

## Forms
_How does this session stuff fix the forms?_

By default, we'll have a _main_ session with a cookie path at `/`. This will act the same as before. We'll also have a second _csrf_ session, which will strictly be for forms. The cookie for this session will be set to the current path, that way each form has it's own cookie.

# Caveats
There are a couple pieces of obscure functionality missing with this implementation (some intentionally). 

- Transparent sid - This is considered insecure and disabled by default
- Upload progress - I've never seen this used. It has a slew of prerequisites as well (PHP 5.4, not FastCGI, request buffering disabled (Nginx), etc.). This is probably the biggest "feature" that isn't possible, as opposed to not implemented.
- There are a couple of situations where there were enforced restrictions that are no longer there. For instance, **cache limiter**, could send anti-caching headers. This could be implemented later however.
- There are several session **options** that are not configurable in the options array. 
	- Some correspond to the things mentioned above: `use_trans_sid`,  `url_rewriter.tags`, `upload_progress.*`, `cache_*`. 
	- Some are old: `bug_compat_42`, `bug_compat_warn`. 
	- Some are implied: `use_cookies`, `use_only_cookies`. 
	- Options relating to `Generator` and `Serializer` which are configured via DI: `serialize_handler`, `entropy_*`, `hash_function`, `hash_bits_per_character`.

I'm just trying to cover my bases here. These are all edge cases and shouldn't be worried about. 

## Backwards compatibility
As this sits behind an interface, backwards compatibility isn't that big of a concern. Where it is a concern is in configuration. This is what faces the rest of bolt and has changed a bit (see [`Application::initSession`](https://github.com/CarsonF/bolt/blob/feature/csrf-sessions/src/Application.php#L81-L100)). I've taken steps to ensure backwards compatibility is maintained in this area, such as pulling options from ini by default.

-----

If you've made it this far, then take a look at the code and give me your thoughts. Also, huge thanks to @gawainlynch for his help and guidance with this as always! :beer:.

## TL;DR:

----
@bolt/contributors Hang off on merging. I just have a couple more clean ups I want to do. Almost there!


bobdenotter:
omm-monkey

CarsonF:
Ok I think this is ready to go now! At least for this round :) :shipit:

GawainLynch:
:shipit:

bobdenotter:
Whoop-tie-doop!

GDmac:
C'est beaucoup bonne

PR #3915, opened 2015-08-03
[2.2] Publishing queries speed up

GawainLynch:
Just a slight performance boost in the front end on a default install…

**Before**
before
**After**
after

…and for a more complex site…

**Before**
before
**After**
after


bobdenotter:
Nice! :shipit:

PR #3919, opened 2015-08-03
Use cookie lifetime from config and remove session_user_storage_handler from config

CarsonF:


CarsonF:
Oh travis :unamused: :shipit:

PR #3924, opened 2015-08-05
Add a copy of the MIT license

Calinou:
It is good practice to add a copy of the license to the repository.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+# The MIT License (MIT)
+
+Copyright (c) 2015 Bolt
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


bobdenotter:
I wouldn't be against this.. It's not like we _hide_ the fat that we're MIT licensed, but it wouldn't hurt to make it more explicit.

@GawainLynch, thoughts?

GawainLynch:
Well, we're MIT licensed… So it is not required, as that is a free-for-all license that simply serves the purpose of creating a limited liability for the copyright holders… It is simply a point of avoiding ambiguity that has no solid legal precedent… Copyright law is the important part here, not the licence… But IANAL.

That said, apart from a CDO hit over root litter, no objections :+1:

GawainLynch:
Just to add to this discussion, there is already a licence file in [`app/LICENSE.txt`](https://github.com/bolt/bolt/blob/master/app/LICENSE.txt) that could use some attention (read: line breaks) too.

bobdenotter:
In the root is better, I think. I'd say we merge this in, and fix the link in our 'about' page as well as remove the one from `app/`.

GawainLynch:
:+1:

bobdenotter:
Top notch!

PR #3926, opened 2015-08-05
Allow the 'stack' column to be `null`, for backwards compatibility. Fixes #3925

### LINKED ISSUE #3925: Fatal error in SQL query when logging in..
To reproduce (on my system at least)

1 Set up an install of Bolt 2.2. Add a user, log on and **don't** add any files to the stack. 

2 Update to 'master'

3 Run some commands: 

```
$ php app/nut database:check                                                     
Modifications required:
 - Table `bolt_users` is not the correct schema: invalid column `stack`, invalid column `roles`
One or more fields/tables are missing from the Database. Please run 'nut database:update' to fix this.
$ php app/nut database:update                                                                             
Modifications made to the database:
 - Updated `bolt_users` table to match current schema.
Your database is now up to date.
$ php app/nut database:check                                                
The database is OK.
```

4 Attempt to log in to Bolt.

5 See error: 

screenshot 2015-08-05 21 07 30

bobdenotter:


bobdenotter:
Hmm.. I thought I changed the `bolt.db` to match the tests, but apparently not.. 

```
+'Modifications required:
+ - Table `bolt_users` is not the correct schema: invalid column `username`, invalid column `password`, invalid column `lastseen`, invalid column `lastip`, invalid column `displayname`, invalid column `enabled`, invalid column `shadowpassword`, invalid column `shadowtoken`, invalid column `shadowvalidity`, invalid column `throttleduntil`, invalid column `email`
+One or more fields/tables are missing from the Database. Please run 'nut database:update' to fix this.'
```

How is this file created?

src/Storage/Database/Schema/Table/Users.php
```diff
@@ -21,7 +21,7 @@ protected function addColumns()
         $this->table->addColumn('lastseen',       'datetime',   ['notnull' => false, 'default' => null]);
         $this->table->addColumn('lastip',         'string',     ['length' => 32, 'default' => '']);
         $this->table->addColumn('displayname',    'string',     ['length' => 32]);
-        $this->table->addColumn('stack',          'json_array', ['length' => 1024]);
+        $this->table->addColumn('stack',          'json_array', ['length' => 1024, 'notnull' => false]);
```
GawainLynch: You can't do this for json_array… It breaks some platforms.

bobdenotter: Hmm.. Any suggestions on how we should fix this differently?

GawainLynch: Yes… By ignoring me, partially… It was the `default` array key that broke stuff.

The bigger question is, however, why is this needed? Is this just something that popped up while stack was broken in master (I only fixed it in the session merge)?

GawainLynch:
> Hmm.. I thought I changed the `bolt.db` to match the tests, but apparently not..

Oh, look… Travis has magically started re-running… I suspect that :koala: might have something to do with it :wink:

src/Storage/Database/Schema/Table/Users.php
bobdenotter: The main reason I added this, is that I was getting errors when trying to log in, that were caused when switching between 2.2 and 2.3. The value of the 'stack' column happened to be `null` on my 2.2 setup and when I switched, I was getting fatal errors, and couldn't log in. There's a screenshot in #3925

GawainLynch:
Travis looked into it :8ball: and now says :+1:

PR #3931, opened 2015-08-05
Prevent MD5 collisions in hashes. Fixes #3930

### LINKED ISSUE #3930: Assets in extensions have colliding MD5 hashes, if created at same time.
If you have an extension, that adds two assets, like this: 

```
            $this->addJavascript('assets/handsontable.full.min.js', true);
            $this->addJavascript('assets/start.js', true);
```

If these files are created at the same time on the filesystem, their hashes will clash. This means only the last one will be added to the output.

bobdenotter:


src/Provider/AssetServiceProvider.php
```diff
@@ -36,9 +36,18 @@ function ($app) {
             $fullPath = $app['resources']->getPath('root') . '/' . $fileName;
 
             if (is_readable($fullPath)) {
-                return substr(md5($app['asset.salt'] . (string) filemtime($fullPath)), 0, 10);
+                $path = $fullPath;
             } elseif (is_readable($fileName)) {
-                return substr(md5($app['asset.salt'] . (string) filemtime($fileName)), 0, 10);
+                $path = $fileName;
+            } else {
+                $path = false;
```
CarsonF: `return false;` here. Skip the if/else below

GawainLynch: This is wrong… At the very least add the full path to the hashed string.

bobdenotter: @GawainLynch I'm not sure I follow.. When I use this, $path already is a full path, that's used for the hash: 

ie: 

```
"/Users/bob/Sites/bolt-origin///extensions/vendor/bobdenotter/seo/assets/seo.css"
"/Users/bob/Sites/bolt-origin///extensions/local/bobdenotter/gridviewfield/assets/handsontable.full.min.css"
"/Users/bob/Sites/bolt-origin///extensions/local/bobdenotter/gridviewfield/assets/handsontable.full.min.js"
"/Users/bob/Sites/bolt-origin///extensions/local/bobdenotter/gridviewfield/assets/start.js"
```

What should be added?

Issue #3936, opened 2015-08-06
Error generated with attempting to use Liquibase with BOLT


mhodges-cainc:
To manage database changes, we're using Liquibase, which injects 2 tables into the bolt database: DATABASECHANGELOG & DATABASECHANGELOGLOCK.

The latter includes a bit field. When I try to bring up bolt after these tables have been create, I get the error:

Unknown database type bit requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it.

To fix the issue, I had to add the following line to Database/initListener.php (at line 32):
$db->getDatabasePlatform()->registerDoctrineTypeMapping('bit', 'integer');

2 questions:
Why would Bolt care about non-Bolt tables?
Can you incorporate this fix into the Bolt source?

GawainLynch:
> Why would Bolt care about non-Bolt tables

@bobdenotter can you remember the history on this one?

bobdenotter:
> Why would Bolt care about non-Bolt tables?

I believe this is a bug, plain and simple. We've noticed it before, but we never got round to fixing, yet.

bobdenotter:
@mhodges-cainc Your fix might fix it in your case, but I think this is a good time to look into why it crawls and compares _all_ tables, and not just the Bolt ones.

GawainLynch:
Related to #2769

GawainLynch:
Fixed in #3949

PR #3949, opened 2015-08-10
Filter db tables

### LINKED ISSUE #2769: Listing all tables can make DBAL mad
DBAL's Schema Manager tries to create objects for all tables in a database, but fails if it doesn't recognize a type ([here](https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L423)).

Bolt should *only* care about tables its own tables (with the prefix specified).

Unfortunately the only way I can find to get around this is by overridding the platform classes.
`getListTablesSQL` needs another where condition for our prefix.

[SqlitePlatform](https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L435)
[MySqlPlatform]
(https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L349)
[PostgreSqlPlatform](https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L259)

CarsonF:
Look what I found! DBAL has a config option to filter tables!

Should fix #2769 and #3936.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -169,6 +169,24 @@ class Application extends Silex\Application
         );
         $this->register(new Database\InitListener());
 
+        // Filter tables to only those bolt cares about (based on table name prefix).
+        // This prevents schema parsing errors from tables we don't need to worry about.
+        $app = $this;
+        // This key will make more since in version 2.3
+        $this['schema.tables_filter'] = function () use ($app) {
+            $prefix = $app['config']->get('general/database/prefix');
+            return "/^$prefix.+/";
+        };
+        $this['db.config'] = $this->share(
+            $this->extend('db.config',
+                function ($config) use ($app) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
awwweesssooommee

CarsonF:
:shipit:

bobdenotter:
Shouldn't this be applied to `master` as well?

GawainLynch:
Yes, but we're taking a slightly different approach… /me also nags @CarsonF :stuck_out_tongue:

bobdenotter:
Should we then merge this into master, only to make sure we don't forget?

GawainLynch:
Nah, it's headed for the SP in master

PR #3952, opened 2015-08-11
Issue 3951 theme config override

### LINKED ISSUE #3951: Allow themes to override config parameters
While porting a theme, I had to change the value of `listing_records` to match my theme layout. I tried to change it using the config.yml in theme, but it seems it isn't supported. 

So, themes should be able to override, at least, this parameter. To keep it organized and safe, I would place theses overrides in a specific section in the config file, like, `override`.

pygillier:
Hi, 

this PR aims to provide a theme override for `listing_records`, `listing_sort` and `search_results_records`.

Fixes #3951


bobdenotter:
Looks good.. Could you add these as examples to `config. yml` in `theme/`?

GawainLynch:
:+1:

Issue #3950, opened 2015-08-10
Translations not loading at all throughout the back end. Think I've found the cause.


philbates35:
I have installed the SimpleForms extension, which means I need to set up `mailoptions` in config.yaml. There's a helpful flash message that tells me this at the top of every screen in the backend.

Today I ran `vendor/bin/nut cache:clear` and no matter what I did after doing so, no translations were loading throughout the back end, and instead just showing `label.key.name.example` instead. After some debugging, I'm pretty sure I've found the reason:

Summary:
* The flash message about setting 'mailoptions' tries to load a translation.
* However, all resources have not yet been added.
* As part of the label retrieval process, it gets what it thinks are all the available resources (which is completely empty) and caches it.
* Subsequent requests for translation files (after the resources have been added, and the translations are theoretically available) see that there is a cached translation file (which upon manual inspection contains no translations) and tries to load it from the cache.
* As the cache is empty, it can't find a match and therefore just displays the translation key.

Temporary Fix to prove cause:
* Comment out the translation [request](https://github.com/bolt/bolt/blob/v2.2.8/src/Application.php#L434)
* Clear the cache using `vendor/bin/nut cache:clear`
* Reload the backend and all translation load successfully.

Quick explanation of the backtrace:
* The [flash message](https://github.com/bolt/bolt/blob/v2.2.8/src/Application.php#L433) in question is set in `Bolt\Application::initMailCheck()`.
* Setting the flash message results in a call to `Bolt\Translation\Translator::__($error)` [here](https://github.com/bolt/bolt/blob/v2.2.8/src/Application.php#L434).
* However - the crucial point is that the translation resources aren't loaded until `Bolt\Provider\TranslationServiceProvider::boot()` is called. In terms of the application, `Bolt\Application::initMailCheck()` is called before the [translation resources have been added](https://github.com/bolt/bolt/blob/v2.2.8/src/Provider/TranslationServiceProvider.php#L36).

GawainLynch:
See #3948… do you have MenuEditor installed, and any other extensions?

philbates35:
Ah dammit I searched and couldn't find a similar issue! Sorry about that.

However in my case I am not using MenuEditor. The only extension I am using is bolt/simpleforms, as I mentioned in the OP. I am using bolt/bolt 2.2.8.

I'm pretty sure that the issue is in the bolt core rather than the extension though, for the following reason:

[...]

PR #3953, opened 2015-08-12
Move initMailCheck() call to a before() handler

### LINKED ISSUE #3948: [2.2.5+] Bolt generates empty translator cache when installed with menueditor
### Description
When bolt is installed in conjunction with menueditor (and presumably any extension using app['translator']) it generates an empty translation cache. Not even the fallback locale is included in the cache.

#### Details

This only happens with other locales than en_GB (tried en_GB, sv_SE and nl_NL).
This does not happen if the extension is installed via /bolt/extend.

Clearing the cache via app/nut does not fix the issue.
Clearing the cache via the backend route /bolt/clearcache seems to fix the issue until one changes the locale, then it reappears again.

The issue is in release/2.2 and versions 2.2.8, 2.2.7, 2.2.6 and 2.2.5.

2.2.4 and down does not have the issue.

Master does not seem compatible with the menueditor.

Perhaps https://github.com/bolt/bolt/pull/3753 is to blame? The timeline fits.

What the empty cache looks like (in case of sv_SE):

    <?php
     
    use Symfony\Component\Translation\MessageCatalogue;
     
    $catalogue = new MessageCatalogue('sv_SE', array (
    ));
     
    $catalogueSv = new MessageCatalogue('sv', array (
    ));
    $catalogue->addFallbackCatalogue($catalogueSv);
     
    $catalogueEn_GB = new MessageCatalogue('en_GB', array (
    ));
    $catalogueSv->addFallbackCatalogue($catalogueEn_GB);
     
     
    return $catalogue;

### To recreate:

    git clone git://github.com/bolt/bolt.git "brokencache"
    cd "brokencache"
    git checkout v2.2.8
    composer install --no-dev
    php app/nut extensions:install bacboslab/menueditor 2.1.4
    php app/nut config:set locale "sv_SE"
    
Then go to the login screen and you will see translation keys instead of the proper translation (or the english fallback).

### Workaround:

If you install with some sort of script you can use cURL to clear the cache automatically. Example where $1 is your domain:

    curl -b /tmp/cookie -c /tmp/cookie --data "username=YOURUSERNAME&password=YOURPASSWORD&action=login" http://$1/bolt/login

    curl -b /tmp/cookie -c /tmp/cookie http://$1/bolt/clearcache

GawainLynch:
Fixes #3948 #3950

Already done in master as part of the controller refactor.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -124,7 +124,8 @@ class Application extends Silex\Application
         // Initialize enabled extensions before executing handlers.
         $this->initExtensions();
 
-        $this->initMailCheck();
+        // Mail config checks for extensions
+        $this->before(array($this, 'initMailCheck'));
 
         // Initialise the global 'before' handler.
         $this->before(array($this, 'beforeHandler'));


bobdenotter:
:ship:

PR #3960, opened 2015-08-13
Use embedly without protocol. Fixes #3955

### LINKED ISSUE #3955: Mixed content error on video preview
Having my site behind HTTPS (frontend & backend), when I want to insert a video in an entry, Firefox (IE/Edge/Chrome should do the same, not tested yet) prevents preview to be displayed and console shows a "mixed active content" error.

It's due to HTTP/HTTPS mixing : embedly is called over http only.

Got it !

It's in https://github.com/bolt/bolt/blob/master/app/src/js/modules/video.js#L87
embed.ly API is called specifically on http, It should be called without scheme

(as is it javascript, no PR for me this time :smile: )

bobdenotter:
Fixes #3955


GawainLynch:
Legendary!

PR #3961, opened 2015-08-13
[2.2] Fix/protocol less embedly 22. Fixes #3955

### LINKED ISSUE #3955: Mixed content error on video preview

bobdenotter:
Fixes #3955

### DIFF
--- a/app/src/js/modules/video.js
+++ b/app/src/js/modules/video.js
@@ -82,7 +82,7 @@
     var update = function (key) {
         // Embed endpoint http://api.embed.ly/1/oembed?format=json&callback=:callbackurl=
         // @TODO: Make less dependant on key.
-        var endpoint = 'http://api.embed.ly/1/oembed?format=json&key=51fa004148ad4d05b115940be9dd3c7e&url=',
+        var endpoint = '//api.embed.ly/1/oembed?format=json&key=51fa004148ad4d05b115940be9dd3c7e&url=',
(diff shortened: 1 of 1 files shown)


GawainLynch:
:+1:

cdowdy:
hey all just as a heads up:  

The Protocol Relative URL you are using in the video.js embedly api is considered an "anti-pattern" now -> [Protocol Relative URL Paul Irish](http://www.paulirish.com/2010/the-protocol-relative-url/).  

If the asset is available on https should always serve it with https. It will have no affect to http sites. HTTP served to HTTPS though will cause mixed content warnings - which this pull is meant to fix. Using the protocol relative URL allows for man on the side attacks similar to what github had this year with its outage.

bobdenotter:
Hi @cdowdy, thanks for the pointer! I've added an issue for this, so we can fix all of these at once, sometime in the near future. See #3976.

PR #3979, opened 2015-08-16
Rewrite of getContent (search functionality) to use new storage layer

rossriley:
This PR covers the search functionality for querying implementing the functionality that was formerly in the `Storage.php` file.

The work is now primarily split out into `SearchConfig` and `SearchWeighter` classes which handle the bulk of the functionality. The `SearchQuery` class extends the `SelectQuery` class from the previous PR and adds a few small modifications.
### Native Fulltext Search

Whilst the search weighting functionality is a good cross-platform solution for smaller corpuses of text, doing this in PHP is significantly slower and lacks some of the more advanced features of native fulltext searches. 

So this PR also introduces a new concept of native searches which developers with more advanced user can specify at query time. The api remains similar `setcontent results='pages/nativesearch', {filter: 'lorem ipsum'}` the `nativesearch` directive will switch into fulltext mode where the database platform supports it and all the weightings from the configuration will be applied to the query. 

Same caveats as with the last PR in that some of the finessing will follow in future PRs including pagination and some improvements to the public query API. Also native search currently only supports Postgres, but MySQL will follow shortly, SQLite will fall back to a standard weighted search.

### DIFF
--- a/src/Provider/QueryServiceProvider.php
+++ b/src/Provider/QueryServiceProvider.php
@@ -7,6 +7,7 @@ use Bolt\Storage\Query\Query;
 use Bolt\Storage\Query\QueryParameterParser;
 use Bolt\Storage\Query\SearchConfig;
 use Bolt\Storage\Query\SearchQuery;
+use Bolt\Storage\Query\SearchWeighter;
 use Bolt\Storage\Query\SelectQuery;
 use Silex\Application;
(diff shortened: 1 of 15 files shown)


GawainLynch:
Le awesome!

bobdenotter:
zoidberg

:+1:

PR #4001, opened 2015-08-21
Refactor Wars Episode V: The legacy strikes back

rossriley:
1. Injects a legacy service into all new content entities via an event listener.
2. Adds some empty handlers for missing query directives.


GawainLynch:
star-wars-episode-iv-a-new-hope-973-1284025834603

Issue #3796, opened 2015-07-13
I'm willing to rewrite French translation from scratch


fabschurt:
Hi folks,

I'm building a Bolt website for a client, and I'm not entirely satisfied with the default core French translation. I can tell that a fair amount of work has been put into it, but there are some problems:

* Various misspellings.
* Various language errors and sometimes weird (kind of _old fashioned_) grammatical and syntactical structures.
* Missing translations.
* The translation files' format seems a bit dated, and does not reflect the new format used in the reference `en_GB` translation files (which use YAML hash nesting instead of full-inline string keys).

So for now I've duplicated `vendor/bolt/bolt/app/resources/translations/en_GB/messages.en_GB.yml` to `app/resources/translations/fr_FR/messages.fr_FR.yml` and I'm in the process of fully translating it, starting from scratch.

It's gonna be some bit of work and I think other French Bolt users could benefit from it, so I was wondering if you'd be considering a pull request for this brand new translation or if it's of no interest to you. It doesn't change things so much for me as I'm doing this translation for my client no matter what, but if I know that I'll ultimately push it upstream, I'd put extra care and spell checking into it.

For the record, I have really fair language skills in both English and French, and my relatives regularly call me a _grammar nazi_, so I guess I have the right profile for the job.

rarila:
Your input is really welcome and we happily merge it!

Looking at the French files it looks like the French translation was orphaned and just a leftover from an old Bolt version.

_(I hope I or someone else finds some time to push the conversion to keybased translations forward in the future)_

fabschurt:
Excellent, I'll gladly submit a PR for this work then :+1:

When this is done, I'm willing to offer my help about transitioning to key-based translations.

Just a quick question: can we consider the translation files in the core `app/resources/translations/en_GB/` directory to be up-to-date and can they be used as a reference for other translations? As I mentioned earlier, I copied the `en_GB` files to start from scratch, but maybe it's not the best way to go?

rarila:
Ok, the most up-to-date translation is for sure ``en_GB`` but I doubt that it is 100% complete as devs (I won't exclude me) probably are often lazy and just add some ``__('Whatever text')`` without doing a correct keyed translation. But missing translations can easily regenerated using "Translations/…" in the backend.

[...]

PR #4002, opened 2015-08-21
[I18N] French France translations

pygillier:
Hi, 

here are fr_FR translations based on latest (afaik) en files. 
This PR is linked to #3796


GawainLynch:
@pygillier  We are absolutely migrating to lang_LOCALE, 
@rarila How do we procede? Do we want to just remove the current [French base](https://github.com/bolt/bolt/blob/master/app/resources/translations/fr/) translations?

pygillier:
Good news :) 

I'm checking several tools (ala launchap, lokali.se & others) to see if translations can be automated.

rarila:
@GawainLynch I would say so, as the current one is AFAIK pretty outdated and rather useless.

bobdenotter:
This one is good to go, right?

GawainLynch:
Oui! We just need to swap out the non-locale `fr` files in the process

PR #4011, opened 2015-08-24
Avoid favicon 404's when not set.

SahAssar:


GawainLynch:
`<pre-coffee>`
Won't this cause a (HTML based) override for static favicon.ico users? i.e. shouldn't you do something like this in the web server config (Nginx in this case):
```
    location = /(?:favicon.ico|robots.txt) {
        access_log off;
        log_not_found off;
    }
```
`</pre-coffee>`

SahAssar:
Hmm, yes, I didn't consider the folks who use a favicon but not the config.yml favicon setting. The problem with doing this in the server config is that browsers still try to fetch the favicon, and still get a 404.

Perhaps only do this when the favicon setting is set to false?

GawainLynch:
To be fair, a lot of stuff will 404… Apple stuff, map files and the host. The browser just moves on, and if the site owner is that worried… install a favicon :grin: 

^ Personal opinion

SahAssar:
Yeah, the reason that I only did this for the favicon is because it's the only one of those (that I know of) that is used by all browsers and the only one that impacts pagespeed and related scores (which affects SERP ranking).

Also, it's the only one that we already have a setting for that could be used to decide when to do this. Anyway, this might be more suited in the individual theme and it's config.yml than in bolt proper.

I mainly found this while looking over what changes I've made in my personal bolt version, so if it does not fit in bolt proper feel free to close the PR :smile:

bobdenotter:
Make an extension! ;-)

PR #4016, opened 2015-08-25
Move swiftmailer to its own service provider

CarsonF:


GawainLynch:
Someone had to finally do it! :+1:

Issue #3466, opened 2015-04-30
Can't update non published entries


pygillier:
When I try to update an entry (default content-type) which is in draft state, the updates can't be saved.

After checking and debugging, it seems to be related to endpoint.

In file src/Storage.php, line 1245: check is done on status & endpoint : status is unset and $app['end'] is set to async. So, the condition will validate and set status in query to "status = published".

So, no object will be returned and update will fail.

Tested on 2.1.8 and it seems to be linked to this commit : cfbb0ec9d51563bf3a117acf98d395be61b71b54

bobdenotter:
@pygillier Looking at it, i think you're right. Expect a fix soon.

bobdenotter:
Fix can be found here: #3467

pygillier:
Hi, 

is there a ETA on releasing this patch ? ATM, I can't use my instance because of this bug.

fabschurt:
This problem is still present in `v2.2.4`. I doubt that 0037c11 or 85fe99c are related to this problem at all (even though I concede that I'm having a hard time figuring out the big picture here). All I know is that if I revert cfbb0ec, it solves the problem.

I'm kind of puzzled by cfbb0ec. Its title reads **Make sure we set the status correctly for 'async' requests.**, so I'd expect the commit to fix a problem related to the `async` end, but AFAIK it introduces a bug instead. Moreover, is there actually a point in forcing the `published` status criteria for other ends than `frontend`?

bobdenotter:
Hi,

I assumed this was fixed. I installed a fresh new Bolt 2.2.4 to try it, and I can't reproduce this behaviour. This is what I did: 

![screenshot_2015-07-21_19_51_25](https://cloud.githubusercontent.com/assets/1833361/8808351/602ce4f4-2fe2-11e5-9dfa-123de51a5c4a.png)

![screenshot_2015-07-21_19_51_43](https://cloud.githubusercontent.com/assets/1833361/8808354/63f5a364-2fe2-11e5-867e-15f84ad079e0.png)

![screenshot_2015-07-21_19_51_55](https://cloud.githubusercontent.com/assets/1833361/8808355/66f1e1fe-2fe2-11e5-9b91-d25dafc8becf.png)

Are you sure it's not working for you? If so, could you give me more details on how to reproduce it exactly? :-)

fabschurt:
[...]

PR #4024, opened 2015-08-26
Fix record retrieval for ownership checking

fabschurt:
The missing parameter was preventing `Bolt\Storage::getContent()` to bypass the forced `published` status for retrieved records.

I've seen this "trick" being used in `Bolt\Controllers\Backend` for example.

Please see #3466 for the full discussion.


bobdenotter:
Ah, right!! Good catch, @fabschurt :+1:

bobdenotter:
Green lights, merging in!

PR #4025, opened 2015-08-26
Show profile button on users page, if not allowed to edit other users than yourself.

### LINKED ISSUE #4008: Small improvement: "edit profile" button in "show users" screen.
So, if a user _does_ have permission to view the "users" page, but they _don't_ have the permission toe edit themelves, we should provide a link to the "profile" page, which they _do_ have access to, for convenience. 

screen_shot_2015-08-24_at_15_18_17

bobdenotter:
Fixes #4008.

(The fix for 2.2 is already in. Committed that directly, sorry!)

### DIFF
--- a/app/view/twig/users/_userlist-actions.twig
+++ b/app/view/twig/users/_userlist-actions.twig
@@ -1,4 +1,5 @@
-<div class="btn-group">
+{% if isallowed('useredit') %}
+    <div class="btn-group">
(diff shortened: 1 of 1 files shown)


GawainLynch:
BPFL says BDFL is awesome and can't make mistakes! :wink: 

Either way, tests aren't looking.

PR #4026, opened 2015-08-28
[2.2] Fix changelog single item links

### LINKED ISSUE #4009: Changelog 'next' & 'previous' buttons don't work as expected in Bolt 2.2.x
when trying to navigate between changelog entries, the behavior they got was unexpected.
Essentially, when they try to use the next & previous buttons to navigate, Bolt redirect them to the page with all the changelog entries rather than showing the next or previous change.
After talking to @GawainLynch, my understanding is that this has been solved in 2.3 but not back ported to 2.2.x

GawainLynch:
Fixes #4009


bobdenotter:
Whoop! :+1:

PR #4039, opened 2015-09-01
Don't add base snippets on async routes

### LINKED ISSUE #4034: Slug generator adds extra tail after record saving
i just created a record, then i save it and slug is damn long.

And the second wish : do something with long slugs, controll buttons go under the right panel.

GawainLynch:
Fixes #4034


CarsonF:
All good in the hood

PR #4040, opened 2015-09-01
Nut vs PHPUnit

GawainLynch:


CarsonF:
Shouldnt the color format be closed/reset on that echo timings line?

GawainLynch:
Hmm… Only a problem if you're echoing more on that buffer… But yeah.

GawainLynch:
Both updated

bobdenotter:
Whoop!

PR #4049, opened 2015-09-04
[I18N] Fix for quotes in JS messages

pygillier:
Fix for JS in extensions page. When in french, messages are broken because of quotes.


GawainLynch:
I love the irony of an error message causing an error :grin:

pygillier:
Yay :grin:

PR #4053, opened 2015-09-04
[RFC] Change used date-time format to a more i18n-friendly format

fabschurt:
Hey folks,

I don't consider this a bug or an improvement, it's just some proposed changes.

Basically, the `%b %e, %H:%M` date-time format is used throughout the backend. The "problem" is that it puts the month number before the day number, which is not the way to go in French for example, as we normally put the day number before the month number. I personally don't mind this so much, but I can see how this can be a problem for some of my picky clients.

So I propose to use the `%c` format instead, even though I know it's not ideal, because it's a bit verbose, but it's the only _l10n-friendly_ format available for `strftime()`.

What do you think?

### DIFF
--- a/app/view/twig/editcontent/_aside-status.twig
+++ b/app/view/twig/editcontent/_aside-status.twig
@@ -10,7 +10,7 @@
 
 <p class="lastsaved">
     {% if context.content.id is not empty %}
-    {{ __('Saved on:') }} <strong>{{ context.content.datechanged|localdate("%b %e, %H:%M") }}</strong>
+    {{ __('Saved on:') }} <strong>{{ context.content.datechanged|localdate("%c") }}</strong>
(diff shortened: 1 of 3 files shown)


GawainLynch:
@fabschurt I have tagged this as RFC and am going to absolutely defer to @bobdenotter (or anyone that isn't me) on this.

Basically, I get where you are going and I am nuts about ordering these things, but I totally suck at UI/UX. Bob's back on Monday and we can all nag him then :grin:

fabschurt:
Nice! :sunglasses:

rarila:
… and I'm probably back on wednesday ;-)

bobdenotter:
Looks good to me! The main advantage is that `%c` formats according to the locale. Whatever we set ourselves will be wrong for a number of locales.

bobdenotter:
I'll cherry-pick into 2.2.x, too.

PR #4068, opened 2015-09-08
Improvements to "install extension" screen.

### LINKED ISSUE #3981: [RFC] How to handle extensions with a lot of releases.
For extensions that have a bunch of releases, the list gets quite long.. 

screen shot 2015-08-17 at 16 00 36

How should we handle this? If it were up to me, I'd like to show just one button for the latest release, and a link to show "all other" releases. My gut feeling tells me, 99% of the time people should want to install the latest stable version of an extension.

### LINKED ISSUE #4067: Extensions installer: Clicking "versions" once, triggers the request twice, making it take twice as long.
See screenshot: 

screen_shot_2015-09-08_at_18_03_08

bobdenotter:
- Moved 'install new extension' to the top
- Cleaned up error messages (simpler and hopefully more robust)
- Less ajaxy requests for "Versions" (fixes #4067)
- Shows only "Latest stable release" by default, option for the other version (Fixes #3981)


GawainLynch:
This breaks exception handling:
screenshot from 2015-09-09 03-30-12

GawainLynch:
@bobdenotter a very simple test is to edit `src/Composer/Action/InstallPackage.php` and on line 106 add `throw new \Exception('Oh noes require me!');`

bobdenotter:
> This breaks exception handling:

The real culprit is the `<link>` that's added, but this PR neither causes nor fixes that. I added that try/catch **_because_** of this. Without it, it silently fails with just an error in the console. 

See this screenshot of what happens in `master`: 

screenshot_2015-09-09_08_09_52

As far as I can tell, exception handling was already broken. This PR just makes it more visible, and gives the user feedback, without failing silently.

GawainLynch:
OK… we'll pick on @CarsonF about the other one then!

GawainLynch:
OK, other thing fixed in #4071

PR #4071, opened 2015-09-09
Check the request for isXmlHttpRequest() in SnippetListener

### LINKED ISSUE #4070: [master] Some async routes still adding base snippets
This should have been fixed in #4039 but as noticed when testing #4068 we're still getting problems.

See the end of the trace (which is being sent in JSON so something knows we're on an AJAX route):
screenshot from 2015-09-09 03-30-12

@CarsonF any further ideas on putting a nail in this coffin permanently?

GawainLynch:
Fixes #4070


bobdenotter:
Huzzah!

PR #4073, opened 2015-09-09
Don't sort getContent in listing view, when the contenttype has a taxonomy that has a sortorder.

### LINKED ISSUE #3977: Taxonomy sort ordering on default listing pages incorrect
So I have a content type setup in 'contenttypes.yml' with a taxonomy and no sorting value set and then you setup the taxonomy as follows:
```yml
staff_position:
    name: Staff Position
    slug: staff_position
    behaves_like: grouping
    options: { head: "Head", dept: "Department Head", teacher: "Teacher", other: "Misc" }
    has_sortorder: true
```
When visiting the listing page for that content type which is merely looping around the records such as `{% for record in records %} Do stuff {% endfor %}` it does not apply the custom ordering and will only list them in the order they were created (newest first), despite the sort order and taxonomy values being set correctly when dumping the records that are returned.

If you add a simple `{% setcontent records = 'contenttype' %}` at the start of the listing template then it suddenly applies the sorting order correctly. Although this is sort of a fix, it makes the pagination stuff break obviously, so it's not ideal.

A few things to note:
* The presence of a pager or not seems to have no effect
* Removing all records in the taxonomy table and starting again made no difference
* Applying the taxonomy with only one option as mentioned above makes no difference
* The content type definitely has no sort value set in content types.yml

And the kicker - The back-end view seems to work fine in applying the custom sort ordering and filtering.

P.S  - yes many caches were cleared :hatching_chick: :hatched_chick:

bobdenotter:
Fixes #3977

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -256,6 +256,8 @@ class Frontend extends ConfigurableBase
         if (!$order = $this->getOption('theme/listing_sort', false)) {
             $order = empty($contenttype['sort']) ? null : $contenttype['sort'];
         }
+        // If $order is not set, one of two things can happen: Either we let `getContent()` sort by itself, or we
+        // explicitly set it to sort on the general/listing_sort setting.
         if ($order === null) {
             $taxonomies = $this->getOption('taxonomy');
             $hassortorder = false;
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -254,7 +254,23 @@ class Frontend extends ConfigurableBase
             $amount = empty($contenttype['listing_records']) ? $this->getOption('general/listing_records') : $contenttype['listing_records'];
         }
         if (!$order = $this->getOption('theme/listing_sort', false)) {
-            $order = empty($contenttype['sort']) ? $this->getOption('general/listing_sort') : $contenttype['sort'];
+            $order = empty($contenttype['sort']) ? null : $contenttype['sort'];
+        }
+        if ($order === null) {
+            $taxonomies = $this->getOption('taxonomy');
+            $hassortorder = false;
+            if (!empty($contenttype['taxonomy'])) {
+                foreach($contenttype['taxonomy'] as $contenttypetaxonomy) {
+                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
+                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
+                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
+                        $hassortorder = true;
+                    }
+                }
+            }
+            if (!$hassortorder) {
+                $order = $this->getOption('general/listing_sort');
+            }
         }
 
         $content = $this->getContent($contenttype['slug'], ['limit' => $amount, 'order' => $order, 'page' => $page, 'paging' => true]);


src/Controller/Frontend.php
```diff
@@ -254,7 +254,25 @@ public function listing(Request $request, $contenttypeslug)
...
+        // explicitly set it to sort on the general/listing_sort setting.
+        if ($order === null) {
+            $taxonomies = $this->getOption('taxonomy');
+            $hassortorder = false;
+            if (!empty($contenttype['taxonomy'])) {
+                foreach($contenttype['taxonomy'] as $contenttypetaxonomy) {
+                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
+                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
+                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
+                        $hassortorder = true;
```
GawainLynch: Tired and probably wrong… again!… but shouldn't we `break;` here?

bobdenotter: > Tired and probably wrong… again!… but shouldn't we break; here?

I tried that at first, but couldn't get it to work.. `break;` only breaks out the inner loop. `break 2;` gives this error message: 

```
Fatal error: Cannot break/continue 2 levels in /Users/bob/Sites/bolt-origin/src/Controller/Frontend.php on line 271
Cannot break/continue 2 levels
```

So, i gave up, and did it like this. If you can get it to work, That'd be cleaner than this solution.

GawainLynch: I just tested this here and the only way I could trigger this is due to the break being outside the `foreach` ergo inside the `if` instead.

Again, re-enforcing the Ultra Level Gawain Derp today, but do you have a simple reproducible test case I can start with?

bobdenotter: It throws the error, when i simply do this: 

```
                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
                        $hassortorder = true;
                        break 2;
                    }
```

GawainLynch: Ah, you don't need to `break 2` as you're only in one loop, correct?

This alone should work:
```php
        if ($order === null) {
            $taxonomies = $this->getOption('taxonomy');
            $hassortorder = false;
            if (!empty($contenttype['taxonomy'])) {
                foreach($contenttype['taxonomy'] as $contenttypetaxonomy) {
                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
                        $hassortorder = true;
                        break;
                    }
                }
            }
            if (!$hassortorder) {
                $order = $this->getOption('general/listing_sort');
            }
        }
```

bobdenotter: That just _adds_ the one line with the 'break', right? Doesn't make it much clearer, and the micro-optimisation doesn't make much difference.. 

How 'bout this? ;-)

```
        if ($order === null) {
            $taxonomies = $this->getOption('taxonomy');
            $hassortorder = false;
            if (!empty($contenttype['taxonomy'])) {
                foreach($contenttype['taxonomy'] as $contenttypetaxonomy) {
                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
                        goto end;
                    }
                }
            }
            $order = $this->getOption('general/listing_sort');
        }

        end:
```

GawainLynch: > goto end;

Did we just end up back in 1982 :stuck_out_tongue_winking_eye:

PR #4074, opened 2015-09-10
[2.2] Don't sort getContent in listing view, when the contenttype has a taxonomy that has a sortorder.

### LINKED ISSUE #3977: Taxonomy sort ordering on default listing pages incorrect

bobdenotter:
Fixes #3977 for `release/2.2`.

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -229,7 +229,30 @@ class Frontend
         // First, get some content
         $page = $query->get($pagerid, $query->get('page', 1));
         $amount = (!empty($contenttype['listing_records']) ? $contenttype['listing_records'] : $app['config']->get('general/listing_records'));
-        $order = (!empty($contenttype['sort']) ? $contenttype['sort'] : $app['config']->get('general/listing_sort'));
+
+        if (!$order = $app['config']->get('theme/listing_sort', false)) {
+            $order = empty($contenttype['sort']) ? null : $contenttype['sort'];
+        }
+
+        // If $order is not set, one of two things can happen: Either we let `getContent()` sort by itself, or we
+        // explicitly set it to sort on the general/listing_sort setting.
+        if ($order === null) {
+            $taxonomies = $app['config']->get('taxonomy');
+            $hassortorder = false;
+            if (!empty($contenttype['taxonomy'])) {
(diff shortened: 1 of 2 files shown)


src/Controllers/Frontend.php
```diff
@@ -229,7 +229,30 @@ public function listing(Silex\Application $app, $contenttypeslug)
...
+        // explicitly set it to sort on the general/listing_sort setting.
+        if ($order === null) {
+            $taxonomies = $app['config']->get('taxonomy');
+            $hassortorder = false;
+            if (!empty($contenttype['taxonomy'])) {
+                foreach($contenttype['taxonomy'] as $contenttypetaxonomy) {
+                    if ($taxonomies[ $contenttypetaxonomy ]['has_sortorder']) {
+                        // We have a taxonomy with a sortorder, so we must keep $order = false, in order
+                        // to let `getContent()` handle it. We skip the fallback that's a few lines below.
+                        $hassortorder = true;
```
GawainLynch: What… no `goto` here either :stuck_out_tongue_winking_eye:

GawainLynch:
:+1:

PR #4079, opened 2015-09-11
Ignore theme directory except for base-* and default

GawainLynch:
This prevents the :koala: from committing a theme when working on a core repo…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ extensions/*
 thumbs/
 vendor/
 phpunit.xml
+theme/*
+!theme/base-*
+!theme/default
 
 # Unless you want to store vendor's "state" in your Git repo.
 composer.phar


bobdenotter:
I'm a bit worried this will confuse end-users. People who fork/clone the bolt repo, and build their own project on it, might be confused why their theme isn't commited.. Or worse, They might not notice their theme isn't committed, and the accidentally wipe it.

GawainLynch:
True, but if they are doing a raw pull then they break other things.

GawainLynch:
```
$ git commit theme/my-cool-theme/
error: pathspec 'theme/my-cool-theme/' did not match any file(s) known to git.
```

bobdenotter:
Well, allright then. ;-)

PR #4083, opened 2015-09-11
Flex to the Future! (base-2016, new default theme)

SahAssar:
Flex to the Future

> ## Floats?! Where we're going, we don't need floats!

This is my proposal for a new default theme for bolt. It's based on a flexbox grid, and is compatible with IE11+ and other modern browsers.

Tested on Firefox, Chrome, IE11 and Spartan/Edge on windows.

Tested on Android, iOS and desktop Safari on browserstack.

> #### Marty McFlex's having the time of his life. The only question is -- what browser is it?

It will not work properly on IE10 (because of the combination of borderbox and flex), or any browser that does not support flexbox.

A demo is available at http://testa3.kx.nu/

Changes, comments and insults welcome!


rossriley:
Nice work :beers: 

One minor bug, on the showcases pages eg: http://testa3.kx.nu/showcase/paria-sunt-igitur there seems to be a problem with the aside column dropping below the content rather than floating.

bobdenotter:
Why not use the google-fonts directly from google? Saves a bunch of space in the distro and repo.

SahAssar:
Fixed @rossriley's issue (the code block was too wide.)

SahAssar:
@bobdenotter regarding the fonts, I had this conversation with @GawainLynch on IRC, thoughts?

[18:17] <gawainlynch_> SahAssar: The fonts… a plan?
[18:19] <SahAssar> gawainlynch: google font's have dodgy rendering in chrome on windows (still), using ones own font's kinda fixes this. Plus league spartan isn't on google fonts. But I can swap it out for something else and use google fonts if that's whats wanted.
[18:19] <SahAssar> I do think that a default theme should not depend on that service by default though.
[18:20] <gawainlynch_> Hmm… Well when it comes to the visual I have no preference, but it is ~750KiB in the repo
[18:20] <gawainlynch_> Yeah, true… probably should wait on Bopp to weigh in
[18:20] <SahAssar> gawainlynch: Yeah, That part I definitely understand.
[18:21] <gawainlynch_> A git clone of our repo is already heading for the 50MiB mark
[18:21] <SahAssar> Can I keep league spartan and remove the others? that gets the fonts down to 108KiB.

GawainLynch:
```
Bolt - Fatal error.
Class: Twig_Error_Runtime
Message: Key "3" for array with keys "0, 1, 2" does not exist in "extrafields.twig" at line 3
Code: 0
```

GawainLynch:
`BackendLemmingsCest::dashboardWithoutPermissionRedirectsToHomepageTest()` fails as these are missing (probably a test change):
* Recent Pages
* Recent Entries
* Recent Showcases

GawainLynch:
`FrontendCest::checkMenusTest()` fails as the menus are missing `first` and `active`

SahAssar:
Fonts are gone now, and it should pass all the strict variables tests. If this is changed to the default theme it will also need some slight changes to the tests, those are outlined it the branch tests-base-2016: https://github.com/SahAssar/bolt/commit/ce15589d1a18a3d90adf3aa782993fa095bf457c

fadoe:
A 404-Page is missing in the example:

http://testa3.kx.nu/entry/:

Bolt - Fatal error.
    Class: Symfony\Component\HttpKernel\Exception\NotFoundHttpException
    Message: The page could not be found, and there is no 'notfound' set in 'config.yml'. Sorry about that.
    Code: 0

;-)

GawainLynch:
Ummm…

> The page could not be found, and there is no 'notfound' set in 'config.yml'. Sorry about that.

No, it just means that the [`notfound:`](https://github.com/bolt/bolt/blob/release/2.2/app/config/config.yml.dist#L89) key isn't set.

```yaml
# The default content for the 404 page. Uses the 'record_template' template by
# default.
#
# Make sure this is set to an existing record, otherwise visitors will get an
# error-page when the request a non-existing page.
#
# Note 1: The record specified in this parameter must be set to 'published'
# Note 2: If you are logged on, and debug is set to 'true', you will NOT see the
#         404 page!
notfound: page/not-found
```

fadoe:
@GawainLynch Yes, this is right. And this is the problem why I can't see a 404 page.

SahAssar:
@fadoe I added a page with the slug.

This really isn't an issue with the theme though, since setting up a 404 page is something an editor does with a record.

fadoe:
@SahAssar Yes, it is a bolt configuration thing. But it can be that you created a special 404 page template and this is not displayed because the config key is missing.

Great work on this theme.

SahAssar:
@fadoe: Sorry, didn't get the intonation over the internet :smile:

bobdenotter:
Whoop whoop! :+1:

PR #4084, opened 2015-09-12
Session file security and logging

GawainLynch:
* Make sure we respect `session.save-path` settings for file permissions
* Log errors handling session files

### DIFF
--- a/src/Session/Handler/FileHandler.php
+++ b/src/Session/Handler/FileHandler.php
@@ -153,7 +153,6 @@ class FileHandler implements \SessionHandlerInterface
                 $this->fs->remove($file);
             } catch (IOException $e) {
                 $this->logger->error('Unable to remove session file ' . $file);
-                return false;
             }
         }
 
--- a/src/Session/Handler/FileHandler.php
+++ b/src/Session/Handler/FileHandler.php
@@ -1,6 +1,8 @@
 <?php
 namespace Bolt\Session\Handler;
 
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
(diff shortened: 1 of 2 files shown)


src/Session/Handler/FileHandler.php
```diff
@@ -122,6 +152,7 @@ public function gc($maxlifetime)
             try {
                 $this->fs->remove($file);
             } catch (IOException $e) {
+                $this->logger->error('Unable to remove session file ' . $file);
                 return false;
```
CarsonF: @GawainLynch Can you delete this return line? we shouldn't be stopping for one errored file.

src/Session/Handler/FileHandler.php
```diff
@@ -13,22 +15,33 @@
...
     {
         $this->fs = $filesystem ?: new Filesystem();
+        $this->logger = $logger ?: new NullLogger();
 
+        // @see http://php.net/manual/en/session.configuration.php#ini.session.save-path
+        $depth = 1;
+        $mode = 0600;
         $savePath = $savePath ?: sys_get_temp_dir();
 
         // Handle BC 'N;/path' and 'N;octal-mode;/path` which are not supported here
```
CarsonF: They aren't supported here? ;)

src/Session/Handler/FileHandler.php
GawainLynch: So we should always return `true`?

src/Session/Handler/FileHandler.php
```diff
@@ -13,22 +15,33 @@
...
-     * @param string     $savePath Path of directory to save session files.
-     *                             Default null will leave setting as defined by system temp directory.
-     * @param Filesystem $filesystem
+     * @param string          $savePath   Path of directory to save session files.
+     *                                    Default null will leave setting as defined by system temp directory.
+     * @param Filesystem      $filesystem
+     * @param LoggerInterface $logger
      */
-    public function __construct($savePath = null, Filesystem $filesystem = null)
+    public function __construct($savePath = null, Filesystem $filesystem = null, LoggerInterface $logger = null)
```
CarsonF: Can we flip logger and filesystem? We are way more likely to want to pass in a logger vs a filesystem

src/Session/Handler/FileHandler.php
GawainLynch: Hang on, let me `git blame` that for you ;-)

CarsonF:
Nice work mate :+1:

src/Session/Handler/FileHandler.php
CarsonF: Doesn't matter. SessionStorage doesn't check the returns. Idk what I would do with a false return.

src/Session/Handler/FileHandler.php
CarsonF: Well they weren't until now when you added support for them ha

PR #4087, opened 2015-09-12
Add [ $this, .. ] to callback.

### LINKED ISSUE #4086: In [2.3] snippetcallbacks in extensions no longer work.
In Bolt extensions you can add snippets, like this: 

```
$this->addSnippet(SnippetLocation::END_OF_HEAD, 'headsnippet');
```
In case of the [Sitemap](https://github.com/bolt/Sitemap) extension, it would use `'headsnippet'` as a callback, to insert a piece of generated HTML into the `<head>` section. 

On current master, it no longer recognizes it as a callback, and simply outputs the string `'headsnippet'`. Is this a regression in 2.3?

bobdenotter:
Fixes #4086

### DIFF
--- a/src/BaseExtension.php
+++ b/src/BaseExtension.php
@@ -387,8 +387,13 @@ abstract class BaseExtension implements ExtensionInterface
      */
     public function addTwigFunction($name, $callback, $options = [])
     {
+        // If we pass a callback as a simple string, we need to turn it into an array.
+        if (is_string($callback) && method_exists($this, $callback)) {
+            $callback = [$this, $callback];
+        }
+
         $this->initializeTwig();
-        $this->twigExtension->addTwigFunction(new \Twig_SimpleFunction($name, [$this, $callback], $options));
+        $this->twigExtension->addTwigFunction(new \Twig_SimpleFunction($name, $callback, $options));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/BaseExtension.php
```diff
@@ -450,7 +450,7 @@ public function addSnippet($location, $callback, $extraparameters = [])
             $callback = (string) $callback;
         }
 
-        $this->app['asset.queue.snippet']->add($location, $callback, $this->getName(), (array) $extraparameters);
+        $this->app['asset.queue.snippet']->add($location, [$this, $callback], $this->getName(), (array) $extraparameters);
     }
```
GawainLynch: Ah, you might want to check for a string first, as you can (and I already do) pass in a callback that is not in `Extension.php`.

Also probably worth checking if the string is a method in `$this` as is can also be parsed a pre-built html element.

bobdenotter: I had that at first, but since `addTwigFunction` and `addTwigFilter` don't check for this either, I thought I wouldn't have to here either. I'll fix this, and add checks at the other two functions as well.

GawainLynch: Ah yeah, that's right. That is why I started using `$this->app['twig']->addExtension()`

GawainLynch:
:koala: 0 — @bobdenotter: 1

PR #4081, opened 2015-09-11
Fix to #3991 – Geolocation snaps pin to matched address…

rarila:
Fixes #3991 
@bobdenotter Here to test: Use it that way or remove the snap checkbox and go with what option?

PR #4093, opened 2015-09-15
Geolocation snaps pin to matched address - Porting #4081 to Bolt 2.2

bobdenotter:


bobdenotter:
Labeled as [WIP], because build chain is broken. Modernizr removed their 'modernizr-latest.js' after the update to 3.0. :-/

bobdenotter:
Should be GTG now.

PR #4099, opened 2015-09-16
Let's bump the version to '2.3.0 alpha 2', now we're stabilising.

bobdenotter:


CarsonF:
@bobdenotter Can we set that to `dev` next go round?

PR #4104, opened 2015-09-17
Updated toolchain

rarila:
- All tools bumped to latest version
- Move to plain Modernizr generation
- Use 'modernizr-' prefix for Modernizr classes
- Add hack to prevent postcss/autoprefixer bug from triggering


bobdenotter:
Works like a Mo-Fo! :+1:

Issue #4109, opened 2015-09-18
Pager doesn't work with array as query parameter. 


hugin2005:
When query with array as parameter is provided it seems that pager generates the string value "Array" instead of the parameter values when switching pages. I managed to solve this problem using the http_build_query() and I will submit a pull request with this change.

bobdenotter:
Closed by #4110

PR #4110, opened 2015-09-18
Allow non-strings as query parameters with pager. issue #4109

hugin2005:
Pager doesn't work with array as query parameter. This issue is solved with help of http_build_query().

### DIFF
--- a/src/Pager.php
+++ b/src/Pager.php
@@ -52,14 +52,9 @@ class Pager extends \ArrayObject
         } else {
             unset($parameters['page']);
         }
-        array_walk(
-            $parameters,
-            function (&$item, $key) {
-                $item = "$key=$item";
-            }
-        );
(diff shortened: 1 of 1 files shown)


bobdenotter:
I had never heard of `http_build_query()`. Thanks! :+1:

rarila:
@bobdenotter Oh, it's rather new, just since pho 5 ;-)

hannesl:
It would be lovely if this got ported to 2.2 – how can I help? I believe the code in question is identical in that branch.

bobdenotter:
It's already backported to 2.2.x.. It'll be in the next release! :-)

bobdenotter:
See: 3853d1c132f12c2806f8237a62c0f4a5756e3a34

hannesl:
Ah, thanks!

PR #4111, opened 2015-09-18
[Logger] If exception is a client error reduce log level to warning

CarsonF:
[Logger] If exception is a client error reduce log level to warning. Implemented LoggerAware so the logger can be overridden later.


GawainLynch:
I have :joy: 

One `git rebase master` coming up!

PR #4112, opened 2015-09-19
Moved session configuration to SP

CarsonF:


GawainLynch:
Application.php… 20 lines shorter, thank you!

PR #4113, opened 2015-09-19
Users::token to use csrf provider instead of our own code

CarsonF:
@gawainlynch can this change be applied to `AccessControl` code as well?

### DIFF
--- a/src/Twig/Handler/UserHandler.php
+++ b/src/Twig/Handler/UserHandler.php
@@ -86,6 +86,8 @@ class UserHandler
      *
      * @see \Bolt\Authentication::getAntiCSRFToken()
      *
+     * @deprecated Since 2.3, to be removed in 3.0. Use Symfony forms instead.
+     *
      * @return string
      */
     public function token()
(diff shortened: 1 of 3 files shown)


bobdenotter:
Should we also remove this section from `config.yml`? https://github.com/bolt/bolt/blob/master/app/config/config.yml.dist#L331-L342

Or, now that I think about it, this is the part thats' needed for `AccessControl`, right?

GawainLynch:
> Or, now that I think about it, this is the part thats' needed for AccessControl, right?

Currently correct.

CarsonF:
Which is my question :) Should that code be updated as well or am I missing something?

GawainLynch:
The CSRF tokens I don't trust for authentication purposes myself. One of the things I wanted to achieve in 2.3/2.4 was the breaking out of `AccessControl` as much as possible to make a rip-and-replace of the authentication stack easy in v3.

CarsonF:
Fine with me. Let's ship this :shipit:

bobdenotter:
Whoop!

PR #4115, opened 2015-09-19
Replaced references to Bolt\Application with Silex\Application where possible

CarsonF:
Depends on #4114 

Replaced references to Bolt\Application with Silex\Application where possible. 

Excludes extensions which have it type hinted in the interface :(

`Application::DEFAULT_LOCALE` is next :gun:

### DIFF
--- a/index.php
+++ b/index.php
@@ -42,7 +42,7 @@ if (php_sapi_name() === 'cli-server') {
 }
 
 /**
- * @var \Bolt\Application $app
+ * @var \Silex\Application $app
  */
 $app = require_once __DIR__ . '/app/bootstrap.php';
 $app->run();
--- a/src/AccessControl/Login.php
+++ b/src/AccessControl/Login.php
@@ -2,10 +2,10 @@
 namespace Bolt\AccessControl;
 
 use Bolt\AccessControl\Token\Token;
-use Bolt\Application;
 use Bolt\Storage\Entity;
 use Bolt\Translation\Translator as Trans;
 use Hautelook\Phpass\PasswordHash;
+use Silex\Application;
 use Symfony\Component\HttpFoundation\Request;
 
 /**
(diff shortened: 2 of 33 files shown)


CarsonF:
Go home Scruitinizer you're drunk

GawainLynch:
Scrutinizer is drunk… You want to see what [killed Travis](https://github.com/bolt/bolt/pull/4115/files#diff-d73b469cdd14c7e25186b6050caa0fffR102).

rarila:
@CarsonF Them I'm drunk, too. Because I also see code twice ;-) :recycle:

CarsonF:
Ok reverted that. I'll handle it in the other branch. :shipit:

GawainLynch:
OK, so final thing/question for me. *What are we standardising on, and why?*

I've generally tried to follow — 
* If we're extending a Silex/Symfony class/function, we use `Silex\Application`. 
* The class/function would operate on a Silex Application object independently of Bolt, we use `Silex\Application`
* If I am passing `$app` to an extension function that uses something like `getContent()`, see above
* `use Bol\Application`

GawainLynch:
i.e can we decide and put a short documentation note on the internals section in docs.

CarsonF:
Yeah, sorry for the lack of context. 

I believe `Silex\Application` should be used everywhere. This is because all usages of `Application` are really just a DI container functionality. What we are doing right now is like typehinting `Foo` when we could be typehinting `FooInterface`. 

For GMO, we have some sites that use bolt and some that don't (same repo/framework). We also extend  Application for the same reason as Bolt (to add our services). So we have two Applications now: `Gmo\Application -> Silex\Application` and `Gmo\BoltApplication -> Bolt\Application -> Silex\Application`. All of our services have to be applied to both, and it kinda sucks.

The end goal for me is to have (something like) a `BoltServices` class that connects all of Bolt's SPs. This _single_ class would be injected into `Bolt\Application`, which just makes `Bolt\Application` a convenience class.

GawainLynch:
:+1: from me

bobdenotter:
Then I'm :+1: too.

PR #4123, opened 2015-09-22
Use tag names instead of slugs for autocomplete and tag cloud.

hannesl:
The tag autocomplete widget and the tag cloud both load and show slug values to the user. I'd argue that this is wrong.

Say you add an entry with the tags "naïvité, sjöar, zeeën". All is fine, and the slugs "naivite", "sjoear" and "zeeen" are saved to the DB along with the proper names.

But then, if you add another article, the tag cloud now contains the slugs rather than the proper names. The autocomplete also displays slugs. OK, we can live with that I guess? But what if you actually add these tags to the new article and save? Boom, now the database contains a new tag where the name is equal to the slug. That's a little harder to explain to my client.

I simply changed the async endpoints and JS code to use names instead of slugs. From what I can tell, this won't affect any other parts of Bolt. And it works just like it should.

I'll have a 2.2 version ready in a sec.


hannesl:
Backport for 2.2: https://github.com/hannesl/bolt/commit/f19e209b89d6b36cba2c7b0bc1db911e5b6bc03b

Should I make a separate PR for this?

hannesl:
My bad, forgot about tests. I also forgot to branch the change, so I'm starting over in a new PR.

Issue #868, opened 2014-01-30
Allow Tags with spaces in them


sbonardt:
When using a taxonomy of type 'tags' it is not possible to add tags with a space in it, e.g. 'Den Haag' or 'Belgian Beers'

The tagging function splits it automatically into two tags. Could tags be separated by comma?; or allow creation of tags with space be between quotes like "tags with spaces"?

XusBadia:
Hope this doesn't get forgotten. ;) I'm trying to get tags with spaces and copy and paste sometimes works, but sometimes doesn't, which is a bit frustrating.

bobdenotter:
I - personally - don't give a rat's ass about this. This will either get implemented because a (paying) client needs it, or somebody from the community steps up, and makes a Pull Request for it. \*hint\* \*hint\*

leup:
You could say that this is not a priority and so that you don't - personally - intend to implement it but that you don't give a rat's ass is not a valid argument. I had to go with a different CMS because of that issue and that was unfortunate because I really wanted to give Bolt a chance "in production". That did not happen. It was too much a job at the time to give a go to Bolt over another CMS. As you stated in the last development roadmap that you want Bolt to be as user friendly as possible, this is an issue that is definitely not user friendly (to my opinion). My comment here has nothing to do with the paying customer or community involvement part, I do understand your comment on that. Good luck with the hint ;)

XusBadia:
I understand your position and that there are more important issues to tackle right now, but I still think this is not a niche feature, as well as the Capital letters for tags.

Hope someone follows the *hint*, though!

bobdenotter:
@leup >  I had to go with a different CMS because of that issue and that was unfortunate because I really wanted to give Bolt a chance "in production".

Sorry, but I don't buy that. You're telling me that Bolt fit _all_ your needs perfectly, and you decided _not_ to use it because you couldn't use spaces in tags? That's blackmail, you know it, and I will have none of it. 

> As you stated in the last development roadmap that you want Bolt to be as user friendly as possible, this is an issue that is definitely not user friendly (to my opinion).

I do, but our collective time as developers is unfortunately limited. We choose to spend this time on stability fixes for 2.2, and adding new (often requested) features like inline editing and template-specific fields. You're absolutely right that there's room for improvement, but I'm sure the things we _did_ work on are more important for a lot of people. 

@XusBadia 

> Hope someone follows the hint, though!

[...]

PR #4124, opened 2015-09-22
Optionally allow spaces in tag names.

hannesl:
This change allows spaces in tags. See #868

I added the option `allow_spaces` in taxonomy.yml.dist and set the default to false, in order to not change the default behavior.

A very simple change that seems to work perfectly!


GawainLynch:
Fixes #868

sbonardt:
Nice! :+1:

bobdenotter:
Works like a charm!

hannesl:
Thanks!

@bobdenotter Do you think this belongs in 2.2? With the default being that spaces are not allowed, it doesn't actually change any default behavior.

sachbearbeiter:
Thank you very much!!!

PR #4125, opened 2015-09-22
Use tag names instead of slugs for autocomplete and tag cloud.

hannesl:
New version of #4123

The tag autocomplete widget and the tag cloud both load and show slug values to the user. I'd argue that this is wrong.

Say you add an entry with the tags "naïvité, sjöar, zeeën". All is fine, and the slugs "naivite", "sjoear" and "zeeen" are saved to the DB along with the proper names.

But then, if you add another article, the tag cloud now contains the slugs rather than the proper names. The autocomplete also displays slugs. OK, we can live with that I guess? But what if you actually add these tags to the new article and save? Boom, now the database contains a new tag where the name is equal to the slug. That's a little harder to explain to my client.

I simply changed the async endpoints and JS code to use names instead of slugs. From what I can tell, this won't affect any other parts of Bolt. And it works just like it should.


bobdenotter:
Hi @hannesl, 

I'll test it out tomorrow! If it works well, I'm sure a lot of people will enjoy having this feature in Bolt! Thanks!

GawainLynch:
Awesome stuff @hannesl… this has played with my OCD for a while :+1:

bobdenotter:
Great stuff! Thanks again, @hannesl

PR #4127, opened 2015-09-22
Fix small wrong checks

nantunes:


bobdenotter:
Looking good to me! Thank you! :+1:

PR #4128, opened 2015-09-23
Controller mounting tweaks

CarsonF:
Updated `Application::mount` to forward to our mount event. This ensures low priority (less than 0) controllers are last.

Cleaned up some of the Controller SP as well.

### DIFF
--- a/src/Events/MountEvent.php
+++ b/src/Events/MountEvent.php
@@ -21,9 +21,6 @@ class MountEvent extends Event
     /** @var ControllerCollection */
     protected $collection;
 
-    /** @var array Controllers grouped by priorities */
-    protected $priorities = [];
-
     /**
      * @param Application $app
      * @param ControllerCollection $collection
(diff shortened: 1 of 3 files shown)


CarsonF:
Can I get some thoughts on this? I would like App::mount to work well with our new mounting event, but I don't want to screw up any preconceived ideas. Maybe the connect method should be called immediately, then passed to listener to mount it to the collection?

GawainLynch:
Can you ping me about this tomorrow in IRC.

src/Events/MountEvent.php
```diff
@@ -67,10 +68,47 @@ public function finish()
...
+    /**
+     * Verifies collection is correct type and calls connect on providers.
+     *
+     * Note: This is the same code as {@see Silex\Application::mount}
+     *
+     * @param ControllerProviderInterface|ControllerCollection $collection
+     *
+     * @return ControllerCollection
+     *
+     * @throws LogicException If controllers is not an instance of ControllerProviderInterface or ControllerCollection
```
GawainLynch: @param
@throws
@return

[@see](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-18)

GawainLynch:
A rebase would be awesome if you could… 

Either way, from how I am reading it I think I prefer the current approach… K.I.S.S. at least for now. :+1:

CarsonF:
Ok I'm happy where this ended up :shipit:

GawainLynch:
The OCD is strong in this one!

PR #4129, opened 2015-09-23
Moved randomgenerator to its own service provider

CarsonF:
Moved randomgenerator to its own service provider


GawainLynch:
:koala: says :+1:, so do I

PR #4134, opened 2015-09-23
Updated nut.commands.add shortcut to accept callable. Also added example usage.

CarsonF:
Updated nut.commands.add shortcut to accept callable. Also added example usage.

```php
$app['nut.commands.add'](function ($app) {
    return new Command1($app);
});

$app['nut.commands.add'](function ($app) {
    return [
        new Command1($app),
        new Command2($app),
    ];
});
```


bobdenotter:
:+1:

PR #4141, opened 2015-09-25
Fixed missing variable permissions in _relations.twig. [2.2]

hannesl:
This is a nasty bug that only reveals itself if you have `strict_variables: true`. It appears to have been introduced in 2.2.9 or 10.

How to reproduce:

* Set `strict_variables: true` in `config.yml`.
* Add a relation to `pages` in the `entries` content type.
* Create a page.
* Create an entry that references the page.
* Edit the page. You should get *Variable "permissions" does not exist in "_base/_listing.twig" at line 3*.

I actually haven't tested this in master, so I don't know if it needs to be ported.

Happy friday!

### DIFF
--- a/app/view/twig/editcontent/_relations.twig
+++ b/app/view/twig/editcontent/_relations.twig
@@ -21,7 +21,7 @@
                 {% setcontent record = reltype ~ '/' ~ id %}
                 {% if record %}
                     {% set editable = isallowed('edit', record) %}
-                    {% include ['_sub/_listing.twig'] with {'excerptlength': 280, 'thumbsize': 54, 'compact': true, 'content': record} %}
+                    {% include ['_sub/_listing.twig'] with {'excerptlength': 280, 'thumbsize': 54, 'compact': true, 'content': record, 'permissions': context.permissions} %}
                 {% endif %}
             {% endfor %}
         </table>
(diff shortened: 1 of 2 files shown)


hannesl:
Seeing Travis spinning off a test job makes me think that the test environment should really run with `strict_variables: true` – right?

GawainLynch:
Yep, we set it for the acceptance tests right [here](https://github.com/bolt/bolt/blob/master/tests/codeception/extensions/CodeceptionEventsExtension.php#L148)

hannesl:
Nice! So in this case I guess we're missing an acceptance test.

GawainLynch:
Yep… Help much appreciated if you have the time. But I do plan to expand on those when I have some time myself.

GawainLynch:
:+1:

GawainLynch:
The issue doesn't seem to present itself in master… I'll try to get a test in soon.

hannesl:
2.2.12 is going to be amazing! :smile:

GawainLynch:
Yeah, we're pretty happy all around with the way 2.2 has turned out… 2.3 is the one that is truly amazing, but we have a big job to stabilise that over the next few weeks!

I am sure @bobdenotter won't be too far away from tagging 2.2.12 for everyone though.

PR #4142, opened 2015-09-25
Update version constraints for Symfony components

GawainLynch:
Guessing that Composer has changed something upstream as Monday's composer version just pulled Symfony 2.7 into my Bolt 2.2 install


CarsonF:
That seems like a bug no?

GawainLynch:
In which part?

CarsonF:
The change with composer

GawainLynch:
Ah, right… Context is everything :stuck_out_tongue_winking_eye: 

Yeah, the usual jumping around with upstreams though… Only so much benefit on not adapting in the interim.

bobdenotter:
I'm not sure what this would change.. On my test-setup i'm getting Symfony 2.6.11 components either way.

GawainLynch:
Maybe there was something wrong with Monday's Composer build that's now fixed.

PR #4162, opened 2015-09-29
Tweaking translation.

bobdenotter:


GawainLynch:
Disgaurding CI… They won't find a thing… and we have :koala: power

PR #4167, opened 2015-09-30
Update fonts, fixes #3586

### LINKED ISSUE #3586: Backend font doesn't have cyrillic letters, so it looks ugly
I suggest change Source Sans to any of sans-serif fonts from https://www.google.com/fonts , but with cyrillic support. I suggest Roboto or popular Open Sans. They both have nice unicode support.

Intendit:
This updates our Source Sans Pro font files to version 2.020 from [here](https://github.com/adobe-fonts/source-sans-pro/releases/tag/2.020R-ro/1.075R-it).

This fixes #3586, where there was some issues with Cyrillic characters.

### Before:
before-example
before-glyphcount

### After:
after-example
after-glyphcount


Intendit:
In the future we should perhaps also consider including/loading Source Han Sans for the appropriate locales.

rarila:
@Intendit Did you just copy fonts from there or are there any further steps needed? I think we should add this to our grunt toolchain, we same as we do with Fontawesome.

SahAssar:
@rarila: Just copied and renamed from `.ttf.woff2`/`.ttf.woff` to `.woff`/`.woff2`, but we could just use the names that adobe uses.

rarila:
@SahAssar Renaming is no problem

SahAssar:
@rarila Yeah, just thinking it makes more sense to keep the names the same as upstream, but whichever way works best for you :)

rarila:
I have that implemented in a local branch now. is that normal that the size of those 6 fonts went up from 158kB to 530kB? Or have i done something wrong?

SahAssar:
Yep, that's because they actually include all glyphs now. We could strip out different parts if we want, but then this issue is just going to reappear with those languages later.

I guess we could split up the font into their respective languages (russian, greek and so on) and then load them based on locale, if you think it's worth it.

GawainLynch:
I am wondering if this isn't well placed as a set of Composer packages? Just a thought… ideas?

rarila:
Splitting is probably not a good idea as there isn't a 1:1 relation from locale to font then and that gives a lot of trouble finding out wich one to use with what locale.

GawainLynch:
More so thinking `bolt/fonts-base` and `bolt/fonts-extra`, but yeah, either way it probably isn't workable.

rarila:
Another related question: why do we have Fontawesome in 6 different formats but Sans Open in only two?

SahAssar:
@rarila: no good reason, Woff and Woff2 covers our backend compatability, right? We don't support android 4 - 4.3, right?

rarila:
@SahAssar Don't know. I'm not into fonts that much. But if woff/woff2 is enough, we could shave off 653kB from FontAwesome.

rarila:
PRed #4169 - if someone want's to check if I included the correct fonts.

cdowdy:
> I have that implemented in a local branch now. Is it normal that the size of those 6 fonts went up from 158kB to 530kB? Or have i done something wrong?

@rarila if you're worried about large file size you should look into using [unicode-range](https://developer.mozilla.org/en-US/docs/Web/CSS/%40font-face/unicode-range) or subsetting the font (which you all probably cannot do because of the use of font-squirrel or using a font service)

SahAssar:
@rarila: looks good to me. Only difference is that you are using the `otf` sources where I used `ttf`, but that shouldn't make any difference.

SahAssar:
@cdowdy Yeah, but even then the question becomes what to remove. AFAIK Source Sans does not have a lot of "extras" like smallcaps or ligatures, and most of the filesize is from the sheer number of languages that are covered, which we want.

cdowdy:
@SahAssar yeah and if you use unicode-range it will only download the characters needed for that page or language

rarila:
@SahAssar Ah, no, it makes a difference. The ``ttf`` ones are quite a bit smaller. So what's the difference, should I use those?

SahAssar:
@rarila woff is just a container/compression format, so the difference is if it is a ttf or an otf font that is inside the woff. If the ttf's are smaller use them, both "formats" have equal support.

cdowdy:
> the ttf ones are quite a bit smaller. So what's the difference, should I use those?

the otf fonts have all the open type features included. the ttf ones dont

rarila:
For the other Fontawesome formats used: http://caniuse.com/#search=woff

As nobody should use the Android Browser for security reasons anyway, I think we are save with ``woff`` and ``woff2`` for 2.3. Gonna remove them in a later PR then.

rarila:
@SahAssar Ah, I understand

@cdowdy As you seem to know, you could save me from looking it up. Do these features justify those extra 55kB or is this just stuff we won't benefit from in Bolt?

SahAssar:
@cdowdy I didn't think Source Sans used/had any opentype features, but obviously they do... You know a lot more than me about this, so I'll just sit back and listen ;)

cdowdy:
@rarila that depends. @bobdenotter was concerned about ligatures and kerning and other things earlier which is why the original issue was/is still open. but Open Type typically has better support for international character sets, and of course ligatures (which source sans has little of)  

Really the decision comes down to what features and broad spectrum of languages you wan to support.  

[Adobe Open Type ](http://www.adobe.com/products/type/opentype.html)  

this is [for .net but gives a good example of how open type works[(https://msdn.microsoft.com/en-us/library/vstudio/ms745109%28v=vs.100%29.aspx)

cdowdy:
@SahAssar no you're right..... it has very few ligatures. but if you're dealing with multiple languages you'll want the glyph support of open type.   

haha see dealing with i18n is toooooooough. Especially when you are trying to strike a balance between file weight, no external services and broad support. Currently there isn't a magic bullet unless you use an external service which then complicated proxies or being behind the great firewall

rarila:
@cdowdy Hmm, most of the stuff described there seems to be of no use for a web page (e.g. Titling Capitals). The thing needed are ligatures - ttf hasn't them?!

cdowdy:
@rarila ttf has ligatures. but OTF has a broader range.. Give me a tic I'll find a post on this for you..  

as an aside here are the ligatures/glyphs in the [OTF font of source sans](http://www.fontsquirrel.com/fonts/source-sans-pro)

SahAssar:
@cdowdy Yeah, I know to basics of webfonts, but someday I need to sit down and properly read up on the different features and formats. 

My vote goes for the `otf`, since Adobe's article that was linked to specifically mentions better support for languages that we probably want to support such as Turkish, Polish, Cyrillic and Greek.

SahAssar:
Just leaving this here so that nobody merges this. #4169 should be merged instead, that one fixes the problem permanently.

PR #4169, opened 2015-09-30
Add Source Sans Pro fonts using grunt workflow

rarila:
<del>[WIP] Either #4167 has to be merged/closed first. On this one the commit of the font binaries is missing atm.</del>

Can be merged now


rarila:
I have used ``git://github.com/adobe-fonts/source-sans-pro.git#2.020R-ro/1.075R-it`` as source and not the proposed ``…#release``, to have a bit more control about the version used. Too bad that ``source-sans-pro`` isn't correctly registered so we don't get update hints by bower. But at least an update is easier in the future now.

bobdenotter:
Cool. We should check if these are "good" versions, though. Are the filesizes of these comparable to what we're shipping with Bolt now?

SahAssar:
@bobdenotter We had a bit of discussion about that in #4167, but these versions are straight from adobe-fonts github releases, so they should be good.

rarila:
@bobdenotter Sizes are increased, see the other thread

bobdenotter:
Whoop whoop!

PR #4172, opened 2015-09-30
Fix secondary nav element's class not being output in certain cases

fabschurt:
So, I don't know if there was a reason for the element's `class` not being output when `allowedamount == 1`, but it produces the following visual bug:

Illustrative screenshot

This fixes it, but I hope it doesn't have any side effects?

I sent a PR for `release/2.2` as well: #4173

### DIFF
--- a/app/view/twig/nav/_macros.twig
+++ b/app/view/twig/nav/_macros.twig
@@ -43,15 +43,13 @@
 
     {# Show stuff! #}
     {% if allowedany %}
-        {% if allowedamount == 1 and not force_submenu|default(false) %}
-            {% set item = allowedsingle %}
-            <li>
+        <li{% if class %} class="{{ class }}"{% endif %}>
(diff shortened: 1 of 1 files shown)


GawainLynch:
> This fixes it, but I hope it doesn't have any side effects?

It just crashed my Commodore 64, but the tape drive might be broken :grin:

fabschurt:
That's kind of a relief, as I was actually expecting nuclear-explosion-related issues :sweat_smile:

fabschurt:
`[BOOTING TROLLING ROUTINE…]` And that's the kind of reasons why I preferred the Amiga by the way :floppy_disk:

GawainLynch:
Pfft… you modern kids and your floppy disks! In my day, we booted servers uphill… both ways!

bobdenotter:
:+1: 

Thanks @fabschurt!

PR #4173, opened 2015-09-30
[2.2] Fix secondary nav element's class not being output in certain cases

fabschurt:
Please see #4172 for details.


bobdenotter:
Merging in, thanks!

Issue #3823, opened 2015-07-17
Field type time


errogaht:
Im think this is good idea to create new field type "Time"

rarila:
Probably a good idea.

geoff-winner:
https://github.com/bolt/bolt/pull/4186

Made a PR for this Feature

SahAssar:
Sorry, but this feels like an extension feature for now... It'd be pretty easy to make as an extension too. Closing this issue as part of a cleanup, if you feel this really should be part of bolt core please feel free to reopen! :)

PR #4186, opened 2015-10-02
Add the fieldtype 'time' with an hourly range dropdown menu, issue #3823

geoff-winner:
Suggested by issue #3823 https://github.com/bolt/bolt/issues/3823

A field type that has a drop down of different hourly times to choose from.

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -142,6 +142,34 @@ showcases:
             type: date
             default: "first day of last month"
             variant: inline
+        time:
+            type: select
+            values:
+              - 12 AM
+              - 1 AM
+              - 2 AM
+              - 3 AM
+              - 4 AM
+              - 5 AM
+              - 6 AM
+              - 7 AM
+              - 8 AM
+              - 9 AM
+              - 10 AM
+              - 11 AM
+              - 12 PM
+              - 1 PM
+              - 2 PM
+              - 3 PM
+              - 4 PM
+              - 5 PM
+              - 6 PM
+              - 7 PM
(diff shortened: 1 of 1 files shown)


geoff-winner:
Failed Travis CI... I'll look into it more tomorrow

GawainLynch:
Don't stress, those failures are not related to this. Travis is being, um, difficult for some strange reason :frowning:

bobdenotter:
Hi @geoff-winner, 

Thanks! 

This isn't really a field _type_, but more of a convenient example of how to use the existing `select`-fieldtype as a configurable solution for the lack of a dedicated `time`-field. I'm not sure this should be in 'core' like this. How about we add a section for this example to this page in the docs? https://docs.bolt.cm/howto/select-field-type

geoff-winner:
@bobdenotter,

Yeah I think that is a good idea. Let me know if you need me to help with that.

 I kind of just threw that in there after reading for a couple of minutes and figured I could get some feedback. Still figuring out how bolt works.

I figure I should mention I don't like brown M&M's but the fact of the matter is I actually do haha

bobdenotter:
> Let me know if you need me to help with that.

Sure thing. If you'd want to help, gladly! The documentation of Bolt is on Github as well: https://github.com/bolt/docs/blob/master/source/howto/select-field-type.md

Feel free to add this somewhere in the page as you seem fit as a PR, and we'll get it merged in!

> I figure I should mention I don't like brown M&M's but the fact of the matter is I actually do haha

;-)

GawainLynch:
Handled in https://github.com/bolt/docs/pull/252

PR #4189, opened 2015-10-02
Refactor Wars Episode VI: Return of the Koala

GawainLynch:
A long time ago in an RFC far, far away…

**Episode VI**

**Return of the Koala**

Ross Skywalker has returned to
his home planet of Tatooine in
an attempt to rescue his
friend Bopp Solo from the
clutches of the vile gangster
Dropbear the Hutt.

Little does Ross know that the
GALACTIC EMPIRE has secretly
begun construction on a new
armoured release v2.3 even
more powerful than the last stable
dreaded Bolt v2.2.

When completed, this ultimate
weapon will spell certain doom
for the small band of rebels
struggling to restore monolithic
principles to CMS development...

refactor wars


bobdenotter:
I remember reading something the other day about "It's best to keep your PR's small and modular, for easy review". ;-)

rarila:
PR rejected, description insufficient! Closed! :scream:

GawainLynch:
> I remember reading something the other day about "It's best to keep your PR's small and modular, for easy review". ;-)

Yes, and I wholeheartedly agree… But this branch was supposed to take a Saturday morning… some time back in August :grin:

jadwigo:
<iframe width="1152" height="698" src="http://play.starwars.com/html5/starwars_crawlcreator/?refOverride=http%3A%2F%2Fwww.starwars.com%2Fgames-apps%2Fstar-wars-crawl-creator&coreGameId=4fb065cca018b30157985123?null" frameborder="0" scrolling="no"></iframe>

GawainLynch:
Also this will require a `composer update` after merge if you're not @rossriley or @GawainLynch :grin:

bobdenotter:
Here goes!

app/bootstrap.php
```diff
@@ -34,6 +34,9 @@ function () {
         foreach ($autodetectionMappings as $autoloadPath => $configType) {
             if (file_exists($autoloadPath)) {
                 $loader = require $autoloadPath;
+                // Register a PHP shutdown function to catch early fatal errors
+                register_shutdown_function(['\Bolt\Exception\LowlevelException', 'catchFatalErrorsEarly']);
```
CarsonF: Moved into a `for` loop?

GawainLynch: It's what we call a :koala: derp…

src/Storage/Entity/EntitySerializeTrait.php
```diff
@@ -0,0 +1,44 @@
...
+        $data = [];
+        foreach ($this as $k => $v) {
+            if (strpos($k, '_') === 0) {
+                continue;
+            }
+            if (in_array($k, $this->_internal)) {
+                continue;
+            }
+            $method = 'serialize'.$k;
+            $data[$k] = $this->$method();
```
CarsonF: Should we note that `EntitySerializableTrait` can't be used without `MagicAttributeTrait`?

src/Storage/Entity/TemplateFields.php
```diff
@@ -0,0 +1,21 @@
...
+     * Getter for templates using {{ content.get(title) }} functions.
+     *
+     * @param string $key
+     *
+     * @return mixed
+     */
+    public function get($key)
+    {
+        $accessor = 'get'.ucfirst($key);
+        return $this->$accessor();
```
CarsonF: Is this for BC? Why not ``{{ content.title }}`?

src/Storage/Field/Type/DateTimeType.php
```diff
@@ -12,6 +17,31 @@ class DateTimeType extends DateType
     /**
      * @inheritdoc
      */
+    public function __construct(array $mapping = [], EntityManager $em = null)
```
CarsonF: This shouldn't work. The parent constructor is `final`.

src/Storage/Field/Type/DateType.php
```diff
@@ -38,7 +51,24 @@ public function query(QueryInterface $query, ClassMetadata $metadata)
...
+     * {@inheritdoc}
+     */
+    public function persist(QuerySet $queries, $entity, EntityManager $em = null)
+    {
+        $key = $this->mapping['fieldname'];
+        $value = $entity->get($key);
+
+        if (!$value instanceof \DateTime && $value !== null) {
+            $value = new Carbon($value);
+            $value::setToStringFormat('Y-m-d');
```
CarsonF: Probably shouldn't set the static string format here, because it will apply globally. Can you call `toString` manually with that format instead?

src/Storage/Field/Type/FieldTypeBase.php
```diff
@@ -52,36 +80,55 @@ public function persist(QuerySet $queries, $entity, EntityManager $em = null)
         $type = $this->getStorageType();
 
         if (null !== $value) {
-            $value = $type->convertToDatabaseValue($value, $qb->getConnection()->getDatabasePlatform());
+            $value = $type->convertToDatabaseValue($value, $this->getPlatform());
```
CarsonF: Doesn't DBAL handle this mapping internally?

src/Storage/Field/Type/DateTimeType.php
GawainLynch: class DateTimeType extends DateType
class DateType extends FieldTypeBase
abstract class FieldTypeBase implements FieldTypeInterface

src/Storage/Field/Type/DateType.php
GawainLynch: Too long ago to remember and you're the Carbon expert… I do remember there was something weird coming from DBAL with this though, but I can't remember what.

src/Storage/Field/Type/FieldTypeBase.php
```diff
@@ -97,4 +144,21 @@ public function getStorageType()
...
+     * @param mixed $value
+     *
+     * @return boolean
+     */
+    protected function isJson($value)
+    {
+        if (!is_string($value)) {
+            return false;
+        }
+        json_decode($value);
```
CarsonF: Do we need to stub out the error handler here so warnings aren't triggered?

src/Storage/Field/Type/DateTimeType.php
CarsonF: Derp

src/Storage/Field/Type/FieldTypeBase.php
GawainLynch: You're expecting PHP to act rationally? `json_decode()` just fails silently and you need to check `json_last_error()`… hence the reason `json_last_error()` is called belonw :stuck_out_tongue_winking_eye:

src/Storage/Field/Type/DateType.php
CarsonF: Actually you shouldn't even need to do it at all. The format is pulled from the Platform ([here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/DateTimeType.php#L57))

src/Storage/Field/Type/FieldTypeBase.php
CarsonF: Whatttttttt. I don't believe you. I've had to silence those calls before

CarsonF: https://github.com/composer/composer/blob/master/src/Composer/Json/JsonManipulator.php#L201

GawainLynch: http://php.net/manual/en/function.json-decode.php

CarsonF:
Awesome work @rossriley and @GawainLynch! :beers:

PR #4191, opened 2015-10-02
Don't override "templatechosen" if previously set.

### LINKED ISSUE #4190: The "Templates nub" in the toolbar gives the wrong results, sometimes.
See this screenshot with default contenttypes.yml, and the base theme.

screen_shot_2015-10-02_at_16_39_19

bobdenotter:
Fixes #4190.


GawainLynch:
:+1:

PR #4195, opened 2015-10-02
Repeating Fields Part 1: DB Structure

rossriley:
Keeping these nice and small, in separate parts. This handles the new tables and fields for repeating fields.


GawainLynch:
We are go! This is probably going to be the most asked for feature in this release :+1:

PR #4200, opened 2015-10-03
Timed storage forward port

GawainLynch:
Forward porting this to prevent regressions if the new layer's parts aren't cut over in this cycle.


bobdenotter:
Merging in!

PR #4203, opened 2015-10-03
Slug field: multiline and bootstrap conform

### LINKED ISSUE #4035: Slug field in backend doesn't wrap text inside
Probably it should be moved to textarea with rows='1' and javascript snippet will track the length of slug. 
Or maybe it could be done with hidden field and div: move text from input to div, since it's not used as input and all editings are being made with a popup.

rarila:
- Slug field gets now multiline when the slug is too long
- Bootstrap conform (no longer two buttons in input group)
- Full width
- Design changes

Fixes #4035


bobdenotter:
Schwoooop!

PR #4208, opened 2015-10-04
Fix error in taxonomy listing

### LINKED ISSUE #3803: Error in taxonomy listing
I use categories taxonomy in articles, but now it stop working listing of particular taxonomy. I get this error.
```
Controller "Bolt\Controller\Frontend::taxonomy()" requires that you provide a value for the 
"$request" argument (because there is no default value or because there is a non optional 
argument after this one).
```
It has something to do with routing I guess, but I have no idea, what to do. Because I didn't change routing for taxonomy listings. It's default.

Does anybody know what is the issue? Thanks.

GawainLynch:
Fixes #3803

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -299,7 +299,7 @@ class Frontend extends ConfigurableBase
      *
      * @return BoltResponse|false
      */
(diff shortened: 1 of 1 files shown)


bobdenotter:
Green lights!

PR #4218, opened 2015-10-04
Unlock slug on new record

### LINKED ISSUE #4217: Slug locked by default on new record (on master).
On master, simply create any new record with a slug to reproduce.  Notice that on the new record edit page, the slug is locked by default, and needs to first be unlocked to successfully add the record, otherwise it fails with a database error (slug's a not-null column).

From IRC for reference:

&lt;slick0&gt; rarila: ah, what is it?
&lt;rarila&gt; slick0: if (fconf.isEmpty) {field.lock.trigger('click');}
&lt;rarila&gt; slick0: but now we have discrete lock and unlock buttons => field.unlock.trigger('click');

Cheers!

rarila:
- Fixes #4217
- Make new dropdown localizable


GawainLynch:
Très bien!

PR #4230, opened 2015-10-05
Fix user ownership changes

### LINKED ISSUE #4226: Can't change ownership for a record from root account
There are 2 users registered -  Me (Mikie) and my client. I have root role:

When i'm trying to change record owner  for a record, created by another user and trying to save — it doesn't allow me. Error is "changing ownership is not allowed".

GawainLynch:
Fixes #4226

### DIFF
--- a/src/Storage/RecordModifier.php
+++ b/src/Storage/RecordModifier.php
@@ -109,16 +109,16 @@ class RecordModifier
         $formValues = Input::cleanPostedData($formValues);
         unset($formValues['contenttype']);
 
+        $user = $this->app['users']->getCurrentUser();
         if ($id = $content->getId()) {
             // Owner is set explicitly, is current user is allowed to do this?
             if (isset($formValues['ownerid']) && (integer) $formValues['ownerid'] !== $content->getOwnerid()) {
-                if (!$this->app['permissions']->isAllowed("contenttype:{$contentType['slug']}:change-ownership:$id")) {
+                if (!$this->app['permissions']->isAllowed("contenttype:{$contentType['slug']}:change-ownership:$id", $user)) {
                     throw new AccessControlException('Changing ownership is not allowed.');
                 }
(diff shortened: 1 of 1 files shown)


src/Storage/RecordModifier.php
```diff
@@ -109,16 +109,16 @@ private function setPostedValues(Content $content, $formValues, $contentType)
         $formValues = Input::cleanPostedData($formValues);
         unset($formValues['contenttype']);
 
+        $user = $this->app['users']->getCurrentUser();
         if ($id = $content->getId()) {
             // Owner is set explicitly, is current user is allowed to do this?
             if (isset($formValues['ownerid']) && (integer) $formValues['ownerid'] !== $content->getOwnerid()) {
-                if (!$this->app['permissions']->isAllowed("contenttype:{$contentType['slug']}:change-ownership:$id")) {
```
CarsonF:
```php
$this->app['users']->isAllowed();
```
better or no?

GawainLynch: During the controller refactor we decided that Bolt\Users needed to be free of such things…

CarsonF: Ah right. Can you mark that as deprecated?

GawainLynch: I can… But I don't know that I made a complete accurate representation of our discussion in my comment, ergo I don't know we are specifically deprecating it or not…

The discussion went something along the lines of "request cycle in Bolt\Users === bad… stop using outside classes/things/stuff" and we then moved a lot/most of the permission calls directly to the permission class. 

Knocking it on the head without a clear direction feels a bit wrong to me… so we probably need to step back and look at that… and not in a tiny bug-fix branch… but more than happy to hash it out!

CarsonF: Point taken :shipit:

Issue #2275, opened 2014-12-05
showimage function: width and height attribute


nlemoine:
When using the `showimage` template tag. The image attributes used are set to the size passed to the function. For example, I want a maximum width of 800 pixels:
```
record.image.filename|showimage(800, 9999, 'r')
```
Will output:
```
<img src="http://path/to/image.jpg" width="800" height="9999" />
```

As the "r" filter resizes given a maximum width or height, the other attribute should be set accordingly, preserving original aspect ratio. If I have a 1600x700 image with the same parameters as the example above, the actual function will ouput:
```
<img src="http://path/to/image.jpg" width="800" height="9999" />
```
Whereas it should output:
```
<img src="http://path/to/image.jpg" width="800" height="350" />
```
<hr />
I also noticed a little mistake in the 2.0 docs:

![capture d ecran 2014-12-05 a 15 13 27](https://cloud.githubusercontent.com/assets/2526939/5316895/513b8e46-7c91-11e4-9e66-0ddcf7009a01.png)

You can't pass an image array directly, you have to pass the filename:
```
{{ record.photo.filename|showimage(800, 600) }}
```
It would actually be nice to pass the image array instead of filename. Or both.

SahAssar:
Just commenting here since I just opened a PR regarding this:

To get proper height/width variables in a showimage has been supported for some time it seems, but only when calling with a filename. For example `record.image.filename|showimage(800, 9999, 'r')` Would work if called like this instead even before the "fix": `record.image.filename|showimage(800, 0, 'r')`. What the fix changes is that you can call it with an image from an imagelist directly (since it checks for an array), or with `record.values.image|showimage(800, 0, 'r')` if you want it to output a proper alt tag.

bobdenotter:
Closed by #4231

PR #4231, opened 2015-10-05
Fix alt and title tags for popup & showimage, height/width for showImage

### LINKED ISSUE #2866: image ALT attributes
Currently if you use  
```html
{{ record.image|popup(100, 100, "r") }}  
<!-- markup below -->
<a href="/bolt/thumbs/1000x750r/filename.jpg" class="magnific" title="Image: filename.jpg">
  <img src="/bolt/thumbs/100x100r/filename.jpg" width="100" height="100">  
</a>
```
 or 
```html
{{ record.image|showimage(800, 600) }}   

<img src="/bolt/thumbs/800x600c/filename.jpg" width="800" height="600">
```  
There are no alt attributes present. We need these for accessibility purposes and a less significant need of being a valid image tag.  

Most instances a file name can be used without the .jpg/png/gif extentions (if the file name actually describes the image ie. alt="bolt logo") but that isn't very descriptive and in some cases not valid alt attribute text. There is also no place to add a description for the image. Only an option to rename the image.   

I understand this gets into the murky waters of media management and there are no quick fixes ( besides possibly adding a description field to the image in the database).  

more info on alt attributes for anyone interested:  
[web aim atl text](http://webaim.org/techniques/alttext/)
[w3 accessibilty](http://www.w3.org/standards/webdesign/accessibility)

### LINKED ISSUE #4022: twig popup() while $filename is an array
twig imagehandler popup expects a string,
however docs suggest to serve it an image field (array) ``{{ image|popup(200,200,'c') }}``

The imagehandler does pass the filename (array) to its thumbnail method
https://github.com/bolt/bolt/blob/release/2.2/src/Twig/Handler/ImageHandler.php#L183

The thumbnail method extracts the name if it is an array
https://github.com/bolt/bolt/blob/release/2.2/src/Twig/Handler/ImageHandler.php#L298

And then back in popup method, it tries to use filename (still array) for title
https://github.com/bolt/bolt/blob/release/2.2/src/Twig/Handler/ImageHandler.php#L187

SahAssar:
Fixes #4022 and #2275.

All credit for for the title fix to @GDmac which I grabbed from #4022 since I got impatient after waiting for a month ;)

This also seems to fix #2866 if I am not wrong...

### DIFF
--- a/src/Twig/Handler/ImageHandler.php
+++ b/src/Twig/Handler/ImageHandler.php
@@ -248,6 +248,8 @@ class ImageHandler
                 $alt = $filename['alt'];
             } elseif (isset($filename['title'])) {
                 $alt = $filename['title'];
+            } else {
+                $alt = '';
             }
 
             if ($width === 0 || $height === 0) {
@@ -255,7 +257,7 @@ class ImageHandler
                     $filename = isset($filename['filename']) ? $filename['filename'] : $filename['file'];
                 }
 
-                $info = $this->imageInfo($filename);
+                $info = $this->imageInfo($filename, false);
 
                 if ($width !== 0) {
                     $height = round($width / $info['aspectratio']);
--- a/src/Twig/Handler/ImageHandler.php
+++ b/src/Twig/Handler/ImageHandler.php
@@ -200,7 +200,7 @@ class ImageHandler
             
             if (is_array($filename) && isset($filename['alt'])) {
                 $alt = $filename['alt'];
-            }else {
+            } else {
                 $alt = $title;
             }
 
@@ -246,7 +246,7 @@ class ImageHandler
 
             if (isset($filename['alt'])) {
                 $alt = $filename['alt'];
-            }elseif (isset($filename['title'])) {
+            } elseif (isset($filename['title'])) {
                 $alt = $filename['title'];
             }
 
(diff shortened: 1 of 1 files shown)


src/Twig/Handler/ImageHandler.php
```diff
@@ -184,16 +184,34 @@ public function popup($filename = '', $width = 100, $height = 100, $crop = '', $
...
+                    } else {
+                        // fallback to filename as provided (string)
+                        $title = sprintf('%s: %s', Trans::__('Image'), $filename);
+                    }
+                }
+            }
+            
+            if (is_array($filename) && isset($filename['alt'])) {
+                $alt = $filename['alt'];
+            }else {
```
GawainLynch: Should be:
```php
} else {
```

src/Twig/Handler/ImageHandler.php
```diff
@@ -226,7 +244,17 @@ public function showImage($filename = '', $width = 0, $height = 0, $crop = '')
             $width = intval($width);
             $height = intval($height);
 
+            if (isset($filename['alt'])) {
+                $alt = $filename['alt'];
+            }elseif (isset($filename['title'])) {
```
GawainLynch: Should be:
```php
} elseif (isset($filename['title'])) {
```

bobdenotter:
Looks good to me!

PR #4232, opened 2015-10-05
Remove the extensions config menu as per #3461

### LINKED ISSUE #3461: Request simpler extension menu
I think the extension menu item (under Extra) can be 'much' simpler. I always click the wrong item when I want to configure something or install a new extension. Why not replace Extra with Extensions?

The submenu now contains an item to view/install extensions and a second item for configuring extensions. But I can also choose view/install extensions to configure them. I think it is an improvement when these two items are combined in to one. On the overview page I can view information about different versions, a manual, a button to configure extensions or uninstall them. That is all a user needs. An addition could be a link to the extension page on the extension website.

SahAssar:
Fixes #3461


bobdenotter:
Some extensions, like Gawain's BoltBB add a menu option to the extensions menu. I think that this change is a tiny bit too drastic, and would actually break that functionality.

GawainLynch:
> Some extensions, like Gawain's BoltBB add a menu option to the extensions menu.

Same with MenuEditor which is the number 1 downloaded extension.

SahAssar:
When there are extensions the "extras" submenu will appear, and when there are not it will be a main menu item. Just having an extension with a menuitem will make the "Extras" menu item appear, but when you have no menu items the "Extras" submenu will collapse into the only submenu item there is, "Extensions".  The change in handling of single-item submenus was made in #3323.

I just tried installing menueditor, and the only thing that does is expands the "Extensions" menuitem into a submenu that contains "Extensions" and "Menu editor".

GawainLynch:
Yes, that indeed is the case in my local test… However what might be a UX issue is that the main menu is "Extensions" when there are no submenus, but "Extras" when there is…

We're in UX territory now… so I am stepping back :grin:

bobdenotter:
Ah, ok.. as long as that's been taken into account, i'm cool with this.. I'll check it out tomorrow, and if nothing breaks we merge it in!

bobdenotter:
Looks good to me! :+1:

PR #4233, opened 2015-10-05
Handle save status transitions

### LINKED ISSUE #3244: Error when saving content due to insufficient permissions
**Context**
When a user has permission to create content without the permission to publish and the contenttype is published by default you will get this error. I would suggest a fallback to status draft.

**Steps to reproduce**
* Create a contenttype with default_status: published (contenttypes.yml)
* Give a role in the permission to create the contenttype (permissions.yml)
* Make sure the role doesn't have permission to publish that contenttype (permissions.yml)
* Create a user with that role and use it to login
* create a contentitem

**Expected result**
The contentitem should be saved with a fallback status of draft and/or remove the option `published` from the status pulldown when the user has insufficient permissions.

**Actual result**
Error message: There was an error saving this {contenttypename}

GawainLynch:
Fixes #3244

### DIFF
--- a/src/AccessControl/Permissions.php
+++ b/src/AccessControl/Permissions.php
@@ -724,11 +724,12 @@ class Permissions
                 return empty($fromStatus) ? null : 'depublish';
 
             case 'timed':
+            case 'publish':
             case 'published':
                 return 'publish';
 
             default:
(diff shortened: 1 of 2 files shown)


CarsonF:
:+1: :shipit:

PR #4242, opened 2015-10-07
Record list overhauled

### LINKED ISSUE #3548: Request: entry delete button also on top op entry list
When I want to delete a entry, for example the first one in a list of twenty entries, I have to go to the bottom of the entry list to click the delete button. Is it possible that the delete button also appears on top of the list when an entry is selected? That makes deleting entries as easy as deleting an entry that is at the bottom of the list (no scrolling needed).

rarila:
This is some work, that have to be continued in later releases. Sofar done:

- Various little bugs fixed
- Fixes #3548 (entry delete button also on top op entry list) - Now we have a dropdown at the top
- Start to fix upcoming BC issues with custom lists by isolating what variables are available
- Intruduced ``buic-listing``, that as universal list class should replace dashboardlisting in the long term
- ``buic-listing`` also should be a first try to solve possible BC problems of styles changing that widgets or extensions use. So, as soon as it is mature, it should be documented is frre to use by those. Other element should follow.
- Striping now is fixed in that as it starts with a darker row if it is the first one and a white row if a (dark) header line is before.
- Improvements in how the delete button is attached to the table
- Sort order in the header row is now indicated
- Sort order now is tri state. That way now there's a defined way to remove ordering
- Sorting now works for links with more that the order parameter
- …

As usual, still much things to do. Comments and input welcome!

(Hint: Lists are used on dashboard, overviews, relatedto and omnisearch at the moment)


bobdenotter:
Looks good at first glance. (haven't actually tried it out yet, because tapping away on my ipad, but i did scroll through the code)

> As usual, still much things to do.

Do you want to add things to this PR, or should we merge in this one, and tweak as we go along?

rarila:
@bobdenotter :shipit:

PR #4255, opened 2015-10-10
[2.2] Check for an array before iterating

### LINKED ISSUE #4238: Errors in loading extension's info from git breaks JSON request by showing PHP error
The bolt UI's built in tools to query info on extensions from git break when the git request does not return a proper response.

For example, if you do the following request on a git repo where you are missing access rights, the $info ->version object will be null and not a proper array. In this case the next line and it's foreach command will generate a PHP Warning for accessing an invalid array.
This PHP Warning is shown inside the response thus making the generated JSON response invalid and breaking the calling UI.

In my case the reason for inaccessible git repo was a private GitHub repository. Since bolt is running the git commands as the webserver's user and since this one doesnt have any user/token setup, bolt was unable to get anything from GitHub.

Sample Request:
http://bolt.local/bolt/extend/installInfo?package=xyz

Sample Response:
```
<br />
<b>Warning</b>:  Invalid argument supplied for foreach() in <b>/private/var/www/bolt/root/src/Controllers/Extend.php</b> on line <b>288</b><br />
{"dev":[],"stable":[]}
```

Dump of $info after this line:
```
$info = $app['extend.info']->info($package, $app['bolt_version']);
```
````
object(stdClass)#804 (2) {
  ["package"]=>
  bool(false)
  ["version"]=>
  bool(false)
}
```

**Bolt 2.1.8**

GawainLynch:
Fixes #4238

### DIFF
--- a/src/Controllers/Extend.php
+++ b/src/Controllers/Extend.php
@@ -286,7 +286,7 @@ class Extend implements ControllerProviderInterface, ServiceProviderInterface
         $package = $request->get('package');
         $versions = array('dev' => array(), 'stable' => array());
         $info = $app['extend.info']->info($package, $app['bolt_version']);
-        if (isset($info->version)) {
+        if (isset($info->version) && is_array($info->version)) {
             foreach ($info->version as $version) {
                 $versions[$version->stability][] = $version;
             }


bobdenotter:
Same here.

PR #4252, opened 2015-10-10
Omnisearch listing reloaded

rarila:
- Simplified how the landingpage is remove from search results
- Show all results again, not only contenttypes and files edits
- Added a $fonts-family-monospace sass variable
- Fixed wrong theme file edit path in omnisearch

PR #4256, opened 2015-10-11
"undefined" on omnisearch search popup

rarila:
Fixes glitch reported [here](https://github.com/bolt/bolt/pull/4252#issuecomment-147174220).


rarila:
Wait on this…
Did some further investigation and it turns out that select2 wants ``Object {disabled: true, loading: true, text: "Searching…"}`` to be formatted. I wonder why that one even hits ``templateResult()`` and isn't handled internally and if there has any behavior changed.

rarila:
@bobdenotter That's better now, ready to go

bobdenotter:
Topkek! :+1:

PR #4257, opened 2015-10-11
Password hashing

GawainLynch:
This replaces the forked version of PHPASS maintained by Hautelook with Anthony Ferrara's PasswordLib.

This gives us the following benefits:
* Existing PHPASS hashed passwords can still be verified
* New accounts & password saves will update the hash to bcrypt that is compatible with PHP 5.5+ password hashing functions


GawainLynch:
@bobdenotter the fails on the Composer installs are due to the master install still having the old libraries. So GTG if you're happy.

bobdenotter:
As mentioned on IRC, my main concern would be BC: People not being able to log on, using their old passwords. Since this takes care of that, I don't see any downsides to switching! :+1:

GawainLynch:
Yes, and this means we are up to standard on this part of our security... We have a great record on security, but it would be well appreciated if any silent lurkers (or core team members) could start testing master for authentication functionality and break stuff. :grin:

PR #4262, opened 2015-10-12
Add dev flag to composer.json for passwordlib

rossriley:


GawainLynch:
:koala: :gun:

rossriley:
I'm pretty sure the travis failure is what this fixes... because it's broken on master and the script calls `composer create-project bolt/bolt:dev-master` then this merge still won't pass.

GawainLynch:
Potentially, or should we be using `1.0.x-dev` as the constraint?

bobdenotter:
:leaves: :leaves: :leaves: :leaves:

GawainLynch:
Tested locally, and if we (temporarily) require passwordlib prior to doing the create-project it installs… sending in and we can debate `@dev` versus `1.0.x-dev` when we're bored and needing a OCD fix :grin:

PR #4267, opened 2015-10-13
Use correct mount prefix for upload controller

### LINKED ISSUE #4241: Upload permissions not working for non-root users
In this case, an upload with a standard `chief-editor` role
access control derpy

As reported by @boorj

GawainLynch:
Oh, look… :koala: 

Also fixes #4241

**Edit:** Ended up adding a fix for table checks as I thought that was flooding the logs… Turns out it was the news site being down :grin:

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2819,26 +2819,11 @@ class Storage
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


bobdenotter:
Whoop, whoop! :leaves:

PR #4270, opened 2015-10-13
Repeating fields fieldtype

rossriley:
This is part 2 of 4. 

Sets up a new Repeater field type that handles writing to the database.


GawainLynch:
:+1:

PR #4273, opened 2015-10-14
Renamed example repeat field.

### LINKED ISSUE #4272: "repeat" is a reserved word in MySQL
Installed from master today and I'm seeing the following on the Dashboard: 

You're using 'repeat' as a field name, but that is a reserved word in mysql. Please fix it, and refresh this page.

This looks to be down to the fact that the `showcases` entry in `contenttypes.dist.yml` includes a `repeat` field: Renaming it resolves the issue (I will PR this shortly).

chriscollinsboxuk:
This was causing errors to be thrown on the dashboard, due to the fact
that "repeat" is a reserved word in MySQL.

Fixes #4272.

Mentioning brown M&Ms, as per the contributing doc!

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -165,7 +165,7 @@ showcases:
             type: templateselect
             filter: '*.twig'
             group: meta
-        repeat:
+        repeater:
             type: repeater
             limit: 3
             fields:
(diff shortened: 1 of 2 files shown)


GawainLynch:
Awesome job of mentioning the M&Ms! :grin:

PR #4276, opened 2015-10-15
Increase z-index, so popups cover `{{ dump() }}` output.

bobdenotter:


GawainLynch:
Whoa… hang on! Won't this invoke a #1555 like world ending event?!?!?! :rocket:

PR #4282, opened 2015-10-16
Update .htaccess 78 characters for terminal

brammittendorff:
If you copy paste this in a standard terminal it will put the text "the section below" to a new line. This will result in a fault in .htaccess. That's why "the section below" is maby better on a new line.


GawainLynch:
This might be as dangerous as #1555… So let's merge it before things go really wrong :grin:

bobdenotter:
Cherry-picked into 2.2.x

PR #4285, opened 2015-10-18
Repeating Fields UI [Part 1]

rossriley:
I was going to PR the ui all in one go, but after getting nearly there, there's a bit of work to do on improving the general js initialisers to support dynamically added fields, and this is probably better addressed in a more specific PR rather than being included in this one.

This PR contains the UI to render repeating fields along with corresponding styling. It also includes a fix for repeating fields showing up in the db check.


bobdenotter:
Awesomeballs!

GawainLynch:
> <rarila> gawainlynch: I guess we do it the usual way: merge first, fix later ;-)

Sounds good to me! :+1:

PR #4287, opened 2015-10-19
Catch TableNotFoundException exceptions on Bolt\Users lookup and handle gracefully

GawainLynch:
Fixes a corner-case bug found by @rarila where an extension that checks for users will fatal if the database has been removed.


bobdenotter:
Ha, nice!

PR #4288, opened 2015-10-19
Put right cannonical link into html head on paging content request 

### LINKED ISSUE #3995: Pagination has incorrect canonical
I have a listing of a content type that doesn't have individual pages - that means that the listing page is how one would access all of the content.

I have created the route in `routing.yml`

```yaml
bookspage:
    path: '/books'
    defaults:
        _controller: 'Bolt\Controllers\Frontend::listing'
        contenttypeslug: books
```

And if I go to `http://mysite/books?page=2`, I get the second page *but* the canonical does not have the page query string in it.

```html
<link rel="canonical" href="//mysite/books">
```

I'm aware there may be some debate as to whether the query string should be added in all cases... I'm not sure myself. I'd appreciate if someone could point me towards a workaround because search engines indexing the pages as unique is important. As a last resort, I'd disable the canonical.

rixbeck:
Fixes #3995 for now

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -366,10 +366,15 @@ class ResourceManager
         $this->setRequest('hostname', $hostname);
         $current = $request->getBasePath() . $request->getPathInfo();
         $this->setUrl('current', $current);
-        $this->setUrl('canonicalurl', sprintf('%s%s', $this->getRequest('canonical'), $current));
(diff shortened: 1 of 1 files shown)


bobdenotter:
Top! :+1: 

This should be cherry-picked into 2.2.x as well, right?

bobdenotter:
I cherry-picked this into 2.2.x! :-)

PR #4299, opened 2015-10-20
[Tests] Twig handler text

GawainLynch:
For reference the missing lines are "should never get there" and Windows.

coverage


bobdenotter:
Choo choo!

Issue #4286, opened 2015-10-18
[Tracker] Improving JS to support repeating fields


rossriley:
Pinging @rarila and @SahAssar especially, but any help will be appreciated.

## Overview

Our frontend handlers for fields are very much anchored around initialising on page load and then retiring for a beer.  In trying to get repeating fields working I've run into the problem of needing to initialise fields subject to a context and additionally we need to start considering how we allow custom / extension fields to hook into the `Bolt.fields` object too.

## Todos 

So to get this working we need to adjust each initialiser to take a context and then we need to ensure that they can be called ad-hoc depending on events. Ideally we'd like to support all fields as member of a repeating field set. For reference, in the recent PR for repeating fields I've added a context to `Bolt.fields.init()` and ideally calling `bolt.fields.init(element);` where element is any jQuery element should re-initialise for that node only.

Easy (I think these should all work, or with minor modifications)
- [x]  Geolocation
- [x]  Select
- [ ]  Templateselect
- [x]  Text

Harder Ones
- [x]  Image
- [x]  Imagelist
- [x]  File
- [x]  Filelist
- [x]  Markdown
- [x]  Video

Not Sure
- [x] HTML

Other
- [x] Allow other fields handlers to be registered on `Bolt.fields`

~~I must admit, I've combed through the code and can't quite work out how the ckeditor ones are initialised at the moment, they seem to need a class of `ckeditor` on the textarea.~~

SahAssar:
Closed cause #4422

PR #4301, opened 2015-10-20
Remove hardcoding of Bolt Fields in Bolt.fields

rossriley:
Related to #4286 This removes the hardcoded field names and looks to see wether a handler is registered on the Bolt.fields object.


CarsonF:
Such swag :shipit:

rarila:
Now the "Not implemented yet." information has gone. At least adding it in a comment would have been nice.

PR #4314, opened 2015-10-22
Group "template" translation

rarila:
Nothing to see…


GawainLynch:
Other languages don't need the default key? Or is that auto-something-or-other?

rarila:
@GawainLynch Up to the translators, defaults to EN then. (Sorry, I'm not a language genius;-)

GawainLynch:
Oh no, just was checking if I remember what you just explained correctly :+1:

PR #4315, opened 2015-10-22
Changing strict visibility of method isPagingRequest(), accidentally set

rixbeck:
...why not _public_ ... ?


GawainLynch:
> why not public?

Does it need to be part of the API?

rixbeck:
It depends. I think it is not a disadvantage if we have a reusable method to check actual request is paging. But at least should be _protected_ I think.

PR #4316, opened 2015-10-22
contenttypes.yml.dist reorder

rarila:
- Templateselect: meta -> template group
- Extra group for repeaters
- Switch repeaters and template group position


bobdenotter:
Sounds good to me!

PR #4320, opened 2015-10-23
hide viewless from record browser popup #4120

### LINKED ISSUE #4120: Resource contenttypes should be invisible in link-tool
I noticed that the link-tool popup gives the opportunity to link to viewless contenttypes, which would result in an error in the frontend. If possible, it would be better not to show these.

In this example the quotes and the libraryitems:
files_ _bolt_-_2015-09-21_09 47 13

royallthefourth:
Check each record type that appears in the record browser popup to make sure it's not viewless before adding its links to the list. Prevents appearance of brown M&Ms.

**Edit**
Fixes #4120

### DIFF
--- a/src/Controller/Async/FilesystemManager.php
+++ b/src/Controller/Async/FilesystemManager.php
@@ -232,14 +232,16 @@ class FilesystemManager extends AsyncBase
         $results = [];
 
         foreach ($this->storage()->getContentTypes() as $contenttype) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
Thank you kindly! :+1: 

Only thing was that I've un-nested the inner loop and skipped it with a `continue` in the viewless check. Just makes testing a bit easier.

PR #4321, opened 2015-10-23
[Tests] TwigExtension

GawainLynch:
Too many tests to fit on the image… But this brings our Twig **extension** tests to complete coverage!
coverage

…and total unit tests to 933


GawainLynch:
Oh and just so it's on the record… the handler tests are covering functionality, this just brings in the remaining contract methods to test and shifts the remaining methods to check that they call the correct handler.

bobdenotter:
Green lights!

PR #4326, opened 2015-10-24
[2.2] Handle save status transition

### LINKED ISSUE #4323: Bug on new install: Editor submits an entry
Hi, a bug report:

With Bolt 2.2.13, a completely new install.
When an 'editor' level user submits an entry, he gets an error page instead.
If that same user gets 'chief-editor' level permissions, there is no error page and the page gets saved all right.

So editors cannot save entries in this version. But chief-editors can. there's a permissions problem.

Error Message:

 Exception
 Invalid content status transition: -> publish 
 at articles/src/Permissions.php, lines 725-734:

                   return 'depublish';
                    }
                    break;
                case 'timed':
                case 'published':
                    return 'publish';
                default:
                    throw new \Exception("Invalid content status transition: $fromStatus -> $toStatus");
            }
        }

GawainLynch:
Backport of #4233 to `release/2.2`
Fixes #4323

### DIFF
--- a/src/Controllers/Backend.php
+++ b/src/Controllers/Backend.php
@@ -2,6 +2,7 @@
 
 namespace Bolt\Controllers;
 
+use Bolt\Content;
 use Bolt\Helpers\Input;
 use Bolt\Library as Lib;
 use Bolt\Permissions;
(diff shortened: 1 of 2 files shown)


bobdenotter:
:+1:

PR #4334, opened 2015-10-26
Asset handling fixes/tweaks

### LINKED ISSUE #4332: $this->addJavascript and addCss doesn't add more than one js file.
Version: current master.

I have two local  extensions installed. Both of them want to include js files into the backend like this:

```
    public function initialize()
    {
        if ($this->app['config']->getWhichEnd() == 'backend') {
            $this->addCss('assets/css/file1.css');
            
            $this->addJavascript('assets/js/file1.js', true);
            $this->addJavascript('assets/js/file2.js', true);
        }
    }
```

and

```
    public function initialize()
    {
        if ($this->app['config']->getWhichEnd() == 'backend') {
            $this->addCss('assets/css/file2.css');
            
            $this->addJavascript('assets/js/file3.js', true);
        }
    }
```

Only `file1.css` and `file2.js` get included into the backend. If i comment out `file1.js`, `file2.js` and `file1.css`, than `file3.js` and `file2.css` show up in backend.

Pretty wired! Any ideas?

GawainLynch:
**Summary**
* Fixes #4332
* Allows file asset objects to be created with file name, skipping fluency if desired

**Documentation**
This cleans up and fixes some derps from the asset refactor, and widget implementation branches. Below is an example of what now **can** be done, 2.0 & 2.2 behaviour is still valid.

* In the front-end "zone" we simply inject a JavaScript and CSS files with default options, exactly the same as `addCss('assets/css/foo.css')` `addJavascript('assets/js/foo.js')` 
* In the back-end "zone" we create the object and set whatever other options we do/don't want.

```php
use Bolt\Asset\File\JavaScript;
use Bolt\Asset\File\Stylesheet;
use Bolt\BaseExtension;
use Bolt\Controller\Zone;

class Extension extends BaseExtension
{
    public function initialize()
    {
        $this->app->before([$this, 'before']);
    }

    public function before(Request $request, Application $app)
    {
        // Simple asset injection
        if (Zone::isFrontend($request)) {
            $this->addCss(new Stylesheet('assets/css/foo.css'));
            $this->addJavascript(new JavaScript('assets/js/foo.js'));
        }

        // With all the bells and whistles!
        if (Zone::isBackend($request)) {
            $css = new Stylesheet();
            $css
                ->setFileName('assets/css/bar.css')
                ->setPriority(7)
            ;
            $this->addCss($css);

            $js = new JavaScript();
            $js
                ->setFileName('assets/js/bar.js')
                ->setLate(true)
                ->setPriority(99)
                ->addAttribute('defer')
                ->addAttribute('async')
            ;
            $this->addJavascript($js);
        }
    }
}
```
**Note:** Using objects is optional for convenience, calling the functions with 2.0/2.2 signatures will still work *exactly* the same as we just build the objects before adding them to the queue.

### DIFF
--- a/src/Asset/File/FileAssetBase.php
+++ b/src/Asset/File/FileAssetBase.php
@@ -22,6 +22,16 @@ abstract class FileAssetBase implements FileAssetInterface
     protected $cacheHash;
 
     /**
+     * Constructor.
+     *
+     * @param string $fileName
+     */
+    public function __construct($fileName = null)
+    {
+        $this->fileName = $fileName;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function getType()
--- a/src/Asset/File/FileAssetBase.php
+++ b/src/Asset/File/FileAssetBase.php
@@ -16,7 +16,7 @@ abstract class FileAssetBase implements FileAssetInterface
     protected $late;
     /** @var integer */
     protected $priority;
-    /** @var string */
+    /** @var array */
     protected $attributes;
     /** @var string */
     protected $cacheHash;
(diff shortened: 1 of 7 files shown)


bobdenotter:
:+1:

PR #4337, opened 2015-10-26
Document cookies_no_frontend in config.yml.dist

pbnPierre:
The parameter cookies_no_frontend added in bolt#3425 is not documented in config.yml.dist


pbnPierre:
I did not see cookies_no_frontend in config.yml.dist so I made a PR.

GawainLynch:
Yes, and it won't be documented as it is a very very very broken work around that doesn't exist any longer in master.

If you are using something like varnish you should use that to strip cookies when required.

PR #4341, opened 2015-10-26
Retrieve existing relations before setting them from POST. Fixes #4044

### LINKED ISSUE #4044: No related entries on preview
Hello,

I have contenttype `A` and contenttype `B`.

On `A` I can select one of `B`. On the template for `B` I can fetch `A` with related().. But When I preview `B` the relation to `A` is `false`.

Looks like a bug to me.

bobdenotter:
Fixes #4044

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -190,6 +190,11 @@ class Frontend extends ConfigurableBase
 
         // First, get the preview from Post.
         $content = $this->storage()->getContentObject($contenttypeslug);
+
+        if ($id = $request->get('id')) {
+            $content = $this->storage()->getContent($contenttype['slug'], array('id' => $id, 'returnsingle' => true));
+        }
+
         $content->setFromPost($request->request->all(), $contenttype);
 
         $liveEditor = $request->get('_live-editor-preview');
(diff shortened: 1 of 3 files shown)


bobdenotter:
Same here, please hold off on merging... I think I'm overlooking something, and want to make sure.

GawainLynch:
Yeah, this doesn't look right for master and I think will break when the other parts of the relationship stuff is pushed… but I might just be missing the obvious myself :grin:

bobdenotter:
What would break? It seems to work correctly on master now, and since it uses `->setRelation()` now, I'm not sure it would break _more_ than before..

GawainLynch:
I'm guessing, as it's still "fluxy"… We're fiddling with those objects at the moment, but if it fixes, it fixes… give me a few, I just want to follow up on this one as there is potential overlap with the changes I am doing right now for that other thing.

GawainLynch:
Oh… Read the code, Gawain… this is all in legacy! Nothing to see here! But more-so that should have been cut-out be our changes and may be why something else is breaking! :grin:

bobdenotter:
Works on PHP 5.4 now!

bobdenotter:
\*twiddles thumbs patiently\* ;-)

GawainLynch:
About [when this lands](https://github.com/bolt/bolt/tree/hotfix/relationship-structure) as a PR… @rossriley is having a look at present.

GawainLynch:
The eagle has landed!

PR #4343, opened 2015-10-26
Fixed priority issue in assets. Fixes #4319

### LINKED ISSUE #4319: Adding assets through an Extension when are rendered are previously sorted
After adding the extension Editable, I realize that ckeditor.js is added after startup.js, the author has added the javascript with ckeditor.js first.
Searching the code I find in "src/Extensions.php" the following line (l.834) (look at the last line):
```php
    public function processAssets($html)
    {
        foreach ($this->getAssets() as $type => $files) {

            // Use http://en.wikipedia.org/wiki/Schwartzian_transform for stable sort
            // We use create_function(), because it's faster than closure
            // decorate
            array_walk($files, create_function('&$v, $k', '$v = array($v[\'priority\'], $k, $v);'));
            // sort
            sort($files);
```
This provokes that ckeditor for any reason is put at the end, so startup.js does not know CKEDITOR variable

dactivo:
Priority did not work because it was converted to boolean if the variable priority was set. The methods affected are addCss and addJavascript. 

A dump of the variable $files in the method  processAssets shows  this:
```
array:2 [▼
  "ebb081dd3fe3d1e6b08d5f11c5e01ac3" => array:4 [▼
    "filename" => "/extensions/vendor/bolt/editable/assets//ckeditor/ckeditor.js"
    "late" => true
    "priority" => true
    "attrib" => false
  ]
  "dcb1cb7e3145c3508a7ee333d2803159" => array:4 [▼
    "filename" => "/extensions/vendor/bolt/editable/assets//ckeditor/startup.js"
    "late" => true
    "priority" => true
    "attrib" => false
  ]
]
```

### DIFF
--- a/src/Extensions.php
+++ b/src/Extensions.php
@@ -548,7 +548,7 @@ class Extensions
 
             $options = array(
                 'late'     => isset($args[1]) ? isset($args[1]) : false,
-                'priority' => isset($args[2]) ? isset($args[2]) : 0,
+                'priority' => isset($args[2]) ? $args[2] : 0,
                 'attrib'   => false
             );
         }
@@ -578,7 +578,7 @@ class Extensions
 
             $options = array(
                 'late'     => isset($args[1]) ? isset($args[1]) : false,
-                'priority' => isset($args[2]) ? isset($args[2]) : 0,
+                'priority' => isset($args[2]) ? $args[2] : 0,
                 'attrib'   => false
             );
         }


CarsonF:
@dactivo Nice catch! The :koala: must have got loose. 

Would you mind PRing to master as well?

dactivo:
@CarsonF Thx, I would say master hasn't got this bug

CarsonF:
Derp, you're right

PR #4348, opened 2015-10-28
Add back the namespaced path without namespace. For BC reasons. Fixes #4347

### LINKED ISSUE #4347: Namespacing twig breaks extensions that add pages to the backend.
screen shot 2015-10-28 at 08 59 45

This is because the old

    {% extends '_base/_page-nav.twig' %}

Is now: 

    {% extends '@bolt/_base/_page-nav.twig' %}

We should probably add the old path back in as deprecated, for BC reasons.

bobdenotter:


GawainLynch:
:+1:

PR #4349, opened 2015-10-28
Prevent notice if working in strict mode.

bobdenotter:
Prevent this notice: 

    Notice: Undefined offset: 1 in /[..]/src/Extensions/AssetTrait.php on line 129


GawainLynch:
Ha! I had this on my TODO list today… it was driving me nuts with logs!

PR #4350, opened 2015-10-28
Correctly get templatefields.

bobdenotter:


GawainLynch:
Change triggers a fatal error on the resource CT in the acceptance tests.

bobdenotter:
Ok, the following  works and it passes the acceptance tests locally: 

```
        if ($content->getTemplatefields()) {
            $templateFields = $content->getTemplatefields()->getContenttype()->getFields() ?: [];
        } else {
            $templateFields = [];
        }
```

Would that be a good way to check if `getTemplatefields()` is OK?

GawainLynch:
Yeah, they're already injected into the class at that point, so no extra overhead that way… and easy to read.

bobdenotter:
Ok, fixed.

GawainLynch:
Dank je :smiley_cat:

PR #4351, opened 2015-10-28
Uploads revisited

rarila:
To make it short:
- Big refactor of upload stuff. Now only ``upload-files.js`` is missing.
- The fileselector now uses its native validation. You get much imore information about what did go wrong. We also can easily restrict the number of uploads and the maximum upload size by fieldset now.
- We now have a progress bar buic component that unifies upload progress meters in one place with much improved event based handling.


GawainLynch:
Adding here so it doesn't get lost:
* Progress bar gets to 100% but never completes
* Doesn't handle HTTP 413 from the web server.

GawainLynch:
Manually merged with a ninja fix for the tests :+1:

PR #4352, opened 2015-10-28
Getter for a record's 'title' field.

### LINKED ISSUE #4259: Relation contenttype with no title column
**Summary**
A relation with a contenttype that have no (well-known) title column throws an exception on contenttype edit page.

**Details**

* Request URI: /bolt/editcontent/persons
* Exception:

> Doctrine \ DBAL \ Exception \ SyntaxErrorException
> An exception occurred while executing 'SELECT id, as title FROM bolt_buildings buildings ORDER  BY number ASC': SQLSTATE[HY000]: General error: 1 near "as": syntax error

* Bolt Version: current master (33e82875fd69d38e0de5bea1c5ef8b18a2a455a1)
* Database: SQLite

**Steps to reproduce**

* Create this two contenttypes:
```yaml
persons:
    name: Persons
    singular_name: Person
    fields:
        name:
            group: person
            type: text
            class: large
            label: "Nachname"
            variant: inline
            required: true
        firstname:
            type: text
            class: large
            label: "Vorname"
            variant: inline
            required: true
        slug:
            type: slug
            uses: [ name, firstname ]
        image:
            type: image
    relations:
        buildings:
            multiple: true
            label: Buildings
            order: -number
            format: "#{{ item.number|escape }} {{ item.address|escape }} <span>(ID {{ item.id }})</span>"

buildings:
    name: Buildings
    singular_name: Building
    fields:
        address:
            type: text
            class: large
        slug:
            type: slug
            uses: [ address ]
        number:
            type: text
            pattern: "[0-9]{1,4}"
            variant: inline
        information:
            type: html
```
* Go to bold backend and try to create or edit a new person.

**Debug**

* The relation lists are get in `Bolt\Storage\RecordModifier`, method: `getRelationsList`
* The relation list is get via `Bolt\Storage\Repository\ContentRepository`, method: `getSelectList($contentType, $order)`
* SQL Query is built in same class in method `querySelectList($contentType, $order)`
```php
       $qb = $this->createQueryBuilder($contentType);
        $qb->select('id, ' . $this->getTitleColumnName() . ' as title')
            ->orderBy($order, $direction)
        ;
```
* `$this->getTitleColumnName()` search for a well-known definition of a 'title' column:
```php
      $names = [
            'title', 'name', 'caption', 'subject', // EN
            'titel', 'naam', 'onderwerp',          // NL
            'nom', 'sujet',                        // FR
            'nombre', 'sujeto'                     // ES
        ];
```
* Problem: Title column of contenttype 'buildings' should be the first text field, here 'address'. The function `getTitleColumnName()` return in this case currently nothing, and result in a wrong sql query.

### LINKED ISSUE #4339: Editing an existing record shows 'New ..' instead of 'Edit .. » ..'.
For example, see screenshot: 

screen_shot_2015-10-26_at_18_18_32

Steps to reproduce: 

Create a contenttype. The key is to add a 'title' field, not named `title`: 

```
dummies:
    name: Dummies
    singular_name: Dummy
    fields:
        the_title:
            type: text
            class: large
            group: content
        slug:
            type: slug
            uses: the_title
        body:
            type: html
            height: 300px
```

Then, update the DB, and add some loripsum records.. So far, so good: 

screen shot 2015-10-26 at 18 24 28

Notice that the 'guessed' titles here are correct. It uses the `->getTitle()` method as it should. 

Now, if you edit one of them, it'll say "New Dummy" instead of "Edit Dummy". If I edit `app/view/twig/editcontent/editcontent.twig`, like so: 

```
    \Content :
    {{ dump(context.content) }}

    context.content.the_title :
    {{ dump(context.content.the_title) }}

    context.content.getTitle()
    {{ dump(context.content.getTitle()) }}

    context.content.herpderpnonexistant()
    {{ dump(context.content.herpderpnonexistant()) }}
    ...

```

This is the output: 

screen shot 2015-10-26 at 18 33 32

So, `context.content` _is_ `\Bolt\Storage\Entity\Content`, as it should. It just doesn't work. And that's when I got lost. 

Halp? ¯\\(°_o)/¯

GawainLynch:
Fixes #4339 
Fixes #4259 

**Edit:** Updated and confirmed fix for 4259

### DIFF
--- a/src/Storage/ContentRequest/Edit.php
+++ b/src/Storage/ContentRequest/Edit.php
@@ -29,7 +29,6 @@ class Edit
     protected $users;
     /** @var Manager */
     protected $filesystem;
-
     /** @var LoggerInterface */
     protected $loggerSystem;
     /** @var FlashLoggerInterface */
@@ -65,21 +64,21 @@ class Edit
      * Do the edit form for a record.
      *
      * @param Content $content     A content record
-     * @param array   $contenttype The contenttype data
+     * @param array   $contentType The contenttype data
      * @param boolean $duplicate   If TRUE create a duplicate record
      *
      * @return array
      */
-    public function action(Content $content, array $contenttype, $duplicate)
+    public function action(Content $content, array $contentType, $duplicate)
     {
-        $contenttypeSlug = $contenttype['slug'];
+        $contentTypeSlug = $contentType['slug'];
         $new = $content->getId() === null ?: false;
         $oldStatus = $content->getStatus();
         $allStatuses = ['published', 'held', 'draft', 'timed'];
         $allowedStatuses = [];
 
         foreach ($allStatuses as $status) {
-            if ($this->users->isContentStatusTransitionAllowed($oldStatus, $status, $contenttypeSlug, $content->getId())) {
+            if ($this->users->isContentStatusTransitionAllowed($oldStatus, $status, $contentTypeSlug, $content->getId())) {
                 $allowedStatuses[] = $status;
             }
         }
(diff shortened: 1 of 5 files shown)


bobdenotter:
Whoooop!

PR #4362, opened 2015-10-29
Environment update handling

### LINKED ISSUE #3551: Composer install issues (was "Missing js files after updating from 2.1 to 2.2")
Hi all,

My setup: I used composer install `create-project`
Now I ran an update to *2.2-dev* by changing the composer.json file.

Then I started an `composer update` and the commands provided in the update section in the docs.

Everything seemed to be fine but in the backend I see in browser console:

    GET http://cms.dev/bolt-public/view/js/maps/lib.min.js.map 404 (Not Found)
    GET http://cms.dev/bolt-public/view/js/maps/bolt.min.js.map 404 (Not Found)

I think there are some post scripts missing for composer update?

GawainLynch:
* New class to manage install environment/checks
* Moved version check / cache flush from `Bolt\Cache`
* If web root is writeable the relevant `app/view/` directories are synchonised
* Added nut command where web user has no write access:
* Multi-site set up just got a lot easier and smoother updates.
* Fixes #3551 for composer install updates

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -104,6 +104,9 @@ class Application extends Silex\Application
         // Initialize the rest of the Providers.
         $this->initProviders();
 
+        // Do a version check
+        $this['config.environment']->checkVersion();
+
         // Calling for BC. Controllers are mounted in ControllerServiceProvider now.
         $this->initMountpoints();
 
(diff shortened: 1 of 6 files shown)


CarsonF:
Love it! :heart:

was there a reason for the `realpath` calls? `getPath` should return an canonicalized path

GawainLynch:
> was there a reason for the `realpath` calls? `getPath` should return an canonicalized path

Yes…
```php
$ git rebase -i master --enable-derp
```

bobdenotter:
Awesome!

PR #4366, opened 2015-10-29
JS Refactor for repeating fields [Part 3]

rossriley:
More sequels than Police Academy....

This one takes care of fields that are also BUIC interface elements.


GawainLynch:
td0t4

PR #4368, opened 2015-10-29
Regex that insert snippets not working when </head> does not starts the line

dactivo:
It is related to #4367 in release 2.2 branch


bobdenotter:
I've added tests for this in #4372.

PR #4374, opened 2015-11-01
JS Refactor for repeating fields [Part 4]

rossriley:
Continuing from previous PRs, this one looks after changes to the video field to allow them to be initialised on demand.

I've also removed the need for unique IDs to be used on the fields so each input element can be selected by type within a fieldset.

@rarila apologies if this clashes with what you're working on, but this at least works for now and you can build on it as required.


rarila:
@rossriley No problem, merge first, change later ;-)

PR #4376, opened 2015-11-02
 I'm all 'bout that js, 'bout that js, (no) trouble …

rarila:
What started as a branch to make the four upload fields compatible with repeaters ended in a Chuck-Norris-Roundhouse-JS-Kick. :punch:

Let's see, what's in there:
- Some changes to initialisation of repeaters.
- All id's on cloned fields are replaced now.
- Id-generator implemented in twig (``buid()`` ) and js (``bolt.app.uid()``) which get's you a new unique id on each call.
- ``file``, ``image``, ``filelist``, ``imagelist`` refactored to not use keys/ids anymore. Id's should be used for label/for usage only and a generated id should be used for fields from now on (not yet implemented).
- ``filelist`` and ``imagelist`` handling integrated into ``uploads.js``.
- Browse on server handling simplified and functionality put into ``filebrowser.js``.
- Select from server now only uses one modal (sofar every upload field had its own in code). The last opened url is remembered on upload button basis.
- Stack handling overhauled.
- Stack now has a upload progress bar.
- No more data-actions for stack and filebrowser.
- Muchos simplifed lots of things.
- Error display now for list uploads.

I hope this is a solid foundation for further work on the repeaters and prepares for using jQuery ui widgets for some stuff as a next step. I guess that will further ease field handling.

:warning: Extension devs: Backbone.js and Underscore.js are deprecated now and will be removed from ``bolt.js`` after 2.3! Be prepared.


GawainLynch:
Unsure if the test breakage is just to do with the way we set up the test environment… So I'll merge this and we can catch it up later… :bomb:

PR #4378, opened 2015-11-03
Coding style

rarila:
Coding style police :cop: is back and has a brand new car :oncoming_police_car: 

Time to bring the error count down to get codesniffer in a usefull state again. (Thise time down from about 900 to under 200)


GawainLynch:
Someone didn't do this against an up-to-date branch :stuck_out_tongue_winking_eye:

PR #4379, opened 2015-11-03
Making buttons line out a bit more consistent. Tiny bit more contrast in labels.

bobdenotter:


GawainLynch:
If tests fail on stuff we're not testing… We'll just blame the :cat:

PR #4390, opened 2015-11-05
Allow to encode password when editing/adding users.

### LINKED ISSUE #4373: Users added with app/nut have their password written to DB in plaintext
In master if you add a user with app/nut their password is written to the DB in plaintext instead of being properly hashed/encrypted.

wysow:
Tell me if you think that we should use the second parameter when hashing the password.
Also tell me if I'm not using the right implementation you were thinking of...

Fixes #4373

### DIFF
--- a/src/Nut/UserAdd.php
+++ b/src/Nut/UserAdd.php
@@ -2,6 +2,7 @@
 
 namespace Bolt\Nut;
 
+use Bolt\Storage\Entity;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


GawainLynch:
This approach is not correct. We are doing the [hashing here](https://github.com/bolt/bolt/blob/master/src/EventListener/StorageEventListener.php#L77-L87) based on the [check here](https://github.com/bolt/bolt/blob/master/src/EventListener/StorageEventListener.php#L45-L47).

#4373 occurs as the Nut command still [builds a legacy object](https://github.com/bolt/bolt/blob/master/src/Nut/UserAdd.php#L40-L41) and that needs to be moved to the new storage layer.

wysow:
Thx for this useful tips @GawainLynch I'm on it...

GawainLynch:
:+1:

src/Users.php
```diff
@@ -54,6 +55,10 @@ public function saveUser($user)
             $user = new Entity\Users($user);
         }
 
+        $hasher = new PasswordLib();
+
+        $user->setPassword($hasher->createPasswordHash($user->getPassword()));
+
```
GawainLynch: This one still needs to go, or you will end up with hashed password hashes, that have hashes in their hash-hash… it would be like Hashception :grin:

wysow:
@GawainLynch done... I should squash right?

GawainLynch:
> I should squash right?

You just put the biggest smile on my face! Yes please that would be nothing short of **awesome**!

Also, while you're doing that, can you clean out the `use PasswordLib\PasswordLib;` in `Users.php` as well… I think the :koala: snuck it back in.

wysow:
@GawainLynch do you have a clue about: https://travis-ci.org/bolt/bolt/jobs/89422322#L1776 ?

GawainLynch:
Thank you :wine_glass: :beers:

PR #4395, opened 2015-11-06
Increase 'useragent' field length from 128 to 512

### LINKED ISSUE #4394: 'useragent' field of 'bolt_authtoken' table is too short
#### Class: Doctrine\DBAL\Exception\DriverException

`Data too long for column 'useragent' at row 1`

```
An exception occurred while executing 'INSERT INTO bolt_authtoken 
(username, token, salt, lastseen, ip, useragent, validity) VALUES(?, ?, ?, ?, ?, ?, ?)' 
with params ["princeton-adrian", "e3faf73ee19981603da29a46606db55a", 
"CarEJNduz0S1dTwkqZK0CcZUYX52n680", "2015-11-06 02:16:31", "172.17.0.5", 
"Mozilla\/5.0 (X11; Linux x86_64) AppleWebKit\/537.36 (KHTML, like Gecko) Ubuntu 
Chromium\/47.0.2526.35 Chrome\/47.0.2526.35 Safari\/537.36", "2015-11-20 02:16:31"]: 
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 
'useragent' at row 1
```

Reproducible by setting your user-agent string to something longer than 128 chars.

adrianguenter:
Fixes #4394

### DIFF
--- a/src/Storage/Database/Schema/Table/AuthToken.php
+++ b/src/Storage/Database/Schema/Table/AuthToken.php
@@ -20,7 +20,7 @@ class AuthToken extends BaseTable
         $this->table->addColumn('salt',      'string',   ['length' => 128]);
         $this->table->addColumn('lastseen',  'datetime', ['notnull' => false, 'default' => null]);
         $this->table->addColumn('ip',        'string',   ['length' => 32, 'default' => '']);
-        $this->table->addColumn('useragent', 'string',   ['length' => 128, 'default' => '']);
+        $this->table->addColumn('useragent', 'string',   ['length' => 512, 'default' => '']);
         $this->table->addColumn('validity',  'datetime', ['notnull' => false, 'default' => null]);
         // @codingStandardsIgnoreEnd
     }


adrianguenter:
http://stackoverflow.com/questions/654921/how-big-can-a-user-agent-string-get

It appears the consensus in the above SO thread is that it's preferable to store it in a TEXT blob, within a separate table where you don't care about fixed-width records, then linking with a foreign key. But < 256 is typical, so 512 seems a reasonable compromise short of creating a whole new table IMO.

What say ye?

yamitenshi:
Forgive me if this is a silly question, but why compromise instead of going all the way? The same SO thread you linked shows examples of user agent strings upwards of 8000 characters, and with some plugins having a tendency tendency to append things to the user agent I can easily imagine some instances going well beyond 512 characters.

GawainLynch:
> Forgive me if this is a silly question, but why compromise instead of going all the way?

Not a silly question. But I would argue that simply trimming the string on INSERT/UPDATE might make more sense. This fields purpose is to provide hash data to sessions, and by the time you're past 128 characters of input, it's not really adding a lot… so to speak.

bobdenotter:
Agreed with @GawainLynch. It's for hashing, so let's just trim it in order to prevent database errors, and be done with it. :-)

PR #4396, opened 2015-11-06
Fix/updating slug field

bobdenotter:
There's no easily implementable way to make the slug look good, when wrapped over two lines: 

 - Either it looks "weird", because the button is _way_ too high, and it looks out of place
 - The CSS would need to be hacky, and wonky browsers interpret the hacky css differently.

To circumvent this, I'm proposing to make the 'slug' field take up the full width, and no longer allowing it to wrap over more than one line. 

 - Wrapping is an edgecase, so doesn't occur often.
 - Not that it's full width, it can be a fair bit longer before wrapping happens.

screen shot 2015-11-06 at 10 47 35

New features: 

  - Doubleclick the slug to unlock
  - Added a 'copy to clipboard' option in the drop down.

### DIFF
--- a/app/resources/translations/en_GB/messages.en_GB.yml
+++ b/app/resources/translations/en_GB/messages.en_GB.yml
@@ -474,6 +474,7 @@ field:
             preview: "Preview"
     slug:
         button:
+            copy: "Copy"
             edit: "Edit"
             lock: "Lock"
             unlock: "Unlock"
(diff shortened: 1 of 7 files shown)


rarila:
I guess a look at the [following lines](https://github.com/bolt/bolt/blob/master/app/src/js/.jshintrc#L20)  will get you on track :-)

bobdenotter:
I'm happy with it now! See #4396

rarila:
``field.slug.message.unlock`` needs removing

PR #4403, opened 2015-11-07
Don't break if Jquery is not present.

bobdenotter:
Ideally, we'll make this script "vanilla", so it doesn't require jquery at all. Until that time, this will have to do. At least it doesn't break when JQuery isn't present.


GawainLynch:
@bobdenotter didn't I tell you to put up a :koala: detector! :stuck_out_tongue_winking_eye:

PR #4404, opened 2015-11-08
Welcome jQuery UI Widgets to Bolt!

rarila:
With this PR I like to introduce jQuery UI Widgets to Bolt.

They offer nicer interfaces, better decoupling and overall a nicer experience in working with them. Especially having instances bound directy to DOM nodes makes things easy and intuitive as you can directly interact with them without the need of juggling keys or ids. Also things should be faster as references to elements can be held inside the instance instead of excecuting selectors over and over. Data structures can allso be held inside (e.g. listingpart hold a array of selected ids instead recalculating them everytime they are needed)

A few simple sample usages:
```javascript
$(foo).bar(); // Initialize all elements of the collection to be a bar widget
$(foo).bar('set', 'value'); // Invoke a method on bar widget
$(':bolt-bar').each(…); // Find all bar widgets and do something with them
```
Sofar all ``buic`` elements are moved to be widgets:

- ``checkbox``
- ``listing`` and ``listingpart``
- ``moment``
- ``progress``
- ``select``

While at it, ``checkbox`` is now valid html and a label can be added right of it.

An open question is namespacing. jQuery Widgets allow only one level of namespacing. So at the moment all live in the ``bolt`` namespace. The namespace is not even used, when invoking them (see first two samples above). Question is, should they in any way prefixed with something like ``buic``, ``field`` (I don't know at the moment if uppercase or underscore is possible, e.g. ``fieldFilelist``, or ``field_filelist``) or if there should be multiple namespaces ``boltbuic``, ``boltfield``, … polluting the global namespace or if it isn't needed at all and you simply have to know what ``listing`` or ``text`` is …


rarila:
Just did some investigation: camelcase is possible so something like ``bolt.fieldFilelist`` and ``bolt.buicCheckbox`` looks good to me. This also lowers the possibilty to have conflicts with other external widgets (we already use ``select2()`` and have ``select()`` in this PR)

bobdenotter:
:+1:

PR #4408, opened 2015-11-09
Fix environment view sync for all install types

### LINKED ISSUE #4386: Environment.php tries to sync unexisting folders and throws an error.
calling to @GawainLynch  :leaves: 

I assume this code tries to sync my `root\app\view\*` folders with `root\web\app\view\*`
```php
    // \Bolt\Configuration\Environment::syncViewDirectory
    protected function syncViewDirectory($dir)    {
        //...
        $source = $this->srcRoot . '/app/view/' . $dir; 
        $target = $this->webRoot . '/app/view/' . $dir;
        //...
        $this->filesystem->mirror($source, $target, null, ['override' => true, 'delete' => true]);
        //...
    }
```
So.. points are:
1. in my setup there's no `root\app\view` folder. I've got one in `root\vendor\bolt\bolt\app\view`. 
2. all bolt assets are layed in root\web\bolt-public. So it's strange check.

So working if i turn `syncViewDirectory` off, otherwise :skull: **it dies** with an error in `\Symfony\Component\Filesystem\Filesystem::mirror`

GawainLynch:
This PR fixes #4386 by introducing a `view` path into `ResourceManager`, so we now properly sync both composer-install and multi-site set ups back-end assets on a version change. 

Side note… What this fixes bit me when I set up a Composer install to test this. Starting with a 2.2 install,  then updating that to a master via composer leaves 2.2 `lib.min.js` where in 2.3 `lib.js` is needed.

So thank you @Boorj for nagging me, and @rossriley for his suggested approach.

### DIFF
--- a/src/Composer/BootstrapGenerator.php
+++ b/src/Composer/BootstrapGenerator.php
@@ -11,7 +11,7 @@ class BootstrapGenerator
 {
     public $templateStart = <<<'EOD'
 <?php
-require_once "%s";
+require_once '%s';
 $configuration = new Bolt\Configuration\Composer(%s);
 
 EOD;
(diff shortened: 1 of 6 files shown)


bobdenotter:
Green lights across the board!

PR #4409, opened 2015-11-10
HTML fixes

rarila:
> Document checking completed. No errors or warnings to show.

… at least for the pages checked.

@bobdenotter:
In an cloak-and-dagger operation @CarsonF and I decided that ``form_div_layout.html.twig`` is superfluous. We really, really wanted to consult you before, but as you insist on having sleep at night :last_quarter_moon_with_face: while others do the hard work … don't complain :stuck_out_tongue_winking_eye:


bobdenotter:
If I'm not present, I don't get a vote. Democracy in action, whoohooo!

PR #4411, opened 2015-11-10
Hotfix role handling

GawainLynch:
Fixes:
* Some POST saves of users are generating associative arrays
* The "everyone" role is needlessly persisted


bobdenotter:
Whoop!

PR #4416, opened 2015-11-10
Fix #4196 - ckeditor on chrome in android is not working

### LINKED ISSUE #4196: ckeditor on chrome in android is not working
If you use the backend on an android device the ckeditor component does not work. If you use the same on firefox (or iOS) it works.

According to the documentation the compatibility flags are changed in ckeditor 4.5 and the ckeditor should work. http://docs.ckeditor.com/#!/guide/dev_browsers - http://docs.ckeditor.com/#!/guide/dev_unsupported_environments

I'm happy to help test a fix, but I'm lost in trying to locate why it's not working.

rarila:
[WIP] Don't merge until :+1: from @jadwigo 

Was it really that easy?

Fixes #4196

### DIFF
--- a/app/view/twig/editfile/editfile.twig
+++ b/app/view/twig/editfile/editfile.twig
@@ -78,30 +78,28 @@
         </div>
     </div>
 
-    {% if not ismobileclient() %}
-        {% if context.filetype in ['twig', 'html', 'htm'] %}
-            {% set codemirror = ['xml', 'javascript', 'css', 'htmlmixed'] %}
-        {% elseif context.filetype == "php" %}
-            {% set codemirror = ['matchbrackets', 'javascript', 'css', 'htmlmixed', 'clike', 'php'] %}
-        {% elseif context.filetype == "yml" %}
-            {% set codemirror = ['yaml'] %}
-        {% elseif context.filetype == "md" %}
-            {% set codemirror = ['markdown'] %}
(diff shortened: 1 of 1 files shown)


rarila:
Ok, the issue itself is fixed…

But I still would say we should merge it as I don't see a need to not use cke for file edit on mobile while we use it for html fields. Opinions?

bobdenotter:
This is for Codemirror, on the "Edit file" pages.. 

The issue was for CKeditor on "Edit record" pages.

I don't think this is correct.

rarila:
@bobdenotter As I said, the issue itself is solved. I thought, now that the PR is there, to merge it anyway as that behavior makes no sense for me

bobdenotter:
Does codemirror work on android? IMHO the codemirror is only a nuisance when i'm on iphone or ipad.

rarila:
@bobdenotter hmm, ok, didn't thought that is a problem. Can't try at the moment as I can't login from android :-(

bobdenotter:
Try it on codemirror.net.. If you think it adds something worthwhile on mobile, i'll merge this in. ;-)

rarila:
@bobdenotter In a short test I've seen nothing that makes it worse than a plain textarea, but it at least adds syntax coloring which I consider as helpful.

rarila:
@bobdenotter said :+1:

PR #4417, opened 2015-11-10
Automatically create upload directories for contenttype fields - Fixes #4405

### LINKED ISSUE #4405: auto-creation of subfolders in /files/ for contenttype uploads
##### Feature suggestion
I suggest this feature: create directory, pointed in contenttype for upload fields.
So if i set
```yml
image:
    type: image
    upload: 'articles'
```
bolt should create that directory. At the moment it just shows `upload not allowed`.

It should be made via command in menu or everytime on backend login or elsewhere..

yamitenshi:
This should remove the need to manually create file upload directories. Before, if the directory did not already exist, the upload button would be disabled for the affected field, with the text "Upload is not allowed". This change will automatically create the needed directory, and only show the "Upload is not allowed" text if creation fails.

### DIFF
--- a/src/Storage/ContentRequest/Edit.php
+++ b/src/Storage/ContentRequest/Edit.php
@@ -176,6 +176,7 @@ class Edit
 
     /**
      * Test write access for uploadable fields.
+     * Autocreates the desired directory if it does not exist.
      *
      * @param array $fields
      *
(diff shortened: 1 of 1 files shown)


GawainLynch:
The problem with this approach is that if the web server doesn't have permission to create the directory, Bolt will :fire:

yamitenshi:
I've actually tried that, and didn't have any issues - just got a disabled button saying "Upload is not allowed". What do you mean by :fire: exactly?

GawainLynch:
> What do you mean by :fire: exactly?

Crash and burn.

> I've actually tried that, and didn't have any issues

Ah, yeah… @CarsonF when you FS stuff lands is that valid?

rarila:
I translate that to :bomb: :boom: :fire:

GawainLynch:
OK, talked this over with @CarsonF on IRC. We'll go with this for now, but the upcoming file system changes will impact here… So :shipit: :+1:

PR #4418, opened 2015-11-10
Handle ordering / saving for taxonomies.

### LINKED ISSUE #4210: Taxonomy order not saved in record editor

rossriley:
Fixes #4210 

This introduces collections of entities as a way to reduce the handling code in the controller and shift it to a class that can normalize data from various sources.

The controller code should now be reduced to:

```
$taxonomies = $this->em->getCollectionManager()->create('Bolt\Storage\Entity\Taxonomy');
$taxonomies->setFromPost($formValues, $content);
```

The collection object is an instance of `Bolt\Storage\Collection\Taxonomy` and the form values can be passed straight from the post.

The goal is to discourage (and perhaps by typehinting make it impossible) to set anything other than a collection of taxonomies as a value, each taxonomy entity should be complete, if it has an id it is an existing taxonomy and if not, then we no to create it.

Note this doesn't do the actual saving yet, still WIP.


src/Storage/Field/Type/TaxonomyType.php
```diff
@@ -20,24 +20,6 @@ class TaxonomyType extends FieldTypeBase
...
-    {
-        $field = $this->mapping['fieldname'];
-
-        foreach ($query->getFilters() as $filter) {
-            if ($filter->getKey() == $field) {
-                // This gets the method name, one of andX() / orX() depending on type of expression
-                $method = strtolower($filter->getExpressionObject()->getType()) . 'X';
 
                 $newExpr = $query->getQueryBuilder()->expr()->$method();
                 foreach ($filter->getParameters() as $k => $v) {
```
GawainLynch: Not meant as code review, just flagging… But it looks like some :koala: copy/pasta

rossriley: eek, I don't think I meant to touch that file, must have rebased something wrong etc.

GawainLynch:
> This introduces collections of entities as a way to reduce the handling code in the controller and shift it to a class that can normalize data from various sources.

:+1: 

> and perhaps by typehinting make it impossible

My vote is type hinting :+1:

GawainLynch:
:tada:

PR #4425, opened 2015-11-10
Don't get latestactivity when there's no target

rarila:
:bug:


GawainLynch:
Someone found the command prompt :grin:

rarila:
hehe, no, gone the clicki-way … use the GUI and :shipit: please :-)

PR #4432, opened 2015-11-13
Fixes problem of web profiler bundle template location by using reflection. Fixes #4431

### LINKED ISSUE #4431: Web Profiler Bundle template location
When Symfony is installed along side Bolt, the location of the web profiler bundle changes. [Symfony uses Composer's `replace` functionality](https://github.com/symfony/symfony/blob/2.8/composer.json#L77), which moves the web profiler bundle to `vendor/symfony/symfony/src/Symfony/Bundle/WebProfilerBundle`.

During the Bolt application's initialisation phase, Bolt attempts to load the web profiler bundle's templates into the twig engine, but in doing so, assumes they are always in the same place; `vendor/symfony/web-profiler-bundle/Symfony/Bundle/WebProfilerBundle/Resources/views`.

This causes Bolt to crash with a `Twig_Error_Loader` exception and the message:

> The "/var/www//vendor/symfony/web-profiler-bundle/Symfony/Bundle/WebProfilerBundle/Resources/views" directory does not exist.

emarref:
Use reflection to determine the location of the web profiler bundle, thereby removing the guesswork of configuring twig with the location of the web profiler bundle's templates.

I also don't like brown M&Ms.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -292,8 +292,10 @@ class Application extends Silex\Application
             $this->extend(
                 'twig.loader.filesystem',
                 function (\Twig_Loader_Filesystem $filesystem, Application $app) {
+                    $refProfilerClass = new \ReflectionClass('Symfony\Bundle\WebProfilerBundle\WebProfilerBundle');
+                    $webProfilerPath = dirname($refProfilerClass->getFileName());
                     $filesystem->addPath(
-                        $app['resources']->getPath('root') . '/vendor/symfony/web-profiler-bundle/Symfony/Bundle/WebProfilerBundle/Resources/views',
+                        $webProfilerPath . '/Resources/views',
                         'WebProfiler'
                     );
                     $filesystem->addPath($app['resources']->getPath('app') . '/view', 'BoltProfiler');
--- a/src/Application.php
+++ b/src/Application.php
@@ -292,8 +292,8 @@ class Application extends Silex\Application
             $this->extend(
                 'twig.loader.filesystem',
                 function (\Twig_Loader_Filesystem $filesystem, Application $app) {
-                    $refProfilerClass = new \ReflectionClass('Symfony\Bundle\WebProfilerBundle\WebProfilerBundle');
-                    $webProfilerPath = dirname($refProfilerClass->getFileName());
+                    $refProfilerClass = new \ReflectionClass('Symfony\Bundle\WebProfilerBundle\DependencyInjection\WebProfilerExtension');
+                    $webProfilerPath = dirname(dirname($refProfilerClass->getFileName()));
                     $filesystem->addPath(
                         $webProfilerPath . '/Resources/views',
                         'WebProfiler'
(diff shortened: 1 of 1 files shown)


GawainLynch:
For a reason I haven't had the time to look into yet, this change breaks Composer installs:

```
Class 'Symfony\Component\DependencyInjection\ContainerAware' not found
```
```
…/­vendor/­symfony/­http-kernel/­Symfony/­Component/­HttpKernel/­Bundle/­Bundle.php:29
```

GawainLynch:
OK, sorry about the delay.

What's going on here is that creating the ReflectionClass from `Symfony\Bundle\WebProfilerBundle\WebProfilerBundle` tried to invoke the whole `Bundle` chain which ultimately requires `ContainerAware` which we don't include.

Easiest fix would probably be if you changed to use the profiler Twig extension class to reflect and just use a double `dirname()` like:
```php
$refProfilerClass = new \ReflectionClass('Symfony\Bundle\WebProfilerBundle\Twig\WebProfilerExtension');
$webProfilerPath = dirname(dirname($refProfilerClass->getFileName()));
```

If you want to update that, tests should pass and we're good to go.

emarref:
The last build failed due to composer failing to download some dependencies. I'm closing to re-open which will trigger another build.

emarref:
I tried using composer directly to find the location of the bundle class, but the tests kept failing. I've reverted to your suggested solution of testing the extension class now.

bobdenotter:
I'm not at the computer right nor so i can't take a better look, but I do know that Travis has been grumpy today. The failure might not be with your code necessarily. I'll take a look tomorrow, unless somebody beats me to it. :-)

emarref:
Looks like composer is failing to download some dependencies. I'll postpone this fix 24hrs. I think using composer's `findFile()` method would be the tidiest solution. Better than a reflection call on every request anyway.

GawainLynch:
> The last build failed due to composer failing to download some dependencies.

Nope, it is just the GitHub limit being hit, so it pulls in source rather than dist.

The error is easy to reproduce on a Composer install, and this fixes it:
```php
$refProfilerClass = new \ReflectionClass('Symfony\Bundle\WebProfilerBundle\Twig\WebProfilerExtension');
$webProfilerPath = dirname(dirname($refProfilerClass->getFileName()));
```

CarsonF:
@emarref `Symfony\Bundle\WebProfilerBundle\DependencyInjection\WebProfilerExtension` has the same problem as the `Bundle` class. We don't have Symfony's DependencyInjection component, so using that for reflection won't work.

As far as composer, we don't really have composers class loader available. I know there's mention of it in `ResourceManager` but I think that case is rarely used.

and @GawainLynch beat me to the rest of it :)

emarref:
Thanks @CarsonF, are you able to confirm the approach reflecting the extension class fails for you? @GawainLynch confirms it is working for him.

CarsonF:
@emarref Yeah reflecting the `Twig\WebProfilerExtension` class works for me too

emarref:
Ah sorry, just noticed the `Twig` part. I'll update. Thanks.

GawainLynch:
:+1:

PR #4434, opened 2015-11-13
Updated Footer.

bobdenotter:
The ever-present footer bar was starting to get on my nerves. It's always there, taking up screen estate, and "boxing in" the page. This PR introduces a smaller, more subtle one, as well as some other minor changes to the "about" page. 

Screenshot: 

screen shot 2015-11-13 at 12 13 58

If you have set the 'branding' in your `config.yml`, it will still display the old, large bar:

screen shot 2015-11-13 at 12 27 11


GawainLynch:
:+1:

GawainLynch:
I still think we should have read `app/view/twig/nav/` to `app/view/twig/koalas_can_not_read_when_jetlagged/`… just sayin'

PR #4440, opened 2015-11-13
Prevent fatal error, if we're trying to fetch the querySelectList for a contenttype, that hasn't explicitly defined a sorting order.

bobdenotter:


GawainLynch:
:+1:

PR #4443, opened 2015-11-14
Fixing typo. Because "editontent" isn't really a thing.

bobdenotter:


rarila:
Edit on :tent: – where's the problem? :smile_cat:

GawainLynch:
We need to erect the :koala: a :palm_tree: so that we can keep it away from our code base!

rarila:
@GawainLynch I don't want to comment on that here! :scream:

Issue #4422, opened 2015-11-10
[Tracker] Repeating Fields Push to the Finish


rossriley:
1. How to see repeating fields in action? Add this to the bottom of your contenttypes.yml.

https://gist.github.com/rossriley/da5a932841286d5f23bb

### Still To Do.
-  [X] Javascript: As new groups are added and on load the name attribute of the html elements need to be indexed. The format should be `name="<repeatfieldname>[<index>][<elementname>]`
- [X] Javascript: Need to prevent adding more than the limit specified in `limit:`
- [x] PHP: Validate the YAML file to disallow certain configurations. Disallow unsupported field types.
- [x] PHP Controller: Posted values need normalising and passing to `->save()` method is same as used in tests here: https://github.com/bolt/bolt/blob/master/tests/phpunit/unit/Storage/FieldLoadTest.php#L88-L95
- [x] PHP / Frontend - Once this is working make sure that they are re-displayed on subsequent edits.
- [X] Set default values for fields so that editcontent works with ``strict_variables: true`` again
- [x] Select & multi-select currently populate first option as a default value

rarila:
- [ ] Set default values for fields so that editcontent works with ``strict_variables: true`` again

GawainLynch:
- [ ] Select & multi-select currently populate first option as a default value

rarila:
How important from an UIX standpoint is duplicating? I see here some extra work todo

Boorj:
@GawainLynch, i think your comment references also this issue: #4481 (*Set default option for type: select*)

GawainLynch:
@Boorj no it is a different issue

PR #4420, opened 2015-11-10
Make 'required' for `type: select` fields work. 

bobdenotter:
As @rarila pointed out: 

> The first child option element of a select element with a required attribute and without a multiple attribute, and whose size is “1”, must have either an empty value attribute, or must have no text content.

This PR should fix that. Fixes #4353

PR #4447, opened 2015-11-15
Partial fixes for #4422

rarila:
- Indexing like ``name="<repeatfieldname>[<index>][<elementname>]``
- Rescpect limit. We can also set a maximum or minimum limit
- ``strict_variables: true`` possible 
- Duplicate now just uses the template. Todo: copy data
- Partial backport of #4420 (empty select value now ``''``, not ``0``)


GawainLynch:
Test value fixed here

PR #4454, opened 2015-11-15
Don't list queries in the profiler that are only seen as noise.

bobdenotter:
Only the two "information_schema" ones are added. I just moved it around a bit. 

before: 

screenshot 2015-11-15 14 28 59

after: 

screenshot 2015-11-15 14 21 47


rossriley:
Volkswagen would be proud of that one :-)

rarila:
@rossriley If it only was as simple as adding a space for them :)

PR #4460, opened 2015-11-15
[Travis] Only run Sqlite acceptance tests on Composer install

GawainLynch:


CarsonF:
Faster tests? I'm all about this!

PR #4462, opened 2015-11-16
Don't continue processing a table after exception is hit

GawainLynch:
Under certain conditions, Sqlite can lose data if table processing is allowed to continue…


bobdenotter:
Disregard Scrutinizer!

PR #4463, opened 2015-11-16
Reverse `$type` and `$location`. fixing mixup in #4452

bobdenotter:


bobdenotter:
Don't merge yet! I need to fix the tests, still.

PR #4472, opened 2015-11-16
Use correct i18n js file for ckeditor

phoenixgao:
Ckeditor use bcp-47 language tags for lang files, previously bolt use short locale tag to include ckeditor's lang file, it causes wrong file has been selected for some languages.

This pull request convert posix language tag like "en_GB" to bcp 47 tag like "en-gb" then determine if en-gb.js exists for ckeditor, if not, fallback to use short locale "en.js".


bobdenotter:
Hi @phoenixgao,

This looks good to me! If @rarila agrees, we can merge it in. Thank you!

rarila:
:+1:

PR #4473, opened 2015-11-16
Sort widgets on render() not process()

GawainLynch:
Widget::process() only inserts a JavaScript snippet if any of the
widgets are deferred.

Whereas unlike files, the output is done in render() and the sorting
should be done here.


bobdenotter:
:+1:

PR #4476, opened 2015-11-16
Fix js initialisation on list fields

### LINKED ISSUE #4469: Select from server & Upload image buttons not working
In the record editor, both of them add the image to the stack, but not the record field.

@rarila I have a feeling this might have snuck in recently with that last set of module changes

rossriley:
Fixes #4469

Fixes the bug in processing result of upload fields.


GawainLynch:
Works for me! :+1:

PR #4484, opened 2015-11-17
Fixes ~~two~~ a notice

bobdenotter:
Notice: Undefined index: seo in /xxx/Storage/Field/Type/FieldTypeBase.php on line 97

and:

    Notice: Undefined index: sortorder in /xxx/Storage/Field/Type/TaxonomyType.php on line 201

### DIFF
--- a/src/Storage/Field/Type/FieldTypeBase.php
+++ b/src/Storage/Field/Type/FieldTypeBase.php
@@ -94,9 +94,8 @@ abstract class FieldTypeBase implements FieldTypeInterface
     {
         $key = $this->mapping['fieldname'];
         $type = $this->getStorageType();
-        $val = $data[$key];
-        if ($val) {
-            $value = $type->convertToPHPValue($val, $this->getPlatform());
+        if (!empty($data[$key])) {
+            $value = $type->convertToPHPValue($data[$key], $this->getPlatform());
             $this->set($entity, $value);
         }
     }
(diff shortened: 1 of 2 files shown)


CarsonF:
@bobdenotter I think it would be better to validate `sortorder` when `mapping` is initialized, instead of where it is used.

GawainLynch:
Good catch @CarsonF… :+1: 

Also centralising will help with what we're planning for 2.4 config objects and validation.

bobdenotter:
> I think it would be better to validate sortorder when mapping is initialized, instead of where it is used.

Sure thing! Dumb question: Where's that? :-/

CarsonF:
Not sure...maybe @rossriley can point us in the right direction?

bobdenotter:
Rebased, removed the second one (see #4496). Should be GTG now.

GawainLynch:
Hang back on this… @rossriley is looking deeper.

bobdenotter:
*twiddles thumbs patiently* ;-)

GawainLynch:
@bobdenotter just relaying the conversation from IRC here…
```
<rossriley> gawainlynch: i’m pretty sure it’s not a problem now… 
<gawainlynch> rossriley: As in fixed after Bopp PRed?
<rossriley> I just installed the extension, got a notice to update the db and then loaded / 
a page and saved some SEO data and it all got written ok without a notice
<rossriley> that’s on current master
<gawainlynch> Hmm… OK, I'll leave a note for Bopp to re-test then and pull the PR if needed
<rossriley> so, yes I’m not sure if we’ve already sorted it, or I did something different 
to him
```
Can you recheck/retest that you're still getting these notices as we'd prefer to not hide them.

bobdenotter:
I've tested it out, and the other one is not an issue anymore either. Closing this one.

PR #4488, opened 2015-11-17
Database check debug mode

GawainLynch:
This adds the ability to use `debug` as a query parameter in the URL when viewing the **Database check / update** page in order to display the SQL that will be applied when the update is attempted

screenshot from 2015-11-18 06-36-19


bobdenotter:
Loving it! :+1:

PR #4492, opened 2015-11-18
fix wrong guzzle curl options

phoenixgao:
When news.bolt.cm is unreachable, php-fpm timeouts because connection_timeout of curl options does not work.

according to guzzlehttp 5.3 documents, it shoule be set like this:
http://docs.guzzlephp.org/en/5.3/http-messages.html#request-config

and curl options should be constants instead of string.

Fixed :D

Also I found in PackageManager.php it uses guzzle request options to set timeout time, but when fetch bolt news it use raw curl options, maybe should be consistent?

### DIFF
--- a/src/Composer/PackageManager.php
+++ b/src/Composer/PackageManager.php
@@ -395,7 +395,7 @@ class PackageManager
         }
 
         try {
-            $this->app['guzzle.client']->head($uri, [], ['query' => $query, 'exceptions' => true, 'connect_timeout' => 5]);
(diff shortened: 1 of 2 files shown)


bobdenotter:
Hi, 

The change itself looks good to me.. I'll see if i can figure out why the tests are failing.

bobdenotter:
Travis was giving failures, because I made a mistake in an unrelated part this afternoon. 

Thanks for the contribution, @phoenixgao!

PR #4495, opened 2015-11-18
Move echo out of constructor into shutdown function

GawainLynch:
This prevents multiple exceptions being output in a single request cycle… Bring on the lifting of feature freeze.

screenshot from 2015-11-20 06-12-48


bobdenotter:
Whoop! :traffic_light: :vertical_traffic_light:  :traffic_light: :vertical_traffic_light:  :traffic_light: :vertical_traffic_light:

PR #4498, opened 2015-11-19
Use db charset and collate set in config file to generate table creation sql

phoenixgao:
No matter what charset was set in config, bolt created tables in utf8 and utf8_unicode_ci in mysql during installation, because it's doctrine dbal default:

https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L478

btw utf8mb4 and utf8mb4_xxx_ci works fine.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -909,6 +909,7 @@ class Config
                 'dbname'         => 'bolt',
                 'prefix'         => 'bolt_',
                 'charset'        => 'utf8',
+                'collate'        => 'utf8_unicode_ci',
                 'randomfunction' => '',
             ],
             'sitename'                    => 'Default Bolt site',
--- a/src/Controller/Backend/Users.php
+++ b/src/Controller/Backend/Users.php
@@ -355,13 +355,13 @@ class Users extends BackendBase
         }
 
         $login = $this->login()->login($request, $userEntity->getUsername(), $form->get('password')->getData());
-
         $token = $this->session()->get('authentication');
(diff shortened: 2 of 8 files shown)


src/Config.php
```diff
@@ -909,6 +909,7 @@ protected function getDefaults()
                 'dbname'         => 'bolt',
                 'prefix'         => 'bolt_',
                 'charset'        => 'utf8',
```
phoenixgao: These defaults could change to utf8mb4 in future if you want to support multi byte characters out of box for non-technical users.

GawainLynch: Isn't `utf8mb4` a MySQL only work around for their own design mistakes? 

We support Sqlite, PostgreSQL and MySQL is all.

GawainLynch:
Very clean PR by the way… Nice work :+1:

src/Config.php
phoenixgao: @GawainLynch ah, you're right. I forgot this.
ignore my above note ;p

I don't have experience with databases other than mysql, is it ok to have a default "collate" here?

GawainLynch: > is it ok to have a default "collate" here?

I believe it is OK. If I understand correctly, the parameter just gets ignored for other platforms.

CarsonF:
Agreed, nicely done @phoenixgao!

bobdenotter:
Thanks again, @phoenixgao !

PR #4500, opened 2015-11-19
Fix codeception tests

CarsonF:
We were asserting that we **see** text which is in an `<input value="">`.
These started getting stripped https://github.com/Codeception/Codeception/issues/2249 https://github.com/Codeception/Codeception/issues/2461.
**seeInSource** doesn't strip tags https://github.com/Codeception/Codeception/pull/2465


GawainLynch:
:shipit:

Issue #4223, opened 2015-10-05
Clearing a templatefields selection doesn't clear the warning


GawainLynch:
Set a templatefields template, then clear it:
![template-problem](https://cloud.githubusercontent.com/assets/1427081/10279047/4f00eea6-6b5a-11e5-9522-a8aeefaae8b4.png)

rarila:
That's not an easy one. The whole process is a bit clunky. Not only having to save to get the additional templates field in the beginning but also requiring to reload the page is more than suboptimal. We could either render all possible templates fields as hidden ones and (de-)activate them on selection or get them ajaxy. If you could provide me with an ajaxy hook …
Still not an easy task then with having to add an tab at the right place …

SahAssar:
Well, the reload is okay in my mind, the issue is (*was*) that if you set a template with templatefields and then (without a save) reset it to the "default template" option it will still display the warning. I can't reproduce on current master though. The issue here seems to be that it no longer reloads the page when saved and there is a templateselect with template specific fields, no?

SahAssar:
Closing since fixed, more fixes in this area coming in #4502 :)

PR #4502, opened 2015-11-19
[WIP] Smoothen UI for templatefields toggle.

bobdenotter:
This is a tentative fix for #4223.

If the templatefields are filled when resetting it, it shows the warning as usual.. Short video:

http://recordit.co/LtHrCTAnKr

If the fields are cleared, it will hide the "template" tab silently, and it doesn't show the message. Short video:

http://recordit.co/shIEW9vR12

This "fix" is done vary hackishly. Especially this part is flawed, and needs input from @rarila to do properly:

```
$('input[name^="templatefields"]').each(function(){
    if ($(this).val() !== '') {
        templatefieldsEmpty = false;
    }
});
```


GawainLynch:
Functionally… LGTM

bobdenotter:
*bumping @rarila *

rarila:
From watching the video it solves the issues' issue but the bigger UIX problem (save-refresh) persists.

Technically one could fiddle with a few things, e.g . ``$('#tabindicator-tab-template')`` could be done like ``$('[href="\\#tab-template"]').closest('li')`` or ``$('[href="\\#tab-template"]').parent()`` to save some ids.

Looking at the "hackishly" part… At least it misses select fields, so probably just ``$('[name^="templatefields"]')`` will do the trick. Then selects could return ``object`` for ``$(this).val()``… I guess we can catch this with just ``if ($(this).val()) {`` – To be tested what happens if there's a zero in a textfield and if that's casted to ``number``  ``0`` then … WAT?!
Then there are not only selects, there also checkboxes … And it will break on fields with default values set … Much things to take into account, so:

A much cleaner aproach would be probably to add a ``hasChanged()`` or ``changed()`` method to our fields, taking advantage of the fact that they are now instances. Adding those like ``empty()``, ``changed()`` and ``validates()`` was on my list anyway. One of the nice things widgeted fields brings us. :smile:

bobdenotter:
I made some tweaks, and it's working OK now. There's room for improvement, but it's currently better than it was, and it will likely be redone in the future. So, for now I declare this "Good enough,  ¯\_(ツ)_/¯".

GawainLynch:
OK… let's go for go-go-mobile!

Issue #4470, opened 2015-11-16
Change lock symbol on slugs to better reflect editability 


GawainLynch:

rarila:
If I don't miss something, that is not a bug, but a user (UIX) error ;-)
"Lock" means that it doesn't reflect changes in the title field (or whatever it is defined in ``uses``).

Probably that should be renamed to
```
     Copy
"[X] Connect to <Title+Subtitle>
     Edit
```
with ``fa-link`` and ``fa-unlink`` as icon.

Alternatives: 
- "Linked to …"
- "Generated by …"
- "Build with by …"

Native speakers can certainly come up with better alternatives.

GawainLynch:
Yeah, and I don't think it's the first time I have made this mistake myself… But yeah, I think "linked" works better. The lock says to me that I can't edit it.

bobdenotter:
I think the "lock" should be interpreted as "doesn't reflect changes automatically". Unlocked means it does. 
I think i'm the one who made this originally, but i'm sure it's a fairly common pattern. it's just "locked" against unintentional changes. 

Personally, i'd think "link" might be more confusing to average users. It's _always_ linked to the same field, it's just not always _updated_ from that field.

GawainLynch:
Sure… but the current behaviour doesn't stop you changing it when locked… I keep confusing that.

rarila:
> it's just "locked" against unintentional changes.

can't test right no, but isn't it just so that you **can** change it if it is locked?

bobdenotter:
How can you change it?

GawainLynch:
Click the pencil button, fill in new data, click save.

Boorj:
i think you have enough data even to print with small grey font : "this field is linked to *title*" and if it's unlinked = it could be "is unlinked from title". It even increase understanding for casual users (for ex. - my customers), who doesn't understand the logic of linking slug to a field.

jadwigo:
I would offer the alternative: 
automaticcally synchronized with title
edit manually (as action on a button)
locked (meaning slug has been automatically or manually set, but the linked field has changed since then)

rarila:
@jadwigo Hmm, automatically synced as default wouldn't work.

Technically the slug is just a string in the db. The state (syncing or not) is not saved in any way, it is just something that persists in that record editing session. So what we have now is just some comfort setting, it could also be a button "generate" on the input that generates it from the linked fields by command.

PR #4517, opened 2015-11-23
K.i.s.s. the slug!

rarila:
- Streamlined interface
  - No drop down anymore
  - Direct slug editing
  - Less styles
  - Less javascript
  - Double click removed
- Display of linked fields
- Use widget internal (un-)binding

(See #4470 for discussion. For now I have left the :lock: symbol in)


bobdenotter:
I'm still not sure why this needs work.. I mean, I like this things that are "less", but now the icon says "locked" and you can edit it. I really don't like that part. Slugs should _not_ be editable. especially not on existing records, but we can't get away with disabling it altogether. Right now it's _way_ too easy though. 

Displaying what it's linked to doesn't really make sense to me. If it's up to we we just leave that out.

rarila:
But that is exactly the annoying thing what @GawainLynch and I were talking about. Also in its present incarnation it means:

- :unlock:
  - The slug is autogenerated from linked fields.
  - Calling ``edit`` on it, changes the state to :lock:.

- :lock:
  - The slug *is* editable

bobdenotter:
Ok IMHO, the slug should _never_ be editable.

GawainLynch:
> Ok IMHO, the slug should never be editable.

I humbly disagree… If for example you "duplicate" a record, you're left with `old-slug-1`… and editors do weird stuff all the time… mine drive me mental.

bobdenotter:
That's why i said "but we can't get away with disabling it altogether." I know we'll have to provide _some_ method to modify them, but we should make it absolutely harder than having an "open" form field to change it on a whim. 

What we had before was good: Either a button, or a double-click to trigger the prompt.

GawainLynch:
Yeah, and this is my problem… it is too easy to change and even still with this, while better, it is still confusing…

* An unlocked symbol means I can't change it and hitting the backspace key takes me off page
* A locked symbol means I can edit it…

I am not trying to put forward opinions on UI/UX (Gawain's Third Rule), but as an end user this is *still* confusing… it is confusing in 2.2, it is confusing in master as well.

bobdenotter:
How about we change it to a single tri-state button: 

 - Locked (won't change, can't be edited, default position after opening an existing record)
 - Auto (changes according to the `uses: ` value. default postion on _new_ records)
 - Edit (can be edited, by people who've [never read the classics](http://www.w3.org/Provider/Style/URI).)

GawainLynch:
@bobdenotter I think that reply has made my morning… and nearly awoken the house with the laughing it inspired… "…by people who've never read the classics." :+1: 

The end-user in me would be most pleased with that approach.

rarila:
@bobdenotter I think you hit the point! We were trying to press a three state problem into a two state solution.

This addition now fixes that and I'm quite happy with it. It makes a round an logical solution now. All annoying popups are gone, you still be warnend and editing mode is even auto ended to get at locked state as soon as possible.

GawainLynch:
Slight flaw… "lock" and "link to" both follow "linked" field regardless.

bobdenotter:
Glad you both like it. I think this is a good solution. Will try it later, when i'm back at a desk, and i have time to breathe .:-)

rarila:
@GawainLynch Oh, I thought I had that one squashed out :-(

rarila:
Ok, bug fixed, some cleanup, spinner added, no mor focus outline in readonly mode, …

bobdenotter:
Looks good to me.. One tiny nitpick: 
screen shot 2015-11-24 at 18 06 53

Could you put "Link to:" and "Title" on the same line, plox?

rarila:
@bobdenotter Could get a wide popup with ``uses: [title1, title2, title3]`` and other non short locales (e.g. German: "Verknüpfen mit")

rarila:
@bobdenotter perhaps some css magic with max-width, text-overflow, … will do the trick.

bobdenotter:
> perhaps some css magic with max-width, text-overflow, … will do the trick.

I think that would work for me. :-)

rarila:
Done with it

bobdenotter:
Looks good to me! :shipit:

PR #4521, opened 2015-11-23
Explicitly getting '2.8.0@beta' for some symfony dependencies.

bobdenotter:
.. because otherwise it still pulls in the 2.7.x STABLE versions. As a side-effect this makes the the version resolution a bit faster.


GawainLynch:
Damn it… now I have to rebase my Symfony 3 branch as I did the same thing, just put them at the bottom so I could remove them later :grin:

bobdenotter:
@GawainLynch Perhaps we should leave them it. It looks like the more you specify the dependencies, the less time the resolution takes. So, leaving them in, shaves off a few seconds on every single composer install / update.

GawainLynch:
Oh yeah, I only had mine temporary and lined up for deletion as I didn't have enough data to justify the change… and was lacking time as-is to gather it.

PR #4530, opened 2015-11-25
Add isset, to prevent notices when creating new records with `debug_error_level: -1`

bobdenotter:
Fix as suggested by Ross on IRC. 

Screenshot of the 'before' situation: 

screen shot 2015-11-25 at 16 29 11

PR #4534, opened 2015-11-25
Fix field defaults when creating new content / Removes notices from create content

### LINKED ISSUE #4481: Allow default option for type: select
Documented way to set default option in *select* field for new record is `default : foo`.
Not sure this code is using `default`
https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/fields/_select.twig#L43
```twig
{% for id, value in values %}
    {% set is_array = (value is iterable and (value | length) > 1) %}
    {% set options = options|merge([{
        value:     id,
        text:      is_array ? value[0:]|join(' / ') : value,
        selected:  id in selection or (not onlyids and (is_array ? value[0] : value) in selection),
    }]) %}
{% endfor %}
```

### LINKED ISSUE #4532: "Default" values in contenttypes.yml aren't used.
If you do this in contenttypes.yml: 

```
        datetime:
            type: datetime
            default: "2000-01-01"
            variant: inline
        date:
            type: date
            default: "first day of last month"
            variant: inline
        integerfield:
            type: integer
            index: true
            default: 666
        floatfield:
            type: float
        selectfield:
            type: select
            values: [ none, foo, bar ]
        multiselect:
            type: select
            multiple: true
            postfix: "Select your favourite turtle(s)."
            default: "Donatello"
```

New records use those default values in 2.2.13: 

screen_shot_2015-11-25_at_16_56_04

In 2.3 however, these are ignored: 

screen_shot_2015-11-25_at_16_56_33

rossriley:
Fixes #4532 
Fixes #4481 
Replaces #4530

### DIFF
--- a/src/Storage/Entity/Builder.php
+++ b/src/Storage/Entity/Builder.php
@@ -138,7 +138,8 @@ class Builder
             if ($handler) {
                 call_user_func_array($handler, [$entity, $data[$key]]);
             } else {
-                call_user_func_array([$fieldType, 'set'], [$entity, $data[$key]]);
+                $val = isset($data[$key]) ? $data[$key] : null;
+                call_user_func_array([$fieldType, 'set'], [$entity, $val]);
             }
         }
 
--- a/src/Storage/Field/Type/FieldTypeBase.php
+++ b/src/Storage/Field/Type/FieldTypeBase.php
@@ -115,7 +115,7 @@ abstract class FieldTypeBase implements FieldTypeInterface
     public function set($entity, $value)
     {
         $key = $this->mapping['fieldname'];
-        if (!$value) {
+        if (!$value && isset($this->mapping['data']['default'])) {
             $value = $this->mapping['data']['default'];
         }
         $entity->$key = $value;
--- a/src/Storage/Field/Type/FieldTypeBase.php
+++ b/src/Storage/Field/Type/FieldTypeBase.php
@@ -115,6 +115,9 @@ abstract class FieldTypeBase implements FieldTypeInterface
     public function set($entity, $value)
     {
         $key = $this->mapping['fieldname'];
+        if (!$value) {
+            $value = $this->mapping['data']['default'];
+        }
         $entity->$key = $value;
     }
 
--- a/src/Storage/Field/Type/ListTypeBase.php
+++ b/src/Storage/Field/Type/ListTypeBase.php
@@ -21,7 +21,7 @@ class ListTypeBase extends FieldTypeBase
         $key = $this->mapping['fieldname'];
         $value = $entity->get($key);
 
-        if ($value !== null) {
+        if ($value !== null && !empty($value)) {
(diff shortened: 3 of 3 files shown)


bobdenotter:
giphy 1

CarsonF:
@rossriley Isn't `call_user_func` more performant than `call_user_func_array`?

CarsonF:
http://paul-m-jones.com/archives/182

bobdenotter:
> 2005-09-22

Could be that the difference is smaller, now that we're ten years further along. ;-)

Still might be good to look into, just in case.

GawainLynch:
OK, so a test case of a loop with a million iterations:
```php
class Derp
{
    private $arg1 = 'foo';
    private $arg2 = 'bar';
    private $func = 'derpMe';
    
    public function loopyInline()
    {
        $arg1 = $this->arg1;
        $arg2 = $this->arg2;
        $func = $this->func;
        for ($i = 0; $i < 1000000; $i++) {
            $this->{$func}($arg1, $arg2);
        }
    }
    
    public function loopyUserFunc()
    {
        $arg1 = $this->arg1;
        $arg2 = $this->arg2;
        $func = $this->func;
        for ($i = 0; $i < 1000000; $i++) {
            call_user_func([$this, $func], $arg1, $arg2);
        }
    }
    
    public function loopyUserFucArray()
    {
        $arg1 = $this->arg1;
        $arg2 = $this->arg2;
        $func = $this->func;
        for ($i = 0; $i < 1000000; $i++) {
            call_user_func_array([$this, $func], [$arg1, $arg2]);
        }
    }
    
    public function derpMe($arg1, $arg2)
    {
        return $arg1 . $arg2;
    }
}
```

#### Executing on PHP 5.6

Average on 10 runs ~1.2 seconds
```php
(new Derp())->loopyInline();
```

Average on 10 runs ~2.9 seconds
```php
(new Derp())->loopyUserFunc();
```

Average on 10 runs ~3.0 seconds
```php
(new Derp())->loopyUserFucArray();
```

#### Verdict

The difference in using `call_user_fuc()` versus `call_user_fuc_array()` is minimal and only relevant in a *very* hot path… if you're worried about performance the inline function call if more than twice as fast and should be used in hot paths…

#### Disclaimer

No :koala: were harmed in the making of this test suite.

CarsonF:
Thanks @gawainlynch for that thorough answer :)

rarila:
@GawainLynch Just for the interest, tried on php 7? (only if it also doesn't harm any :koala:)

GawainLynch:
#### PHP 7
Inline ~1.1 seconds
User func ~1.2 seconds
User func array  ~1.2 seconds

rarila:
cool, no reason to inline then

rarila:
… although it looks nicer anyway

GawainLynch:
I am personally a fan of the inline.

Also I don't think we're raising our minimum required PHP version to 7 just yet… I wish though :wink:

PR #4537, opened 2015-11-26
Fix/more spinners and missing icons

bobdenotter:
This PR returns the icons from "login" and "create first user" that went missing. As an added bonus, it adds nice spinners, to give the user the idea that something is happening.


GawainLynch:
:+1: :shipit:

Issue #4444, opened 2015-11-14
[Tracker] Missing functionality for widgets.


bobdenotter:
Our new widgets functionality works fine, but its missing a few specific pieces of functionality. I think we should get these done for 3 .0

  - [x] Add ``{% if haswidgets('..') %}`` functionality 
  - [x] Add `{{ listwidgets('..') }}` function, to get a list of currently registered widgets
  - [x] Make sure `priority` works. Or does it already? 
  - [x] Rename `widget` to `widgets`
  - [x] Make the widget insert script Vanilla Javascript instead of Jquery (see `app/theme_defaults/widgetjavascript.twig`)
  - [x] Make `{{ widgets() }}` use named parameters, and have 'frontend' be default. So, `{{ widgets('aside_top') }}` should work.
  - [x] Allow for a way to define the wrapper for the widget and widgetholder. This should be possible ~~both in code, as well as~~ in the theme.
  - [x] Add "Latest [x] from [x]" and "search" as widgets. 
  - [x] Write documentation
  - [ ] Add widget hooks on the user edit forms (Moved from #4966)

Do at some later point in time: 
  - [ ] Add permission check for widgets, perhaps?
  - [ ] Allow for a max number of widgets in a given widget area

GawainLynch:
> Add permission check for widgets, perhaps?

Can I vote against that as it is starting to produce extra complexity that really doesn't belong here.

> Allow for a max number of widgets in a given widget area

Valid, absolutely… Question on my mind though, are we inviting race type conditions with people setting bigger and bigger priorities to out do each other… `**cough**` WordPress `**cough**`

bobdenotter:
> Can I vote against that as it is starting to produce extra complexity that really doesn't belong here.

Ok, let's scrap that for now, at least. 

> Valid, absolutely… Question on my mind though, are we inviting race type conditions…

Let's put that one on the backburner too. We can always introduce it, if it becomes an issue.

bobdenotter:
Documentation is in PR. 

Other changes will be bumped to 3.1.

bobdenotter:
Closing!

PR #4541, opened 2015-11-28
Feature/widget configurable wrapper

bobdenotter:
Allows themes to set the widget-wrapper and widget-holder HTML. Last functional part of #4444.


src/Asset/Widget/Queue.php
```diff
@@ -156,23 +156,23 @@ public function countItemsInQueue($location, $type = 'frontend')
...
         }
 
         if ($html !== null) {
-            $html = sprintf(
-                '<div class="widgetholder widgetholder-%s">%s</div>',
-                $location,
-                $html
-            );
+            if (empty($wrappertemplate)) {
+                $wrappertemplate = 'widgetholder.twig';
```
CarsonF: Why not define this in the parameter definition instead of empty string? Same for below as well

GawainLynch: And that should be done at the Twig Extension end… the API parameters should be required.

src/Asset/Widget/Queue.php
```diff
@@ -156,23 +156,23 @@ public function countItemsInQueue($location, $type = 'frontend')
      *
      * @return string|null
      */
-    public function render($location, $type = 'frontend')
+    public function render($location, $type = 'frontend', $wrappertemplate = '', $holdertemplate = '')
     {
```
GawainLynch: Small favour… can we go for camelCase on `$wrapperTemplate` and `$holderTemplate` to keep in-line with PSR-2 and the way we're taking our code standards.

src/Asset/Widget/Queue.php
CarsonF: Yes, I would argue there, since it is the entry point, but also here since this could be called from php.

GawainLynch: On that last one, yes… but `$holderTemplate` should be required or default to `null`, and certainly be required in the protected function.

bobdenotter:
Ok, it's currently broken, and I can't figure out why. I have this in my `_aside.twig` on line 7: 

            {{ widgets('aside_top') }}

and I get this error: 

screenshot 2015-12-01 20 32 31

Where it _actually_ breaks, is in the `render()` here: 

```
    protected function addWidgetHolder(WidgetAssetInterface $widget, $holderTemplate)
    {
        $html = $this->render->render(
            $holderTemplate,
            [
                'widget' => $widget,
                'html'   => $widget->isDeferred() ? '' : $this->getHtml($widget),
            ]
        );

        return $html;
    }
```

Debugging tells me `$holderTemplate` is `widgetholder.twig` at that point, and the file exists and is readable. What the hell?!

GawainLynch:
@bobdenotter we've had a look, test is fixed… but @CarsonF think a few things need to change before we deploy API.

1. We should rename the 'Type' use to 'Zone' as that is consistent with how we're doing it in the controllers
2. We should only use one template so we don't sub-render.
  * The exception you're seeing is due to the template shifting around and needed elements from teh widget object that don't exists outside the loop

CarsonF:
@GawainLynch Can we change `Widget` `type` property to `zone` as well?

@bobdenotter I haven't quite figured out how to combine templates yet, so until then carry on.

I am wondering why the twig template files need to be configurable at this level. Isn't the precedence of twig paths enough? I mean that's the whole reason why this folder is called `theme_defaults`.

GawainLynch:
>  Can we change Widget type property to zone as well?

That was what I was asking

GawainLynch:
@bobdenotter happy to offer our service here too… just watching where my foot is in proximity to toes :grin:

bobdenotter:
Green lights! :-)

GawainLynch:
Amanhã… Precisar dormir!

src/Asset/Widget/Queue.php
```diff
@@ -96,7 +96,7 @@ public function process($html)
...
         /** @var WidgetAssetInterface $widget */
         foreach ($this->queue as $widget) {
-            if ($widget->getType() === 'frontend' && $widget->isDeferred()) {
+            if ($widget->getZone() === 'frontend' && $widget->isDeferred()) {
```
CarsonF: Can you use the `Zone::FRONTEND` constant instead of the string?

GawainLynch:
Ninja added the constants… time to :shipit:

PR #4544, opened 2015-11-30
Fix preview unpublished content

marcin-piela:


bobdenotter:
Ah, I see! Thanks for the fix! :+1:

PR #4552, opened 2015-12-02
Making buic checkbox a bit more subtle and consistent with other fields.

bobdenotter:


bobdenotter:
Before: 

screen shot 2015-12-02 at 16 05 28


After: 

screen shot 2015-12-02 at 16 12 26

PR #4554, opened 2015-12-03
Add the possibility to configure accepable image formats in config.yml

pbnPierre:
I need in my project to add svg support in an image attribute so here is a PR to support it.
Are you ok with this feature ?

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -223,6 +223,9 @@ accept_file_types: [ twig, html, js, css, scss, gif, jpg, jpeg, png, ico, zip, t
 # instead. It includes file types that are harder to exploit.
 # accept_file_types: [ gif, jpg, jpeg, png, txt, md, pdf, epub, mp3, svg ]
 
+# The following is the default list of file-types that can be uploaded through the 'image' field.
+accept_image_file_types: ['gif', 'jpg', 'jpeg', 'png']
+
 # If you want to 'brand' the Bolt backend for a client, you can change some key
 # variables here, that determine the name of the backend, and adds a primary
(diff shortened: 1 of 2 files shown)


CarsonF:
@pbnPierre Remind me, do svgs require different html syntax?

@GawainLynch I think we should default this to image extensions pulled from `Bolt\Filesystem\Handler\Image\Type::getExtensions`. Agree?

GawainLynch:
@CarsonF yeah, that'd be my preference, as is not enabling the setting by default… which I think is the subtle hint you're giving there anyway.

SahAssar:
@CarsonF no, SVG's can be used with the same `` syntax as a png or jpg (but can also be used in an object tag to allow for interactivity/animation). A related issue is that bolt/thumbs can't parse them though, see bolt/bolt-thumbs/issues/22

pbnPierre:
@CarsonF same answer as @SahAssar with img tag you can display SVG files. And I also noticed the problem with the image resizer you mentioned in bolt/bolt-thumbs#22

SahAssar:
@pbnPierre I think the issue that I am having with this PR is that the expectation is that any file allowed in an image field should be able to be parsed by bolt's image functions. @CarsonF and @GawainLynch seems to agree with this since they want the allowed filetypes to be determined by what bolt-thumbs supports rather than being hardcoded. Also, if you **really** want to support SVG in an image field you can simple configure it use the `extensions` property like so (although all image functions will be broken, and no preview will appear):

        image:
            type: image
            extensions: ['svg']

SahAssar:
Alright, closing this one per the previous comment. Reopen if you feel you have a better idea :)

PR #4561, opened 2015-12-06
Filesystem / Thumbnail Refactor Phase I

CarsonF:
Over the last few months I've completely rewritten our [thumbnail code](https://github.com/bolt/bolt-thumbs), and recreated a [filesystem abstraction](https://github.com/bolt/filesystem) (based on Flysystem). 
My goal is to have a unified filesystem layer that is both usable for our everyday needs and supports remote filesystems, such as S3.

As such, the new filesystem code comes with the following:
* Symfony Finder-like functionality
* Robust Handlers (File and Directory objects)
* Image Info and EXIF abstractions
* Exceptions for error handling
* Interfaces for all the things
* Manager class that fully implements a filesystem interface
* Support for including / loading PHP files

The new thumbnail code includes several OO design changes and is built on top of this filesystem interface.

These changes here are the first round of the integration of these two libraries.
Huge thanks, of course, to @GawainLynch for helping finish this out in core, and in filesystem repo.


GawainLynch:
Just fore reference, the failures at this point are simply due to package changes that are not picked up in the composer install due to it not knowing about the PR tree changes.

i.e. could someone please :shipit:

CarsonF:
Let's do this!

PR #4562, opened 2015-12-06
Fix for save / overwrite error on repeating fields

rossriley:
When flattening the existing values, index the array keys numerically so that subsequent groups dont overwrite the previous.


GawainLynch:
On limited bandwidth right now, but it looks like something it triggering a lot of system log messages, causing the acceptance test to fail on that step.

GawainLynch:
:shipit:

Issue #4573, opened 2015-12-09
Sidebar Height Resize Issue


draconusdesigns:
The left sidebar in the admin area resizes to match the page content. It gets bigger, but not smaller. I will submit a pull request with the fix for this shortly.

GawainLynch:
Fixed in #4575 & #4576

PR #4575, opened 2015-12-09
Issue #4573: added default height of 0 to the navpage secondary outer (2.2)

draconusdesigns:
There is an issue where the sidebar resizes larger as the content size increases, but the content size decreasing is not resizing the sidebar. This fix sets the sidebar height to 0 before calculating the sidebar height to ensure that the document height is pulled in correctly for the calculation.

This is a fix for issue #4573 and should be pulled into master, 3.0 and 2.2


bobdenotter:
Same here. :shipit:

PR #4580, opened 2015-12-11
Restore Symfony fluent-interface chained call syntax

GawainLynch:
This restores the :scream: inflicted in #4378 and brings calm to Symfony addicts and their war on anti-perfection!

@rarila: I have updated the code sniffer rules so this won't increase the count :+1: 
@CarsonF: Any Belgian :beers: are fine :grin: 

**No :koala: were harmed in the making of this PR.**

### DIFF
--- a/src/AccessControl/Password.php
+++ b/src/AccessControl/Password.php
@@ -186,7 +186,8 @@ class Password
             ->setReplyTo($from)
             ->setTo([$userEntity['email'] => $userEntity['displayname']])
             ->setBody(strip_tags($mailhtml))
-            ->addPart($mailhtml, 'text/html');
+            ->addPart($mailhtml, 'text/html')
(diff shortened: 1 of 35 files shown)


CarsonF:
Thank you mate :100:

rarila:
So we now have a sniffer rule that makes sure the semicolon is at the correct position?

GawainLynch:
Technically we always did… The rule set we based ours on allows it, one of the Squiz rules over rides it.

These sill fails test:
```php
    $moo = 'cow' ;

    $foo->bar()
        ->baz()
;
```

rarila:
:leaves: 

Didn't saw a PR on that because of ninja edit

GawainLynch:
I don't know what you mean :wink:

PR #4582, opened 2015-12-12
[Travis] Require pass from PHP 7

GawainLynch:
We have been passing tests on PHP 7 since PHP 7-RC4 and 7.0.0 is now GA.


bobdenotter:
Whoop!

PR #4598, opened 2015-12-16
Cache update

GawainLynch:
I have split this out of the branch that @CarsonF and I are working on for extensions as it should be reviewed separately.

This brings the cache a bit closer in-line with Doctrine's interface in preparation for work that we'll do during the 3.x-dev life cycle.

@bobdenotter of note to you, we've removed the override of `getFilename()` as upstream has "fixed" their behaviour in 1.5.0.


bobdenotter:
Whooop!

PR #4599, opened 2015-12-16
add in cache-control max age to go along with s-max-age directive

### LINKED ISSUE #4600: Page Render Cache-Control
1) __Bug Summary__.  
If you set caching for GET requests to true the page you are on/visiting doesn't get served from the cache if you visit/return within the duration set in the config.  

2) __Specifics__.  
* Bolt Version:  
  * All versions of Bolt.  
  * installed through composer, git, and download from site.  
* Browsers affected... All modern ( Edge, Webkit, Blink Based browsers and Firefox )  

3) __Steps to reproduce__:  

set caching for GET requests to true. 
```yaml  
caching:
    config: true
    templates: true
    request: true
    duration: 10
    authenticated: false  
```  
visit the page. Here's a good ol gif of that in process.    
  
s-maxage-chrome  

4)__Expected Result__  
page should be served from the cache.  

------------------------  
The browser looks for the "max-age" in the cache control header. The s-maxage is used for shared caches.  

Adding in a max-age cache-control directive will fix this. Here is another gif of that in process with a max age and s-maxage cache-control directives.  
chrome-max-age  

As you can see with the max-age directive when you return to the previously rendered/visited page within the duration set in the caching config settings the page is served from the cache.  

Since you all are now using symfony's response and setcache methods this might also open up areas for extensions to either create a static page and serve those instead of calls to the db to render or serve the page with a 304 similar to http://symfony.com/doc/current/book/http_cache.html#optimizing-cache-validation  

 
I've gone ahead and created a pull request to "fix" this ( #4599 ) . If it's not what you are looking for then I'll just use it in my own deploys.  

Thanks!

cdowdy:
Add in ```max-age`` cache control directive to allow pages to be served from the cache 

Fixes #4600

### DIFF
--- a/src/Render.php
+++ b/src/Render.php
@@ -108,10 +108,10 @@ class Render
                 // only 50% max, but likely less
                 // 's_maxage' sets the cache for shared caches.
                 // max_age sets it for regular browser caches
-                $response->setCache(array(
-                    'max_age'       => $this->cacheDuration() / 2,
-                    's_maxage'      => $this->cacheDuration() / 2
-                ));
+
+                $age = $this->cacheDuration() / 2;
+
+                $response->setMaxAge($age)->setSharedMaxAge($age);
             }
         }
 
--- a/src/Render.php
+++ b/src/Render.php
@@ -106,7 +106,12 @@ class Render
                 // maximum duration, otherwise a proxy/cache might keep the
                 // cache twice as long in the worst case scenario, and now it's
                 // only 50% max, but likely less
-                $response->setSharedMaxAge($this->cacheDuration() / 2);
+                // 's_maxage' sets the cache for shared caches.
+                // max_age sets it for regular browser caches
+                $response->setCache(array(
+                    'max_age'       => $this->cacheDuration() / 2,
+                    's_maxage'      => $this->cacheDuration() / 2
+                ));
             }
         }
 


GawainLynch:
Yeah, I was in two minds about this myself when I last updated that section. But I think my hesitation came down to being from Shared-Cache-Only Land. :grin:

src/Render.php
```diff
@@ -106,7 +106,12 @@ public function fetchCachedRequest()
...
                 // cache twice as long in the worst case scenario, and now it's
                 // only 50% max, but likely less
-                $response->setSharedMaxAge($this->cacheDuration() / 2);
+                // 's_maxage' sets the cache for shared caches.
+                // max_age sets it for regular browser caches
+                $response->setCache(array(
+                    'max_age'       => $this->cacheDuration() / 2,
+                    's_maxage'      => $this->cacheDuration() / 2
+                ));
             }
```
GawainLynch: Could of points:

* master (to be v3) is PHP 5.4+ and we've switched to sort array syntax
* We're also sticking with fluent interfaces for things like this… can you switch to something like:
```php


                $age = $this->cacheDuration() / 2;
                $response->setMaxAge($age)->setSharedMaxAge($age);
```

cdowdy:
Done and Done. Funnily enough I originally used short array syntax but then I thought....WAIT.... WHAT ABOUT 5.3 haha..  Thanks for the pointers @GawainLynch !

GawainLynch:
There was a PHP 5.3? Is that anything like PHP 6… Oh… wait…

PR #4616, opened 2015-12-21
Fix/checkbox: checkbox input and label after, removed button (semantics)

sbonardt:
Put the label *after* the input type checkbox and removed the button. Now it's just an input `type="checkbox"` and a `label` (with an `<i>` in it). Had to adjust the other fields so the label is written out in a block. The label for the checkbox isn't in the block, but after the input. Adjusted the styling...

screen shot 2015-12-21 at 17 04 39

Check the showcase content type. It has a checkbox in the "other" tab


bobdenotter:
:+1:

rarila:
Don't merge for now, I need to further investigate this …

rarila:
Ok, checked things out …

As I see it right, all the changes here have one purpose: having the label on the right side of the checkbox instead of left in the label column?!

Some problems with this patch:

- The ``button`` was in here for a simple but very important reason: allowing keyboard access. With this patch the ability to set focus on the checkbox with tab is gone.

- ``buic.checkbox`` has right label already build in, so the same goal could easily achieved with just a few [changes](https://github.com/bolt/bolt/compare/master...rarila:checkbox-label-right).

---

The interesting question is, what is the best label position. I see usecases where the current solution is best and where the label on the right is best. I can even see cases where you want a label on both sides (short on the left what's it is about and on the right a more descriptive text), eg something like that:

```
Status:    [X] Merged
```

So why not allow an array for ``label`` in contentttype.yml making this possible:

- Label left:
```yaml
checkbox:
    type: checkbox
    label: "Beer empty"
```
```
| Beer empty:    [X]
```
- Label right:
```yaml
checkbox:
    type: checkbox
    label: ["", "Beer empty"]
```
```

 |                [X] Beer empty
```
-  Label left and  right:
```yaml
checkbox:
    type: checkbox
    label: ["Beer", "Empty"]
```
```
 | Beer:          [X] Empty
```

That way the developer can decide to use the position that fits best. What do you think?

sbonardt:
O man, seems lik you planned to built it right like it should! :D 

I like the way you're able to decide label left or right with the change, but do the two labels both get associated with the input when in use?

Furthermore, the documentation you've written out clarifies it all. And we could have made an issue before just changing this without reference. Sorry again. As we in Holland say, "too short through the bend" (too blunt...)

Let's revert this @bobdenotter

rarila:
No, I'm sorry. You put so much effort into this …

> O man, seems lik you planned to built it right like it should! :D 

Hehe, I do my best. the ``buic`` stuff is planed to be used throughout the codebase and (when being documented) also from extensions. So it has to be layout a bit more general.

Getting the checkbox component into that state took quite some hours and a few PRs. Accessibility stuff is missing, but I don't want to add such things without having someone in the team who really knows about it. Otherwise there's a chance to make it even worse instead of better.

> I like the way you're able to decide label left or right with the change, but do the two labels both get associated with the input when in use?

The ``for`` attribute of the labels point both of course to the checkbox's ``id`` then.

> Furthermore, the documentation you've written out clarifies it all.

As I always say, documentation is wasted time  :wink:

Ok, so as soon as others give the ok to double label idea I'll add that to my branch and PR it.

bobdenotter:
:+1: from me.

GawainLynch:
So… um… yeah… :penguin:

PR #4628, opened 2015-12-23
[NotFoundListener] Check if template exists instead of string ending with ".twig"

CarsonF:
Can't find where @Boorj was complaining about this, but this is less hacky.


GawainLynch:
:+1:

Boorj:
thank you Carson @CarsonF  !

PR #4638, opened 2015-12-29
Prevent exception empty extension config files

### LINKED ISSUE #4636: Exception if extension config is empty
I installed a bolt extension with empty config file (like clientprofiles). When I want to open this file in the backend, I get a exception "The file 'extensions/clientprofiles.bolt.yml' doesn't exist, or is not readable.". The problems are here:

* https://github.com/bolt/bolt/blob/master/src/Controller/Backend/FileManager.php#L81
* https://github.com/bolt/bolt/blob/release/2.2/src/Controllers/Backend.php#L1849

The checks should be:
```
if (!$file->exists() || false === $file->read()) {
```
$file->read() returns false if file can't be read. At the moment $file->read() returns a empty string and this is checked as false.

fadoe:
When I want to open a empty extension config file in the backend, I get a exception "The file 'extensions/clientprofiles.bolt.yml' doesn't exist, or is not readable."

$file->read() returns false if file can't be read. At the moment $file->read() returns a empty string and this is checked as false.

Fixes #4636

### DIFF
--- a/src/Controller/Backend/FileManager.php
+++ b/src/Controller/Backend/FileManager.php
@@ -78,7 +78,7 @@ class FileManager extends BackendBase
         $type = Lib::getExtension($file->getPath());
 
         $contents = null;
-        if (!$file->exists() || !($contents = $file->read())) {
+        if (!$file->exists() || false === $file->read()) {
             $error = Trans::__("The file '%s' doesn't exist, or is not readable.", ['%s' => $file->getPath()]);
             $this->abort(Response::HTTP_NOT_FOUND, $error);
         }


GawainLynch:
Thank you :+1: 

Merged locally due to test failures, as the `$content` variable still needs a value. Also for master (potentially no way you've seen this yet) but `bolt/filesystem` now throws exceptions :heavy_check_mark: 

…so just [caught those].(https://github.com/bolt/bolt/commit/b1906c664d4066b0fb6b62e8278be2b7bbe9e918)

fadoe:
Okay, the next time I will run all tests

Issue #4627, opened 2015-12-22
[RFC] Refactoring current Pager implementation


rixbeck:
Bolt 3.0 is going over a great rejuvenation and had been reorganized beside some important viewpoint. I thought let's touch those parts too - if you all think so - that should have a more elegant face lifting. One of those our current paging (Pager) solution.

We should have ``Bolt\Provider\PagerServiceProvider`` and ``Bolt\PagerManager`` or ``Bolt\Pager\PagerManager``

### class PagerManager ...

- is going to be a centralized service that would be instantiated lazily by ``PagerServiceProvider``.
- Furthermore each currently applied ``Pager::`` based, global smelling static call would be served by the manager to fit TDD principle.
- The service will be accessible via ``$app['pager']`` then.

#### Role of the manager

- manages Pager elements which are atomic objects of a paging reference and roughly corresponding current ``Bolt\Pager`` objects
- responsible for decoding/encoding pager objects from/to query parameters
- centralizing all pager related operations like ``$page = ($request->query) ? $request->query->get($param, $request->query->get('page', 1)) : 1;`` occurred redundantly some places in code ATM
- a pager element would be reached like ``$app['pager']['search']``
- ``(string) $app['pager']`` for encoding HTML query rather than ``Pager::makeLink()``
- no more ``Bolt\Legacy\Storage::GetContent()`` (and others) need to receive ``&$pager`` as argument, which is not so clear enough

At first sight... Thoughts?

rarila:
:+1:

GawainLynch:
> Bolt\Provider\PagerServiceProvider

:+1: 

> Bolt\PagerManager or Bolt\Pager\PagerManager

I am in two minds on namespace location… need more :coffee: first… and I am sure others will chime in.

> (string) $app['pager'] for encoding HTML query rather than Pager::makeLink()

Yes please… I already know the `__toString()` :cop: will give us a hard time, but solutions damnit!

> Furthermore each currently applied Pager:: based, global smelling static call would be served by the manager to fit TDD principle.

And pretty please can we also enhance tests on these at the same time… I know a start has been made, but it is a part of the code that is really under tested.

> no more Bolt\Legacy\Storage::GetContent() (and others) need to receive &$pager as argument, which is not so clear enough

:100: :100: :100:

SahAssar:
Fixed by #4648

PR #4642, opened 2015-12-30
Refactoring current Pager implementation

rixbeck:
Targeting #4627


src/Pager/Pager.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: rix
+ * Date: 2015.12.21.
+ * Time: 12:44
+ */
+
+namespace Bolt\Pager;
```
GawainLynch: Can you nuke these, or turn them into proper PHPDoc class blocks just above the class declaration, please :+1:

src/Legacy/Storage.php
```diff
@@ -1831,12 +1817,11 @@ protected function executeGetContentQueries($decoded)
...
      * @param string $textquery
      * @param string $parameters
-     * @param array  $pager
      * @param array  $whereparameters
      *
      * @return array
      */
-    public function getContent($textquery, $parameters = '', &$pager = [], $whereparameters = [])
+    public function getContent($textquery, $parameters = '', $whereparameters = [])
     {
```
GawainLynch: Not sure that changing the signature on this is a good idea. This whole class is deprecated, but we're trying to make the transitional in 3.x and remove it all in 4.0

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,321 @@
...
+ *
+ * Practical:
+ * ----------
+ *  - Variable ``pager`` injected into templates contains a member ``manager`` furthermore, so PagerManager API can be
+ *  accessed via
+ *  - Instantiating PagerManager - like reaching ``$app['pager']`` - decodes http page parameters and can be addressed
+ *  by their context id. So ``$app['pager']['entities']`` returns a Pager object decoded from query parameters was
+ *  ``page_entities=N`` originally
+ *
+ * @package Bolt\Pager
```
GawainLynch: :fire: `@package` use

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * Parameter Id builder.
+     * Http query parameter names of pager objects are built of PAGE constant and a name of pager that is refers
+     * to its context.
+     *
+     * @param $contextId
+     * @return string
+     */
+    public function makeParameterId($contextId = '')
```
GawainLynch: This, and the other uses should be defaulting to `null`

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,321 @@
...
+    {
+        $contextId = ($contextId !== '') ? '_'.$contextId : '';
```
GawainLynch: As above, comparison should be against a `null`, or whatever type `$contextId` should be.

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,321 @@
...
+     * Use for calling from template to build up paginated URL link.
+     * It preserves each http query parameters except which is requested and returns a http GET query string.
+     * Last parameter appended to the end of the string so its value can be just concatenating.
+     * Against deprecated Pager::makelink() this is will build link based on initialized pagers.
+     *
+     * @param string $linkFor [optional] Id of pager the link should be built for. With empty argument passing
+     *          the link will be built for the first initialized pager object found.
+     * @return string GET query string
+     */
+    public function makelink($linkFor = '')
```
GawainLynch: PSR-2:

```php
public function makeLink($linkFor = null)
```

src/Pager/Pager.php
```diff
@@ -0,0 +1,36 @@
...
+     *
+     *  public $for;
+     *  public $count;
+     *  public $totalpages;
+     *  public $current;
+     *  public $showing_from;
+     *  public $showing_to;
+     *
+     *  public $manager;
+     */
```
GawainLynch: Are these supposed to be `@method` docs?

src/Pager/Pager.php
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     *
+     * @param string $linkFor
+     * @return mixed
+     */
+    public function makelink($linkFor = '')
+    {
+        return $this->manager->makelink($linkFor);
+    }
```
GawainLynch: `$this->manager` isn't a property and if "magic" should be checked as an object before use.

src/Storage/EntityManager.php
```diff
@@ -423,20 +423,4 @@ public function __call($method, $args)
...
-     * @param array  $pager
-     * @param array  $whereparameters
-     *
-     * @return \Bolt\Legacy\Content|\Bolt\Legacy\Content[]
-     */
-    public function getContent($textquery, $parameters = [], &$pager = [], $whereparameters = [])
-    {
-        return $this->legacy()->getContent($textquery, $parameters, $pager, $whereparameters);
-    }
 }
```
GawainLynch: I am pretty sure you didn't want to remove this… it is the legacy bridge for `$app['storage']` to that function.

src/Twig/SetcontentNode.php
```diff
@@ -29,7 +29,7 @@ public function compile(\Twig_Compiler $compiler)
 
         if (!is_null($this->getNode('wherearguments'))) {
             $compiler
-                ->raw(', $pager, ')
+                //->raw(', $pager, ')
                 ->subcompile($this->getNode('wherearguments'));
```
GawainLynch: If it needs to go, can we remove please.

src/Legacy/Storage.php
```diff
@@ -2931,42 +2915,4 @@ protected function findContent($tablename, $contentId)
...
-            }
-        } else {
-            return static::$pager;
-        }
-    }
-
-    public function isEmptyPager()
-    {
-        return (count(static::$pager) === 0);
-    }
```
GawainLynch: I  know we've ripped the head off a lot of stuff, but this class is kind of central. Probably better to leave these as empty functions so that they don't throw an exception on deprecated use.

Can we :fire: this class already :grin:

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,315 @@
...
+         */
+        if ($this->link) {
+            return $this->link;
+        }
+
+        $pagerid = $this->findPagerId($linkFor);
+        $saved = $this->pagers[$pagerid];
+        unset($this->pagers[$pagerid]);
+
+        $qparams = $this->app['request']->query->all();
```
GawainLynch: Can you use `$app['request_stack']->getCurrentRequest()` and check if it is `null` before using to prevent use out of cycle.

GawainLynch: …and below

src/Storage/EntityManager.php
rixbeck: Yess

src/Pager/PagerManager.php
```diff
@@ -0,0 +1,315 @@
...
+
+    protected $app;
+    protected $link;
+    protected $pagers = [];
+
+    public function __construct(Application $app)
+    {
+        $this->app = $app;
+        $this->pagers = $this->decodeHttpQuery();
+    }
```
GawainLynch: The constructor seems to only need `$app['request_stack']`. Also that request itself almost certainly doesn't exist at the point this is instantiated and we shouldn't be building class parameters now, only as needed later in the cycle.

src/Pager/PagerManager.php
rixbeck: It is reasonable..

src/Legacy/Storage.php
rixbeck: I'm not sure what do you suggest. Because it is a reference we'll have declared but unused variables on calls.

GawainLynch: Sure, but if someone does:

```php
$textquery = 'pages';
$parameters = 'foo';
$whereparameters = ['status' => 'hello'];

$content = $app['storage']->getContent($textquery, $parameters, [], $whereparameters);
```

They are going to lose their `$whereparameters`. One of many reasons to pass around objects and not arrays. Unless someone has a brilliant idea, I would think we'd need it for BC until we remove that class.

src/Storage/EntityManager.php
rixbeck: Wait! No, this was for purpose because magic __call() will work on if we don't use ``&$pager`` as argument anymore. So I would remove that parameter from signature... Thoughts?

src/Twig/SetcontentNode.php
rixbeck: If we can agree should we use or not that ``&$pager`` parameter then will remove... or not. :-)

PR #4645, opened 2016-01-01
fix Catchable fatal error: Object of class Bolt\Content could not be …

fadoe:
…converted to string in default record template


fadoe:
When I activate the default template and show a single page, this exception will be thrown.
This fixed this exception in record.twig

bobdenotter:
Good catch! Thank you, @fadoe! :+1:

PR #4648, opened 2016-01-01
Refactoring current Pager implementation, Fixes #4627

### LINKED ISSUE #4627: [RFC] Refactoring current Pager implementation

rixbeck:
...continuing the accidentally closed #4642 
:clap: 

Fixes #4627


bobdenotter:
Green lights on Travis! Does this mean we're Good to Go? :-)

rixbeck:
Finially it is, Bob! :-)

src/Legacy/Storage.php
```diff
@@ -940,7 +937,7 @@ public function searchContentType($contenttypename, array $parameters = [], &$pa
 
         // If we're allowed to use pagination, use the 'page' parameter.
         if (!empty($parameters['paging']) && $this->app->raw('request') instanceof Request) {
```
CarsonF: Can you change the second part of this check to:
```php
$this->app['request_stack']->getCurrentRequest() !== null
```

Or maybe even:
```php
$app['pager']->isInitialized()
```

rixbeck:
... then screwtinizer?

GawainLynch:
> ... then screwtinizer?

Restarted for the sake of seeing it prove itself as useless :-1:

GawainLynch:
@CarsonF: Are you calling this GTG now?

GawainLynch:
Thanks @rixbeck … let's get this in and keep the train a moving!

PR #4654, opened 2016-01-02
Fix selects with empty values

rarila:
In buic-select an empty option was added to non multiple selects regardless of if there already ways an empty option included. That produced an output like this:

```html
<select><option value="" label="–"></option><option selected></option>…
```

Now we check for empty values for correct output.


bobdenotter:
Whoop whoop! :shipit:

PR #4656, opened 2016-01-03
Set depth limit on file/directory look up

GawainLynch:
Remember kids, :koala:s are dangerous… Code with care!


bobdenotter:
Whoop! :shipit:

PR #4660, opened 2016-01-03
Set depth limit on file/directory look up

### LINKED ISSUE #4659: Async File Listing Bug
Again on master only.

An async browse files request on the default directory, eg: `/async/browse` seems to do a recursive listing of all files rather than just a listing of the files within the base folder.

Once you navigate to a folder then behaviour seems to return to normal, eg when the url shows: `/async/browse/files/2016-01` then the listing appears to behave correctly again, showing only files in that level folder.

GawainLynch:
Fixes #4659
Matches #4656
Requires :beers:


bobdenotter:
:shipit:

PR #4678, opened 2016-01-06
[Feature] Added capability to set a record amount for specific kind of taxonomy

ScreamZ:
Hey guys, thanks for this great CMS.

I'm facing a case where I want to define differents amounts of records for multiple kind of taxonomies.

For my articles I have a taxonomy called **Thematic**, for which I want 2 records fetched in the full view of the taxonomy **Thematic**, but I also have a taxonomy **levelofDifficulty** and for that one I want 6 records per pages.

Therefore I made a little fix that override the theme's specific or theme's global config. Do you agree about that ? Or do you have a better way to handle that ?

PS : I wanted also include the listing option, but I contents doesn't have the same fields, it can make some issues. So nevermind about listing order. ^

PS2 : Carry on buddy !


bobdenotter:
@ScreamZ Good catch, thank you! :+1: 

(I will backport this to `release/2.2` as well)

PR #4680, opened 2016-01-08
Remove a ghost from the past …

rarila:
Looks like this was never removed although not needed anymore.

@rarila blames Zuul.


GawainLynch:
:ghost: :gun:

PR #4683, opened 2016-01-08
Clear 'removedColumns' attribute from ContentType table diffs to prevent accidental data removal

GawainLynch:


bobdenotter:
:shipit:

PR #4684, opened 2016-01-08
Look for slugs and singular ones too

### LINKED ISSUE #4674: contenttype name with _ (underscore) not allowed
If you create a content type with an underscore in the identifier the database update will break. The following will result in a lot of notices like `Notice: Undefined index: custom_things in /[]/html/src/Storage/Database/Schema/Builder/ContentTables.php on line 33`

```yml
custom_things:
  name: Things
  singular_name: Thing
  fields:
    title:
      type: text
      required: true
      label: "Title"
    slug:
      type: slug
      uses: title
    body:
      type: textarea
      label: "Body"
```

Also the fields `title` and `body` will not be created, so the database update will keep nagging you that the database is broken.

GawainLynch:
Fixes #4674

### DIFF
--- a/app/view/twig/editcontent/fields/_slug.twig
+++ b/app/view/twig/editcontent/fields/_slug.twig
@@ -62,7 +62,7 @@
 {% block fieldset_controls %}
     <div class="col-sm-12">
         <div class="input-group {{ mode }}">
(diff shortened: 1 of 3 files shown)


bobdenotter:
:shipit:, too.

Issue #4686, opened 2016-01-10
"Low level errors" are broken in current master. 


bobdenotter:
To reproduce, edit a contenttype to have neither a name nor slug.

```
Fatal error: Uncaught exception 'Bolt\Exception\LowlevelException' with message
'In contenttype projects, neither 'singular_name' nor 'singular_slug' is set.
Please edit contenttypes.yml, and correct this.' in /Users/bob/Sites/bolt-
origin/src/Config.php:452 Stack trace: #0 /Users/bob/Sites/bolt-
origin/src/Config.php(365): Bolt\Config->parseContentType('projects', Array,
Array) #1 /Users/bob/Sites/bolt-origin/src/Config.php(222):
Bolt\Config->parseContentTypes(Array) #2 /Users/bob/Sites/bolt-
origin/src/Config.php(84): Bolt\Config->getConfig() #3 /Users/bob/Sites/bolt-
origin/src/Config.php(78): Bolt\Config->initialize() #4 /Users/bob/Sites/bolt-
origin/src/Provider/ConfigServiceProvider.php(16):
Bolt\Config->__construct(Object(Bolt\Application)) #5 /Users/bob/Sites/bolt-
origin/vendor/pimple/pimple/lib/Pimple.php(126): Bolt\Provider\ConfigServiceProv
ider->Bolt\Provider\{closure}(Object(Bolt\Application)) #6 /Users/bob/Sites
/bolt-origin/vendor/pimple/pimple/lib/Pimple.php(83):
Bolt\Application::{closure}(Object(Bolt\Applicatio in /Users/bob/Sites/bolt-
origin/src/Config.php on line 452
```

bobdenotter:
Fixed in #4688

PR #4688, opened 2016-01-10
Fix fatal error when missing "fields" in contenttype.

### LINKED ISSUE #4687: When "fields" is not set in a contenttype, PHP throws a fatal error, instead of a nice readable error.
for example: 

```
Notice: Undefined index: fields in /Users/bob/Sites/bolt-origin/src/Config.php on line 497

Catchable fatal error: Argument 1 passed to Bolt\Config::parseFieldsAndGroups()
must be of the type array, null given, called in /Users/bob/Sites/bolt-
origin/src/Config.php on line 497 and defined in /Users/bob/Sites/bolt-
origin/src/Config.php on line 527
```

bobdenotter:
Fixes #4687

(or rather: It will, after #4686 has been resolved)

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -452,6 +452,12 @@ class Config
             throw new LowlevelException($error);
         }
 
+        // Contenttypes without fields make no sense.
+        if (!isset($contentType['fields'])) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
This need a cherry-pick into 2.2?

bobdenotter:
> This need a cherry-pick into 2.2?

Might as well. :-)

PR #4689, opened 2016-01-10
Fix fatal error when missing "fields" in contenttype.

### LINKED ISSUE #4687: When "fields" is not set in a contenttype, PHP throws a fatal error, instead of a nice readable error.

bobdenotter:
Fixes #4687 for Bolt 2.2

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -418,6 +418,12 @@ class Config
             throw new LowlevelException($error);
         }
 
+        // Contenttypes without fields make no sense.
+        if (!isset($contentType['fields'])) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Is already cherrypicked. No need for this.

PR #4691, opened 2016-01-10
Added capability to set a record amount for specific kind of taxonomy

GawainLynch:
Back port of #4678 into an existing class file… :koala:


bobdenotter:
Ha, looks like I derped there, this afternoon. :-)

GawainLynch:
:wink:

PR #4693, opened 2016-01-11
fix multiple repeater fields

### LINKED ISSUE #4692: Multiple repeater fields SQL error
Hi!

A model with multiple repeater fields currently cannot be queried:
The SQL generated in src/Storage/Field/Type/RepeaterType.php:load contains the fixed alias "f" leading to a failing unique constraint. 
Moreover getPlatformGroupConcat uses the alias "fields"  such that even when the first alias is changed, only the values of the last repeater field is available.

I'll create a PR wtith a hackish fix.

Thanks!

d-m-:
Fixes #4692 

Still creates redundant SQL: there is one GROUP_CONCAT per repeater field, each returning the same values (all field values of all repeater fields of the respective contenttype).

Only tested on sqlite.

### DIFF
--- a/src/Storage/Field/Type/RepeaterType.php
+++ b/src/Storage/Field/Type/RepeaterType.php
@@ -37,12 +37,14 @@ class RepeaterType extends FieldTypeBase
             $alias = $from[0]['table'];
         }
 
+        $dummy = "f_" . $field;
+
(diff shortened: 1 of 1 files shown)


bobdenotter:
Hi @d-m-, 

Looks good to me, but @rossriley and @GawainLynch know most about this part of the code, so I'll leave it to either one of them to merge in.

PR #4698, opened 2016-01-12
Bump PHP required version to 5.5.9

GawainLynch:


CarsonF:
scrutinizer is so stupid sometimes. :+1:

PR #4704, opened 2016-01-15
Stay on target… stay on target…

GawainLynch:
This is one for the OCD-IDE users mainly, as it will give reliable code completion.

For everyone else, we'll document the snake-case ones in the documentation.


bobdenotter:
Whoop!

PR #4709, opened 2016-01-16
Make base 2016 default

SahAssar:
I'm proposing making base-2016 the default (as discussed on IRC this morning/lunchtime) now so that we actually find any rough edges before we get into beta/RC. This PR also includes all the necessary test changes (I hope).

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -18,7 +18,7 @@ payoff: The amazing payoff goes here
 #
 # Don't edit the templates in theme/base-2014, because they _will_ get updated
 # in next releases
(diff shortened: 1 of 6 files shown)


GawainLynch:
:+1:

PhillippOhlandt:
:+1:

bobdenotter:
What's Travis' problem with this?

GawainLynch:
That the Composer install is partially "faked" and *some* things are rsync'ed but not themes. They are safe to ignore in this case.

bobdenotter:
Allright. Let's go then! :+1:

PR #4718, opened 2016-01-17
Add Twig `{% switch %}` statement.

### LINKED ISSUE #4717: We need a `{% switch %}` statement in Twig.
Like this: https://craftcms.com/docs/templating/switch

@see: https://gist.github.com/maxgalbu/9409182

bobdenotter:
Fixes #4717

### DIFF
--- a/src/Twig/SwitchNode.php
+++ b/src/Twig/SwitchNode.php
@@ -12,7 +12,7 @@ namespace Bolt\Twig;
  */
 class SwitchNode extends \Twig_Node
 {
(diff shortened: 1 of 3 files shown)


bobdenotter:
Usage is: 

```
{% set foo = 1 %}
{% switch test %}
    {% case 1 %}
        Foo was equal to the number one.
    {% case 2 %}
        Foo was two.
    {% default %}
        This is the default fallback.
{% endswitch %}
```

PR #4719, opened 2016-01-18
Add matching paths to 'twig.loader.bolt_filesystem' as are in 'twig.loader'

GawainLynch:
@CarsonF I know we need to probably approach this better prior to release, just haven't been able to schedule time with you to finalise.


bobdenotter:
Boom! :boom:

PR #4720, opened 2016-01-18
[2.2] Add js events to editfile/editcontent ajax saving

SahAssar:
Add js events to allow extensions and custom fields to know about the savestate of content/files.


rarila:
See https://github.com/bolt/bolt/pull/4721#issuecomment-172585889

SahAssar:
Waiting to see if there is more to do/rename/fix in #4721, so this is not gtg yet.

PR #4721, opened 2016-01-18
Add js events to editfile/editcontent ajax saving

SahAssar:
Add js events to allow extensions and custom fields to know about the savestate of content/files. Oh, and 	port #4682 to master.

### DIFF
--- a/app/src/js/init.js
+++ b/app/src/js/init.js
@@ -42,14 +42,14 @@ var init = {
      * @param {object} data
      * @returns {undefined}
      * 
-     * @fires start.bolt.editfile.save
-     * @fires done.bolt.editfile.save
-     * @fires fail.bolt.editfile.save
-     * @fires always.bolt.editfile.save
+     * @fires start.bolt.file.save
+     * @fires done.bolt.file.save
+     * @fires fail.bolt.file.save
+     * @fires always.bolt.file.save
      */
     bindEditFile: function (data) {
         $('#saveeditfile').bind('click', function () {
-            $(Bolt).trigger('start.bolt.editfile.save');
+            $(Bolt).trigger('start.bolt.file.save');
 
             // If not on mobile (i.e. Codemirror is present), copy back to the textarea.
             if (typeof CodeMirror !== 'undefined') {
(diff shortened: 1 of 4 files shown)


rarila:
A few thoughts:

I wonder if ``start.bolt.content.save`` and ``start.bolt.file.save`` would do or is that not precise enough? In other words, can it happen that we save content or a file if we are not in edit mode?

And then I would probably name the gmaps ones ``xxx.bolt.gmaps``(without ``field``) as there is nothing field specific in there. Just a request to load the Google maps api that could also be used elsewhere. 

Name it ``gmapsapi`` instead? I  think ``gmaps`` will do to be short.

Last thing here: shouldn't it be ``start.bolt.gmaps.load``, … to be in line with the other event names?

So we would endup ``status``.``namespace``.``subject``.``verb`` with our custom events.

Hey, I hate designing things that are BC relevant:-)
Opinions!?

GawainLynch:
There is something strange about this PHP… I can't put my finger on it…

SahAssar:
@GawainLynch I found some nice, calming PHP for you to read: http://bit.ly/1ntINPy

GawainLynch:
Oh, I like this one
```php
captial_SCREWYOU_dangit()
```

:wink:

SahAssar:
@rarila I renamed them, your way seems more logical. You mentioned on IRC about going more generic and putting more things into extra parameters, but personally I think that it's better if extension developers have somewhat simple events to listen to and that they don't have to write boilerplate code like
```
$(Bolt).on('start.bolt', function(event, data){
    if(data.type === 'content' && data.action === 'save'){
        /* do stuff here */
    }
});
```
But I'm open to whatever feels like it will work best :) How do you usually do it?

rarila:
@SahAssar you're probably right. It might be shorter to just have longer event names instead lots of conditional code here. But I'm still missing the overall vision inside me here and wonder in what direction this eventing will go in the end. So we probably have to do our best now and simply bite the bitter bc breaking pill in the future if it happens we come to the conclusion that we were wrong.

PS: I still wonder about ``gmap`` vs. ``gmapapi``. Did I tell you that I hate that naming stuff :-)

SahAssar:
@rarila Is there some way that you can think of that gives us the equivalent functionality and that you are more comfortable with?

I went with gmapapi since it only loads that api, it dosen't actually create a "gmap".

rarila:
@SahAssar no, I guess it is ok that way. It's probably the way software debelopment works. You add stuff after stuff and at some point you wonder why you didn't had it done in another way and them shout "refactor" or "rewrite" :smile: 

gmapi: Ok … eh …:bulb: … how about ``gmapi``?

jadwigo:
what's wrong with `googlemapsapi` or something else which is actually telling you what it is without needing to guess

rarila:
@jadwigo we devs like it mysterious, are lazy and serve this code clientside… every byte counts :sunglasses:

jadwigo:
@rarila I know, but http://www.gmapi.com/ exists ...

SahAssar:
Isn't gmapapi pretty unambiguous? Although I have no problem changing it to googlemapsapi if that seems better.

jadwigo:
I would prefer `googlemapsapi` :) just because the provider is `google`, and not `gstar` or anything else that starts with a `g`.

Similarly I would prefer `foursquarelocationapi` over `flocapi` - some extra letters are cheap and figuring out if `f` stands for foursquare or facebook would be annoying.

SahAssar:
@jadwigo Fixed :)

rarila:
Sorry

GawainLynch:
:shipit: and :wrench: later.

PR #4722, opened 2016-01-18
Add {{ fields() }} Twig tag

### LINKED ISSUE #4714: [RFC] We need a `{{ outputfields() }}` for outputting all fields.
Currently, our `record.twig` has a bunch of boilerplate to basically output each field sequentially. When we have 'repeating fields' in, this will only increase. 

At the same time, I've often seen people use the default `record.twig`, and the code in there is fugly.If we have a Twig tag to do this, we could clean up this mess in the `record.twig` templates, and open it up for use to more people. 

(I'm not too happy with the name `outputfields()` though. Any suggestions?)

bobdenotter:
Fixes #4714


bobdenotter:
I'll be tweaking this a bit more. Please hold off on merge for a bit. :-)

bobdenotter:
Added docblock / comments. Should be GTG now.

PR #4724, opened 2016-01-19
add condition to image field in theme

jadwigo:
this will prevent some warnings and other ugly output

### DIFF
--- a/theme/base-2016/record.twig
+++ b/theme/base-2016/record.twig
@@ -17,7 +17,7 @@
                             </div>
                         {% endfor %}
                     </div>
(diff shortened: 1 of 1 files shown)


SahAssar:
Is that still needed after #4716?

jadwigo:
The rest of the fields also check for `value is not empty` so it's weird to not do it for images.
Also it does not hurt if we don't call {{ popup }} if we know we don't have anything to display anyway. 

And I got notices in my theme with an alpha3 build that was made yesterday. (with debug mode on, and a server that displays all notices)

GawainLynch:
:+1:

PR #4726, opened 2016-01-19
Add repeater fields to legacy content objects

rossriley:
This makes repeating fields available in templates. Note that this PR doesn't actually implement the output of these fields, just makes them available in the resulting `record` object.

The field values are different to standard value based fields in that the field contains an iterable collection, so you will need to do two for loops one to loop the sets and then a further inner loop to list each of the fields within a set.


GawainLynch:
:+1:

PR #4727, opened 2016-01-19
Update Composer.json for Whoops 2.0.

bobdenotter:
Shiny and new! http://filp.github.io/whoops/


bobdenotter:
Herp derp.. We shouldn't update this for 2.2.

PR #4729, opened 2016-01-19
VarDumper, Casters, and Image Info! Oh my!

### LINKED ISSUE #4170: Twig dump() does not output the current context.
Tested on Bolt 2.2.11, Chrome 45.0.2454.101, OSX 10.10.3, Nginx 1.8.0, PHP 5.6.13_3, Xdebug v2.3.3, MariaDB 10.10
## Description:

Calling Twig dump() results in **Warning: Missing argument 1 for Bolt\Twig\TwigExtension::printDump()**. 

http://twig.sensiolabs.org/doc/functions/dump.html says that calling dump() is supposed to return the current context, but it does not.

## To Recreate:

Insert {{ dump() }} into index.twig(or equivalent).  Run the app.  Get the Warning.

## The Issue:

Bolt extends Twig's dump() function with it's own printDump function in [https://github.com/bolt/bolt/blob/master/src/Twig/TwigExtension.php](https://github.com/bolt/bolt/blob/e21f6d8248cd50115034875d071efb5cce72196f/src/Twig/TwigExtension.php#L440)

printDump() requires a single argument, $var but does not return the context when the argument is empty.

### LINKED ISSUE #4679: [RFC] Imageinfo changes in 3.0
The output of the imageinfo function has changed quite a bit in master, A bunch of "nice to have" data seems to have been removed, like aspectratio and landscape/portrait/square.

It would be nice if we could agree on what we want to include in it, and hopefully make it as compatible with 2.2 themes as possible.

Example of 2.2 output:

	array:12 [▼
	  "width" => 1391
	  "height" => 1852
	  "type" => "jpeg"
	  "mime" => "image/jpeg"
	  "aspectratio" => 0,75107991360691
	  "filename" => "2016-01/19142717931-30117fa7ef-o.jpg"
	  "fullpath" => "/var/www/test/files/2016-01/19142717931-30117fa7ef-o.jpg"
	  "url" => "/files/2016-01/19142717931-30117fa7ef-o.jpg"
	  "exif" => array:5 [▼
		"latitude" => ""
		"longitude" => false
		"datetime" => false
		"orientation" => false
		"aspectratio" => 0,75107991360691
	  ]
	  "landscape" => false
	  "portrait" => true
	  "square" => false
	]

Example of master/3.0 output:

	Info {#2418 ▼
	  #dimensions: Dimensions {#2095 ▼
		#width: 1391
		#height: 1852
	  }
	  #type: Type {#1856 ▼
		-id: 2
		-name: "JPEG"
	  }
	  #bits: 8
	  #channels: 3
	  #mime: "image/jpeg"
	  #exif: Exif {#2096 ▼
		#data: array:4 [▼
		  "FileSize" => 492026
		  "MimeType" => "image/jpeg"
		  "height" => 1852
		  "width" => 1391
		]
		#rawData: array:8 [▼
		  "FileName" => "9k="
		  "FileDateTime" => 0
		  "FileSize" => 492026
		  "FileType" => 2
		  "MimeType" => "image/jpeg"
		  "SectionsFound" => ""
		  "COMPUTED" => array:4 [▼
			"html" => "width="1391" height="1852""
			"Height" => 1852
			"Width" => 1391
			"IsColor" => 1
		  ]
		  "IPTC" => []
		]
	  }
	}

CarsonF:
Working backwards: This fixes #4679. The `imageinfo` twig function now returns a filesystem image, which has image info as one of its properties. Note that BC is mostly kept with magic calls. [`NullableImage`](https://github.com/CarsonF/bolt/blob/bugfix/twig-dump/src/Filesystem/Handler/NullableImage.php) was needed to return an empty `Image\Info` where there is a problem, since twig cannot catch exceptions.

The other problem @SahAssar was bringing up is that `dump()` doesn't show all the properties he is expecting (lazy getters and getters that don't have properties). Symfony's VarDumper has a nifty way to extend the data pulled from objects called _Casters_. I created [`FilesystemCasters`](https://github.com/CarsonF/bolt/blob/bugfix/twig-dump/src/Debug/Caster/FilesystemCasters.php) to populate these lazy properties and getter methods without properties.

This caster needs to be injected into `VarCloner`, so DI was needed for the entire object tree (see [`DumperServiceProvider`](https://github.com/CarsonF/bolt/blob/bugfix/twig-dump/src/Provider/DumperServiceProvider.php)).

This also needs to be passed into Twig. Symfony has a `DumpExtension`. I've hooked in, and that fixes #4170. However, the dumper cannot be passed in, only the cloner. So I extended that to allow the dumper to be passed in. Hopefully, I can get that change upstreamed and we won't have to reimplement all of it.

All in all it was lots of stuff, but this is a good foundation for more casters, which will make `dump` more helpful.

Before: 
before
After: 
after

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,6 @@
     ],
     "license" : "MIT",
     "require" : {
-        "bolt/dumper" : "~2.0",
         "bolt/filesystem" : "^2.0@dev",
         "bolt/package-wrapper": "^3.0",
         "bolt/pathogen" : "~0.6",
--- a/src/Application.php
+++ b/src/Application.php
@@ -181,6 +181,8 @@ class Application extends Silex\Application
             $this->register(new WhoopsServiceProvider());
         }
 
+        $this->register(new Provider\DumperServiceProvider());
+
         // Initialize Web Profiler providers
         $this->initProfiler();
     }
--- a/src/Debug/Caster/AbstractCasterProvider.php
+++ b/src/Debug/Caster/AbstractCasterProvider.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Bolt\Debug\Caster;
+
+/**
+ * Abstract class providing casters.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ */
+abstract class AbstractCasterProvider
+{
+    /**
+     * @return callable[]
+     */
+    protected static function defineCasters()
+    {
+        return [];
+    }
+
+    public static function getCasters()
+    {
+        $cls = get_called_class();
+
+        return array_map(
+            function ($func) use ($cls) {
+                if (method_exists($cls, $func)) {
+                    return [$cls, $func];
+                }
+
+                return $func;
+            },
+            static::defineCasters()
+        );
+    }
+}
--- a/src/Debug/Caster/FilesystemCasters.php
+++ b/src/Debug/Caster/FilesystemCasters.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Bolt\Debug\Caster;
+
+use Bolt\Filesystem\Handler\DirectoryInterface;
+use Bolt\Filesystem\Handler\FileInterface;
+use Bolt\Filesystem\Handler\HandlerInterface;
+use Bolt\Filesystem\Handler\Image;
+use Symfony\Component\VarDumper\Caster\Caster;
+use Symfony\Component\VarDumper\Cloner\Stub;
+
+/**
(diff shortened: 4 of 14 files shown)


bobdenotter:
My precious `bolt/dumper`!

CarsonF:
@bobdenotter muhahaha

GawainLynch:
For the record, and for those not in on the joke. `bolt/dumper` was deprecated in 2.0-dev and is being removed in a major version bump.

rarila:
:+1:

PR #4731, opened 2016-01-20
fix Multibyte String BUG

cx-emerge:
fix Multibyte String BUG

### DIFF
--- a/src/Twig/Handler/TextHandler.php
+++ b/src/Twig/Handler/TextHandler.php
@@ -147,6 +147,10 @@ class TextHandler
      */
     public function ucfirst($str)
     {
+        if (!function_exists('mb_ucfirst') && function_exists('mb_substr')) {
+            return mb_strtoupper(mb_substr($str, 0, 1)) . mb_substr($str, 1);
+        }
+        
         return ucfirst($str);
     }
 }


bobdenotter:
Hi @cx-emerge, 

In practice, how likely is it for a PHP install to have `mb_substr`, but not `mb_ucfirst`?

rarila:
```php
if (function_exists('mb_strtoupper') && function_exists('mb_substr')) {
     return mb_strtoupper(mb_substr($str, 0, 1)) . mb_substr($str, 1);
} else {
     return ucfirst($str);
}
```

… sounds more reasonable to me.

cx-emerge:
``` PHP
if (extension_loaded('mbstring')) {
	return mb_strtoupper(mb_substr($str, 0, 1)) . mb_substr($str, 1);
} else {
	return ucfirst($str);
}
```
sorry, In Windows have this problem, but in Linux do not have this problem.
I think it might be a file encoding issues
http://php.net/manual/en/function.ucfirst.php#84122

CarsonF:
Can we not just deprecate this function and forward it to `title` function?

GawainLynch:
> Can we not just deprecate this function and forward it to `title` function?

:+1:

CarsonF:
@bobdenotter Do you have a problem with using the `title` function instead?

GawainLynch:
Closing as this is handled correctly by Symfony and we should not be implementing our own.

PR #4732, opened 2016-01-20
Make showimage/popup do an empty check

SahAssar:
to catch unfilled fields and getImage


GawainLynch:
Sending in before Texas comes online :stuck_out_tongue_winking_eye:

PR #4733, opened 2016-01-20
Fix a few base-2016 things

SahAssar:
* Add widgetchecks where appropriate
* Remove randomimage stuff
* Make index.twig able to show both listings and record
* Simplify index.twig
* Simplify title & description in head
* Use getImage for headerimages and preview images

The frontpage_image setting won't work until the themebase issue is solved.


jadwigo:
less random - me like!

bobdenotter:
Looks good to me! :+1:

PR #4734, opened 2016-01-20
fix dashboardnews 504 Gateway Time-out

cx-emerge:
fix dashboardnews 504 Gateway Time-out

### DIFF
--- a/src/Controllers/Async.php
+++ b/src/Controllers/Async.php
@@ -127,24 +127,27 @@ class Async implements ControllerProviderInterface
                 base64_encode($hostname)
             );
 
+            // Define guzzle options
+            $guzzleOptions['config']['curl'] = array();
+            $curlOptions = &$guzzleOptions['config']['curl'];
+
             // Options valid if using a proxy
             if ($app['config']->get('general/httpProxy')) {
-                $curlOptions = array(
-                    'CURLOPT_PROXY'        => $app['config']->get('general/httpProxy/host'),
-                    'CURLOPT_PROXYTYPE'    => 'CURLPROXY_HTTP',
-                    'CURLOPT_PROXYUSERPWD' => $app['config']->get('general/httpProxy/user') . ':' .
-                                                $app['config']->get('general/httpProxy/password')
-                );
+                $curlOptions[CURLOPT_PROXY] = $app['config']->get('general/httpProxy/host');
+                $curlOptions[CURLOPT_PROXYTYPE] = 'CURLPROXY_HTTP';
+                $curlOptions[CURLOPT_PROXYUSERPWD] = $app['config']->get('general/httpProxy/user') . ':' .
+                                                       $app['config']->get('general/httpProxy/password')
+                ;
             }
 
             // Standard option(s)
-            $curlOptions['CURLOPT_CONNECTTIMEOUT'] = 5;
+            $curlOptions[CURLOPT_CONNECTTIMEOUT] = 5;
 
             try {
                 if ($app['deprecated.php']) {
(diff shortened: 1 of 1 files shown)


src/Controllers/Async.php
```diff
@@ -127,24 +127,27 @@ public function dashboardnews(Silex\Application $app, Request $request)
...
 
             // Standard option(s)
-            $curlOptions['CURLOPT_CONNECTTIMEOUT'] = 5;
+            $curlOptions[CURLOPT_CONNECTTIMEOUT] = 5;
 
             try {
                 if ($app['deprecated.php']) {
-                    $fetchedNewsData = $app['guzzle.client']->get($url, null, $curlOptions)->send()->getBody(true);
+                    $fetchedNewsData = $app['guzzle.client']->get($url, $guzzleOptions)->send()->getBody(true);
                 } else {
```
GawainLynch: This is incorrect, as it is for Guzzle3 and has the function signature:
```php
public function get($uri = null, $headers = null, $options = array())
```

src/Controllers/Async.php
```diff
@@ -127,24 +127,27 @@ public function dashboardnews(Silex\Application $app, Request $request)
...
-                    $fetchedNewsData = $app['guzzle.client']->get($url, array(), $curlOptions)->getBody(true);
+                    $fetchedNewsData = $app['guzzle.client']->get($url, $guzzleOptions)->getBody(true);
                 }
```
GawainLynch: This one is correct :+1:

src/Controllers/Async.php
```diff
@@ -127,24 +127,27 @@ public function dashboardnews(Silex\Application $app, Request $request)
...
-                    'CURLOPT_PROXY'        => $app['config']->get('general/httpProxy/host'),
-                    'CURLOPT_PROXYTYPE'    => 'CURLPROXY_HTTP',
-                    'CURLOPT_PROXYUSERPWD' => $app['config']->get('general/httpProxy/user') . ':' .
-                                                $app['config']->get('general/httpProxy/password')
-                );
+                $curlOptions[CURLOPT_PROXY] = $app['config']->get('general/httpProxy/host');
+                $curlOptions[CURLOPT_PROXYTYPE] = 'CURLPROXY_HTTP';
+                $curlOptions[CURLOPT_PROXYUSERPWD] = $app['config']->get('general/httpProxy/user') . ':' .
+                                                       $app['config']->get('general/httpProxy/password')
+                ;
```
GawainLynch: You can keep the array, only the `'` needs to go on the contants

src/Controllers/Async.php
```diff
@@ -127,24 +127,27 @@ public function dashboardnews(Silex\Application $app, Request $request)
...
             // Options valid if using a proxy
             if ($app['config']->get('general/httpProxy')) {
-                $curlOptions = array(
-                    'CURLOPT_PROXY'        => $app['config']->get('general/httpProxy/host'),
-                    'CURLOPT_PROXYTYPE'    => 'CURLPROXY_HTTP',
-                    'CURLOPT_PROXYUSERPWD' => $app['config']->get('general/httpProxy/user') . ':' .
-                                                $app['config']->get('general/httpProxy/password')
-                );
+                $curlOptions[CURLOPT_PROXY] = $app['config']->get('general/httpProxy/host');
+                $curlOptions[CURLOPT_PROXYTYPE] = 'CURLPROXY_HTTP';
```
GawainLynch: `CURLPROXY_HTTP` is also a constant and needs to be changed

PR #4735, opened 2016-01-20
fix PackageManager ping 504 Gateway Time-out

cx-emerge:
fix PackageManager ping 504 Gateway Time-out

### DIFF
--- a/src/Composer/PackageManager.php
+++ b/src/Composer/PackageManager.php
@@ -544,14 +544,23 @@ class PackageManager
             $query = array();
         }
 
+        $guzzleOptions = array(
+            'query' => $query,
+            'config' => array(
+                'curl' => array(
+                    CURLOPT_CONNECTTIMEOUT => 5
+                )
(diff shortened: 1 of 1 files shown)


src/Composer/PackageManager.php
```diff
@@ -544,14 +544,23 @@ private function ping($addquery = false)
...
+                )
+            )
+        );
+
         try {
             /** @deprecated remove when PHP 5.3 support is dropped */
             if ($this->app['deprecated.php']) {
                 /** @var $response \Guzzle\Http\Message\Response  */
-                $response = $this->app['guzzle.client']->head($uri, null, array('query' => $query))->send();
+                $response = $this->app['guzzle.client']->head($uri, $guzzleOptions)->send();
```
GawainLynch: As with #4734 this is for Guzzle 3 and is not correct

GawainLynch:
Merging and fixing locally.

PR #4736, opened 2016-01-21
fix iis rewrite extend 404

cx-emerge:
fix iis rewrite extend 404

admin/extend/install?package=sahassar%2Ffoxycart&version=0.1
404


GawainLynch:
Cherry picked into master as well

PR #4737, opened 2016-01-21
Move noscript message to a danger flash message

### LINKED ISSUE #4615: [RFC] Make Javascript mandatory for the backend
At the moment we display a warning

> JavaScript is currently disabled in your browser. Most functionality in Bolt will work without it, but for greater ease of use we recommend you enable JavaScript in your browser.

in the sidepanel of dashboard and omnisearch. I haven't testet it but I guess it will range more between some things will probably work and you're in danger of corrupting things. Too much things depend on Javascript to not only work more comfortable but to work at all. Even if 50% of all pages would work, the warning would have to be displayed on every page or at least on every page where switching it off can make problems, not just on two pages.

So my suggestion is to make Javascript mandatory for the backend. Lets display the warning at the login screen and make the login Javascript based.

SahAssar:
Move noscript message to a danger flash message, which makes it appear on every page and also make it more clear that JS isn't really optional. Fixes #4615, but will require updates of all translations.

### DIFF
--- a/app/resources/translations/en_GB/messages.en_GB.yml
+++ b/app/resources/translations/en_GB/messages.en_GB.yml
@@ -557,7 +557,7 @@ field:
 generic:
     noscript:
         headline: "JavaScript disabled"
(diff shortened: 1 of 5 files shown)


GawainLynch:
:shipit:

PR #4746, opened 2016-01-24
Prevent notices when hydrating from partial data, fixes #4624

### LINKED ISSUE #4624: Notices in backend, when logging in:
screen shot 2015-12-22 at 17 24 09

I get these right after logging on. `composer update`'d and I have `debug_error_level: -1` in `config.yml`.

jkazimir:
:koala:


GawainLynch:
@rossriley I might need more :coffee:, but weren't these part of a larger uncaught problem in the mapper?

…maybe I am just 2 days without :beer:

PR #4748, opened 2016-01-25
Block access to .git folders in .htaccess.

bobdenotter:
I'm sure we had this in at some point, but it got lost again.


GawainLynch:
I think we added it to the docs for Nginx

PR #4756, opened 2016-01-26
Fix date js init in repeaters

rossriley:
ping @rarila to check this is OK. The only thing I'm not sure about is the performance implication of calling app.run, since elsewhere we limit to only the updated context.


rarila:
pong @rossriley Won't just calling ``bolt.datetime.init();`` do?

rossriley:
It might do, but what other fields / widgets aren't refreshed? Is it only dates and datetimes that are not moved to the new widgets?

rarila:
Hmm, probably add ``bolt.ckeditor.init();``

PR #4758, opened 2016-01-27
[Extension] Add a trait to simplify extensions to add tables

GawainLynch:
Allows:

```php
    /**
     * {@inheritdoc}
     */
    protected function registerExtensionTables()
    {
        return [
            'my_table' => new MyTable(),
        ];
    }
```

**DISCLAIMER:** This is the shorthand version (that will still work), however tables should ideally be DI objects.

### DIFF
--- a/src/Extension/DatabaseSchemaTrait.php
+++ b/src/Extension/DatabaseSchemaTrait.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Bolt\Extension;
+
+use Bolt\Storage\Database\Schema\Table\BaseTable;
+use Pimple as Container;
+
+/**
+ * Database schema modification.
+ *
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
+trait DatabaseSchemaTrait
+{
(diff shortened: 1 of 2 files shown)


PhillippOhlandt:
That's cool. Can it also override existing table schemas to add columns? (Like for the users table)

GawainLynch:
Not specifically with the trait right now… but extending/overriding existing `Pimple`s will work fine.

i.e. Very doable, needs documentation, and agreements what we're publishing as supported for v3 :grin:

PhillippOhlandt:
Ah ok. The ability to add contenttypes itself would be cool to, if not implemented yet.

GawainLynch:
That *can* be done now, we load early enough now that a ContentType can be appended to the `Config` array… but you want to know what you're doing :grin:

PhillippOhlandt:
That's nice! I can use it at work to do as much as I can inside my Extension. That would be also good for Extensions like BoltBB, right?

bobdenotter:
Aight. :+1:

PR #4759, opened 2016-01-27
Fix select fields in templatefields (fixes #4358)

### LINKED ISSUE #4358: select fieldtype as a templatefield does not save value
You can easily re-create this in a docker.

In the 'demo' theme, edit config.yml and add a field under the 'extrafields.twig' like this:

```
templatefields:
    extrafields.twig:
        teaser:
            type: select
            values: [1,2,3]
```

Creating a page with the extrafields.twig template and choosing a value for this field of the newly created page, you can see there is no value saved.

In extrafields.twig, adding {{ dump(record.templatefields) }} confirms this.

If it lasts long enough to save anyone time: http://bolt_1x0xe5.bolt.dockerfly.com/page/test

SahAssar:
Fixes #4358

This works for me but it seems too "simple" for it to be a proper solution... :wink:

Can anyone think of any glaring issues with it?

### DIFF
--- a/src/Content.php
+++ b/src/Content.php
@@ -210,6 +210,8 @@ class Content implements \ArrayAccess
                     case 'select':
                         if (is_array($this->values[$field])) {
                             $newvalue[$field] = json_encode($this->values[$field]);
+                        } else {
+                            $newvalue[$field] = $this->values[$field];
                         }
                         break;
 


GawainLynch:
> Can anyone think of any glaring issues with it?

Not enough :cow: :bell:

rossriley:
Hmm, that block of code looks like it may have some other bugs too..
Look at the HTML block, I assume the same would happen should `general/wysiwyg/ck/allowNbsp` be set to true, then the value would be empty.

bobdenotter:
@rossriley Going to merge this one in, and i'll see if `allowNbsp` works as advertised. If not, i'll open a separate issue for that.

PR #4762, opened 2016-01-27
Fix thumbnails loaded/created from theme folder

SahAssar:


CarsonF:
For 7d61405 change. Thanks @SahAssar :+1:

Issue #4764, opened 2016-01-28
Bug in mailservice provider


geggleto:
This should go along with code styling... but

https://gist.github.com/geggleto/04f187024efe25ca315d

This is why you do not do assignments in if statements :)

https://github.com/bolt/bolt/blob/master/src/Provider/MailerServiceProvider.php#L29

GawainLynch:
> This is why you do not do assignments in if statements :)

Highly dependent IMHO :grin: 

Fixed in #4765

geggleto:
Bhahaha ;) its so easy to forget to wrap in () @GawainLynch :P better to be safe than sorry. Be a good coder and wrap your if assignments ;) tgif.

PR #4765, opened 2016-01-28
Update MailerServiceProvider.php

geggleto:
- BugFix Removed assignments from if's
- Related to #4764


bobdenotter:
Because i'm dense I didn't get this straight away, but I agree that the current coding style is confusing. :+1: from me.

Boorj:
that produced an error, actually :(

GawainLynch:
Nothing functionally changes here.

What are you trying to achieve? As I think that if you've resolved something locally, this is not to root cause.

GawainLynch:
nvm… I see, the second one isn't wrapped in `()`. This, people, is why you shouldn't review PRs before :coffee: in the morning! :grin:

Issue #4014, opened 2015-08-25
[Feature] Configurable route binding for each taxonomy


CarsonF:
_taxonomy.yml_

``` yaml
categories:
    ...
    route: categorieslink
    ...
```

_routing.yml_

``` yaml
categorieslink:
    path: /categories/{slug}
    defaults:
        _controller: Bolt\Controllers\Frontend::taxonomy
    requirements:
        slug: [Bolt\Controllers\Routing::getTaxonomyRequirement, [categories]]
```

CarsonF:
This should also be applied to `content types` as well

SahAssar:
I might very well be wrong, but testing on a 2.2.17 site this seems fixed. I used this routing:

    taxonomybinding:
        path: '/{slug}'
        defaults:
            _controller: 'Bolt\Controllers\Frontend::taxonomy'
            taxonomytype: 'categories'

So I'm going to close it, please reopen if it isn't fixed.

CarsonF:
@SahAssar You miss understand. The feature adds `taxonomybinding` or any route name to the `taxonomy type` in `taxonomy.yml`. Right now we assume the route you want is called `taxonomybinding`.

SahAssar:
Alright, so after some explaining on IRC @CarsonF set me straight. Currently bolt will always generate urls for taxonomies in the pattern '{{taxonomyslug}}/{{taxonomykey}}', we want it to be able to use custom routes as canonical urls and when linking to them...

bobdenotter:
@CarsonF Let's close this one, unless you realistically expect to be able to work on this anytime soon.

SahAssar:
Closing per the above comment. I'm guessing this will get done when there is time to do it :)

PR #4766, opened 2016-01-28
Hack slug prefix to "look" right until real fix can be applied with #4014

CarsonF:
Hack slug prefix to "look" right until real fix can be applied with #4014

### DIFF
--- a/app/view/twig/editcontent/fields/_slug.twig
+++ b/app/view/twig/editcontent/fields/_slug.twig
@@ -62,7 +62,8 @@
 {% block fieldset_controls %}
     <div class="col-sm-12">
         <div class="input-group {{ mode }}">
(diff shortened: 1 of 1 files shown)


bobdenotter:
In the spirit of "Let's get things working and ship it", i'd think we should merge this in. I've added issue #4767 as a reminder that we should fix it properly, though. 

Could you add a comment in the code that refers to this issue, so people who stumble upon it can have a clue what's going on in there?

CarsonF:
@bobdenotter Does that work?

bobdenotter:
For me it does, but Travis disagrees. Sorry, mate. :-)

bobdenotter:
****bumps @CarsonF****

CarsonF:
@bobdenotter ready

PR #4769, opened 2016-01-29
PHPDoc update

GawainLynch:


src/Storage/Repository.php
```diff
@@ -6,6 +6,7 @@
 use Bolt\Events\StorageEvent;
 use Bolt\Events\StorageEvents;
 use Bolt\Storage\Entity\Builder;
+use Bolt\Storage\Entity\Entity;
```
bobdenotter: If it didn't break without the `use`, is it really needed, then?

GawainLynch: It gets the PHPDoc in-line so that when you're extending or typehinting, it doesn't show as an error in the IDE

bobdenotter: Ah, all good then!

bobdenotter:
:+1:

PR #4771, opened 2016-01-30
Default values for fields in contenttypes

### LINKED ISSUE #4526: Ability to set default map position for geolocation
It would be lovely to have the ability to set default **zoom**, **latitude** and **longitude** for geolocation field type.
Now it's hardcoded to.. (is it Bopp's house? i think i can see him there)
https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/fields/_geolocation.twig#L58-L59
Some of my clients live not in Europe and sometimes there's no address yet (real estate agency), so it takes too long to scroll. 
I know there are other way to do that, but for routine work it will save time. I suppose there are some people who will agree.

### LINKED ISSUE #4743: Placeholder and default properties in contenttype fields
Probably i'm wrong (using v2.3a6), but when i was looking into 
https://github.com/bolt/bolt/blob/release/2.2/app/view/twig/editcontent/fields/_text.twig 
https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/fields/_text.twig
and some other fields, i found that:
* Textarea is missing **placeholder** and needs:
```twig
placeholder: context.contenttype.fields[key].placeholder|default(''),
```
* lots of fields (integer, float, text, textarea, …) are missing the usage of **default** value. Here's the fix that i use:
```twig
{# for text #}
set value = context.content.get(contentkey)|default(context.contenttype.fields[key].default)|default('')
{# for numbers: float and integer #}
set value =  1 * context.content.get(contentkey)|default(context.contenttype.fields[key].default)|default(0)
```

Boorj:
### Summary:
- partially fixes #4743
- planning to fix #4526 (default value for **geolocation**)

### Comments
I see that default values are generally unused, but i often need that. PR adds an usage of `default` property of field setup in contenttypes (not so verbose). In **date** and **datetime** it was implemented but using incorrect variables.

Also i'm planning to add `placeholder` and `default` value for **markdown** and, probably, **html** fields. Does anyone have objections?

By the way, it would be nice to mention in docs, that for date and datetime you can use "now" as default value (`"now"|date("Y-m-d")` works)

### DIFF
--- a/app/view/twig/editcontent/fields/_date.twig
+++ b/app/view/twig/editcontent/fields/_date.twig
@@ -6,7 +6,7 @@
     label:     field.label,
     info:      field.info|default(''),
     options:   field.options.datepicker|default(''),
-    default:   (context.content.id == null and field.default) ? context.content.get(contentkey)|date('Y-m-d') : null,
+    default:   (context.content.id == null and field.default) ? field.default|date('Y-m-d') : null,
     required:  field.required|default(false),
     errortext: field.error|default(''),
 } %}
--- a/app/view/twig/editcontent/fields/_datetime.twig
+++ b/app/view/twig/editcontent/fields/_datetime.twig
@@ -6,7 +6,7 @@
     label:     field.label,
     info:      field.info|default(''),
     options:   field.options.datepicker|default(''),
-    default:   (context.content.id == null and field.default) ? context.content.get(contentkey)|date('Y-m-d H:i:s') : null,
+    default:   (context.content.id == null and field.default) ? field.default|date('Y-m-d H:i:s') : null,
     required:  field.required|default(false),
     errortext: field.error|default(''),
 } %}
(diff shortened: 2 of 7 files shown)


Boorj:
@rarila and @CarsonF mentioned that it could be solved in php part, but this could be a temporary fix for that. And, anyway, date and datetime were set, but not properly.

bobdenotter:
This is your first PR, right? Anyhow, good work, @Boorj 

I'll review this one and merge in tomorrow, unless someone beats me to it! :+1:

Boorj:
not sure about
```twig
text:     format|twig({'item' : item, 'title' : item.title, 'id' : item.id}),
```
at first i made as `format|twig({'item' : item})` so it corresponds to documentation. But probably somebody has uset format with 'id' as `id`, not as `item.id`

bobdenotter:
Looks good to me! Thanks, @Boorj

PR #4773, opened 2016-01-30
Fix Handling of Bi-directional relations

rossriley:
This takes care of maintaining backwards compatibility with Bi-directional related records. For example where pages is related to entries and entries related to pages. 

The current behaviour operates on a first-come first-saved system for determining the direction of the relation and once saved there's no distinction between a page that's related to an entry, or an entry related to a page.

This PR attempts to replicate this, whilst leaving open the ability to deprecate in future when we are ready to move to uni-directional relations.


bobdenotter:
Cool stuff! :+1:

PR #4775, opened 2016-01-30
move onbeforeunload init and remover, fixes #4640

### LINKED ISSUE #4640: Viewing a preview of a record clears the unsaved record warning
If I make a new record, or edit an existing one, edit some fields, and then try to leave the page without saving I will get a warning if I try to navigate away from the page.

If I preview the record (via the preview button in the editor page), and *then* try to navigate away from the page, I will no longer get such a warning. The expected behaviour is that I would still get a warning.

Tested on version `2.2.15`.

SahAssar:
fixes #4640

### DIFF
--- a/app/src/js/modules/editcontent.js
+++ b/app/src/js/modules/editcontent.js
@@ -80,8 +80,6 @@
 
                     return false;
                 }
-                // Submitting, disable warning.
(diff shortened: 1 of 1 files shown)


bobdenotter:
Noice! :+1:

PR #4778, opened 2016-01-30
Set checkboxes with prop (fixes #4356)

### LINKED ISSUE #4356: Checkbox back-end unexpected update behavior
I found some unexpected behavior with checkboxes in the back-end for content types.
When i edit a checkbox types in my back end, if i don't return to records list, the value of the checkbox isn't update, in other words, if i keep changing the checkbox value and saving whit the default option "save" this field its not updated in the database, until i go back the records list or select the "Save & return to overview" option, this field is updated.

I don't know if this behavior is what an user expect, if he is editing multiple times a record, whit out going back to the list of records.

Thanks for all your hard work !

SahAssar:
Fixes #4356

### DIFF
--- a/app/src/js/modules/editcontent.js
+++ b/app/src/js/modules/editcontent.js
@@ -302,8 +302,14 @@
                                     // so we're catching arrays and ignoring
                                     // them, someone else can fix this!
                                 } else {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Whoop whoop

PR #4783, opened 2016-01-31
Add 'ispath()' twig function, to test for an existing path / route.

bobdenotter:


Boorj:
:bow: thanks, lightspeed coding master) :santa:

Boorj:
Library usage is a dirty workaround, `app['routes']->get($routename)` is a proper way, i suppose (since it's used in original `UrlGenerator`).



 And, maybe, 'route_exists' or 'is_route' :panda_face:?

src/Twig/Handler/RecordHandler.php
```diff
@@ -313,4 +314,26 @@ public function selectField($content, $fieldName, $startempty = false, $keyName
...
+     * @param bool   $relative
+     *
+     * @return bool
+     */
+    public function isPath($name, $parameters = array(), $relative = false)
+    {
+        $success = true;
+
+        try {
+            $path = Library::path($name, $parameters, $relative);
```
GawainLynch: This is introducing a new dependency on a deprecated class/function.

Would be much better to use the SP directly.

src/Twig/Handler/RecordHandler.php
```diff
@@ -313,4 +314,26 @@ public function selectField($content, $fieldName, $startempty = false, $keyName
...
+        } catch (\Exception $e) {
```
GawainLynch: Also, just came back and noticed this, we should catch `\Symfony\Component\Routing\Exception\RouteNotFoundException` as a general exception could point to other problems in the stack that would be silently buried.

Issue #4770, opened 2016-01-29
Numeric slugs


d-m-:
Currently numerical slugs are not allowed, although they might be useful for some
custom contenttypes (say, yearly events, with URL "/event/2016"). 
Is there any reason not to allow them? 
I did not encounter any problems after commenting out 
https://github.com/bolt/bolt/blob/master/src/Legacy/Storage.php#L2836-L2839.

SahAssar:
@d-m- IIRC they are not allowed since they conflict with the id's that can also be used to access content.

d-m-:
The frontend router first tries to fetch by slug and falls back to IDs, 
{{ record.link }} uses slugs, and the editlinks use only IDs, as far as I know...
So the only conflict I could think of would be if someone uses a theme with explicit
ID-links. 

Would a PR allowing numeric slugs if an option is set in contenttypes.yml have any chance of getting merged?

C4Grey:
I solved this problem with a custom controller(I needed this feature to make a news archive available by dates presented in url). You can make your own route like "archive" (or override a route for your entity) and process those URLs in controller assigned to this route.

SahAssar:
@d-m- Perhaps conflict is the wrong word, but it would things slightly less predictable. Any record can be accessed by both it's slug and it's ID, so a user might make a record with the slug 2016 and later expect to be able to access a record with the ID 2016 using the ID, which would then return the record with the slug 2016.

Whether a change would be accepted I don't really know (I don't really make those calls)... If it was changed I'd prefer it to be an option rather than the default, since it might screw with some existing sites otherwise. It has been discussed previously in #4327 and #3001 though, so clearly some people want this to be an option.

d-m-:
@C4Grey Thanks, this is basically what I'm doing right now.

However, I think that writing a controller for this is more complicated than necessary, especially since this probably is the expected behavior.
I guess most users don't use ID routes and and it should be fairly simple for developers building a site to prevent conflicts.

I opened a small PR that makes this an option.

padawan:
I concur with @d-m- and our use case is a conference website that already uses /YYYY (e.g /2016) as the entry for the event year. Having to write a controller for this is indeed more complicated than necessary.

If you allow this instead and clarify in the docs the potential conflict that may exist with ID (in edge cases, not generally, if I'm not mistaken), then it's best of both worlds for everybody, isn't-it?

GawainLynch:
Done in #4784

CarsonF:
[...]

Issue #3001, opened 2015-03-03
Ability to make slugs not add prefixes for numbers


Pinpickle:
If have a record (say, of contenttype page) with a slug that is generated from a number, it will give you a slug: `page-<x>`. To me this is unexpected behaviour, especially if you are manually entering the slug by clicking on the pencil and editing it.

While I would like the default to be to add no prefix at all, I understand that this feature is here for a reason. Could there be an option for slug fields to not add a prefix?

bobdenotter:
Bolt really likes for slugs to be unique. I'm having a hard time understanding why you'd want this. :-)

GawainLynch:
To add to what @bobdenotter said. This would, in a few ways, be a backwards compatibility break. 

We currently assume that a record is available by it's `contenttype/id` as well as its `contenttype/slug`.  Something that Bolt avoids by doing this (that drove me nuts — pun intended — in other systems was the `/index.php?1234` junk)

The id/slug consistency is also used in our query processing and the like… So, to say the least, this would be a *very* big job to implement :crying_cat_face:

Pinpickle:
Ah fooey. The use case was a house contenttype that had a lot field which I wanted to be its slug - so you'd get:
`/house/<lotno>`

At the very least could we gain the ability to customise what that prefix is? Because `house/house-<x>` looks silly in my opinion. I would be happy with a method to change it to `house/lot-<x>`

GawainLynch:
You should be able to do `contenttype/prefix-N` now. Bolt will, by default, create the initial slug out of what you tell it to in `contenttypes.yml`.  Our default is `uses: title`, but you should be able to "unlock" and "edit" it. 

This did temporarily break at some point during 2.0, so if it's not working for you right now can you update to 2.0.6 (even better if you could quickly check on 2.1-beta2) and open an issue if the problem persists… We'll send the Bolt Medical Team rushing in :ambulance:

Pinpickle:
I know myself (or a client) could manually change it to `lot-<x>` but it's not super intuitive if you don't have any idea of what a slug is, or how urls work, or if you even care what your urls look like. I'll roll with this for now, and can see the feature as being a low-priority but here's how I see it working:

```
type: slug
uses: lot
number_prefix: lot
```

I suppose a hackish way to do this would be to add a select field with a default value, and only one value, and that being the prefix you want, and then use that field as well as the number field to generate the slug. But that's quite a hack.

GawainLynch:
[...]

PR #4784, opened 2016-01-31
option to allow numeric slugs

d-m-:
Introduces an option `allow_numeric_slugs` in contenttypes.yml. 
See issues #3001 #4770 .

Docs (Contenttypes and records > Defining contenttypes > available options):

* `allow_numeric_slugs`: By default, Bolt prefixes slugs that would be purely numeric with the contenttype slug in order to prevent conflicts with routes of form /{contenttypeslug}/{id}. If this option is set to `true`, numeric slugs remain unprefixed. Care has to be taken not to use routes with IDs in the frontend.

Or something like this.

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -1315,7 +1315,10 @@ class Storage
             // like 'entry/12' or '/page/12345'
             $decoded['contenttypes'] = $this->decodeContentTypesFromText($match[1]);
             $decoded['return_single'] = true;
-            $ctypeParameters['id'] = $match[2];
+            // if allow_numeric_slug option is set on contenttype, interpret number as slug instead of id
+            $contenttype = $this->getContentType($decoded['contenttypes'][0]);
+            $field = ( $contenttype['allow_numeric_slugs'] === true ? 'slug' : 'id' );
+            $ctypeParameters[$field] = $match[2];
         } elseif (preg_match('#^/?([a-z0-9_(\),-]+)/search(/([0-9]+))?$#i', $textquery, $match)) {
             // like 'page/search or '(entry,page)/search'
             $decoded['contenttypes'] = $this->decodeContentTypesFromText($match[1]);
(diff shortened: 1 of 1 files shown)


bobdenotter:
I'd (personally) be hesitant to use this, but to each his own. If you use `/entries/3`, will the 'slug' or the 'id' get precedence? 

Anyhow, could you add a small explanation of this setting to the docs, at bolt/docs, please?

src/Legacy/Storage.php
```diff
@@ -2795,8 +2795,8 @@ public function getUri($title, $id = 0, $contenttypeslug = "", $fulluri = true,
         $id = intval($id);
         $slug = $this->app['slugify']->slugify($title);
 
-        // Don't allow strictly numeric slugs.
-        if (is_numeric($slug)) {
+        // Don't allow strictly numeric slugs, unless allow_numeric_slugs options is set
+        if (is_numeric($slug) && $contenttype['allow_numeric_slugs'] != 'true') {
```
GawainLynch: Can you please not use loose comparison, and `true` should not be a string but a boolean, e.g.
```php
if (is_numeric($slug) && $contenttype['allow_numeric_slugs'] !== true) {
```

d-m-:
@GawainLynch You are totally right, i'll correct that, and add the explanation to the docs.
In particular, concerning your first comment, I'll add that the frontend router first tries slugs (which is the reason why this actually works and does not cause trouble unless a theme designer has chosen not to use record.link).

d-m-:
@bobdenotter I am sorry that I forgot about the shorthand notation for setcontent queries, which always interprets numbers as ids. I added a check for the allow_numeric_slugs option to overwrite this behaviour. 

Basically I think that this is consistent: If one wants to use numeric slugs, it does not make sense to query 'contenttype/id', since the shorthands mimic the standard frontend routes.

d-m-:
added docs: https://github.com/bolt/docs/pull/302

bobdenotter:
@d-m-: Thank you for providing docs for this change! :+1:

CarsonF:
Dang I wish I remembered to comment on this. This change was only applied to the *deprecated* code, hence the `Legacy` namespace. It needs to be written for the new storage code as well. And tests should be written as well. Else we are just providing ross with more work. Either way this pushes us farther away from release.

PR #4785, opened 2016-02-01
Fixing not working buic browser.

Boorj:
This error didn't allow to **select from server** image, when clicking.


GawainLynch:
Darn it, @rarila… who was feeding the :koala: again!

rarila:
:leaves:

PR #4786, opened 2016-02-01
buic browser styling : making file info easier to read

Boorj:
Please, zoom in this picture:

* removed repeating *px*
* different style for size and dimensions (easier to find needed info by eyes)


rarila:
Avoid inline styles, please.

Boorj:
thought about some other styles - like bold for filesize, or vice versa... but the suggested one is the lightest.

denied options:

Boorj:
ok, no problem

rarila:
Best add styling to [browser.scss](https://github.com/bolt/bolt/blob/master/app/src/sass/buic/_browser.scss) and do it like
```html
<div class="filesize">…<div><div class="dimension>…<div>
```
Ii would say there's no semantic value in ``<small>`` and ``<br>``

bobdenotter:
It really should look the same as the ones in the "regular" files screen: 

screen_shot_2016-02-01_at_15_43_28

Boorj:
maybe i can make regular browser look like that async browser?

CarsonF:
@Boorj Wow please no. There's no need to parse the dimensions. It was supposed to be a shortcut. 

You can use `{{ file.info.width }}` and `{{ file.info.height }}` instead.

CarsonF:
Also `{{ file.sizeFormatted }}` should be used as well, which gives: `2.2 MiB`

Boorj:
but anyway, could it be as temporary make-up? 
this one really looks better!

bobdenotter:
No, sorry mate. They should look exactly the same. Icon size, filesize, and image size. I'm open to finetuning how they exactly look, but they *must* look the same.

Boorj:
Ok, Bopp, gotcha, but now it looks the same as i did, but with same text styles and with repeating px.
I think i will use `.syntax` class to make all syntax (as px, × and other) look brighter than surrounding text. 
Could you please give me the full list of parts that should be restyled; from now i see only file browser and async browser. Somewhere else?

bobdenotter:
Yes, i believe those two locations are the only ones.

bobdenotter:
Reopening this one. I think it should be fixed, and made to look the same. :-)

rarila:
> The large icons there have to go

@bobdenotter Au contraire, the other way round. I made them bigger for a reason. No need for icons when you can't recognize what they stand for because they are too small. And there's really enough space for icons you don't need a magnifier for. :eyeglasses:

bobdenotter:
Sorry, rarila, but I disagree. The icons there are currentyl a huge visual distraction. They aren't meant to be so you can have a close-up look at what's going on in there, but merely as a 'visual anchor' in the row of similarly placed icons. 

The way they are now, they are too big and distracting and clunky. :-)

CarsonF:
Both pages should definitely be consistent. I'm not really sure why this twig code is duplicated, it should be from the same abstracted twig file.

And I think I agree with @bobdenotter.

Boorj:
Could you please tell me the path, where this _abstracted twig file_ should be? So i could put that templates in that place. Or you can merge PR and process unification by yourself.

I agree with @rarila that bigger icons are ok - we can make them brighter, but they do have a rhythm, so it's not distracting so much. But they definetely should have a sense. If you make different icons for png and jpgs - that could be a sense. _My clients sent me several photos to upload, i made that, and those were some png photos for 1mb each. I found that later, and had to convert that and re-attach. We could avoid that, having icons, showing different file types.

Boorj:
and, btw, i disagree about "KiB" and "MiB".
1. Those [KiBiBytes](https://en.wikipedia.org/wiki/Kibibyte) and MeBiBytes could be replaced with KB and MB
2. They could be done with semi-transparent fonts to focus attention on number and be non-distrative
AND 
3. if you do change bytes to KB, please, put also a `title='…'` attribute with size in _Bytes_.
4. Also we could put upload date to the `title`

Boorj:
Is it ok?

Boorj:
What is the concideration? What should i do?

GawainLynch:
@CarsonF @bobdenotter @rarila ?

bobdenotter:
I think it's OK, but i'd like to take a closer look. Will do so later, and merge in! I'll self-assign.

GawainLynch:
OK… but can we make the "MiB/KiB" optional if you're going to default to MB/KB… the latter is flat out wrong and would get in my way, personally speaking.

rarila:
MB/KB is ok as long as it is calculated on a decimal base

GawainLynch:
> MB/KB is ok as long as it is calculated on a decimal base

But that isn't the problem for my workflow. 

If I am looking at something that is 2654 bytes, that is 2.654 KB and 2.592 KiB… so I will be seeing different numbers in the UI compared to what I will be seeing when I query file system forcing me to manually, or mentally, calculate the difference.

bobdenotter:
Gawainlynch: Would it sit better with you, if we made it `~ 2.2 mb`, and then a tooltip that shows `2.253.425 bytes`?

GawainLynch:
I think the tilda would confuse as many people as MiB would… and I am totally on board with making things more friendly.

The pain for someone like me who is looking at things for very different reasons to our editor, or potentially frontender, target group(s) is that I am looking at things forensically to try and see why the previous two groups of colleagues are not seeing "the right thing"… they only need the decimal suffix, I am specifically interested in the IEC suffix.

So maybe the workable solution is just the tool tip… the target group I belong to could work with that .

bobdenotter:
Sounds good to me! :-)

bobdenotter:
Ok, going to merge this in, and work on some additional fine-tuning and consistency in a separate PR. I'll keep everyone's feedback in mind. :-)

Boorj:
Bopp, maybe we'd rather use dotted underline, than dashed? Ususally dotted - is a standard for tooltip presence indication, like in <abbr> html tag. No?

bobdenotter:
@Boorj I used the same as the tooltip for the 'moments' things, like 'A month ago'. In my opinion it makes sense that they're the same.

Boorj:
Usually dashed line is assumed to be an interactive control, that opens a popup. That's why, probably, moment.js is using that. But... http://html5doctor.com/the-abbr-element/

CarsonF:
Oy! Why was this merged? I specifically said to not do the dimensions like that.

Boorj:
i agree that dimensions could be lighter than the number... so it's easier to read number information

app/view/twig/async/browse.twig
```diff
@@ -106,9 +106,9 @@
                                 </td>
 
                                 <td class="info">
-                                    {{ file.size }}
+                                    <div class="filesize">{{ file.size|e|number_format(0, '.', '<small>&thinsp;</small>')|raw }}</div>
                                     {% if file.image %}
-                                        <br> {{ file.dimensions }}
+                                        <div class="image-dimensions">{{ file.dimensions|e|replace({'px' : '', ' × ' : '<span class="times">×</span>'})|trim|raw }}</div>
```
CarsonF: I meant change this to:
```twig
<div class="image-dimensions">{{ file.width }}<span class="times">×</span>{{ file.height }}</div>
```

Boorj:
And, by the way, we can swap date and dimensions:

bobdenotter:
> I specifically said to not do the dimensions like that.

@CarsonF Tons of people have said things in this thread, hard to keep up with the plethora of personal preferences. What's wrong with that we have?

> Usually dashed line is assumed to be an interactive control, that opens a popup.

It looks and behaves consistent with the moments.js popups we've had for ages. If you'd like to propose something else, and put in the work, feel free to make an RFC for it. 

> And, by the way, we can swap date and dimensions:

Yes, we can. We can also leave it like it is. <- that's me wearing the BDFL hat.

CarsonF:
Because you don't need to pull the dimensions apart by doing a string replace, we already have them separated. It's hacky and unnecessary.

PR #4788, opened 2016-02-02
Check app.routes for preview route

CarsonF:
There's no need to go through the process of generating a url for a route just to see if the route exists. Much less do we need yet another twig function.


GawainLynch:
:+1: from me
:+1: from @bobdenotter 
:+1: from :koala:

PR #4791, opened 2016-02-02
Update MenuBuilder.php

### LINKED ISSUE #4790: Menu function performs unnecessary hydration
I've noticed that with time every page on my site become extremly slow, so I spent some time with a profiler and found next:
image
All my content types somehow are fetching on every page, it is always a few items, but it takes a huge amount of time, especially for the "programs" content type. So I've looked into template parts that used on every page...and found that when I removed the "menu" function call, my pages began to render in 200-300 ms! Indeed, all those huge blocks on the timeline specifies IDs of programs presented as links in a main menu. [This line of code](https://github.com/bolt/bolt/blob/aa99c8e914bdfc9f09e6f41eb049b9d473c4b843/src/Storage.php#L2566) takes all that time.
P.S. I also think that #4672 is related to this issue.

C4Grey:
Remove unnecessary "hydration" for a menu item

Fixes #4790

### DIFF
--- a/src/Helpers/MenuBuilder.php
+++ b/src/Helpers/MenuBuilder.php
@@ -174,7 +174,7 @@ class MenuBuilder
     private function populateItemFromRecord(array $item, $path)
     {
         /** @var \Bolt\Content $content */
-        $content = $this->app['storage']->getContent($path);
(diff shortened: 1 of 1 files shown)


GawainLynch:
LGTM

bobdenotter:
To me too. :+1:

PR #4794, opened 2016-02-02
Fix canonical link in source to prevent duplicate content.

### LINKED ISSUE #4793: "index.php" shows up in canonical, while it shouldn't.
Silex allows getting URLs with `index.php` in them, in the routing. This is fine, but the *canonical* shouldn't show it: 

screenshot_2016-02-02_18_53_17

bobdenotter:
Fixes #4793


rarila:
… or just a simple all-in-one regex?

```php
$link = preg_replace('%^/?index\.php/%', '', $content->link());
```

CarsonF:
Sorry, but this isn't the problem we can or should fix. This is webserver misconfiguration. There's no way we know that path would work without the `/index.php/` part.

I offered my own apache config on #4793.

bobdenotter:
No, it is not.

GawainLynch:
I agree with @CarsonF

bobdenotter:
This is an issue that needs fixing in the code. We _might_ be able to mitigate it in the htaccess or webserver config, but that's not the point. The point is that we have duplicated content, and it needs to show the proper URL for that in the canonical. That is the only thing this PR fixes, and it fixes it correctly.

GawainLynch:
OK then, lets build BoltFirewall while we're at it… :stuck_out_tongue_winking_eye: 

We have been trying to move away from web server specific code in core since 2.0-dev and things like this don't belong at the code layer.

We just introduce **less* flexibility, more internal dependencies and things that **we** are responsible for, not the server administrator. Sure, I'll grant it is unlikely, and pretty stupid, to want to have URLs with `index.php` in them… but that shouldn't be **our** call to make.

bobdenotter:
You and @CarsonF are both not getting the point. This has nothing to do with server configuration, but with *reducing duplicate content*. Let's turn it around, example time: 

Let's say you have an entry 'Lorum ipsum', with id 1. In config.yml you've set `canonical: https://example.org`. What should the `<meta canonical>` tag say for the following URLs, regardless of server configuration: 

 - http://www.example.org/entry/lorum-ipsum
 - http://example.org/entry/lorum-ipsum
 - https://www.example.org/entry/lorum-ipsum
 - https://example.org/entry/lorum-ipsum
 - http://www.example.org/entry/1
 - http://www.example.org/index.php/entry/lorum-ipsum

I'm confident it should be `https://example.org/entry/lorum-ipsum` for all six of these.

GawainLynch:
Then, unless I'm missing something **really* obvious in this PR… specifically, how does:
```php
            if (strpos($link, 'index.php/') === 0) {
                $link = str_replace('index.php/', '', $link);
            }
```
Turn `http://www.example.org/entry/lorum-ipsum` into `http://example.org/entry/lorum-ipsum`

The above code does not do that, it simply tests for the string `index.php/` as position zero, and if present does a replace.

That aside… Your assertion is otherwise correct, and I agree wholeheartedly, but not covered in the current PR code.

bobdenotter:
> The above code does not do that, it simply tests for the string index.php/ as position zero, and if present does a replace.

The current code is already behaving as expected for all the examples i've listed, except for the one with `index.php` in it. This PR brings that last case in line with the others.

bobdenotter:
@CarsonF How are those changes you were working on coming along? I'd like to get this one sorted out.

bobdenotter:
Fixed by #4941. Thanks, @CarsonF, @GawainLynch!

See: 

screenshot_2016-03-02_21_22_59

PR #4806, opened 2016-02-05
Fix extension's relative urls

### LINKED ISSUE #4802: Incorrect directory separator in extension assets path
There's problem with directory separators.

getAssetPath function returns path, taken from extension's json, and it has backslashes in its path string. So it returns:

(example of how it **is now** and how it **has to be**).

So [here in AssetTrait](https://github.com/bolt/bolt/blob/master/src/Extension/AssetTrait.php#L255) we can use
```php
return str_replace('\\', '/', $this->getRelativeUrl()) . $fileName;
//instead of 
return $this->getRelativeUrl() . $fileName;
```
which fixes the problem. But also it could be fixed in other places, depending on what is the purpose of 'path' usage.
Probably
- [here in Extension\Manager](https://github.com/bolt/bolt/blob/master/src/Extension/Manager.php#L216) 
- or [here in PackageDescriptor](https://github.com/bolt/bolt/blob/master/src/Composer/EventListener/PackageDescriptor.php#L128)
- or somewhere [here in Asset Injector](https://github.com/bolt/bolt/blob/master/src/Asset/Injector.php#L220)
- or even directly [in Asset/File/JavaScript.php](https://github.com/bolt/bolt/blob/master/src/Asset/File/JavaScript.php#L21)

I could make a PR (because it fixes my problem), but don't know where to put a fix properly.

CarsonF:
Fixes #4802

### DIFF
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -242,7 +242,7 @@ class Manager
         $extension = new $className();
         if ($extension instanceof ExtensionInterface) {
             $baseDir = $this->filesystem->getDir($descriptor->getPath());
-            $relativeUrl = sprintf('/extensions/%s/web/', $descriptor->getPath());
+            $relativeUrl = sprintf('/extensions/%s/web/', str_replace('\\', '/', $descriptor->getPath()));
             $this->add($extension, $baseDir, $relativeUrl, $descriptor->getName())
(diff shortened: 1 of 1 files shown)


GawainLynch:
Duuuuuuuuuude! :frowning:

PR #4811, opened 2016-02-06
Improve process for adding custom fields to new storage layer

rossriley:
This is mainly some helper syntax to make it  easier to add custom fields.

Bolt 3 needs to keep both the new and the old FieldManager in operation since not everything has been ported over to the new system.

So this PR adds access to the old manager to the new one, so that a call to `$app['storage.field_manager']->addFieldType('mycustomfield', $myCustomFieldObject)` will take care of adding to both the new and the old one.

One thing to note, that shouldn't have any impact on BC, is that this PR also adds a hydrate pass of new custom fields to the legacy content hydration. So if you add a new-style field that does something fancy in the hydration step then it will show up in old-style `Content` objects.

### DIFF
--- a/src/Provider/StorageServiceProvider.php
+++ b/src/Provider/StorageServiceProvider.php
@@ -66,7 +66,7 @@ class StorageServiceProvider implements ServiceProviderInterface
 
         $app['storage.field_manager'] = $app->share(
             function ($app) {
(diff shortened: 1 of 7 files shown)


rossriley:
Looks like I spoke too soon.....

GawainLynch:
:+1:

PR #4750, opened 2016-01-25
Fixes "non-interactive configuration" of composer-install by providing extra vars in composer.json

skript-cc:
So here goes my first pull request, please help me out if I'm doing something wrong.

I posted the issue this weekend in the bolt/composer-install repository, see https://github.com/bolt/composer-install/issues/7. I also described the solution to this problem put forward by this PR.

**Quick summary**
*Problem*: "I want to install Bolt in separate directory, but I can't do this when doing a composer install non-interactively."
*Solution*: "Do a composer install instead with a slightly modified version of the composer.json file, where an extra option named 'bolt-separate-web-dir' can be set."

This PR also modifies the default values so that by default the web root is installed as a separate directory. This way the code reflects the recommendations from [the documentation](https://docs.bolt.cm/installation-advanced), which states:

> As of version 2.0 the location of app resources is completely configurable so you only need to store public assets inside the web root directory. We would strongly recommend that you use this strategy if you have control over your hosting environment.

PR #4812, opened 2016-02-07
Enable non-interactive configuration of composer-install

GawainLynch:
Forward port of #4750 with credit to @skript-cc


bobdenotter:
Whoop! :+1:

Issue #4816, opened 2016-02-08
[RFC] custom formatting of listing item


Boorj:
##### Issue
Sometimes i need to customize the way, how my list looks like. Maybe it's interesting to have an ability to setup, how bolt formats each row in listing? Now it looks like heavy usage of `title` field and some other random fields. We can add a new property, that defines, how will Bolt format this string.

##### Probably it could look like:
```yml
persons:
    singular_name: Person
    editcontent_listing_format: 'item.last_name|default('Pleast set the last name')'
    fields: 
        …
```
we can use another property name

##### additionally
There are different types of listings. Should we allow user to customize them all?

Boorj:
I have a gallery, where each item has some properties, but no title. It looks this way:
![2016-02-08 13-08-14](https://cloud.githubusercontent.com/assets/2145313/12879962/fb0109da-ce64-11e5-93a7-eaa885045ac0.png)

By the way, **the second issue is:** listing string is escaped somewhere in *.twigs.

I can make this part of code by my own, only need opinion of the community.

bobdenotter:
This is already possible. Take a look at this for an example: 

https://github.com/bolt/bolt/blob/release/2.2/app/view/twig/custom/listing/pages.twig.dist

The second issue (the escaping), we will fix.

Boorj:
Can it be done for exact entity in `contenttypes.yml`?

Boorj:
and, by the way, does calculation of excerpt length concern about extra formatting symbols ?

bobdenotter:
> Can it be done for exact entity in contenttypes.yml? The idea was to make it simply editable, as format: in relations.

Yes. 

> and, by the way, does calculation of excerpt length concern about extra formatting symbols ?

No, by design. Your layout should really, really not depend on something being the exact guesstimated length.

GawainLynch:
Fixed in #4818

Boorj:
> No, by design. Your layout should really, really not depend on something being the exact guesstimated length. 
There are some html tags which are calculated as symbols here
https://github.com/Boorj/bolt/blob/master/app/view/twig/_base/_listing.twig#L165
```twig
{{ content.excerpt(excerptlength - title|length) }}
```
Maybe we should strip tags at first?

Boorj:
And @bobdenotter , how can I customize listings directly from `contenttypes.yml` ? not with this magic:

[...]

PR #4818, opened 2016-02-08
Escaping 'title' in listing views.

bobdenotter:


bobdenotter:
Fixes the second part of #4816.

PR #4823, opened 2016-02-09
[2.2] fix listing_records in taxonomies

### LINKED ISSUE #4822: Impossible to set a page size for a taxonomy listing
Although we can specify a count of entries per page via the "listing_records" parameter, it is impossible to set a different amount of entries in a taxonomy listing template(or, what is necessary sometimes, to set a page size per one taxonomy entry). I don't see a way to make it simply via extension, since a few core features should be re-written...Correct me, if I wrong, please.

SahAssar:
I'm guessing we need to set the amount before the query :wink: Fixes #4822

### DIFF
--- a/src/Controllers/Frontend.php
+++ b/src/Controllers/Frontend.php
@@ -299,12 +299,13 @@ class Frontend
         $page = $query->get($pagerid, $query->get('page', 1));
         $amount = $app['config']->get('general/listing_records');
         $order = $app['config']->get('general/listing_sort');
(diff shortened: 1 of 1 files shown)


SahAssar:
This is already fixed in master, so no more PR incoming.

GawainLynch:
We can literally blame the :koala: again… c.f. #4691

SahAssar:
That darn little :smiling_imp: :koala:

Issue #4274, opened 2015-10-14
Context-sensitive excerpts and search term highlight for search results


Eiskis:
One of the most requested features from my end-users and content editors is context-sensitive preview of page content in search results, with the actual hits highlighted.

I believe in its simplest form, this could be implemented as another `excerpt` function that would build an array or concatenated string of the hits for the search term (passed as input). You'd use this on whatever content you'd display of your records on the search result page.

The actual highlighting is trivial to do with JS, but it can only use the content it has been passed or is already in the DOM. Right now that's what I'm doing, but the end result is slighly confusing since the search term might be included on the page a hundred times at the end without being visible in the excerpt I'm showing to the user now.

Now of course the search functionality uses tags and other pieces of content as well for the hits, so in practice you might have to take more things into account. We could also think of an entirely different way to implement this, for example the search result nodes including some kind of context information in addition to the relevance ranking we have now.

I might look into building an extension myself if I have the time but would love to hear some thoughts, I doubt I'm the first one working on this.

sbonardt:
For the highlighting part.... I ran into this one as well. Look at below code. If you put the search variable containing your entered term(s) in brackets you can use it in a twig replace

'<p>{{ record.excerpt(250)|replace({(search): "<strong>" ~ search|escape ~ "</strong>" })|raw }}</p>'

ps: with thanks to @bobdenotter

SahAssar:
Moved to the forums since this in it's current form isn't very actionable or even sure if it should be in core or in an extension. See here: https://discuss.bolt.cm/d/139-context-sensitive-excerpts-and-search-term-highlight-continuation-of-4274

SahAssar:
Fix for this on the way in #4825 courtesy of @bobdenotter!

PR #4825, opened 2016-02-09
Making better excerpts.

bobdenotter:
This PR makes better excerpts of entries / arrays / strings: 

 - Less duplicated code
 - Centers content around a search term, so the excerpt is more relevant
 - Highlights search terms


bobdenotter:
Screenshots: 

screenshot_2016-02-09_20_33_36

screenshot_2016-02-09_20_35_47

GawainLynch:
So there is a new Bob-charge of 4 Belgian beers per-`empty()` call… I am not going to be able to walk for weeks!

bobdenotter:
@GawainLynch As far as i'm aware, i'm only using `empty` in cases where i'm _actually_ checking for empty strings. Like, empty titles, for example. In these cases using `empty` should be OK, right?

GawainLynch:
No… `empty()` is an PHP "easy mode" that really should dodo… 

From the [PHP manual](http://php.net/manual/en/function.empty.php)… The following things are considered to be empty:
* "" (an empty string)
* 0 (0 as an integer)
* 0.0 (0 as a float)
* "0" (0 as a string)
* NULL
* FALSE
* array() (an empty array)
* $var; (a variable declared, but without a value)

If you want to check for an empty string, you should check for an empty string… if the variable is null, you know (because you haven't sinned with `public $foo = 'bar';` in the first place that the variable hasn't been set… ever.

If it should have been, when you `dump()` or debug, you can immediately see that there is a missing value and it gives a more indicative hint to what the programmer has missed.

bobdenotter:
As @SahAssar pointed out. This PR fixes #4274.

bobdenotter:
@GawainLynch Ok, so now there's one `empty()` left. 

```
        if (empty($focus)) {
            $excerpt = Html::trimText(strip_tags($excerpt), $length);
        } else {
            $excerpt = $this->extractRelevant($focus, strip_tags($excerpt), $length);
        }
```

In this case i'd want to check if it's either an empty array, or an empty string. replacing it with `(is_array($focus) && count($focus) === 0) || (is_string($focus) && $focus === '')` seems messier. What should I do with that, then?

GawainLynch:
> In this case i'd want to check if it's either an empty array, or an empty string, or null. replacing it with (is_array($focus) && count($focus) === 0) || (is_string($focus) && $focus === '') seems messier. What should I do with that, then?

That one is a valid use :+1:

GawainLynch:
Rule #1: Don't let :koala:s near Twig!

PR #4826, opened 2016-02-10
Use a relative path to src/Composer/EventListener

GawainLynch:


bobdenotter:
Whoohoo, green lights!

PR #4827, opened 2016-02-11
Config caching

GawainLynch:
Backstory… the PHP file that config is cached to is set to 644 and can break nut in stricter permission installs.

This also removed the last vestiges of the load/save serialisation functions in core… :fire:


bobdenotter:
Probably one of the very last pieces of ancient PivotX code to go the way of the dodo. :bird:

GawainLynch:
> Probably one of the very last pieces of ancient PivotX code to go the way of the dodo. :bird:

There is more… the :koala: is hungry too… :stuck_out_tongue_winking_eye:

bobdenotter:
Whoop!  :bomb:  :boom:

PR #4828, opened 2016-02-11
Lock cocur/slugify to 1.4

GawainLynch:
Slugify changed the function signature of Slugify::create() in 1.4.1 and we can't (yet) know if they intend to revert in 1.4.2 so we'll lock on a known-good version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "bolt/pathogen" : "~0.6",
         "bolt/thumbs" : "~2.0",
         "brandonwamboldt/utilphp" : "~1.0",
-        "cocur/slugify" : "~1.0",
+        "cocur/slugify" : "1.4",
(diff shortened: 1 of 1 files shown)


GawainLynch:
For the benefit of keeping history… The Slugify change was merged in November last year into their 2.0 branch… then the 2.0 branch was merged into master soon after… I think they've screwed up their tagging.

bobdenotter:
Seems like a reasonable assumption, because they now have `[]` instead `array()` all of a sudden. :-)

GawainLynch:
Yeah, I have logged an issue with them as it looks like an honest mistake that they'll no doubt want to fix :grin: 

https://github.com/cocur/slugify/issues/94

GawainLynch:
Upstream have rectified the problem… Nothing to see here :grin:

PR #4832, opened 2016-02-13
DB Query Optimisations

rossriley:
This PR adds some per-request database query caching which should reduce unnecessary multiple queries for the same item.

To do this it uses the built-in functionality of Doctrine DBAL, which is triggered by passing in a `QueryCacheProfile` instance to the standard `executeQuery` method.

I've also built in the future potential to allow user-level opt-in to the cache (via Bolt's standard `$app['cache']`) although this is for now a secret feature because it warrants some more testing at production scale.


GawainLynch:
Awesome mate, thank you!

bobdenotter:
Good stuff! :-)

PR #4835, opened 2016-02-13
Fixes for multiple filesystem namespaces

### LINKED ISSUE #3831: Potential BC/Functionality Break in Uploads
Pinging @GawainLynch and @CarsonF 

In the move around of the controllers I just noticed that the previous services for both `$app['upload']` and `$app['upload.container']` have been moved inside a Pimple `$app->share()` declaration.

see: https://github.com/bolt/bolt/blob/release/2.2/src/Controllers/Upload.php#L25
vs: https://github.com/bolt/bolt/blob/master/src/Provider/UploadServiceProvider.php#L21

I've not got a test case handy, but I'm pretty sure that this will disallow a lot of runtime modification, for example (when we get around to making it) modification of the upload prefix based on contenttype setting... `upload: uploads/%contentype%`` for instance.

rossriley:
Fixes #3831 

This tidies up a couple of bugs that were causing breakages on uploads to multiple locations.

### DIFF
--- a/src/Controller/Backend/Upload.php
+++ b/src/Controller/Backend/Upload.php
@@ -185,13 +185,13 @@ class Upload extends BackendBase
         $this->app['upload.prefix'] = $prefix;
 
         // Do the upload
-        $result = $this->handleUploadFiles($request, $namespace);
+        $fullResult = $this->handleUploadFiles($request, $namespace);
 
         array_shift($handler);
         $original = $namespace;
 
-        if (count($result)) {
-            $result = $result[0];
+        if (count($fullResult)) {
+            $result = $fullResult[0];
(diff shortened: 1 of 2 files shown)


GawainLynch:
Thanks so much mate!

One note, I added a75a512 as the logic check was a bit misleading and prone to give us problems in the longer term… 

One problem is that `$this->filesystem->createDir($path);` *now* returns a `void`, so doing a logic check against it is meaningless.

CarsonF:
Also visibility has nothing to do with whether one can upload or not

GawainLynch:
How? if a user can't write to the directory… they can't upload.

CarsonF:
Visibility doesn't not define whether a user can write to a directory

Issue #4291, opened 2015-10-19
Uploading files too large gives incorrect error


svivian:
I'm trying to upload PowerPoint files (some are `.ppt`, some are `.pptx`). The config option `accept_file_types` includes both of those (it's just the default one, I haven't touched it).

But when I upload I get the error (emphasis mine)

> File does not have an acceptable extension (TWIG, HTML, JS, CSS, SCSS, GIF, JPG, JPEG, PNG, ICO, ZIP, TGZ, TXT, MD, DOC, DOCX, PDF, EPUB, XLS, XLSX, **PPT, PPTX**, MP3, OGG, WAV, M4A, MP4, M4V, OGV, WMV, AVI, WEBM, SVG)

Can't see any reason why they wouldn't be valid PPT files. Other files upload fine (haven't tried them all but `jpg`, `png`, `pdf`, `doc` and `docx` all upload fine).

Using latest Bolt, 2.2.13.

**UPDATE:** the error was occurring due to `upload_max_filesize` being too small, so there is no problem with file types but the wrong error message is displayed.

GawainLynch:
You haven't [installed this](http://extensions.bolt.cm/view/ed895319-603e-11e4-866c-43d66b93a270) by any chance have you?

SahAssar:
@GawainLynch We all know that Clippy loves Microsoft Office files, so it can't be him. Also, clippy never hurt anyone, don't accuse him of stuff that the koala clearly did!

GawainLynch:
@svivian Seriously though, I just tried here on `release/2.2` and `master` and was able to upload on both…

SahAssar:
@svivian Have you checked that it's not just the upload size setting? Could you try a small ppt or pptx?

svivian:
Thanks, looks like that was it. I increased `upload_max_filesize` in php.ini and the files can now be uploaded.

So the bug is really that the error message is incorrect. It says the file type is not allowed, when it should say the file is too big.

Pinpickle:
Weird... this was supposed to be fixed by #3123

A regression, possibly?

rossriley:
I'm closing this for now, as part of #4838 I've tested the upload UI for reporting back potential for the upload to fail in PHP without being alerted in JS before the upload. I've found one edge case where the eventual POST size was a little larger than what the reported post max size was.

In all other cases the UI seemed to alert that a file was too large. So if anyone still finds an issue with this we'll need some more specifics about how to reproduce.

[...]

PR #4838, opened 2016-02-14
Fixing reporting for upload errors

rossriley:
This came out of testing for #4291 and covers the one edge case found.


GawainLynch:
FYI the failures on these tests are unrelated and fixed in #4839… they were caused by #4834 but hidden due to what was fixed in #4837… Test dependency chain or what!

PR #4840, opened 2016-02-14
Schema foreign key constraints

GawainLynch:
This fixes what I would consider a design error by our friendly neighbourhood :koala: 

…and enables table objects to easily add FKCs to tables before we get stuck with BC in stable.


bobdenotter:
Green lights! :-)

PR #4842, opened 2016-02-14
Set correct route permission in before()

### LINKED ISSUE #4789: Not root-account can't upload images
looks the same as #4241 

When this user is uploading:

it gives

when this user does the same:

it gives 302 and returns html back.

**version: v3 master**

GawainLynch:
Fixes #4789


bobdenotter:
:+1:

Boorj:
that made my day. Thanks

PR #4844, opened 2016-02-14
Less entity magic

GawainLynch:
This updates the **base table** entities away from `@method` annotations to using getter/setters, and updates a bunch of PHPDoc in the process… hopefully making debugging easier, and understanding the classes easier for those that aren't @rossriley and @gawainlynch :wink: 

@rossriley when you're done with :beers: for the day, can you give a "yay" or "nay" on 3c652b992d439da77958b6de7895255e7b2cc6a1 please… my preference would be to have it as is, but if there is a valid design decision behind mixing them, I am all ears.

### DIFF
--- a/src/Storage/Entity/Authtoken.php
+++ b/src/Storage/Entity/Authtoken.php
@@ -3,34 +3,129 @@ namespace Bolt\Storage\Entity;
 
 /**
  * Entity for Auth Tokens.
- *
- * @method integer   getId()
- * @method string    getUsername()
- * @method string    getToken()
- * @method string    getSalt()
- * @method \DateTime getLastseen()
- * @method string    getIp()
- * @method string    getUseragent()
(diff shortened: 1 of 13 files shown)


bobdenotter:
Did you have help from PHPStorm on this, or did you manually write 1200 lines of code? ;-)

GawainLynch:
> Did you have help from PHPStorm on this, or did you manually write 1200 lines of code? ;-)

Column A & Column B… There was still a hell of a lot of cross referencing work to be done and some methods were implemented, so it was still a bit of work. But as mentioned, it should make a difference for people "coming up to speed" with Bolt's ORM.

bobdenotter:
> But as mentioned, it should make a difference for people "coming up to speed" with Bolt's ORM.

Yes, I'm liking it! :-)

PR #4849, opened 2016-02-15
Add filesystem mount if not already set

### LINKED ISSUE #4848: Error in parsePath throws an error for custom upload folder
When the system performs this check
```php
        // Test write access for uploadable fields.
        $contentType['fields'] = $this->setCanUpload($contentType['fields']);
```
here: https://github.com/bolt/bolt/blob/master/src/Storage/ContentRequest/Edit.php#L112
it takes upload paths, set for fields in `contenttypes.yml`. 
So, for example, i have a folder called `'downloads'`, which exists : `/files/downloads`.
So when the algorhythm flows further, it goes to [`setCanUpload`](https://github.com/bolt/bolt/blob/master/src/Storage/ContentRequest/Edit.php#L212) and later to [`checkUploadDirectory`](here: https://github.com/bolt/bolt/blob/master/src/Storage/ContentRequest/Edit.php#L218). The last one performs a check with `\Bolt\Filesystem\Manager::parsePath` function. Which throws an error in this check:
```php
        if (!preg_match('#^.+\:\/\/.*#', $path)) {
            throw new InvalidArgumentException('No mount point detected in path: ' . $path);
        }
```
since it doesn't have any `://` instance, and only `'download'` in my example.

'events' is the name of directory in this example:

I suppose it should be prefixed to files://downloads or something like that.

GawainLynch:
Fixes #4848

### DIFF
--- a/src/Storage/ContentRequest/Edit.php
+++ b/src/Storage/ContentRequest/Edit.php
@@ -235,6 +235,9 @@ class Edit
      */
     private function checkUploadDirectory($path)
     {
+        if (strpos('://', $path) === false) {
+            $path = sprintf('files://%s', $path);
+        }
         if ($this->filesystem->has($path)) {
             return $this->filesystem->getVisibility($path) === 'public';
         }


bobdenotter:
Whoop!

PR #4863, opened 2016-02-18
Using 'width' and 'height' instead of 'dimensions'

bobdenotter:
As pointed out by @CarsonF in https://github.com/bolt/bolt/pull/4786#issuecomment-184706715


GawainLynch:
Thanks mate, really appreciated follow up! :+1:

CarsonF:
Thank you @bobdenotter :beer:

PR #4869, opened 2016-02-19
Add a default value for checkboxes

rossriley:
This fixes an error that shows up in newer versions of MySQL which are stricter about types.

Bolt sets a checkbox to a TINYINT field and expects a value of `0` or `1` internally but without a default value set then most browsers will send a value of `on`

With stricter modes for the database this always results in a `0` value being set.


bobdenotter:
I think it's weird that the tests fail on this one. :-/

GawainLynch:
I think the failures are unrelated and already fixed… just dropped focus myself

bobdenotter:
So, let's merge in? :-)

Issue #1914, opened 2014-10-18
[RFC] Dispatchers & listeners


GawainLynch:
@bobdenotter raised the topic of event handlers at BoltCamp.

The vision being that extensions should be able to interface with core Bolt events and respond accordingly.

Location suggestions welcome ;-)

### Suggestions
* Asset creation
* Field type values, e.g. select fields

GawainLynch:
Moving this to the new [forums](https://discuss.bolt.cm/d/9-dispatchers-listeners)

GDmac:
The hasListeners is mostly useful when it might be costly to setup an event object

```php
$data = '';
if ($app['dispatcher']->hasListeners('content_edit.load')) {
    $event = new CostlyEventObject();
    // give data to listeners
    $event->setData($data)
    // listeners might modify data
    $event = $app['dispatcher']->dispatch('my.event', $event);
    // so, retrieve modified data
    $data = $event->getData();
}
// continue normal operation
// $data ....
```

PR #4878, opened 2016-02-20
changed visibility of loginFailed function to protected to allow over…

ToBe998:
The function ```Users->loginFailed($user)``` function is currently marked as ```private```making it untouchable for extending classes.

***Use Case***
The use case is this: Our CMS systems are very frequently hit by attackers and need a more thorough monitoring. I can write a Bolt\Extnsion that replaces the normal Bolt\Users class with a new Users class that extends the Bolt\Users class but also fires certain monitoring events on some functions.
(Yes, doing so from within an Extension might be considered cruede, but it is the only option available without modifying core code.) 

The most important function would be the ```loginFailed()``` function obviously. But this function can not be overridden in a extending class because of it's ```private``` visibility.

Other use cases for this are Bolt Extensions adding LDAP login (Something I also need to do in the near future) or replace other internals with extended versions.

***Compatibility***
Changing it to ```protected```does not introduce incompatibilities to the code and does not introduce any kind of security issues. Anyone able to modify PHP is already way beyond secuity here.

### DIFF
--- a/src/Users.php
+++ b/src/Users.php
@@ -1280,7 +1280,7 @@ class Users
      *
      * @param array $user
      */
-    private function loginFailed($user)
+    protected function loginFailed($user)
     {
         $this->session->getFlashBag()->add('error', Trans::__('Username or password not correct. Please check your input.'));
         $this->app['logger.system']->info("Failed login attempt for '" . $user['displayname'] . "'.", array('event' => 'authentication'));


GawainLynch:
My concern with taking the approach of modifying visibility, and extending the class is firstly, that function no longer exists in master, and the whole class is deprecated for removal in v4.0.

Secondly, the approach really should be to simply have an event dispatched. This would have the advantage of not having to extend a whole class, and making extensions dealing with such things more portable.

There was some discussion in #1914 about what to layout as events. 

Thoughts from others?

ToBe998:
You are right. As I said, there should be a better way and events are one of the best ways for something like this. BUT there are no events in 2 and many people will have to stick with 2 for quite a while.
So i can only repeat the same argument from a past PR, why stop a perfectly compatible feature just because it might be resolved in another solution in the far future?

Another thing, in my oppinion (and yes, this is oppinion and not "the one truth"), no class method should ever need to be private. A fundamental idea of modern software development is to allow modification and extensibility. There simply is no reason to stop this from happening in 99.9% of all cases.

GawainLynch:
> why stop a perfectly compatible feature just because it might be resolved in another solution in the far future?

I am not saying "no" outright… just trying to develop the conversation toward a workable solution. 

That said, my "objection" comes down to BC… it's not a "hard" objection, but one centred around not rushing to the green button to only end up costing more time in support arguments (potentially).

> BUT there are no events in 2 and many people will have to stick with 2 for quite a while.

I am not, personally, against adding the events to v2… it just needs a tiny bit of thought and discussion to make it forward compatible.

> no class method should ever need to be private

We *strongly* disagree there. Non-private methods have BC attached to them… they effectively become part of the API.

ToBe998:
Ok, I think I can work with overriding the other methods for the moment. But I still think there should be a way to extend core classes within reasonable boundaries. And not just in future iterations when the changes would be "soft" enough,

Anyways. Let's see how this one is working out...

GawainLynch:
Well, hard/soft becomes interesting at a point… 2.2 is a long term support **stable** minor release, so technically not open to feature addition, and both of our proposed solutions fall into that category :grin: 

> But I still think there should be a way to extend core classes within reasonable boundaries.

Yeah, that's been a massive part of the work we've done in the upcoming v3. So much more is doable now.

OK… so how would #4883 work for you?

ToBe998:
Looging good. I posted my feedback there.

In the future, we would probably still need a way to replace services, or change service results via events. This would enable us to create other login methods like LDAP or similar SSO systems.

Anyways, this PR is very likely obsolete. I'll close it as soon as #4883 is working.

GawainLynch:
Development meeting of 01/03/16 vote close feature addition to the 2.2 branch, so both this and #4883 are in no-go territory I am afraid.

ToBe998:
I hope that is not a confirmation for a very bad route Bolt may be taking at the moment. One of the big appeals of Bolt was it's simplicity. Very easy and fast to setup and use. Not many features, but you could add any feature you were missing yourself without much effort.

Now Bolt it is getting much more compelexity (which is not a bad thing in itself). It also puts many more barriers in place to stop people from using loopholes in the "official extension concept" to add features that werent strictly possible with allowed means (ex.: OOP designs like this that make it impossible to add functionality that wasnt initially planned). At the same time the code does not provide more or better "official" routes to add missing functionality.

I really hope im mistaken. We'll see...

GawainLynch:
I'm not sure I see your logic here. Both of these PRs would add to the complexity.

You are still free to extend the class, as you are now… yes, you need to re-implement the login functions as that is a bit more code than you'd like… but this is a mature and stable branch, it will be a once-off implementation for you.

If there was going to be a continuation of the features in v2 then the events would have definitely gone in… no two ways about it.

The core developers work damn hard on this project, but we are spread thin and need to manage the amount of support we have to provide, and adding breaking features **very late** in a stable branch's life is asking for the potentiality of more problems.

ToBe998:
I know you guys are working hard. I was just voicing some concerns that appeared after following discussions about this and other issues in the last few weeks in git and on slack. I can understand that new features should not go in at this stage of the release cycle, but based on other comments and issues i saw, i fear that this "no go" stance might remain even after the new version is released.

Like i said, i might be mistaken and i really hope so.

PR #4882, opened 2016-02-21
Return the correct response after deleting a file

### LINKED ISSUE #4854: Error alert when deleting file
When i'm trying to delete a file, Bolt shows an error

but **deletes the file**. But you can't see that file is removed until you reload the page.

robbert-vdh:
Deleting files through the file manager always resulted in an error as the result of the void method `\Bolt\Filesystem\Manager->delete()` was used to check whether the deletion was successful or not. This fixes #4854.

I decided to not include the exception itself in the error message as that would make the error very verbose.

### DIFF
--- a/src/Controller/Async/FilesystemManager.php
+++ b/src/Controller/Async/FilesystemManager.php
@@ -2,6 +2,7 @@
 namespace Bolt\Controller\Async;
 
 use Bolt\Filesystem\Exception\ExceptionInterface;
+use Bolt\Filesystem\Exception\FileNotFoundException;
 use Bolt\Filesystem\Exception\IOException;
 use Bolt\Translation\Translator as Trans;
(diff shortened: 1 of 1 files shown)


GawainLynch:
:+1: 

I don't suppose you could be so awesome as to add an `$app['logger.system']->critical()` to log the exception?

robbert-vdh:
Sure, I'll add it when I get back home!

robbert-vdh:
@GawainLynch I've added exception logging. I've also noticed half of the filesystem tests are commented out so I'll rework those in another pull request.

GawainLynch:
> filesystem tests are commented out so I'll rework those in another pull request.

Yeah, they're now in [bolt/filesystem](https://github.com/bolt/filesystem)

PR #4883, opened 2016-02-21
Enable access control events for login, logout & password reset

GawainLynch:
This is an alternative approach to #4878 that adds forward compatible events for login, logout & password reset

### DIFF
--- a/src/Events/AccessControlEvent.php
+++ b/src/Events/AccessControlEvent.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Bolt\Events;
+
+use Exception;
+use Silex\Application;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\HttpFoundation\Request;
+
+class AccessControlEvent extends Event
+{
+    /** @var string */
+    private $clientIp;
+    /** @var integer */
+    private $dateTime;
+    /** @var string */
+    private $uri;
+    /** @var string */
+    private $userName;
+    /** @var \Exception */
+    private $exception;
+
+    /**
+     * Constructor.
+     *
+     * @param Request   $request
+     * @param Exception $exception
+     */
+    public function __construct(Request $request, \Exception $exception = null)
+    {
+        $this->exception = $exception;
+
+        $this->clientIp = $request->getClientIp();
+        $this->dateTime = $request->server->get('REQUEST_TIME');
+        $this->uri = $request->getUri();
+        $this->userName = $request->request->get('username');
+    }
+
+    /**
+     * Return any trapped exceptions.
+     *
+     * @return Exception
+     */
+    public function getException()
+    {
+        return $this->exception;
+    }
+
+    /**
+     * Return the IP address requesting the access.
+     *
+     * @return string
+     */
+    public function getClientIp()
+    {
+        return $this->clientIp;
+    }
+
+    /**
+     * Return a DateTime object representing the time the request occurred.
+     *
+     * @return \DateTime
+     */
+    public function getDateTime()
+    {
+        $dt = new \DateTime();
+
+        return $dt->setTimestamp($this->dateTime);
+    }
+
+    /**
+     * Return the requested URI of the access event.
+     *
+     * @return string
+     */
+    public function getUri()
+    {
+        return $this->uri;
+    }
+
+    /**
+     * Return the given user name of the access event.
+     *
+     * @return mixed|string
+     */
+    public function getUserName()
+    {
+        return $this->userName;
+    }
+}
--- a/src/Users.php
+++ b/src/Users.php
@@ -2,6 +2,8 @@
 
 namespace Bolt;
 
+use Bolt\Events\AccessControlEvent;
+use Bolt\Events\AccessControlEvents;
 use Bolt\Translation\Translator as Trans;
 use Doctrine\DBAL\DBALException;
 use Hautelook\Phpass\PasswordHash;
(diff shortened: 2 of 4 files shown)


GawainLynch:
@CarsonF quick review? …and yeah, in hindsight the exceptions are probably not relevant for 2.x and 3.0.0 but more of that to come in 3.x

PhillippOhlandt:
That's cool. I had an issue talking about those events some months ago. Will search it when I am on the PC later.

ToBe998:
This is what i was actually looking for in #4878, My extended User service did exactly that. Firing these events inside the service instead the controller might be a slightly better idea thouch:

- Loginc remains inside the service as controller should only do routing and rendering etc
- Ability to differentiate between the different reasons for a failed login. (Username not found, password wrong, user disabled, stalled, etc...)

GawainLynch:
> Loginc remains inside the service as controller should only do routing and rendering etc

Business logic should not be in a service… the only thing worse than putting this sort of logic in a controller is putting it in a service.

I agree that a controller should be only doing routing and rendering, and @CarsonF and I have made some pretty big refactoring passes to move that along with v3, but we still have a few more passes to go.

Nevertheless, my caution comes from keeping this easily forward portable, and wanting to keep the dialogue open as the rest of the core team need to speak up… and for 2.2 we're kinda stuck in terms for approach.

> Ability to differentiate between the different reasons for a failed login. (Username not found, password wrong, user disabled, stalled, etc...)

Yeah, got to agree with you on this one :white_check_mark: 

This was a very quick, almost PoC, attempt… I knew you'd chime in constructively pretty quickly. :smile: 

But…  #4878 only gives you access to the abused use of the application object, and the user array… the latter, for the most, part is less information than this already provides.

Ideas **very much** welcome!

ToBe998:
> Business logic should not be in a service…
Lets just agree to disagree here. This is your app and i really like it. So ok. Let's go on.

> I knew you'd chime in constructively pretty quickly.
Jus tell me if you guys think i should mind my own bussiness. I know i can get a bit stubborn when coding is involved. ;)

I checked the code again, and I think without moving the dispatches into the service, there is no nice method to "enrichen" the fail event without moving this code to far from it's current scheme.
So, since the monitoring service could still get most relevant info and failure cause from checking the user inside the database itself, it's ok for 2.2.

GawainLynch:
BTW… I dropped a word in that debate… what I meant to say was "Business logic should not be in a **service provider**" … so I think we're on the same page, and yes, using the dispatcher in the **service class** is fine with me.

My week is kinda full on so it will be a day or three before I can properly get back to this, but if you have time feel free to PR this branch in my fork.

ToBe998:
No need to rush. I'd rather wait for you to do the clean and nice implementation, possibly with the "login failed Reason" that start implementing now.

src/Events/AccessControlEvent.php
```diff
@@ -0,0 +1,90 @@
...
+{
+    /** @var string */
+    private $clientIp;
+    /** @var integer */
+    private $dateTime;
+    /** @var string */
+    private $uri;
+    /** @var string */
+    private $userName;
+    /** @var \Exception */
```
CarsonF: or null

src/Events/AccessControlEvent.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->clientIp = $request->getClientIp();
+        $this->dateTime = $request->server->get('REQUEST_TIME');
+        $this->uri = $request->getUri();
+        $this->userName = $request->request->get('username');
+    }
+
+    /**
+     * Return any trapped exceptions.
+     *
+     * @return Exception
```
CarsonF: or null

src/Events/AccessControlEvent.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Return a DateTime object representing the time the request occurred.
+     *
+     * @return \DateTime
+     */
+    public function getDateTime()
+    {
+        $dt = new \DateTime();
+
+        return $dt->setTimestamp($this->dateTime);
```
CarsonF: Not following...why do this?

src/Users.php
```diff
@@ -721,10 +723,11 @@ public function resetPasswordConfirm($token)
         $query = sprintf('SELECT * FROM %s WHERE shadowtoken = ? AND shadowvalidity > ?', $this->usertable);
         $query = $this->app['db']->getDatabasePlatform()->modifyLimitQuery($query, 1);
         $user = $this->db->executeQuery($query, array($token, $now), array(\PDO::PARAM_STR))->fetch();
+        $event = new AccessControlEvent($this->app['request_stack']->getCurrentRequest());
```
CarsonF: Can we guarantee request cycle?

CarsonF:
How closely can this implementation be adapted to Symfony's Security Component? Any idea? Would love to get on to that at some point. 

All in all, events are a great way to add extendability without restricting service implementation...as long as they are generic enough that they do not imply else. For example, Symfony Console in v3 now binds input before dispatching command event ([0af1676](https://github.com/symfony/symfony/commit/0af16762932e96500e2e8a143c058dcbfaac8bb9)) which screwed up a listener I had. Anyways...not sure where I was going with that..

Looks fine to me, other than the comments I made on the diffs and the fact that this is a new feature for a release branch but I guess I'll let that slide :P :+1:

src/Events/AccessControlEvent.php
GawainLynch: Far out… I was coding while sober!

Issue #4886, opened 2016-02-21
[BUG/WIP] Controller actions in the async FilesystemManager controller are still using the old API


robbert-vdh:
The FilesystemManager controller is still using the old API and relies on filesystem manager returning booleans instead of throwing exceptions. This causes some file operations to throw errors even though they still succeed. I've already fixed this for `/async/file/delete` in #4882 but I just noticed this is happening in more places. I'm working on a patch (and some much needed tests) right now.

GawainLynch:
Thanks mate. The help is very welcome... Especially on tests.

PR #4887, opened 2016-02-22
Rework the FilesystemManager controller and add tests

robbert-vdh:
I've updated `\Bolt\Controller\Async\FilesystemManager` to work properly with Bolt 3.0's filesystem API and I've improved its responses by returning more proper status codes where possible, logging any errors to the system log (as suggested by @GawainLynch in #4882) and returning a file path instead of an empty JSON object (`{}`) on success. I've also added missing tests so that every controller action is covered.

There are a lot of commits (I was feeling generous) so I could probably squash some of them.

Fixes: #4886

### DIFF
--- a/src/Controller/Async/FilesystemManager.php
+++ b/src/Controller/Async/FilesystemManager.php
@@ -320,13 +320,26 @@ class FilesystemManager extends AsyncBase
         }
 
         try {
-            if ($this->filesystem()->rename("$namespace://$parentPath/$oldName", "$parentPath/$newName")) {
-                return $this->json(null, Response::HTTP_OK);
+            $this->filesystem()->rename("$namespace://$parentPath/$oldName", "$parentPath/$newName");
+
+            return $this->json(null, Response::HTTP_OK);
+        } catch (\Exception $e) {
+            $msg = Trans::__('Unable to rename file: %FILE%', ['%FILE%' => $oldName]);
+
+            $this->app['logger.system']->critical(
+                $msg . ': ' . $e->getMessage(),
+                ['event' => 'status', 'status' => $e]
+            );
+
+            if ($e instanceof FileExistsException) {
+                $status = Response::HTTP_CONFLICT;
+            } elseif ($e instanceof FileNotFoundException) {
+                $status = Response::HTTP_NOT_FOUND;
+            } else {
+                $status = Response::HTTP_INTERNAL_SERVER_ERROR;
             }
 
-            return $this->json(Trans::__('Unable to rename file: %FILE%', ['%FILE%' => $oldName]), Response::HTTP_FORBIDDEN);
-        } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


bobdenotter:
At first glance: Good stuff! It's very refreshing to see someone willing to help out with (boring) tests, instead of hacking together exciting but fragile new features. :+1:

robbert-vdh:
A shame, as writing tests can be quite fun and they're a great way to spot bugs early on. Though I've got to admit that writing shiny new features sounds a bit more exciting.

GawainLynch:
> A shame, as writing tests can be quite fun and they're a great way to spot bugs early on. 

Yeah, I've found a couple of (quietly fixed) doozies that way… but as @bobdenotter said, thanks so much!

robbert-vdh:
Looks like I forgot to clean up the mess I made.

src/Controller/Async/FilesystemManager.php
```diff
@@ -222,15 +240,12 @@ public function duplicateFile(Request $request)
...
-        $extensions = implode('|', explode(',', $request->query->get('ext', '.*')));
+        $extensions = implode('|', explode(',', $request->get('ext', '.*')));
         $regex = sprintf('/.*(%s).*\.(%s)$/', $term, $extensions);
 
-        $files = $this->filesystem()
-            ->find()
-            ->in('files://')
-            ->name($regex)
-        ;
+        $files = $this->filesystem()->find()->in('files://')->name($regex);
```
GawainLynch: FYI: Undid this one

GawainLynch:
Thanks mate!

CarsonF:
Can we revert the usage of [Request::get()](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Request.php#L709-L715)?
> 
     * Avoid using this method in controllers:
     *
     *  * slow
     *  * prefer to get from a "named" source
     *
     * It is better to explicitly get request parameters from the appropriate
     * public property instead (query, attributes, request).

GawainLynch:
Ah crap… I meant to catch that one too! Need more coffee :frowning:

CarsonF:
@robbert-vdh Thanks for your help, this is HUGE! :+1:

robbert-vdh:
@CarsonF Ah, great, I couldn't decide on whether or not the the `$request->query->get()`s were intentional or not and PHPStorm was complaining about uncatched exceptions. :+1:

CarsonF:
I use PHPStorm as well. What was it complaining about?

robbert-vdh:
I didn't look at it closely enough. There shouldn't be any problems as the exception only gets thrown when the path selector itself is malformed. It just stood out to me.

CarsonF:
@robbert-vdh I would just try to ignore that for now. `InvalidArgumentException` is thrown from `get()` only when there's a problem with deep selectors. This is deprecated behavior and removed in Symfony 3.0.

PR #4889, opened 2016-02-22
Don't show Whoops! screens for regular 404's. Fixes #4888

### LINKED ISSUE #4888: [RFC] Should we display '404 pages' instead of error pages, for logged in users?
Currently, when a visitor is _not_ logged in, they get nice 404 pages, like this: 

screenshot 2016-02-22 22 52 10

However, if a user is logged in, they see this, for the exact same URL: 

screenshot 2016-02-22 22 48 21

~~Silex~~ Whoops! does this, because it _assumes_ the 404 is an error. Strictly speaking that's correct, but it also confuses a lot of people. 

Should we fix it so that 404's in the frontend _always_ show the proper 404 page, regardless of logged-in state?

bobdenotter:
Fixes #4888


bobdenotter:
@GawainLynch Nice! Didn't know about that. :+1:

GawainLynch:
Running with this approach :+1:

PR #4890, opened 2016-02-23
Added the search dialog to the standalone file editor interface

xpika:
After Logging in and selecting 

**File Management** > **View/Edit templates**

Doing a Ctrl+F for search does not load the search dialog like it does when editing a page from a content type. As you may know using the native browser search does not work with codemirror for any longer than a couple of pages.

Added the search dialog plugin script resource to the template


GawainLynch:
Travis failure is PHP 5.3 memory use with Composer.

@rarila should this be compiled together, and forward port strategy?

rarila:
@GawainLynch hmmm, I think it can stay this way for now. Forward porting should be no problem.

GawainLynch:
Thanks @xpika :+1:

PR #4893, opened 2016-02-23
Allow for overriding `notfound` in theme.yml

bobdenotter:
In theme.yml, we can currently override most of the templates, optionally: 

```
maintenance_template: maintenance_default.twig

homepage_template: index.twig
record_template: record.twig
listing_template: listing.twig
search_results_template: listing.twig
```

The only one we can't override, is `notfound:`:

```
notfound: notfound.twig
```

This PR makes the behaviour consistent.


rarila:
Seeing the others my ocd says that should have been ``not_found`` from the beginning but I guess we're caught in the BC trap, as so often.

PR #4901, opened 2016-02-24
missing loop.last gives fatal error in strict mode

jadwigo:
when strict variables is on and there is only a single relation  the loop.last is empty.

this sets the loop last to the same as loop first in that case - which prevents the error

### DIFF
--- a/app/view/twig/_base/_listing.twig
+++ b/app/view/twig/_base/_listing.twig
@@ -31,7 +31,7 @@
     nextgroup:      new_group|default(false),
     unordered:      request('order') == '',
     first:          loop.first,
-    last:           loop.last
+    last:           loop.last|default(loop.first)
 }%}
 
 {% set local = {


bobdenotter:
Whooop. :baby_chick:

rarila:
@jadwigo This patch triggers wrong behavior in listings. Can you please tell me tell me how to reproduce the bug you encountered.

``loop`` is a twig internal variable and so always should be set automatically.

jadwigo:
@rarila see the first sentence

loop through an array with length == 1 and strict mode on

in that case loop is set but specifically `loop.last` is not set and if you call `loop.last` in twig that results in a warning

rarila:
Sure, but you also write "single relation", so I wonder where this apears? Simple overview listing like "Overview » Entries" ?

jadwigo:
In the front-end - a record with relations when there is only one relation set.

I think what triggered it was something like:
```
{% set relations = record.relations() %}
{% for relation in relations %}
  {% if loop.last %}say hi{% endif %}
{% endfor %}
```

rarila:
Front-end?! ``_listing.twig`` is used in backend for contentypes listing and dashboard.

As I thought: ``loop.last`` is always set, also for arrays with one element, try:

```twig
{% for letter in ['a'] %} {{ dump(loop) }} {% endfor %}
```

rarila:
Also ``loop.last|default(loop.first)`` wouldn't do, as ``default()`` is not only triggered by undefined values but also by falsey values. So if ``loop.last`` is ``false`` it would be set to ``loop.first`` which isn't really what we want. See:

```twig
{% set foo = false %}
{% dump(foo|default('Not what we expected!')) %}
```

So I think it would be good if you could find where that happened in real code so that we can find the real cause of this and savely revert that change back.

jadwigo:
Yup.. hmm, in that case it probably was in the back-end.

Lets reproduce.

* We have a clean install - bolt 3.0.0 alpha 8
* Two content types are defined ... pages and entries.
* Entries have a relation to pages.
* There is one page in the system, it is published.
* There are some entries in the system.
* One entry has one relation to the existing and published page. The entry is also published.
* Strict mode is on `strict_variables: true` and php is in full annoyance mode `debug_error_level: 30719`
* the template _listing.twig has the following code (without the patch):
```
{% set prop = {
    extended:       not compact,
    has_groupname:  content.group.name is defined,
    nextgroup:      new_group|default(false),
    unordered:      request('order') == '',
    first:          loop.first,
    last:           loop.last
}%}
```
* If you go to http://example.com/bolt/editcontent/pages/1 the following error message turns up

```
Class: Twig_Error_Runtime
Message: Key "last" for array with keys "parent, index0, index, first" does not exist in "@bolt/_base/_listing.twig" at line 34
Code: 0
Twig_Template->getAttribute()
[root]//vendor/twig/twig/lib/Twig/Environment.php(403) : eval()'d code, line 53

__TwigTemplate_256eda7ae11c8d9c054f42c428597beede3f45c4ef6460e60cbb92b87fce3adc->doDisplay()
[root]//vendor/twig/twig/lib/Twig/Template.php, line 381
[...]
```

So it complains about `loop.last` that can't be assigned to `prop.last`, because it has a false-ish value.

Obviously the fix has unintended side effects because I totally borked the logic there.

jadwigo:
*glad I can still reproduce the error though*

rarila:
Ah, in the listing of relationsships… thanks! Bug comfirmed! Further investigation…

rarila:
I think I found it. It seems to be [here ](https://github.com/bolt/bolt/blob/master/app/view/twig/editcontent/_relations.twig#L21).

The problematic part is the if clause in the for loop. If you remove it, it works and you get the full loop object. If it is there you only get the crippled one. Is that a bug in twig or a feature… gonna find out…

rarila:
Ok, twig docu says:

> Using the loop variable within the condition is not recommended as it will probably not be doing what you *expect* it to.

For sure, yes, no one expects the Spanish Inquisition…

I guess we have to filter in front.

PR #4902, opened 2016-02-24
Slightly improve legibility of user agents.

bobdenotter:
Before: 

screen shot 2016-02-24 at 17 56 41

After: 

screen shot 2016-02-24 at 17 58 31


rarila:
``buic-moment``… really? That one should really be only used for display of relative times.

PR #4904, opened 2016-02-25
Add "require" option to select fields

rarila:
On multiple wish of one individual ;-)

### DIFF
--- a/app/view/twig/editcontent/fields/_select.twig
+++ b/app/view/twig/editcontent/fields/_select.twig
@@ -1,10 +1,11 @@
 {#=== OPTIONS ========================================================================================================#}
 
 {% set option = {
+    info:      field.info|default(''),
     label:     field.label,
     multiple:  (field.multiple is defined and field.multiple),
+    required:  field.required|default(false),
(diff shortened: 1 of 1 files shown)


CarsonF:
There's still the empty value. Shouldn't
```twig
{% set values = lookups|selectfield(lookupfield, option.multiple, field.keys|default('id')) %}
```
be
```twig
{% set values = lookups|selectfield(lookupfield, option.multiple or option.required, field.keys|default('id')) %}
```
?

rarila:
[Must](https://www.w3.org/TR/html-markup/select.html#select-constraints):

>A select element with a required attribute and without a multiple attribute, and whose size is “1”, must have a child option element.

>The first child option element of a select element with a required attribute and without a multiple attribute, and whose size is “1”, must have either an empty value attribute, or must have no text content.

CarsonF:
Hmmmm ok. As long as it's enforced though. I can't save it empty right?

rarila:
Car

CarsonF:
Good enough for me :+1:

PR #4910, opened 2016-02-25
Mandrill is dead to us!

SahAssar:
Mandrill is no longer free, so send people to sendgrid.com/free instead.


bobdenotter:
:+1:

dadaxr:
There is also [tipimail](https://www.tipimail.com), with a free 10K mail / month option.
This solution is more close to mandrill than sendgrid.

And the icing on the cake : it's a french company ;)

GawainLynch:
> And the icing on the cake : it's a french company ;)

:wink:

PR #4911, opened 2016-02-25
[master] Mandrill is dead to us!

SahAssar:
Mandrill is no longer free, so send people to sendgrid.com/free instead.


bobdenotter:
:+1:

Issue #4912, opened 2016-02-25
Symfony's Webdeveloper Toolbar should _not_ set flashes..


bobdenotter:
Currently, on most pages, most flashes are set twice. This is because the "regular" page sets it, and then the initialisation of the toolbar sets it again. 

A simple solution is this, is to add something like this to `Flashlogger.php`, in `function add()`:

```
        // Don't show the check if we're in the Webdeveloper toolbar.
        $app = ResourceManager::getApp();
        if ($app['request']->get('_route') === '_wdt') {
            return;
        }
```

But, I'm certain Gawain, Ross or Carson will know of a better place to put that. :-)

CarsonF:
Yes I'll fixed that in my session/csrf branch landing in a day or two

Issue #4899, opened 2016-02-24
[BUG] login page shows database update warning


jadwigo:
If there are pending database updates they are displayed on the login screen.

![image](https://cloud.githubusercontent.com/assets/1488214/13287762/e7e6c44c-db08-11e5-9032-a3ffca48d99f.png)

That message should not be there.

Also, there's some redundancy in the "You have been logged out" and "Please log on" messages.

Details
-------

 - Relevant Bolt Version: 3.0
 - Install type: Composer install
 - PHP version: 5.6
 - Used webserver: Nginx
 - For UX/UI issues: all

GawainLynch:
Redundant redundancy is becoming redundantly redundant!

jadwigo:
Many abundantly clear and descriptive explanations later we still have much information telling us the same thing :)

bobdenotter:
Thinking about this, I think the proper solution would be to only allow _specific_ flash messages on the logon screen. 

I can fix this, and will add the fix to #4881, because it influences the same bit of code.

EJanuszewski:
The database message also stays popped up after updating it(on the database update page), I might open a separate issue for this.

bobdenotter:
Fixed by #4881.

jadwigo:
The following screen pops up _after_ you ran the database update.

![image](https://cloud.githubusercontent.com/assets/1488214/13434609/c52b0080-dfd6-11e5-8b87-d5d473ec8251.png)

So if you need a database update and you go to the update page the notice is hidden and the update button is shown. After you click on the button to update the database the upgrade notice is shown again as well as the 'up to date' notice.

bobdenotter:
@jadwigo The original issue (showing it on the dashboard) is fixed. The one you're seeing now is a duplicate of #4912

PR #4913, opened 2016-02-25
Don't show "Update DB" notice, when we're already updating the DB.

bobdenotter:
As pointed out by @EJanuszewski in #4899 

> The database message also stays popped up after updating it(on the database update page), I might open a separate issue for this.

This PR fixes that, as soon as #4912 is also fixed.

### DIFF
--- a/src/EventListener/StorageEventListener.php
+++ b/src/EventListener/StorageEventListener.php
@@ -3,6 +3,7 @@ namespace Bolt\EventListener;
 
 use Bolt\AccessControl\Permissions;
 use Bolt\Config;
+use Bolt\Configuration\ResourceManager;
 use Bolt\Events\HydrationEvent;
 use Bolt\Events\StorageEvent;
 use Bolt\Events\StorageEvents;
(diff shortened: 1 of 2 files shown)


bobdenotter:
xzibit

src/EventListener/StorageEventListener.php
```diff
@@ -120,7 +121,12 @@ protected function schemaCheck(GetResponseEvent $event)
         $session = $event->getRequest()->getSession();
         $validSession = $session->isStarted() && $session->get('authentication');
         $expired = $this->schemaManager->isCheckRequired();
-        if ($validSession && $expired && $this->schemaManager->isUpdateRequired()) {
+
+        // Don't show the check if we're in the dbcheck already.
+        $app = ResourceManager::getApp();
```
GawainLynch: That is gone very soon… like **very** soon… like 3.1.0 soon.

src/EventListener/StorageEventListener.php
```diff
@@ -120,7 +121,12 @@ protected function schemaCheck(GetResponseEvent $event)
...
+        $notInCheck = !in_array($app['request']->get('_route'), ['dbcheck', 'dbupdate_result', 'dbupdate']);
```
GawainLynch: Kernel event listener event object have the request object

bobdenotter:
@GawainLynch Thanks! Much nicer this way.

Issue #4907, opened 2016-02-25
Twig variable missing in editor for text fields


GawainLynch:
This shows up in master editing a Showcase with Twig set to strict.
```
Key "default" for array with keys "class, label, variant, canUpload, type" 
does not exist in "@bolt/editcontent/fields/_text.twig" at line 32
```

bobdenotter:
On IRC, @EJanuszewski reported something similar: 

```
Impossible to access an attribute ("getEmptySet") on a string variable ("") in "@bolt/editcontent/fields/_repeater.twig" at line 26
```

As far as I can trace it, it seems to have to do with 'default' not being set on repeaters. If I edit the default contenttypes.yml to have: 

```
        repeater:
            type: repeater
            group: repeater
            limit: 3
            prefix: "<p>This allows you to create multiple sets of fields. Use the add button at the bottom to create a new empty set.</p>"
            fields:
                repeattitle:
                    type: text
                    default: 'foo'

```

It works. I have no idea how to fix that though. @rossriley, any ideas?

PR #4914, opened 2016-02-25
Fixes for Twig strict-mode with repeater groups

rossriley:
Pinging @bobdenotter @EJanuszewski for test please.

Hopefully Fixes: #4907 

Adds some auto defaults to the base default array that should prevent Twig choking in strict mode.
Also fixes an access issue for the date field which was also causing a failure in strict mode.

### DIFF
--- a/app/view/twig/editcontent/fields/_date.twig
+++ b/app/view/twig/editcontent/fields/_date.twig
@@ -6,7 +6,7 @@
     label:     field.label,
     info:      field.info|default(''),
     options:   field.options.datepicker|default(''),
-    default:   (context.content.id == null and field.default) ? field.default|date('Y-m-d') : null,
+    default:   (context.content.get('id') == null and field.default) ? field.default|date('Y-m-d') : null,
     required:  field.required|default(false),
     errortext: field.error|default(''),
 } %}
--- a/app/view/twig/editcontent/fields/_repeater-group.twig
+++ b/app/view/twig/editcontent/fields/_repeater-group.twig
@@ -15,7 +15,8 @@
             class:      '',
             label:      '',
             variant:    '',
-            canUpload:  field.canUpload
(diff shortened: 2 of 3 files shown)


GawainLynch:
@rossriley fixes it for me

bobdenotter:
Works for me, too! :+1: 

Let's see if @EJanuszewski chimes in. He had a different error than Gawain and I got.

EJanuszewski:
Oo I feel special getting pinged :D I shall test it out, although it will be a different environment as I got the error at work, I can test here at home and at work in the morning

rossriley:
pinging @EJanuszewski just because.

EJanuszewski:
@rossriley Getting the same error as I did at work, updated to this and still get the same error, maybe just do it as a separate issue though since it seem unrelated?

```
Impossible to access an attribute ("getEmptySet") on a string variable ("") in "@bolt/editcontent/fields/_repeater.twig" at line 26
```

rossriley:
@EJanuszewski can you paste the contenttypes.yml part where your repeaters are defined and I'll try and reproduce. It's probably just a combination of configs I've not tested against.

EJanuszewski:
@rossriley As far as I was aware mines standard from the master branch, here's the repeater section anyway

```        repeater:
            type: repeater
            group: repeater
            limit: 3
            prefix: "<p>This allows you to create multiple sets of fields. Use the add button at the bottom to create a new empty set.</p>"
            fields:
                repeattitle:
                    type: text
                repeatimage:
                    type: image
                    extensions: [ gif, jpg, png ]
                repeatcontent:
                    type: html
```

and whole config if its useful http://pastebin.com/raw/Jhp2uwjy

rossriley:
Thanks, I think I've got it now. The latest push should make sure we catch that error now.

PR #4609, opened 2015-12-19
Feature/password strength indicator

bobdenotter:
This PR adds a password strength indicator, for when you’re adding the first user, or adding new users. No hard cutoffs, just a nudge to let people input a bit more complex passwords.

PR #4918, opened 2016-02-26
Proposal to improve some code

kamesatou:
May I propose to improve some code as follows?
1. File extensions changing by root condition by reading from a config file instead.
2. Optional solution to achieve more hardening a password.
3. Modify some twigs for XSS protection

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -308,8 +308,8 @@ liveeditor: true
 # it, your mail might disappear into a black hole, without producing any errors.
 # Generally speaking, using 'smtp' is the better option, so use that if possible.
 #
-# Protip: If your webhost does not support SMTP, sign up for a (free) Mandrill
-# account at https://mandrill.com/ for sending emails reliably.
+# Protip: If your webhost does not support SMTP, sign up for a (free) Sendgrid
+# account at https://sendgrid.com/free for sending emails reliably.
 
 # mailoptions:
 #     transport: smtp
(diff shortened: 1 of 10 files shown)


rarila:
Isn't there ``autoescape`` set to on per default?

GawainLynch:
**Password change:** 

This is a stable branch and that is a BC break. Not to mention there is more than enough data out there now to show why one shouldn't take this approach. So that is a "no" from me. :grin: 

**Options for file management:** 

This comes up occasionally, but if it is to be changed at any point a) not in a stable branch; b) it should be part of the permissions system, not *another* config option.

bobdenotter:
Hi @kamesatou, 

Thanks for the PR, but we can't merge this in. I agree with the comments above, and additionally:

> 1. File extensions changing by root condition by reading from a config file instead.

This will have no effect. The root user will be able to just go into config, and turn the setting off. They are root after all. 

> 2. Optional solution to achieve more hardening a password.

We _encourage_ longer passwords already (See #4609). Your approach however _lowers_ the total entropy by reducing the number of combinations, by placing restrictions. This is a good read: https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/

> 3. Modify some twigs for XSS protection

This also has no effect. The "hacker" can't add javascript to "titles" anymore, so they'll just add it to the body, any other text field, or directly to the template. 
Additionally, it will break existing sites. I often use titles like `My <em>awesome</em> page!`. That's behaviour people expect. 

So, once again, thanks for the effort, but we're going to close this one.

PR #4920, opened 2016-02-27
Since the Configuration notices are the most "actionable", it makes sense to put them at the top of the aside.

### LINKED ISSUE #4916: [RFC] Configuration notices should be more prominent
Everyone likes their dose of new, granted. But human eyes will not scan past the first or second box on average:
image

bobdenotter:
Fixes #4916


GawainLynch:
:+1:

PR #4922, opened 2016-02-27
Fix checkbox not showing on filelist modal

EJanuszewski:
Fix the checkbox not showing in the modal when using filelist type, this got missed and was only showing for imagelist type.

### DIFF
--- a/app/view/twig/files_async/files_async.twig
+++ b/app/view/twig/files_async/files_async.twig
@@ -83,6 +83,11 @@
                     {% else %}
 
                         <tr>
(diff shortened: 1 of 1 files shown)


GawainLynch:
It is all good on the master version then?

EJanuszewski:
Yep, just checked it, it doesn't have the same check for how it determines what layout to show so works for both imagelist and filelist

GawainLynch:
:koala: :white_check_mark: 
:de:  :white_check_mark:

Issue #4923, opened 2016-02-27
[BUG] Clicking into filelist/imagelist title doesn't focus first time


EJanuszewski:
When clicking the input field in a filelist and imagelist, it doesn't focus into the textbox, you have to click again which is really annoying, it seems to work if its the first click into the list, but not when clicking between each one

Details
-------
 - Relevant Bolt Version: 2.2.18
 - Install type: Git
 - PHP version: 5.5.9
 - Used webserver: Apache version
 - For UX/UI issues: Chrome 48.0.2

I shall investigate this and try put a fix in if we agree this is a bug

bobdenotter:
I'm seeing the same thing: http://recordit.co/zzJ3liOAuy

I guess there's a rogue `blur();` somewhere, that makes it lose focus, when toggling between the different inputs.

bobdenotter:
Note:I'm seeing this _only_ on 2.x.. It works correctly in 3.0.

EJanuszewski:
Good to know its fine in 3.0, I might compare them since I imagine the core is similar in that area

EJanuszewski:
Looking at it in 3.0 it seems the file stuff is handled in /app/src/js/widgets/fields/fieldFilelist.js and there doesn't appear to be anything similar so I can't copy/backport it with the same code.

rarila:
@EJanuszewski half of JS stuff has been completely rewritten for 3.0 (and the other half would have been if I had more spare time)

EJanuszewski:
@rarila yeah I noticed, took me a good 5 minutes to find it in 3.0, you've done a good job :)

rarila:
Thanks! :metal:

PR #4925, opened 2016-02-27
Fix for image/file list blur #4923

EJanuszewski:
Fixes: #4923(please refer to an existing issue number, that this PR fixes)

There's so many different ways to check what the tag is, are you happy with this way?


EJanuszewski:
Hold up on this, it doesn't seem to save the files now if you only change title and try press save button and nothing else

bobdenotter:
Note for others looking along: This issue is only present in 2.x, not in 3.0. So, a 'crude fix' will do, as far as i am concerned. :-)

bobdenotter:
Works for me! 

Thanks, @EJanuszewski :+1:

PR #4926, opened 2016-02-27
[RFC] Set default error reporting to ignore 'warnings'

bobdenotter:
Peter bumped into an issue on bolt 2.2.18, where the composer code was inserting warnings in ajaxy requests.

See screenshot: 

screenshot 2016-02-27 20 16 27

This proposed fix 'resolves' it, but i'm not certain it's the right way to go. The offending code is not ours, so we can't easily fix it.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -249,7 +249,7 @@ debug: true
 developer_notices: false
 debug_show_loggedoff: false
 debug_permission_audit_mode: false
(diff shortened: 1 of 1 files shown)


bobdenotter:
See also: https://discuss.bolt.cm/d/230-problem-with-listing-updating-extensions

GawainLynch:
@bobdenotter pulled into master as well

PR #4927, opened 2016-02-28
[RFC] Adding official UI Bolt events

rarila:
… for backend frontend  :smirk:

Proposed API:
```javascript
// Fire an event:
bolt.events.fire(eventType);
// Subscribe to an event:
bolt.events.on(eventType, handler);
// Subscribe to an event (one shot):
bolt.events.one(eventType, handler);
```

``eventType`` has to be in the form of: "event.&lt;bolt|extensionname&gt;[.subnamespace(s)]" (See examples below)

Extensions could then use ``whatever.myextension.foobar``.

At the moment available system events:

+ On saving content:
  - ``start.bolt.content.save``: Before saving content
  - ``done.bolt.content.save``: Content was saved successfully
  - ``fail.bolt.content.save``: Saving content failed
  - ``always.bolt.content.save``: After saving content (failed or succeeded)

+ On saving an edited file:
  - ``start.bolt.file.save``: Before saving file
  - ``done.bolt.file.save``: File was saved successfully
  - ``fail.bolt.file.save``: Saving file failed
  - ``always.bolt.file.save``: After saving file (failed or succeeded)

+ Loading GoogleMaps API:
  - ``start.bolt.googlemapsapi.load``: Request loading API loading
  - ``done.bolt.googlemapsapi.load``: API loaded successfully
  - ``fail.bolt.googlemapsapi.load``: Loading failed

_This replaces ``$(bolt).trigger(…)`` and ``$(bolt).on(…)``, as it was done until now. So if you were already using this, you have to migrate._


bobdenotter:
Looks good at first glance. 

On a scale of 1 - 11, how absolutely sure are we that there's no chance of this introducing new quirks, this close to version 3.0? Has this been tested on different flavors of IE, Safari and Edge? 

Just being cautious here. :-)

rarila:
@bobdenotter Doesn't depend on browsers in any way. Just plain jQuery, no DOM involved at all. Also except ``googlemapsapi`` no events were used sofar. @SahAssar has used this for some of his extensions.

rarila:
Technically I see no problem here.

Thing to check is if the overall process is ok and naming, especially how namespacing is done. This is because changing things afterwards bring us bc problems. So I named that RFC and did not just a plain PR. We have to often introduced things here in the way of, works, does what it is supposed to be, good, merge it. That without thinking about future implications or nned to extend stuff.

So the question is, do I have overseen something, could some things be named better?

rarila:
Ah, we also could name that here a ``beta-API`` (object to change, use at your own risc) let it out in the wild and se if it solves al cases it is designed for or if something is missing.

bobdenotter:
> We have to often introduced things here in the way of, works, does what it is supposed to be, good, merge it.

Yes, agree completely! That's why I haven't pressed the "merge" button yet. :-)

> Ah, we also could name that here a beta-API (object to change, use at your own risc) let it out in the wild and se if it solves al cases it is designed for or if something is missing.

I'm not in favour of that. It this works as it should, we should make sure all of _our_ events use this, instead of half/half.

rarila:
> make sure all of our events use this

No, I did mean that all use this, just don't guarantee 100% BC for now.


Ok, now some interesting investigations convcerning namespaces:

An event ``type.namespace.subnamespace`` turns out to be interpreted  by jQuery as type ``type`` and namespace ``namespace.subnamespace``.

Subscribing for e.g. ``start.bolt.content`` doesn't get triggered by ``start.bolt.content.save``, but interestingly subscribing for e.g. ``start.bolt.content.save`` gets triggered by ``start.bolt.content`` which doesn't seem to be usefull at all by me.

Removing all ``start`` events or subscribing to all ``start`` events doesn't sound useful to me. What would make sense was e.g. subscribing to all ``bolt.content`` events and handle different cases in that handler. So the question is if we want to stay with jQuerys naming system and do (as we have a separated system) do our own and if yes, if we can find a more usefull system that works inside given constraints.

bobdenotter:
Ok, seems good! 

Additionally, i've done some monkey-testing, and i didn't see any regressions! Let's go! :+1:

PR #4928, opened 2016-02-29
[RFC] Change 'info' to 'success'

bobdenotter:
I know it's pretty late in the cycle for 3.0, and this is a significant change. Perhaps we should bump this to 3.1 or even 4.0? ;-)


GawainLynch:
Don't tempt the Travis Gods… :stuck_out_tongue_winking_eye:

PR #4929, opened 2016-02-29
Better check for templatefields in 'editcontent'

bobdenotter:
This is a bit of an edgecase: Let's say you have a field with `group: template`, but the currently selected template doesn't have any actual 'templatefields'. You'll get this nice error:

screen shot 2016-02-29 at 16 28 43

This PR fixes that.


GawainLynch:
Let's do it!

PR #4942, opened 2016-03-02
Version class

CarsonF:
There's no reason to have the current version in DI container since it doesn't change. This just makes us require an app container in places we may not necessarily have it.

The approach everyone else is doing, if they are, is a `Version` class with a constant in it for the current version.


GawainLynch:
:+1:

PR #4946, opened 2016-03-02
Fix translation resources not being applied

CarsonF:
Although translator should not be invoked before boot, there is a chance that it is. In this case not all of the translations resources have been applied to it. 
This moves them to be added when the translator is invoked, so they are added regardless of when that is.


GawainLynch:
:+1:

PR #4947, opened 2016-03-02
Apply cachehash to includes in the backend.

### LINKED ISSUE #4944: When updating Bolt, you're not always getting the 'new' CSS and JS files.
.. Which causes some people to think they've broken it, until they do a "hard refresh". This shouldn't be necessary.

bobdenotter:
Fixes #4944


GawainLynch:
@rossriley does :beer: merges… :koala: does conference merges :stuck_out_tongue_winking_eye:

PR #4950, opened 2016-03-03
Make PHP icon fit, tweak background color for better fit with Bolt.

bobdenotter:


bobdenotter:
Needs some work for smaller screens.. Marking it as [WIP]

bobdenotter:
Should be GTG now!

Issue #4949, opened 2016-03-03
[FEATUREBUG] app.paths[root] unreachable?


GDmac:
Master:

getPath(root) in config/resourcemanager/getPathObject
actually fetches "rootpath" instead of "root"
because, first check is if (array_key_exists($name . 'path', $this->paths)
and will never return app.paths[root], always returns app.paths[rootpath]

I don't know if this is intentional, or how this is supposed to work.
but IMO this is a confusing situation, ask [root] get [rootpath]
and seeing there is a [root] on dump(paths) but is unreachable?

Changing this might have B.C. consequences because
root = '/boltdir/'
rootpath = '/Users/gdmac/Sites/boltdir/'

Needs some thinking over maybe

GDmac:
The same goes for ``{{paths.app}}`` like used in _page.twig, ``/boltdir/app/``
vs ``getPaths('app')`` which is ``/Users/gdmac/Sites/boltdir/app/``
is it a filesystem path (which the name suggests) or a url path, relative to the domainname?

GDmac:
Is this fixed with your PR @CarsonF ?

CarsonF:
Not sure. Just don't use `{{ app.paths }}` or `{{ paths }}`

rossriley:
Yes, I know about this and whilst it's not ideal it's a remnant that we're stuck with for a while.

Before I did the work on ResourceManager to allow the paths and urls to be configurable they were all put in one array with no differentiation between paths and urls, whenever there was a clash then url had first call on the name and then path had the name with `path` suffix. So that's why we have `root` and `rootpath`.

In the `ResourceManager` class this is now separated, so you can do `$app['resources']->getPath('root')` and `$app['resources']->getUrl('root')` and they are stored separately.

However the one carryover is the `paths` array that is passed into twig, which is still used all over the place and requires the paths and urls to be merged.

So the way we need to attack this is deprecate `paths` for urls and have a separate array for `urls` eg `{{ paths.files }} {{ urls.root }} ` etc.

But obviously this will take a major version deprecation cycle so it won't disappear soon.

PR #4959, opened 2016-03-05
Fix assets

CarsonF:
I asked @jkazimir for an intro:
>Ok, so listen up, y’all. Assets can be like the force, or they can be like Gungans. Currently, assets are handled somewhere along the lines of Ewoks. They work well for the goals, but there’s still something obviously off about them.  This PR increased the assets mediclorians, putting things more in balance with the force.  Something something Darth Maul.

----

While @GawainLynch was refactoring the asset code, we talked about implementing Symfony's new [Asset Component](http://symfony.com/doc/current/components/asset/introduction.html). We had decided to wait until 3.1. But seeing how the code settled and with these bugs that have come up we decided it was time.

### Paths
Paths given to the `cachehash` function had to be the web paths. But they needed to be mapped back to the actual filesystem path to do checks with.
Paths were assumed to be under `root`. To handle different locations `file_exists` checks were done. This is what I call "magic" and is bad news. For example, what happens if both locations had the same file? which one wins?
As @GDmac mentions in #4949 trying to access paths in twig is nightmare right now. I am utterly confused about how it works / doesn't work. (Not pointing the finger or anything, I know there was BC to maintain). The sooner we can get away from using the `paths` object in twig, at least in its current state, the better.

Symfony solves both of these problems with `Packages`. Each `Package` has a base path so the path given can be relative. `Packages` are referenced by name, allowing you to declare which one to use. 

From this: 
```twig
{{ app.paths.app }}view/img/bolt-logo.png
or
{{ cachehash(paths.app ~ 'view/img/bolt-logo.png') }}
``` 
To this: 
```twig
{{ asset('img/bolt-logo.png', 'bolt') }}
```
The difference is subtle, but it's important.

### Versioning
With Symfony's Assets, each `Package` has version strategy. This allows individual packages to do versioning their own way. Currently we create a hash from the file's last modified time, a random string, and the file path. But another one could be swapped out like an empty version or a static version (`?3.1.2`).
"cachehash" or "hash" is a poor name because it's really an implementation of a version strategy. `asset` function name is better because it's further abstracted. You can reference an asset but that doesn't mean it has a version attached to it.

I removed  `cacheHash` and `fileName` from `FileAsset`'s interface and replaced it with `path`, `packageName`, and `url`. `path + packageName` are given to asset packages service which returns the `url`. This also removes the business logic from the `FileAsset` objects.

*Note: This does break code that is currently based on master.*

### Extensions
Gawain and I knew there needed to be some distinction between the directory an extension is installed to and how it is referenced from the web. We came up with `DirectoryInterface $baseDirectory` and `string $relativeUrl`. The problem we have run into is we need to do file operations on assets in the web directory. Trying to map the base directory to the relative url or whatever we were doing was a nightmare and only half working I think.

I replaced `string $relativeUrl` with `DirectoryInterface $webDirectory`. This really makes more sense as it is two directories we are working with. This allows us to remove the magic where we are adding assets from the extension, since those directories have been given to the extension from core.

The other problem @rossriley was talking to me about is these are required even though they may not be needed. I changed the directories so setting them is optional, but if they are requested (with getter) and they have not been set an exception is thrown. I want it this way because I don't want to do null checks every time the getter is called. For all "managed" extensions, ones installed with composer or local, the directories are always given. 
So this only affects extensions manually added:
```php
public function add(ExtensionInterface $extension, DirectoryInterface $baseDir = null, DirectoryInterface $webDir = null, $composerName = null)
```

*Note: This does break extensions that reference `relativeUrl`*

---

All in all this is the best solution as we have better separation of concerns, less magic, are closer to Symfony, and don't have to manage BC.

## What's left
- Fixing tests
- ~~Fix copying extension assets on install (may be a different PR)~~

### DIFF
--- a/app/view/twig/_base/_page-no_nav.twig
+++ b/app/view/twig/_base/_page-no_nav.twig
@@ -24,7 +24,7 @@
         <div class="row">
             <div class="{{ form_class|default('col-md-6 col-md-offset-3 col-sm-8 col-sm-offset-2') }}">
 
-                <img src="{{ app.paths.app }}view/img/bolt-logo.png" width="150" height="62" alt="Bolt" class="logo">
+                <img src="{{ asset('img/bolt-logo.png', 'bolt') }}" width="150" height="62" alt="Bolt" class="logo">
 
                 {{ widgets('login_top', 'backend') }}
 
--- a/app/view/twig/_base/_page.twig
+++ b/app/view/twig/_base/_page.twig
@@ -136,7 +136,7 @@
 
 {# Add the Grunt livereload script, if Grunt livereload is enabled in config.yml. #}
 {% if app.config.get('general/grunt/livereload') %}
-    <script src="{{ paths.hosturl }}:{{ app.config.get('general/grunt/livereloadport') }}/livereload.js"></script>
+    <script src="//{{ app.request.host }}:{{ app.config.get('general/grunt/livereloadport') }}/livereload.js"></script>
 {% endif %}
 
 </body>
--- a/app/view/twig/_base/_page.twig
+++ b/app/view/twig/_base/_page.twig
@@ -23,37 +23,37 @@
 
 {# TODO: Refactor this out. Too much business logic in the template. #}
 {% set page_scripts_inc = [
-    'view/js/lib.js'
+    'js/lib.js'
 ] %}
 
 {# Add locale include for datepicker #}
-{% set localepath = 'view/js/locale/datepicker/' %}
-{% if file_exists(paths.apppath ~ '/' ~ localepath ~ page_locale_long ~ '.min.js') %}
+{% set localepath = 'js/locale/datepicker/' %}
+{% if file_exists(paths.viewpath ~ '/' ~ localepath ~ page_locale_long ~ '.min.js') %}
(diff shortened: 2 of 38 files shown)


GawainLynch:
:+1:

rossriley:
Nice work. looking forward to giving this a whirl :+1:

CarsonF:
Ok I've got composer handlers fixed and updated. Really the problem was the web directory being the same as the root directory. The sooner we can change this for good, the better.

@GawainLynch double check me here. But all that's left now is fixing tests.

cdowdy:
Im a lot late to this but there's a Symyfony Asset Bolt Extension if anyone would like to have a look https://github.com/fadoe/bolt-symfony-asset-extension

GawainLynch:
Thanks @cdowdy that's useful for v2 users for sure.

cdowdy:
yeah @GawainLynch I figured it would at least give some insights on how someone else used it within bolt right now... v3 is a different beast though haha

GawainLynch:
Yeah… I was going to implement Symfony Asset 6 months ago when we did the queue work… but someone talked me out of it… I think said person just wanted to do the PR themselves… 6 months later :stuck_out_tongue_winking_eye:

GDmac:
# 💐

PR #4968, opened 2016-03-07
[RFC] Adding official UI Bolt events - 2nd try

rarila:
*As #4927 was merged before actually comments came in here now the next incarnation of events.*

As DOM events didn't seem to fit very good for our events (namespacing, globbing, …) we now use callback lists.

Event types are now build like this:

> namespace.domain.event.status

(Is it ok how they are named?)

So a valid event would be ``bolt.content.save.always``. At the moment characters from ``a-z``, ``A-Z``, ``0-9``, ``-``, and ``_`` are allowed.  So we could do ``Bolt.GoogleMapsAPI.Load.Done`` instead of just ``bolt.googlemapsapi.load.done``. Change to uppercase, restrict to lowercase, … Opinions?

Now also possible is listening to higher level events and filter in the event handler.

```javascript
Bolt.events.on('bolt.content.save', function (event, data) {
    if (event.status === 'start') {
        ...
    }
    if (event.status === 'done') {
        ...
    }
});
```

At the moment the ``event`` object gives you access to ``namespace``, ``domain``, ``event`` and ``status``.

With a second parameter you have access to specific data that can optionally be added when firing.

*Updates:*
- *``-`` added for event types*
- *fix examples*


SahAssar:
:+1:

rarila:
- Switched to mixed case event types
- Tried to make (better) use of ``@event``, ``@fires`` and ``@listens`` for JsDoc, although I think we are misusing it a bit.
- Add JsDoc for events
- ``Bolt.Content.Save.Done`` now fires ``(… , {form: data})`` instead of just ``(… , data)``. That allows adding of other stuff

rarila:
Ok, I think this is :green_apple: :green_book: :green_heart: :frog: :snake: :four_leaf_clover: :cactus: :pear: to go now. Nobody prevents us from having a "3rd try" :-)

PR #4974, opened 2016-03-08
Backbone.js and Underscore.js removed …

### LINKED ISSUE #4970: [RFC] Removing Backbone.js
We still include ``Backbone.js`` in ``bolt.js`` although we don't use it anymore internally. As extensions have to be refactored for 3.0 anyway I guess we should take the chance and get rid of that 22kb.

Comments?!

EDIT:
Same for ``Underscore.js`` which was mostly there because of ``Backbone.js``. We still have one ``_each()`` and one ``_.has()`` in there, but I'm sure they can be pretty easily replaced.

rarila:
… and some smaller bugs in liveeditor fixed. ``lib.js`` on diet! :tada: 

Fixes #4970


SahAssar:
the libs,
the libs,
the libs are on :fire:.
we don't need no :fire_engine: let the libs :fire:

rarila:
I fell in to a coding :ring: of :fire:
Size went down, down, down
And the commits went higher

And it burns, burns, burns
The :ring: of :fire:
The :ring: of :fire:

PR #4976, opened 2016-03-08
Fix commments not removed on minifying

rarila:
After updating grunt-contrib-uglify to v0.10.0 the ``preserveComments: 'some';`` didn't work anymore as upstream went something wrong. The option ``some`` isn't available anymore and no comments were removed with it.

This PR restores that functionality using some regex and ``lib.js`` alone wents down from 474kB to a whopping 375kB :tada: 

\*sing\* :notes: God shave the lib :notes:


SahAssar:
:paperclip: :fire:

PR #4981, opened 2016-03-09
Prevent duplicate content for paging requests

SahAssar:
/blog?page_blog=1 isn't really paging, it's the same as /blog, so only match numbers higher than 1 when checking if the request is paging.


SahAssar:
Related to #4288, and should be forward-ported to master too.

GawainLynch:
> Related to #4288, and should be forward-ported to master too.

You're on that?

SahAssar:
Yep :smile:

Issue #4980, opened 2016-03-09
Inconsistent results for 'homepage: page/1' and 'homepage: pages' settings.


bobdenotter:
If we set `homepage: page/21` in config.yml, and dump 'record' and 'records':

```
<p>Records:</p>
{{ dump(records) }}

<p>Record:</p>
{{ dump(record) }}
```

We'll get this result: 

![screen shot 2016-03-09 at 12 17 05](https://cloud.githubusercontent.com/assets/1833361/13633724/21d3c98c-e5f1-11e5-97fe-58670336ce1d.png)

If we set `homepage: pages`, we'll get the following result: 

![screen shot 2016-03-09 at 12 16 03](https://cloud.githubusercontent.com/assets/1833361/13633722/1d91b5dc-e5f1-11e5-91ff-92435f8879c9.png)

This is inconsistent, and i'd classify this as a bug. I'd like to make it that in either situation, only 'record' _**or**_ 'records' should have data assigned to it.

SahAssar:
I *knew* we had talked about this! I should have opened an issue about it back then :smile: 
```
[2016-01-20 08:33:23] <SahAssar> Question: Aren't we supposed to make sure that $content is an array here? https://github.com/bolt/bolt/blob/master/src/Controller/Frontend.php#L101
[2016-01-20 08:34:03] <SahAssar> Seems odd that we would set $content to records without checking if it's an array.
[2016-01-20 08:35:02] <Bopp> It being `false` is handled in the template.
[2016-01-20 08:35:33] <Bopp> go to /pages/ -> "No pages were found"
[2016-01-20 08:38:38] <SahAssar> Bopp: Yeah, I just came across this because of some cleanup in base-2016. In index.twig you can have both listings and only a single record, and having it set to both complicates checking if one is on a listing or not.
[2016-01-20 08:39:22] <SahAssar> Basically I thought I'd do {% if records is not defined and record is defined %}{% set records = [record] %}{% endif %}, but that does not work in master, only in 2.2
[2016-01-20 08:43:15] <Bopp> SahAssar: Is that still the case with an up-to-date master?
[2016-01-20 08:44:36] <SahAssar> Bopp: yep, I pulled half an hour ago
[2016-01-20 08:45:30] <Bopp> ok.
[2016-01-20 08:46:05] <gawainlynch> SahAssar: Sorry… getting pulled a few ways here…
[...]

PR #4983, opened 2016-03-09
Add 'single record' as an array to 'records'.

bobdenotter:
This should fix the reported issue, without causing side-effects. This way `records` always contains an actual "array of records", instead of just 'one record'. See screenshot for what `homepage: pages/21` gets you now: 

screen shot 2016-03-09 at 12 55 14

Fixes: #4980


GawainLynch:
Fixes #4980

PR #4985, opened 2016-03-09
Access control events

GawainLynch:
This is a follow up to #4883 & #4878 for master. Late, but low impact.

@CarsonF your concerns from #4883 should have been adequately addressed. As for Symfony Security, that's targeted for Bolt v4.

Ping @ToBe998

### DIFF
--- a/src/AccessControl/Login.php
+++ b/src/AccessControl/Login.php
@@ -2,6 +2,8 @@
 namespace Bolt\AccessControl;
 
 use Bolt\AccessControl\Token\Token;
+use Bolt\Events\AccessControlEvent;
+use Bolt\Events\AccessControlEvents;
 use Bolt\Exception\AccessControlException;
 use Bolt\Storage\Entity;
 use Bolt\Translation\Translator as Trans;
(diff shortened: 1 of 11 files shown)


ToBe998:
Looks great for me, I'll try to get a more thorough check on my use case asap. Tnx

GawainLynch:
> Can we get a failure reason in there?

Yeah, that actually sounds like a good call :+1:

GawainLynch:
@ToBe998: Update mate, can you have another look for me.

Note to merger/reviewer(s)… this looks a bit larger than it is due to PhpStorm "cleaning up" :wink:

ToBe998:
Ah, stumbled about users with empty pw counted as disabled, but that's absolutly ok. I think it's looking good. I will need some days to try my use case, but as far as i see, everything i nwould need is there.

Something that puzzled me a bit however:
You defined your failure reasons to work bitwise, but didnt stack them in the setReason method. Not a feature i would need, but if you set them up as bits, why not also add them together in there? Or just use normal incremental constant?

GawainLynch:
I've created a proof-of-concept [audit logging extension](https://github.com/bolt/Audit) that sends these to syslog.

PR #4973, opened 2016-03-08
Imagehandler updates and bugfixes

SahAssar:
Standardizes the params for all the image functions and makes sure that all the functions uses the config.yml defaults where appropriate. Also some minor restructuring of popup and showimage

fixes #4964

PR #4986, opened 2016-03-09
[master] Fix image defaults/config

SahAssar:
Forward-port of #4973, I've tested that it produces the same results as 2.2.X does now. Ping @CarsonF and @GawainLynch for checking :smile:

### DIFF
--- a/src/Twig/Handler/ImageHandler.php
+++ b/src/Twig/Handler/ImageHandler.php
@@ -97,7 +97,7 @@ class ImageHandler
      *
      * @return string HTML output
      */
-    public function popup($fileName = null, $width = 100, $height = 100, $crop = null, $title = null)
+    public function popup($fileName, $width = null, $height = null, $crop = null, $title = null)
     {
         if (empty($fileName)) {
             return '&nbsp;';
@@ -105,7 +105,7 @@ class ImageHandler
 
         $thumbconf = $this->app['config']->get('general/thumbnails');
         $fullwidth = !empty($thumbconf['default_image'][0]) ? $thumbconf['default_image'][0] : 1000;
-        $fullheight = !empty($thumbconf['default_image'][1]) ? $thumbconf['default_image'][1] : 800;
+        $fullheight = !empty($thumbconf['default_image'][1]) ? $thumbconf['default_image'][1] : 750;
 
         $thumb = $this->getThumbnail($fileName, $width, $height, $crop);
         $largeThumb = $this->getThumbnail($fileName, $fullwidth, $fullheight, 'r');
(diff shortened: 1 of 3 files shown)


SahAssar:
Alright, this is going to need some tests updating that I didn't know about when I opened it...

CarsonF:
Isn't this less intuitive?

SahAssar:
CarsonF: It makes the defaults actually be defaults, instead of them being hardcoded. I don't think that will be less intuitive. For example, the defaults for the a thumbnail are the defaults that have been set, unless you use a popup in which case the defaults are 100x100, which can't be changed. Same goes for cropping on thumbnails. We have a default config so we should actually use it, right?

src/Twig/Handler/ImageHandler.php
```diff
@@ -155,12 +155,20 @@ public function showImage($fileName = null, $width = null, $height = null, $crop
         }
         $thumb = $this->getThumbnail($fileName, $width, $height, $crop);
 
-        if ($width === null || $height === null) {
+        if ($width === null && $height === null) {
+            $thumbconf = $this->app['config']->get('general/thumbnails');
+            $width = !empty($thumbconf['default_image'][0]) ? $thumbconf['default_image'][0] : 1000;
+            $height = !empty($thumbconf['default_image'][1]) ? $thumbconf['default_image'][1] : 750;
+            $thumb->setWidth($width);
+            $thumb->setHeight($height);
```
GawainLynch: I might be missing something here, but isn't this already done in `\Bolt\Helpers\Image\Thumbnail`? 

The point of that class was to abstract, and centralise all of this :wink:

SahAssar: Yeah, but there it assumes that you want the default_thumbnail defaults, here we want the default_image defaults. This is basically the same thing that we already do in the `popup` function.

GawainLynch: Well, yes. That would be why the function is called `showImage()`… :koala:

CarsonF:
@SahAssar I meant having html in the yaml config :/

SahAssar:
@CarsonF Not sure what you mean? There is no html in the yaml AFAIK... Also, this didn't add anything to the config, it only makes the image functions use the defaults set if one uses showimage etc. with no values set.

CarsonF:
@SahAssar Nevermind I'm a derp. I misread your code block on the other issue as yaml instead of results. :+1: :koala:

GawainLynch:
Always the damn :koala: :wink:

SahAssar:
@CarsonF Ah, I probably shouldn't have had the config and the results in the same block :laughing:

PR #4989, opened 2016-03-10
Don't return suffix in Composer version

### LINKED ISSUE #4972: v2 extensions showing up on v3 extend page
image

This is a result of #4942 and IMHO we should only have the `x.y.z` part of the version string.

GawainLynch:
Fixes #4972


SahAssar:
:musical_score:

PR #4998, opened 2016-03-11
Configurable bootstrap

GawainLynch:
Another major release… another re-do of `index.php` and `app/bootstrap.php`… where would we be without them ;-)

This might be a little contentious for some, but we're moving to a Composer based archive install for v3, and want to keep bootstrapping for the average user to catch problems.

This PR changes  `app/bootstrap.php` to return a closure instead of an `Application` object. The closure when called accepts an array of paths and then returns the `Application` object, e.g.:

```php
$bootstrap = require_once '../vendor/bolt/bolt/app/bootstrap.php';
$app = $bootstrap([
    'themebase' => 'public/theme',
    'web'       => 'public',
    'files'     => 'public/files',
    'view'      => 'public/bolt-public/view',
]);

$app->run();
```

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -65,7 +65,6 @@ return function (array $paths = []) {
 
     /** @var \Bolt\Configuration\ResourceManager $config */
     $config->verify();
-    $config->compat();
 
     // Create the 'Bolt application'
     $app = new Application(['resources' => $config]);
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -1,66 +1,77 @@
 <?php
+
 namespace Bolt;
 
 use Bolt\Exception\LowlevelException;
 
 /**
- * Second stage loader. Here we bootstrap the app:
+ * Second stage loader. Do bootstrapping within a new local scope to avoid
+ * polluting the global space.
  *
+ * Here we bootstrap the app:
  * - Initialize mb functions for UTF-8
  * - Figure out path structure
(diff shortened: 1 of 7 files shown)


rossriley:
While you're PRing this we may as well kill the call to `config->compat()` too. That is one piece of legacy that is well and truly destroyed now.

GawainLynch:
On it, thanks for the heads up!

bobdenotter:
Whoop!

CarsonF:
Are we assuming composer installs create their own `index.php` and `nut` files? And if so are the paths config copied for both those files? Or are you thinking each install makes it's own `bootstrap.php` file that pulls from bolt's `bootstrap.php` file?

GawainLynch:
Composer installs have historically not used the bootstrap file at all… we've assumed that people using this method already read the part that said "PHP X.Y.Z+" etc, but for general distribution we want to keep checks in place.

This doesn't change the fact that, as @rossriley put it "I start to twitch if my bootstrap is more than 2 lines", people can and will want to skip/remove all this. It's really about giving git/archive installs a consistent experience and the archive installs provide a simple (and more secure) gateway to using Composer installs.

If you're interested, at this second this is what the [packaging scrpt](https://github.com/bolt/bolt-distribution/blob/bolt-3/package.sh) does based on a ["deafult" layout](https://github.com/bolt/bolt-distribution/tree/bolt-3/staging).

PR #5002, opened 2016-03-11
Invalidate user sessions on disable/delete

### LINKED ISSUE #4965: Users don't get logged off, when their account is revoked..
When a user get disabled or their accounts gets removed, their current sessions should be killed. This isn't happening right now. 

I made a short screencast: http://recordit.co/jTqRKVbWV9

Details
-------

 - Relevant Bolt Version: [ master ]

Reproduction
------------

 - Add a user
 - Log on as that user in a new incognito window.
 - Disable or delete that user. 
 - Toggle back to incognito window, break stuff.

GawainLynch:
Fixes #4965

### DIFF
--- a/src/EventListener/AccessControlListener.php
+++ b/src/EventListener/AccessControlListener.php
@@ -41,6 +41,25 @@ class AccessControlListener implements EventSubscriberInterface
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bobdenotter:
Looks good to me! :+1:

Issue #4975, opened 2016-03-08
We need "Move up" and "move down" buttons for repeaters. 


bobdenotter:
From tonight's IRC meeting: 

 - *rossriley*: it would be nice if a UI person could get move up/down implemented for repeaters
though not a blocker if noone has time
 - *Bopp*: rossriley: will re-ordering them in the HTML do? Or does it need changes under water?
 - *rossriley*: reorder in html and then call the reindex js function

rossriley:
For reference the js that controls it is in `app/src/js/widgets/fields/fieldRepeater.js` and after moving the html it just needs a call to `self._renumber();`

EJanuszewski:
I could potentially look into this, i'll look in a couple of days when i'm back home

bobdenotter:
@EJanuszewski I initially 'self assigned' this, because _someone_ has to do it, but if you have time, you are more than welcome to tackle this one! :+1:

bobdenotter:
Fixed by #5005

PR #5005, opened 2016-03-11
Repeater move buttons

EJanuszewski:
Adds Move Up and Move Down buttons for repeaters, also takes into account if it's the first/last item and disables the relevant buttons.

Fixes: #4975

### DIFF
--- a/app/src/js/widgets/fields/fieldRepeater.js
+++ b/app/src/js/widgets/fields/fieldRepeater.js
@@ -103,6 +103,20 @@
                 self._renumber();
             });
 
+            self.element.on('click', '.move-up', function () {
+                var setToMove = $(this).closest('.repeater-group');
+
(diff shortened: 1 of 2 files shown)


rarila:
@EJanuszewski Haven't seen it live yet, but I think the arrows should have their own ``btn-group``. Probably just having the arrows without Text will do?

EJanuszewski:
@rarila Yeah I did think about doing it in it's own group, @bobdenotter and @GawainLynch suggested with text though, you'll have to fight it out with them :D

rarila:
Then that way. Can be changed later…

bobdenotter:
Hey, ho, let's go! :+1: 

I'm always a fan of getting stuff working, and then we can always improve it later, after some real-world feedback.

Issue #5007, opened 2016-03-11
[BUG] newly created record doesn't get a publish date


Boorj:
Create a record. It gets `'publish'` status, but to publish date. Is it ok?
![2016-03-12 00-27-40](https://cloud.githubusercontent.com/assets/2145313/13711715/3f8924d6-e7ea-11e5-894f-c341fb73ba85.png)

It causes sorting problems and.. seems like it's logically incorrect.

bobdenotter:
It should get a `datepublish` date, absolutely. Not a `datedepublish`, though. :-)

GawainLynch:
Fixed in #5011

Boorj:
Bopp indeed, my mistake )) was angry ande defocused that night)

GawainLynch:
I know that feeling mate, I know that feeling :grin:

PR #5011, opened 2016-03-12
[WIP] Add default date publish

EJanuszewski:
Fixes: #5007 

Add a default publish date in _meta.twig, @GawainLynch is looking into whether we need to change the backend controllers or not for when the status is transitioned.


GawainLynch:
Can't find… :koala: might be eating :leaves:

PR #5013, opened 2016-03-13
Don't filter on empty strings

SahAssar:
By default when filtering on a taxonomy you get a query string like this: "?taxonomy-chapters=main&filter=" which then makes bolt "filter" the content with a empty string, which means that you get this:
skarmklipp 2016-03-13 10 32 38

### DIFF
--- a/src/Storage/ContentRequest/ListingOptions.php
+++ b/src/Storage/ContentRequest/ListingOptions.php
@@ -111,7 +111,7 @@ class ListingOptions
      */
     public function setFilter($filter)
     {
-        $this->filter = $filter ?: null;
+        $this->filter = $filter === '' ? null : $filter;
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


src/Controller/Backend/Records.php
```diff
@@ -116,10 +116,15 @@ public function overview(Request $request, $contenttypeslug)
...
+        $filter = null;
+        if ($request->query->get('filter') !== "") {
+            $filter = $request->query->get('filter');
+        }
+
         $options = (new ListingOptions())
             ->setOrder($request->query->get('order'))
             ->setPage($request->query->get('page_' . $contenttypeslug))
-            ->setFilter($request->query->get('filter'))
+            ->setFilter($filter)
```
GawainLynch: This should be handled in the `ListingOptions` setter.

```
    /**
     * Set the filter.
     *
     * @param string|null $filter
     *
     * @return ListingOptions
     */
    public function setFilter($filter)
    {
        $this->filter = $filter ?: null;

        return $this;
    }
```

GawainLynch:
OK… out voted it would seem, no <3 in the code base… yet! :wink:

Issue #3703, opened 2015-06-17
[RFC] New way of thumbnails generation


Boorj:
### Intro

Current way of thumbnails generation has several disadvantages
**1**. it's insecure to let any visitor to create any image by request. If you have /images/1.jpg, server space can be filled with requesting for 

> /thumbs/**1x1**/1.jpg
> /thumbs/**1x2**/1.jpg
> /thumbs/**1x3**/1.jpg

**2**.  If you have several implementations of your thumbnail size in twig template and willing to change resolution of the thumbs, you have to change this resolution in each place.

> {{ record.photo|showimage(800, 600) }}
> // 800 and 600 has to be changed everywhere

**3**. If you want to do some other transformations (like black and white processing or crop) - you also have to change it.

**4**. Case: you have 100mb serverspace free. Each thumbnail is about 50kb, when you have 20000 thumbnails generated, server space is busy.
Without a thumbnail generation tool (like cache:clear) or nut command - it could happen unexpectedly.

**5**. Folders for thumbnails are heavily hardcoded. It would be have flexibility to setup naming thumbnail folders.

**6**. Case: i want to let user upload huge images, but don't want them to be available in web folder. Only resized versions or versions with some watermarking. How can i restrict access to original images? i think it could be achieved with **5** by setting folder for uploads outside the web wolder.
### Proposal:

Make a thumbnail (resize , process) config  for each image/imagelist contenttype.

``` yaml
#contenttypes:
user_contenttype:
    photo:
        label: "It's a photo, dude"
        type: image
        resizes:
            resize1:
                size: [320,200] # other setting are set by default
            preview:
                size: [800,600]
                upscale: false
                format: 'jpg'
                type: 'stretch' # 'fit' / 'fill' / 'crop'
            thumbnail_renamed_and_processed: 
                size: [200,166]
                filename: "%date%_%dimensions%.%original_extension%"
                folder: "%thumbs_folder%/carrots" # relatively to %web_folder%
```

and so it will be used like

``` twig
{{ record.photo|thumb('resize1')}}
or dynamic filter generation (just a fantasy :)
{{ record.photo|preview}}
or perhaps
{{ record.photo__thumbnail_renamed_and_processed}}
```
### Backwards Compatibility

I think for a while it could be used in special mode in config, that is set by default to old style of thumbnail generation. And if you want to use the new style - just add 

``` yaml
#config.yml
new_thumbs_or_whatever: true
```

rarila:
[...]

PR #5019, opened 2016-03-14
Thumb presets

GDmac:
Some default thumb sizes presets, 
in anticipation of the thumbs update to restrict allowed sizes as discussed in #3703 

- Presets are commented out by default
- When no presets found by thumbs,  
  it falls back to allow all arbitrary sizes (for Backward Compatibility)

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -196,6 +196,10 @@ thumbnails:
     allow_upscale: false
     exif_orientation: true
 #    browser_cache_time: 2592000
+#    presets:
+#      default : [  160,  120 ]
+#      image   : [ 1000,  750 ]
+#      preview : [  300,    0 ]
 
 # Define the HTML tags and attributes that are allowed in 'cleaned' HTML. This
 # is used for sanitizing HTML, to make sure there are no undesirable elements


bobdenotter:
I think I agree with this approach. I do think however that we should leave this be, until after we've lifted the feature-freeze for 3.0, right? So, merge it in, right after the release of 3.0.0.

GDmac:
It could also work the other way around `` {{ imagefield|thumb('presetname') }}`` 
when we implement it in the twig image handler

bobdenotter:
Related discussion, see: #3703

GDmac:
I guess we should change the preset format to allow for future addition of other features like crop, override other settings and future-ready watermark settings as suggested by the original proposal

```yaml
  presets:
    default:
      size: [160,120]
      type: 'stretch' # 'fit' / 'fill' / 'crop'
      upscale: false
      format: 'jpg'
    # or simple sizes only
    image: { size : [ 1000,  750 ] }
    preview: { size : [ 300,  0 ] }
```

CarsonF:
Nothing has been agreed upon in #3703 and it's not going to go into `3.0`. Additionally, if we are adding values in `config.yml`, then they also need to be added in `Bolt\Config::getDefaults` method.

bobdenotter:
While I agree that this should _not_ go into 3.0, I _do_ think this could be valid for - perhaps - 3.1, after some additional work.

GDmac:
"Nothing has been agreed upon in #3703" ... that is a somewhat blunt statement.
I think everyone agrees that runaway thumbnail creation needs to be addressed,
as well in the 2.x (LTS) as in the 3.x branch.

GawainLynch:
> "Nothing has been agreed upon in #3703" ... that is a somewhat blunt statement.
> I think everyone agrees that runaway thumbnail creation needs to be addressed,

Yes, but it **can be** (and IMHO should be) addressed at the web server level, as the routes are known to the site owner. 

Doing this at the application level **is** a feature, a feature that that has **some** consensus on its validity. 

The 2.2 branch has been agreed upon to be feature locked as it was turning into a run-away train. With respect to 3.0… same deal… we need to get it out the door.

bobdenotter:
Re-opening this one. It's a valid issue. Two caveats: 

 - This will *not* go into 2.2, or 3.0. It will go into 3.1 or after.
 - accompanying changes to bolt/bolt-thumbs need to be done first, and those will go according to procedure: RFC, Consensus on the way to do them and finally a PR with the changes _with_ accompanying tests.

GawainLynch:
I willing to go with consensus in terms of adding the feature to both core and `bolt/thumbs`, but shouldn't we agree on the feature's implementation into the upstream library before PRing an unknown implementation?

GDmac:
Closing as the issue and a better described yaml formatting is in the pull request on thumbs here:
https://github.com/bolt/bolt-thumbs/pull/28

PR #5022, opened 2016-03-14
Multiple fixes to related contenttype listings

### LINKED ISSUE #5000: Select All in overview buggy
When using the "select all" in an overview like so:
image

The result is that only the first item is selected:
image

Clearing and reselecting individually is also borked
image

### LINKED ISSUE #5008: [BUG] Extra </tbody> in listing
Looks like in listings (all listings) there's something redundant:

Probably the problem is here

Does it need investigation?

rarila:
First removes the if clause (introduced [here](https://github.com/bolt/bolt/commit/2063fcdc5cb3251386387286e82ef817003388ef)) from the for loop as twig has unpredictable behavior on the loop variable then. That finally is a correct fix for #4901 as that one introduced an almost random generation of ``<tbody>``s which then caused #5000 and #5008 .

Next fix is to decide on the filtered list if the listing should be output and not on the full unfiltered list.

Additional some business logic was moved from twig to php and another bug was fixed that wasn't noticed sofar: Custom listings can have a different column count so it doesn't work to list them in them on in table. Also UIX wise it makes sense to have them in distinct tables.

fixes #4901 
fixes #5000 
fixes #5008

### DIFF
--- a/src/Storage/ContentRequest/Edit.php
+++ b/src/Storage/ContentRequest/Edit.php
@@ -135,7 +135,7 @@ class Edit
             'change_ownership'   => $this->users->isAllowed('contenttype:' . $contentTypeSlug . ':change-ownership:' . $content->getId()),
         ];
         $contextHas = [
-            'incoming_relations' => count($content->getRelation()->incoming($content)),
+            'incoming_relations' => count($incomingNotInverted) > 0,
             'relations'          => isset($contentType['relations']),
             'tabs'               => $contentType['groups'] !== false,
             'taxonomy'           => isset($contentType['taxonomy']),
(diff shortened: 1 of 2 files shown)


src/Storage/ContentRequest/Edit.php
```diff
@@ -108,6 +108,18 @@ public function action(Content $content, array $contentType, $duplicate)
...
+        foreach ($content->getRelation()->incoming($content) as $relation) {
+            if (!$relation->isInverted()) {
+                $record = $this->em->getContent($relation['from_contenttype'] . '/' . $relation['from_id']);
+
+                if ($record) {
+                    $incomingNotInverted[$relation['from_contenttype']][] = $record;
+                }
+            }
+        }
+
```
GawainLynch: The ArrayAccess is for BC and really should be avoided internally (not to mention the nesting)… Try:
```php
        // Build list of incoming non inverted related records.
        $incomingNotInverted = [];
        /** @var Relations $relation */
        foreach ($content->getRelation()->incoming($content) as $relation) {
            if ($relation->isInverted()) {
                continue;
            }
            $fromContentType = $relation->getFromContenttype();
            $record = $this->em->getContent($fromContentType . '/' . $relation->getFromId());
            if ($record) {
                $incomingNotInverted[$fromContentType][] = $record;
            }
        }
```

PR #5024, opened 2016-03-14
Extensions install SSL downgrade

### LINKED ISSUE #5017: [RFC] Your configuration does not allow connection to http://extensions.bolt.cm
I have a feeling that the following will cause a _lot_ of support demand:

> Your configuration does not allow connection to http://extensions.bolt.cm. See 
> https://getcomposer.org/doc/06-config.md#secure-http for details.
> File: /vendor/composer/composer/src/Composer/Util/RemoteFilesystem.php::271

What would be the best way to deal with this?

GawainLynch:
Fixes #5017


bobdenotter:
Works like a charm :sparkles:

PR #5026, opened 2016-03-14
String not complete

horus68:
String value was incomplete - see line 438:
We were unable to generate the theme


bobdenotter:
Hi @horus68 

Good catch, thank you! :+1:

PR #5030, opened 2016-03-16
Crowdin translations 1

horus68:
First Crowdin translations download. Old folders deleted, now with standard language
codes for all.
Translation workflow proposal:
- translation should be done only on Crowdin
- Then downloaded from there to Github (all at same time by a dev)
Project https://crowdin.com/project/bolt


bobdenotter:
I think this is the way to go in the near future, but we are currently in feature freeze for Bolt 3.0. 

Let's do a thourough investigation of the process for Bolt 3.1 or at another point after Bolt 3.0.0 is released.

horus68:
Translations shouldn't impact on core files. There are no changes on English files. 
Also those language files work on v2. 
You shouldn't release a v3 with languages folders (as this will be a major version) that don't follow the same naming convention (as the v2 is).
So at least change the actual folder names to xx_XX language code!

My arguments on a change now:
- Almost all languages lacks strings 
- Almost all languages don't have all language files included (as you can see on master git).
- Major versions tend to attract new users and new installations 
- If the folder names are only changed for 3.1 then people will loose configuration done for the language code they are using on the config file for their directive on ` locale: xx_XX`
- People tend to not delete folders on minor updates (from x.0 to x.1) so they will end up with different language folders for the same language

A simple test from other user could say go now or wait to 3.1 
...and it could a major feature for non-english users!

rarila:
I also think we have to discuss and find a workflow first. There for sure have to be many changes how translations are handled in the next release. Just a few things:

- Theme translations have to be extracted into its own file(s). (and only backend or frontend ones should be loaded as needed)
- The switch to keybased messages should be done (easier done with the files in master being the master data and mass edit, I guess)
- Change the translate function to
  - better handle raw and escaped strings
  - havee some rules for placeholder naming (and switching all translations to use them)
  - use Symfony pluralization
- Having translations in /config/translations with /resources/translations being dist and fallback
- Find a way how themes and extensions store their translations (and how comes Crowdin into play here)
- …

Lots of work in front us and quite some stuff is probably easier to handle with a mass edit (search and replace) which needs either a way to sync both ways or have the repo as master data.

horus68:
I see 2 directions here:
Solve things to make them work now with minimum impact on code and 
I see in this PR and also in the PR  Regenerated translation files by rarila · Pull Request #5031 · bolt/bolt - https://github.com/bolt/bolt/pull/5031

The other direction involves refactor the code and that probably will be left for a version 4 as it will involve to change behaviours on extensions and themes.
And you all know that 3.1 will be just for bugs, 3.2 some fine tunes on code and then 3.3 for security issues and then 3.5 for bugs and so on!

For me is like: Include now just the small changes that should have been in a v2 and that can be in v3.0 with no problem or else it will be only for v4!

GawainLynch:
> The other direction involves refactor the code and that probably will be left for a version 4 as it will involve to change behaviours on extensions and themes.

No, our translations currently don't cover extensions or front-end themes.

> And you all know that 3.1 will be just for bugs, 3.2 some fine tunes on code and then 3.3 for security issues and then 3.5 for bugs and so on!

3.0.1+ will be for bugs, as will 3.1.1+, and 3.2.1+. They are patch releases.

3.1 & 3.2, etc are minor releases and will allow non-BC-breaking feature additions. So there is no reason that we can't work on the translation layer during that time.

CarsonF:
I'm not convinced Crowdin is the right solution over Transifex.

horus68:
@CarsonF What are your issues/advantages you see on both platforms?

I love Transifex. Also love Crowdin. I'm main user and project admin on both (more time on Transifex but also on several other platforms).
Transifex can't work with .YML files from Bolt (do not upload any line).
Crowdin works fine for upload and download.
So you need to contact Transifex to sort things out, create a project and test things.
When it works then we can see what to use then!

GawainLynch:
Closing this one… Let's open a fresh one when we have keywords/etc sorted out and a platform/process agreed upon.

PR #5031, opened 2016-03-16
Regenerated translation files

rarila:
- Many fixes to how html code is used
- Use typografically correct apostrophes on ``en_GB``
- Use typografically correct quotation marks on ``en_GB`` contenttypes
- Switch to non nested output for keybased translations for easier mass editing
- Removing of some unused translations
- Added some missing translations

Another regeneration needed needed, but I will add a nut command to do this for all languages with one keypress.

PR #5032, opened 2016-03-16
Portuguese language split

horus68:
Actual "pt" language renamed to pt_BR (Brazil)

Added: new language files for pt_PT (Portugal)


rarila:
Ah, great! Thanks!

But I’m afraid that we get merge problems after a possible merge of #5031
So I suggest to wait as I believe it’s much easier to redo this one as solving the merge conflict of the other then. What do you think?

horus68:
No problem to me... I also have those regenerated files if you need!
On Crowdin you can download both!

horus68:
A new language that can be uploaded now:
[RFC] Language folders should be named as xx_XX · Issue #5038 · bolt/bolt - https://github.com/bolt/bolt/issues/5038

GawainLynch:
Could you please resubmit with the adjusted strings from #5031?

PR #5035, opened 2016-03-17
Thumbnails save files fix and caching switch

GDmac:
Fixes: https://github.com/bolt/bolt-thumbs/issues/25

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -161,6 +161,7 @@ caching:
     request: false
     duration: 10
     authenticated: false
+    thumbnails: true
 
 # Set 'enabled' to 'true' to log all content changes in the database.
 #
--- a/src/Provider/ThumbnailsServiceProvider.php
+++ b/src/Provider/ThumbnailsServiceProvider.php
@@ -30,14 +30,27 @@ class ThumbnailsServiceProvider implements ServiceProviderInterface
             'files',
             'themes',
         ];
+
+        $app['thumbnails.save_files'] = $app['config']->get('general/thumbnails/save_files');
+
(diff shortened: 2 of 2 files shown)


bobdenotter:
Hi, @GDmac,

A few remarks: 

The first part, with `if ($app['thumbnails.save_files'] === false) {`: Top notch. :+1: 

The second part: `if ($app['thumbnails.caching'] === false) {` is introducing _new behaviour_, and we're deep into feature freeze for Bolt 3.0. We'd be willing to merge that in, but it seems like it falls back to 'false' if it's not in the current config (like on existing sites). Please change that to `true` by default, and I'm going to :+1:  this. :-)

GDmac:
No, it does not fall back to false (triple compare). 
When not set, the value is ``null`` which not === false.
i have manually verified this before commit and again just now.

bobdenotter:
> No, it does not fall back to false (triple compare). 
> When not set, the value is null which not === false.

You are correct! Thank you, for taking our detailed concerns into consideration. ;-)

> How would that check for true have to work when the setting is not set? have an example?

There's a section where the config is loaded. There we can set _defaults_ for values if they aren't set already. In this case not needed, because of the `===`.

CarsonF:
@GDmac What's your use case for disabling caching here?

bobdenotter:
@CarsonF If save_files is `true`, images are "cached" in two locations. While I personally never had an issue with this, I can see how it can be annoying for people who want a higher degree of control over what happens. 

I discussed this with Gawain,btw. He and I agreed to get this in, as long as it 1) fixed the bug and 2) didn't change default behaviour.

PR #5045, opened 2016-03-18
Fix our Cache implementation to follow Doctrine's interface.

CarsonF:
There was a miscommunication between @GawainLynch and I with his last changes to `Bolt\Cache`.
This corrects it by using `flushAll()` from Doctrine's Cache interface, instead of making `doFlush()` public and using it which isn't any better than the deprecated `clearCache()` method. Now the cache implementation can replaced with any Doctrine Cache implementation.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -4,7 +4,7 @@ namespace Bolt;
 
 use Bolt\Filesystem\AggregateFilesystemInterface;
 use Bolt\Filesystem\Exception\IOException;
-use Bolt\Filesystem\FilesystemInterface;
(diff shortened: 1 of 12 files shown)


GawainLynch:
That root cause of test failure is why there was two calls to the finder… there is a lag in the updating of the recursive iterator, or maybe a race… I never tracked it down.

But if you use the Extend page and then try to flush the cache without that approach, you get 
image

For reference the (incorrect) approach of using `->sortByType()` on the finder call does obviously mitigate the problem… so choose your poison if you want this in 3.0.0 as it lands today or goes to 3.1.

GawainLynch:
Issue fixed upstream.

PR #5046, opened 2016-03-19
Only set allowSslDowngrade on a ComposerRepository object

GawainLynch:


bobdenotter:
Looks good to me! :+1:

GawainLynch:
It's corner case, but I hit it with a custom (non-Composer) repo in my `composer.json` and the extensions site set to `http://`

PR #5050, opened 2016-03-20
[Tests] Rebuild extension autoloader after clean up

GawainLynch:
This will just prevent some noise after an acceptance test is run when you have display errors turned up to 11


bobdenotter:
Looks good to me! :+1:

PR #5051, opened 2016-03-20
V2 boltgun cleanup

GawainLynch:
This just cleans up some deprecated extension calls & tests that are only going to cause problems.

Discussed/agreed-upon with @rossriley & @PhillippOhlandt


bobdenotter:
:gun:

Issue #5036, opened 2016-03-17
[BUG] Cannot delete 'last remaining' related contenttype


anketwokings:
Steps to reproduce:
- with 2 related items in same contenttype; delete 1 of those: No problem
- last remaining item: delete this -> Does not delete from DB

This happens across the related contenttypes, so the last one of my 'pressreleases', of my 'contacts' etc. You cannot seem to delete the last one of each.

Details
-------

 - Relevant Bolt Version: [  3.0, a 8 ]

anketwokings:
Great! Thanks!

Issue #5037, opened 2016-03-18
[BUG] cannot delete last remaining tag


jadwigo:
Steps to reproduce:

A record with 2 or more tags, delete all tags but 1: No problem
Try to remove all tags -> Does not delete from DB

You cannot seem to delete the last one.

This is similar to https://github.com/bolt/bolt/issues/5036

Details

Relevant Bolt Version: [ 3.0, a 8 ]

PR #5052, opened 2016-03-20
Fix Deleting last relations and taxonomies

rossriley:
This fixes the issue where the save wasn't being called on the field if an empty post was sent for that field.

There's also a fix for another edge case where large contenttypes (with lots of relations, taxonomies and repeaters) was leading to an overflow of the maximum select column size which on MySQL seems to be 1024 characters. 

Adding distinct to the query, rather than just doing it in the code afterwards will stop this length from getting as big.

Fixes: #5036 #5037


GawainLynch:
Thanks mate!

PR #5057, opened 2016-03-20
Translations correction

rarila:
Various fixups:

- Only use translatables in contenttypes.yml.dist for generation
- Removal of unused messages
- Untranslated messages set to "translated"
- Fix of an test that had checked for an removed key
- Fix detection of key based translations which didn't detect keys with just two parts
- Adding translations for block contenttype


bobdenotter:
Looks good to me! :+1:

PR #5058, opened 2016-03-21
Simplify adding extension repositoiries

GawainLynch:
This simplifies the process for advanced extensions.

### DIFF
--- a/src/Extension/StorageTrait.php
+++ b/src/Extension/StorageTrait.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Bolt\Extension;
+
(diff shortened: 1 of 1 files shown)


bobdenotter:
Merging in! :+1:

CarsonF:
@GawainLynch You're invoking these services instead of extending them...

GawainLynch:
I know… it didn't work on extend and @rossriley was unsure why… this just keeps the implementation in core rather than documenting a way to do it that needs code out of our hands updated.

TBH the process here needs love, I just don't want to put this in the docs

PR #5061, opened 2016-03-21
Portuguese Brazil language

horus68:
Portuguese language for pt_BR (Brazil)
Include: Some untranslated keyword based messages that where ignored by #5031 regenerated files because on v2 translation there was a wrong # in front of the translated value

Just for the records example of the "losted" strings from v2, and now fixed
`"contenttypes.entries.text.new": # "Novo Post"`


GawainLynch:
ping @rarila

rarila:
That's not wrong ``#``, that's a hint in a comment generated from generic messages.

horus68:
but can this be merged?

rarila:
They are automatically created so it is likely that there are grammatically errors in there. So they should be checked before mergin.

horus68:
Before submit I had fixed masculine/feminine concordance errors as plural/singular.
I think its ok to merge!

Also please merge related #5064

bobdenotter:
@horus68 Looks good to me! Thank you for your contibution! :+1:

PR #5064, opened 2016-03-21
Portuguese Portugal language

horus68:
"pt" language revised to be used as Portuguese (Portugal) neutral.
Previous "pt" languages was renamed to pt_BR Portuguese (Brazil) and previously submitted in the PR #5061


bobdenotter:
@horus68 This also looks good to me! Thank you again! :+1:

Issue #4780, opened 2016-01-31
Nut commands don't work with custom directories ([RFC] ?)


Boorj:
### Problem
Sometimes people want to change the structure of their projects. Bolt has nice possibility to customize folders (cache, thumbnails, database, config). Some issues are with `files/` directory (since it's coupled with filesystem in some code parts), but in general: if set folders in index with '$configuration->setPath' it works ok.
But when launching Nut commands it still uses default directories from `\Bolt\Configuration\ResourceManager::__construct`.

### Solution
We discussed that once, and that was sort of a problem of an egg and a hen, but.. Probably we could make a `CustomProjectPaths.php` that is included both in `index.php` and in nut's `bootstrap.php` and performs merging of default paths list and user custom paths. 

### Profit
* System will have much better flexibility.
* Once refactored the system to this solution, we'll get rid of many possible hardcode parts.

##### Comment
Actually, having custom paths sometimes is a hell now, you don't know where the problem will appear from next time. For me (imho) it's convenient to have all project source files in one folder: business logic, configurations, translations, assets. And bolt's files - in another folder (database, cache, log, nut, extensions). That's why i really want to vanish all problems with custom paths.

Boorj:
###### feel free to close it if
solution duplicates #4406 , but @CarsonF told in #4590 to report all weird behaviour.
Seems that i'm that test guy who has both composer install and custom directory structure, and pays his time for discovering of all imperfections in code.

CarsonF:
@Boorj How are you doing a custom directory configuration without using your own `bootstrap.php / index.php` file?

Boorj:
i'm doing it with `index.php` injections. I have a class that provides some custom logic and path settings.

GawainLynch:
Nut just gets it's paths from the same place as the rest of the application, if your path overrides aren't done properly then that will cause you issues.

Boorj:
it takes me to create `cache`, `theme`, `theme_default` and other directories to make nut work (Actually i didn't make it work).. Is it ok?  Theme directory is set inside `config.yml`, maybe we can put directory structure there too? And feed `index.php` and `nut.php` (bootstrap) with this config? IT could solve hen/egg question.

Boorj:
So now i have whole collection of issues that claim to have one file with custom paths configuration , that could be used in:
* `index.php` on application run  (#3594)
* `script handler` on composer update/install (#4406)
* `bootstrap.php` for nut commands (this issue)
[...]

Issue #4406, opened 2015-11-09
Is it ok that ScriptHandler::installAssets uses hardcoded folders?


Boorj:
Script Handler uses, for example, cache folder as hardcoded to 'app/cache': 
https://github.com/bolt/bolt/blob/master/src/Composer/ScriptHandler.php#L74
```php
$filesystem->mkdir($appDir . '/cache/',    $dirMode);
```
And in https://github.com/bolt/bolt/blob/master/src/Configuration/ResourceManager.php#L92 it's being set by path, so it could be changed on init.
```php
$this->setPath('cache', 'app/cache');
```
I put logs and cache to `app/vars/cache`. So the question is : am i missing something in docs? Can i  change cache folder? If so - ok. Probably i could even write my own script handler (i saw that advice [here](https://docs.bolt.cm/installation-advanced))

SahAssar:
@Boorj i'm sorry, but since you have been using dev versions of bolt for quite a while I'm going to ask you to reverify some of the bugs you filed, this being one of them...

Could you check that this is still an issue and in that case update this issue?

GawainLynch:
Fixed in #5069

PR #5069, opened 2016-03-22
Fix bootstrapping for nut and assets path and script handler

### LINKED ISSUE #5023: Extension assets not resolving paths correctly on Bolt Composer installs

CarsonF:
Fixes #5023 #4780


Boorj:
Seems that #4406 has to be revised too, probably it could be closed.

Issue #5071, opened 2016-03-22
Speed - To infinity and beyond


GDmac:
Topic to discuss speed improvements (PHP 5.6)
Running Bolt pages overview with `?XDEBUG_PROFILE=true` and analysing with webgrind

- `src/config/doReplacements()` is very expensive on 5.6  (250ms)
  maybe optimize with `strpos() % !== false` check before substringing?
  alternative, find once keep reference in memory for later easy replace

Other possible optimisations
- if some things can be moved to / or are better suited for backend (and/or for logged in users):
    - Schema check has a hard coded expired constant (1800 seconds), can this maybe be a simple check on contenttypes file mtime and/or other sources of table definitions?
    - Timed publish/depublish. granularity is per minute, but db is queried on every request, possible to combine foreach contenttypes into 1 query?
- ...

- <del>- `Twig_Template->display` and `Twig_Template->displayWithErrorHandling` (110ms)</del>
- <del>`Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch` (90ms)  
  called 9 times, the most expensive one being `HttpKernel->filterResponse` (44ms)</del>

bobdenotter:
As i posted in a separate issue: The newly introduced function doReplacements($value) in app/Config.php gets called 15.000 times for every single page request, causing an overhead of about 25% in time taken on my tests.

Thank you, @GDmac for finding this! Introduced in #3005.

Humble suggestion: :fire: :fire:

GDmac:
i've built a proof of concept, based on ``storage:database:schema:timer`` 
to reduce the Timed publish/depublish call granularity down to every minute,
instead of call it on every request. This saves 40ms of requests in debug mode.

@CarsonF suggested the timed publish be moved to into a cron-able process away from frontend. Together with the option, suggested by @bobdenotter to then be able to disable timed publish from the frontend with a config setting.

For simple sites, that don't use cron (simple sites and shared hosts etc.) i would like to suggest to add to option to still include the (60 sec.) timer functionality as this shaves of a significant amount of processing.
RFC.

SahAssar:
Fixed in #5073, right? Closing, feel free to reopen if it persists.

GDmac:
Nope @SahAssar, especially the Timed publish/depublish is still something i'ld like to tackle.
To bring calls down to the publish granularity of per minute, instead of calling it on every request.
(This saves 40ms per requests in debug mode)

GawainLynch:
[...]

PR #5073, opened 2016-03-22
Improve config replacement performance

rossriley:
This PR offers some optimisation of the config replacement functionality as well as some enhancements to make it much more useful.

References: #5071 

Details
-------

This PR does the compilation of config replacements just once at boot time, rather than on every call to `config->get()`

As an additional bonus it also now respects callables so if your replacement is a callable, eg:

`somepathconfig: %twig.path%`

Then the service will be called and the result returned.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -244,20 +244,20 @@ class Config
         if (is_string($value)) {
             $serviceName = substr($value, 1, strlen($value) - 2);
 
-            if (!isset($this->app[$serviceName])) {
-                return;
-            }
-
             if (strpos($serviceName, ":") !== false) {
                 list($serviceName, $params) = explode(':', $serviceName);
             } else {
                 $params = [];
             }
 
+            if (!isset($this->app[$serviceName])) {
+                return;
+            }
+
             $service = $this->app[$serviceName];
 
             if (is_callable($service)) {
-                return call_user_func_array($service, $params);
+                return call_user_func_array($service, [$params]);
             } else {
                 return $service;
             }
--- a/src/Config.php
+++ b/src/Config.php
@@ -222,7 +222,7 @@ class Config
     {
         if ($value === null) {
             $this->data = $this->doReplacements($this->data);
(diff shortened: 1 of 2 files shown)


CarsonF:
Instead of doing the replacements all at once would it be better to do them when they are asked for and just save the result as the new value?

src/Config.php
```diff
@@ -219,14 +218,23 @@ public function get($path, $default = null)
      *
      * @return mixed
      */
-    protected function doReplacements($value)
+    public function doReplacements($value = null)
```
GawainLynch: Just flagging before tonight's meeing… but if we're going to add this to the API, can we at least mark it `@internal` as there is a :dart: :gun:  on the `Config` class for ^3.1

rossriley:
That was the issue we had before though. For instance at various points of the lifecycle Bolt will ask for different nestings of config.. `$config->get("general")` or `$config->get('general/db')` or even `$config->get('general/db/port')` 

So if you do it on fetch you have to recurse down the array each time and with all the duplication of calls / nesting levels then this results in the speed hit.

src/Config.php
rossriley: done

CarsonF:
Right but this has always been the case. It's only one extra method call per `get()` to check that the string doesn't start and end with `%` then the value is returned.

Wouldn't doing _all_ replacements up front not be performant? Not all values may even be grabbed.

rossriley:
Yes, but take the following scenario:

```
db:
    port: %db.portresolver%
    host: localhost
```

Then when Bolt fetches the config it does `$config->get('general/db')` and because only db is returned then the replacement hasn't happened on port because db is just returned as an array.

CarsonF:
I see, thanks for the example

GawainLynch:
The test fails on this are fixed with #5087… so  co-merge on ths two. :wink:

bobdenotter:
Green lights! Whoop!

PR #5079, opened 2016-03-25
Fix variable check

rossriley:
Fixes error on variable check for AbstractExtension


GawainLynch:
It… never… happened… :wink:

PR #5100, opened 2016-03-30
Don't return nbsp in ImageHandler

SahAssar:
when image/popup is called with an empty filename. It can cause weird breaks in templates when we do, so instead return an empty string.


GawainLynch:
…and tests! Nice work :+1:

Issue #5014, opened 2016-03-13
[BUG] unable to create first user on ip-based server


jadwigo:
If you want to setup the first user on a site that that you get to without a domain name you will get the following error: The CSRF token is invalid. Please try to resubmit the form.

If you do the samething on the exact same site with a domain it works.
## Details
- Relevant Bolt Version:  3.0
- Install type: Zip/tarball install http://bolt.cm/distribution/bolt-3.0.0-alpha8.tar.gz
- PHP version: PHP Version 5.5.9-1ubuntu4.14
- Database: sqlite 3.8.2
- Used webserver: Nginx nginx/1.4.6
## Reproduction

I tried this on a digitalocean lemp droplet and when I accessed it through the ip address of the droplet it did not allow me to setup the first user.

I checked if all folders were writable to the webserver user and I tried different cookie settings for https://github.com/bolt/bolt/blob/master/app/config/config.yml.dist#L343

When I accessed it through a domain it worked.

rixbeck:
I had a similar challenge yesterday. I was unable to create first user with the same CSRF error. My server is a LXC container based VM under my Ubuntu box. Tried with a clean checkout/compser-install/update setup with Nginx first then Apache. I was careful about running the same user with both webserver on the same webroot folder, and file permissions was okay.

Requests look like

* http://bolt.d64:8080/bolt/userfirst in Nginx
* http://bolt.me/bolt/userfirst in Apache

Both of them produced same result repeating CSRF error messages while session entries created under cache folder. Also tried with different browser and cleared browser cookies, until I lowering login session hardening level with 

```cookies_use_httphost: false``` in config.

At this moment the first user has been created with Apache and able to log in under http://bolt.me
but can't with http://bolt.me.

I'm trying to get into deeper...

jeffpfoster:
I can confirm this bug is happening as well.  I was testing the upgrade path to 3.0.x and setup a new server which didn't have an IP address entry yet.  Received the error mentioned above.  I am adding a DNS entry now to test if getting through user setup with the domain name works.  

Testing: verified issue does not exist in version 2 with a fresh install.  Pointed the webroot at a version 3.0.1 and 3.0.4 and reproduced the issue.

rixbeck:
Are you on port 80?

jeffpfoster:
I'm on port 80 with an nginx and php 7 install. Initial setup attempt was on sqlite3.

jeffpfoster:
[...]

PR #5115, opened 2016-04-04
Fix domain cookie value with http port

### LINKED ISSUE #5114: [BUG] Default cookie domain value in header can't contain port number
In case of server addressed via any port number like http://bolt.me:8080 and no explicit _cookies_domain_ value set to _.bolt.me_ then right session storage can't be reached by the app.
This will be producing CSRF errors at first user creation like mentioned in #5014.

Details
-------

 - Bolt Version: [ master ]

Screenshot:
Response header

rixbeck:
Default cookie domain value can't contain server's http port number.
Resolves partially #5014 and resolves #5114

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -325,10 +325,10 @@ class Config
 
         // Make sure the cookie_domain for the sessions is set properly.
         if (empty($general['cookies_domain'])) {
-            if (isset($_SERVER['HTTP_HOST'])) {
-                $hostname = $_SERVER['HTTP_HOST'];
-            } elseif (isset($_SERVER['SERVER_NAME'])) {
+            if (isset($_SERVER['SERVER_NAME'])) {
                 $hostname = $_SERVER['SERVER_NAME'];
+            } elseif (isset($_SERVER['HTTP_HOST'])) {
+                $hostname = reset(explode(':', $_SERVER['HTTP_HOST']));
             } else {
                 $hostname = '';
             }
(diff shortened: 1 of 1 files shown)


bobdenotter:
For me (locally) this breaks in entirely different ways: 

screen shot 2016-04-05 at 12 20 08

src/Config.php
```diff
@@ -325,10 +325,10 @@ protected function parseGeneral()
...
-            if (isset($_SERVER['HTTP_HOST'])) {
-                $hostname = $_SERVER['HTTP_HOST'];
-            } elseif (isset($_SERVER['SERVER_NAME'])) {
+            if (isset($_SERVER['SERVER_NAME'])) {
                 $hostname = $_SERVER['SERVER_NAME'];
+            } elseif (isset($_SERVER['HTTP_HOST'])) {
+                $hostname = reset(explode(':', $_SERVER['HTTP_HOST']));
             } else {
                 $hostname = '';
             }
```
GawainLynch: Might want to go for something like:
```php
            $request = Request::createFromGlobals();
            if ($request->server->get('HTTP_HOST', false)) {
                $hostHeaders = explode(':', $request->server->get('HTTP_HOST'));
                $hostname = reset($hostHeaders);
            } elseif ($request->server->get('SERVER_NAME', false)) {
                $hostname = $request->server->get('SERVER_NAME');
            } else {
                $hostname = '';
            }
```

bobdenotter:
Looks good to me!

PR #5121, opened 2016-04-05
Only skip htmlsnippets if we are returning a cached response

SahAssar:
We set the cache-control here:
https://github.com/bolt/bolt/blob/2236c27440a4c52ca47591b1fba338cbc641f406/src/Render.php#L102 so if it is not set to a max-age we can be sure that we need to add the snippets.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -562,7 +562,7 @@ class Application extends Silex\Application
 
         // true if we need to consider adding html snippets
         // note we exclude cached requests where no additions should be made to the HTML
-        if (isset($this['htmlsnippets']) && ($this['htmlsnippets'] === true) && ($this['config']->get('general/caching/request') == false)) {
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -562,7 +562,7 @@ public function afterHandler(Request $request, Response $response)
 
         // true if we need to consider adding html snippets
         // note we exclude cached requests where no additions should be made to the HTML
-        if (isset($this['htmlsnippets']) && ($this['htmlsnippets'] === true) && ($this['config']->get('general/caching/request') == false)) {
+        if (isset($this['htmlsnippets']) && ($this['htmlsnippets'] === true) && ($response->headers->get('Cache-Control') == "no-cache")) {
             // only add when content-type is text/html
```
GawainLynch: Just say "NO!" to loose comparison. :wink:

SahAssar: Gaaah, I fixed it in our deployment but copypasted my old version :D

bobdenotter:
Only failing on PHP 5.3, so GTG! 

screen shot 2016-04-05 at 14 45 44

Issue #5094, opened 2016-03-28
"Pimple warning" when paging content. 


bobdenotter:
When paging content like `/?&page_entries=2`, I get this notice: 

> Warning: Missing argument 1 for Pimple::{closure}() in /Users/bob/Sites/bolt-origin/vendor/pimple/pimple/lib/Pimple.php on line 122
> Notice: Undefined variable: c in /Users/bob/Sites/bolt-origin/vendor/pimple/pimple/lib/Pimple.php on line 126

See screenshot: 

![screenshot 2016-03-28 15 04 36](https://cloud.githubusercontent.com/assets/1833361/14078513/a8992020-f4f6-11e5-9d99-fc72d30dc15f.png)

Details
-------

 - Relevant Bolt Version: master

bobdenotter:
Partial backtrace, from the Pimple warning: 

```
array:13 [▼
  0 => array:6 [▼
    ”file" => ”/Users/bob/Sites/bolt-origin/src/EventListener/PagerListener.php"
    ”line" => 45
    ”function" => ”{closure}"
    ”class" => ”Pimple"
    ”type" => ”::"
    ”args" => array:1 [▶]
  ]
  1 => array:7 [▼
    ”file" => ”/Users/bob/Sites/bolt-origin/src/EventListener/PagerListener.php"
    ”line" => 45
    ”function" => ”__invoke"
    ”class" => ”Closure"
    ”object" => Closure {#638 ▶}
    ”type" => ”->"
    ”args" => array:1 [▶]
  ]
  2 => array:5 [▼
    ”function" => ”onRequest"
    ”class" => ”Bolt\EventListener\PagerListener"
    ”object" => PagerListener {#1305 ▶}
    ”type" => ”->"
    ”args" => array:3 [▶]
  ]
  3 => array:4 [▼
    ”file" => ”/Users/bob/Sites/bolt-origin/vendor/symfony/event-dispatcher/Debug/WrappedListener.php"
    ”line" => 61
    ”function" => ”call_user_func"
    ”args" => array:4 [▶]
  ]
```

A "fix" for this is to change this in `src/EventListener/PageListener.php`, line 45:

    $manager = $this->managerFactory->__invoke($dummy = null);

But, there must be a cleaner way to do this.

rixbeck:
I'm on cleaning up...

bobdenotter:
Fixed in #5125

PR #5125, opened 2016-04-05
Cleaner pager manager factory with resolving Pimple warning

rixbeck:
In #5094 we have Pimple warnings by invoking raw service factory without parameters. Moreover I found that code wasn't really really clear and self describing in actual implementation. 

Fixes: #5094

### DIFF
--- a/src/EventListener/PagerListener.php
+++ b/src/EventListener/PagerListener.php
@@ -14,16 +14,15 @@ use Symfony\Component\HttpKernel\KernelEvents;
  */
 class PagerListener implements EventSubscriberInterface
 {
-    protected $managerFactory;
+    protected $manager;
 
     /**
      * PagerListener constructor.
-     *
-     * @param \Closure $pagerManagerFactory
+     * @param PagerManager $manager
      */
-    public function __construct(\Closure $pagerManagerFactory)
+    public function __construct(PagerManager $manager)
     {
-        $this->managerFactory = $pagerManagerFactory;
+        $this->manager = $manager;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


CarsonF:
I would just pass `PagerManager` into the `PagerListener` and then call `initialize` there. No need for a separate factory

CarsonF:
Actually that `isPagingRequest` doesn't need to be in the listener either. It should be in the manager.

rixbeck:
This means we're going to create the manager object on each request no matter the request would use paging. It should be about 1-5% of the requests of an average website I bet. If you say it is worth then no `isPagingRequest` check needed at all just blindly creating the manager as a service. Let's do that?

CarsonF:
Yes, we create hundreds of objects on each request. I say drop the check.

GawainLynch:
@CarsonF are you happy with this now?

CarsonF:
Are we not putting spaces in between string concatenations anymore?

Other than that, looks good!

bobdenotter:
> Are we not putting spaces in between string concatenations anymore?

Yes, Symfony Coding Standard says we certainly are. I'll ninja-fix those and merge in.

GawainLynch:
>> Are we not putting spaces in between string concatenations anymore?
> Yes, Symfony Coding Standard says we certainly are. I'll ninja-fix those and merge in.

Actually, we (thankfully) are… Symfony are not.

rixbeck:
Ok, that's an issue in CodeSniffer's ruleset... and in my attention. Going to fix...

joshuan:
Sorry, guys, where now this changes? I have #5094 problem on master branch :(

GawainLynch:
@joshuan OK… there was a bit of a mix up and two hotfix branches got merged together instead of down to master.

I have now merged that PR branch into `release/3.0` (which is the branch you should be tracking now), so that problem should be fixed.

joshuan:
@GawainLynch Thanks!

Issue #4577, opened 2015-12-10
[v2.2.*] Giving your template files a name breaks the filter on templateselect 


PhillippOhlandt:
Hey,

I tested this on 2.2.13 (my current running version) and on 2.2.14 as well, both with the same bug/behavior.

What I did:

1: Installing a fresh copy of v2.2.13
2: Editing the templateselect filter on the `pages` contenttype
```
template:
    type: templateselect
    filter: 'page_*.twig'
```

3: Creating `page_1.twig` and `page_2.twig` in the `base-2014` theme.

Now the templateselect shows me only these two templates.

4: Adding the `templateselect` configuration to the themes `config.yml`
```
templateselect:
    templates:
        - name: 'Template for Records'
          filename: 'record.twig'
        - name: 'Page 1'
          filename: 'page_1.twig'
```

Now I can only select the `record.twig` and the `page_1.twig` (with their given names) on the templateselect. 
But I shouldnt be able to select the `record.twig` template and instead I should be able to select the `page_2.twig` template.

Seems like the filter will be ignored and all templates with a given name will be shown.

SahAssar:
Yeah, ran into this a while ago but forgot to open an issue.

rossriley:
I always assumed this was intended behaviour... If you take over the list in your theme then that replaces the default (which is to search for `*.twig`) with your approved list.

SahAssar:
@rossriley That seems odd to me, since different templates can be meant for different contenttypes. Say you have the following four templates:
```
templateselect:
    templates:
        - filename: "page_gallery.twig"
          name: "Page with gallery"
        - filename: "page_contactform.twig"
          name: "Page with contactform"
        - filename: "entry_map.twig"
          name: "Blogpost with map"
        - filename: "entry_facebook.twig"
          name: "Blogpost with facebook comments"
```

Then I probably want `filter: 'page_*.twig'` in the pages contenttype and `filter: 'entry_*.twig'` in the entries contenttype.

Or is there some other way to restrict named templates to a single contenttype?

PhillippOhlandt:
What @SahAssar said. I use it in that way.

bobdenotter:
I think this is one of those typical "It's not wrong, but probably not what you would expect" type of things. 

In my experience this usually means a specific part of functionality is trying to do too much, and it's time too kill some darlings to make it less confusing. 

[...]

PR #5126, opened 2016-04-05
[RFC] Get rid of broken 'named templateselect' functionality. 🔥

### LINKED ISSUE #5068: [RFC] Consolidate templatefields and template names in theme.yml
Currently we have two different objects in `theme.yml` for adding metadata to template files (names and custom fields)

```yml
templatefields:
    extrafields.twig:
        section_1:
            type: text
            label: Section 1
        section_2:
            type: html
            label: Section 2
        image:
            type: image
    anotherextrafields.twig:
        markdown:
            type: markdown
            label: A markdown field
        geolocation:
            type: geolocation
            label: A geolocation field
        image:
            type: image

# There can be defined the templates for the template chooser
templateselect:
    templates:
         - name: 'Template with extra fields'
           filename: 'extrafields.twig'
         - name: 'Another template with extra fields'
           filename: 'anotherextrafields.twig'
```

This isn't particularly DRY (what if we change the filename of a template?) and will lead to more issues if we come up with more metadata around templates. I think we should bundle them like so:

```yml
templates:
    record.twig:
        name: Default Template
        fields:
            # Insert fields here
```

Details
-------

 - Relevant Bolt Version: 2.2/master

### LINKED ISSUE #5119: Records can't select templates
image

bobdenotter:
Fixes #5119. Closes #5068

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -178,20 +178,7 @@ class RecordHandler
             return null;
         }
 
-        // Get the active themeconfig
-        $themeConfig = $this->app['config']->get('theme/templateselect/templates', false);
         $files = [];
-
-        // Check: Are the templates for template chooser defined?
-        if ($themeConfig) {
-            foreach ($themeConfig as $templateFile) {
-                if (!empty($templateFile['name']) && !empty($templateFile['filename'])) {
-                    $files[$templateFile['filename']] = $templateFile['name'];
-                }
-            }
-
-            return $files;
-        }
 
         $name = $filter ? Glob::toRegex($filter, false, false) : '/^[a-zA-Z0-9]\V+\.twig$/';
         $finder = new Finder();
--- a/theme/base-2016/theme.yml
+++ b/theme/base-2016/theme.yml
@@ -48,10 +48,3 @@ templatefields:
         image:
             type: image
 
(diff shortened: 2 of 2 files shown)


bobdenotter:
Also fixes #4577

GawainLynch:
WFM... Someone with a properly working solution can resubmit :wink:

bobdenotter:
We're going to vote on this during dev-meeting tonight. :-)

GawainLynch:
Well, it causes data loss, so the "feature" is invalid as-is. The only thing we need to vote on is the person that presses the merge button on this PR.

src/Twig/Handler/RecordHandler.php
```diff
@@ -178,21 +178,8 @@ public function listTemplates($filter = null, $safe = false)
...
 
-        // Check: Are the templates for template chooser defined?
-        if ($themeConfig) {
-            foreach ($themeConfig as $templateFile) {
-                if (!empty($templateFile['name']) && !empty($templateFile['filename'])) {
-                    $files[$templateFile['filename']] = $templateFile['name'];
-                }
-            }
-
-            return $files;
```
bobdenotter: Seriously?! If it returns here, it's no wonder the other options are missing. :fire:

GawainLynch:
#5127 wins… good news is, it is :beers: time @bobdenotter

PR #5127, opened 2016-04-05
Fix named templatefields without burning the whole thing down 🚒

### LINKED ISSUE #5119: Records can't select templates

SahAssar:
This makes it so that named templates are just named, everything else is the same as if you hadn't named any templates at all. Filtering works, it shows unnamed templates, and we don't need to burn anything down :wink: 

Fixes #5119, #4577 
Closes #5126

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -178,20 +178,7 @@ class RecordHandler
             return null;
         }
 
-        // Get the active themeconfig
-        $themeConfig = $this->app['config']->get('theme/templateselect/templates', false);
         $files = [];
-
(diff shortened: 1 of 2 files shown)


rossriley:
I'll need to get a refund on these: :fire: :fire: :fire:

bobdenotter:
I could lice with this.. 

Like #5126, this would fixes #5119 and fixes #4577

It would not closes #5068, though.

PhillippOhlandt:
:+1:

GawainLynch:
OK… rejecting this one then as it make too much sense… and @PhillippOhlandt likes it :stuck_out_tongue_winking_eye:

Issue #5084, opened 2016-03-25
[bug] notices in frontend on first pageview after cache clear


jadwigo:
When you clear the cache (after a code change / update to bolt) and you have a configuration that has relations between content types you will get a list of notices at the first page refresh after you have cleared the cache. (There must obviously be some records in the database that have relations set)

The notices will be gone after the first pageview.

It happens in the front-end on any page. Notices look like:

"Notice: Undefined index: fromcontenttype in [..]/vendor/bolt/src/Storage/Field/Type/IncomingRelationType.php on line 68"

![image](https://cloud.githubusercontent.com/assets/1488214/14045467/ca094f90-f29a-11e5-8f32-357d667ac6f6.png)

The notices repeat a few times, and after that the page is rendered normally underneath.

Details
-------

 - Relevant Bolt Version: 3.0 / master 
 - Install type: Github checkout and the others probably
 - PHP version: 5.6
 - Used webserver: Apache and Nginx

bobdenotter:
I cannot reproduce this. Can you, @jadwigo ?

bobdenotter:
I have some additional info on this.. If I do `dump($data)` in `public function hydrate($data, $entity)` in `IncomingRelationType.php`, I get the following: 

![screen shot 2016-04-04 at 17 22 54](https://cloud.githubusercontent.com/assets/1833361/14253027/2dfd9ce4-fa8a-11e5-810c-2017bbb55f11.png)

Where it seemingly goes wrong, is that the `fromcontenttype` values seem to be `unique`. Instead of having three `pages`, two `structures` and one `pressreleases`, I get one of each. See the screenshot with data from the DB: 

![screen_shot_2016-04-04_at_17_27_33](https://cloud.githubusercontent.com/assets/1833361/14253155/aa933782-fa8a-11e5-952e-7abcee6c0fd2.png)

@rossriley: Is this something you could look into?

bobdenotter:
Digging deeper, if I do this in `find` in `Repository.php`

```
        $result = $qb->where($this->getAlias() . '.id = :id')
            ->setParameter('id', $id)
            ->execute()
            ->fetch();

        dump($result);
```

I get this: 

```
  "_incomingrelation_id" => "3,17,36,38,65,66"
  "_incomingrelation_fromid" => "22,27,19,3,1"
  "_incomingrelation_fromcontenttype" => "pages,structures,pressreleases"
```

Seems like that's where the issue is coming from.

[...]

PR #5128, opened 2016-04-05
Improve Select Queries Across Joins

rossriley:
This improves the queries across the relations and taxonomy queries. Originally they could be selected DISTINCT using the db engine but with the addition of extra fields this often gives unwanted behaviour.

So the DISTINCT call has been removed and the cleaning is all done in the `normalizeData()` method instead.

Fixes: #5084


bobdenotter:
Huzzah!

PR #5130, opened 2016-04-05
Update Composer version constraint for stable version

GawainLynch:


GawainLynch:
Fail on this is due to #5701 and fixed in #5135

bobdenotter:
:+1:

PR #5135, opened 2016-04-06
Fix acceptanace tests broken by #5107

GawainLynch:


bobdenotter:
The "red light" is only for PHP 5.3, so :shipit:

PR #5144, opened 2016-04-07
Assets… Remove outdated v2 functionality

GawainLynch:
This was only left in place for 2.3… now that we've ripped out the old extension classes this stuff is just wasted space.


bobdenotter:
:gun:

Issue #5097, opened 2016-03-29
Some minor nitpicks in base-2016


bobdenotter:
Some collected nitpicks for the base-2016 theme. 

 - Vertical rhythm is a bit sloppy, here: 

![screenshot 2016-03-29 21 50 23](https://cloud.githubusercontent.com/assets/1833361/14121550/53acc186-f5f8-11e5-82f1-dcdb3862e5ee.png)

 - Images could do with better margin in single pages: 

![screenshot 2016-03-29 21 51 11](https://cloud.githubusercontent.com/assets/1833361/14121593/763f5114-f5f8-11e5-8785-c4f50215d403.png)

 - Links are not recognisable: (please make them underlined, with a color)

![screenshot 2016-03-29 21 52 07](https://cloud.githubusercontent.com/assets/1833361/14121625/895f5302-f5f8-11e5-9a34-b829df78899b.png)

 - The pagination is _very_ bland. It might be a little more exiting. Perhaps with a sublte background, or an outline?

![screenshot 2016-03-29 21 54 19](https://cloud.githubusercontent.com/assets/1833361/14121697/ce987516-f5f8-11e5-95e4-92178e31733b.png)

 - In general, I think it's a bit _too_ white. Having one 'Spot color', and using that for links and the pager would be nice. Perhaps even make it configurable? 

- Entries in listing views could get some love: vertical rhythm, and the images should get a lightbox/popup/modal. Maybe use http://foundation.zurb.com/sites/docs/media-object.html ?

![screenshot 2016-03-29 21 57 17](https://cloud.githubusercontent.com/assets/1833361/14121842/5ce85ee4-f5f9-11e5-85bb-18f9b0dbe606.png)

 - Videos break the column (needs [flexvideo](http://foundation.zurb.com/sites/docs/flex-video.html) )

![screenshot 2016-03-29 22 09 12](https://cloud.githubusercontent.com/assets/1833361/14122254/3d2b3f52-f5fb-11e5-9e37-25069cec5ee5.png)

 - The 'boxed' layout mode has gone missing ;-)

(Some of these can be fixed in `bolt-foundation` as well, by the way. I'll work on either having changes to the theme flow back to that one, or i'll PR my relevant changes into `bolt/bolt`)

bobdenotter:
@SahAssar: FYI, this fixes the quirk with the video: https://github.com/bobdenotter/bolt-foundation-theme/commit/b1e85461e5d69fe2baac653e4de4800d0ae34c03

[...]

PR #5145, opened 2016-04-07
Theme improvements

SahAssar:
Fixes everything in #5097 except for making it more colorful. I couldn't make it work, and I asked a few designers at work and they couldn't make it "general, but colorful" either. This re-adds magnific popup as a dependency too, since I'm assuming we want to keep using that as the default popup library.


bobdenotter:
Cool! I'll check tomorrow morning, And marge it in! 👍

bobdenotter:
Ugh, stupid phone.

PhillippOhlandt:
I just wondered :D

bobdenotter:
:hear_no_evil: :see_no_evil: :speak_no_evil:

PR #5147, opened 2016-04-08
Fix bug show folders in ck_files template

joshuan:
Bug not created, already fixed.
When in field type html open insert image.


GawainLynch:
:+1:

bobdenotter:
:+1: from me too.

PR #4000, opened 2015-08-20
Refactor Wars Episode IV: A New Hope

GawainLynch:
A long time ago in an RFC far, far away…

**Episode IV**

**A New Hope**

It is a period of refactor war.

Rebel developers, striking from a 
hidden base, have won their first 
victory against the evil 
Monolithic Code Empire.

During the battle, Rebel spies 
managed to steal secret plans to 
the Empire's ultimate weapon, 
Bolt v2.3, a heavily refactored 
release with enough power to 
build an entire web site.

Pursued by the Empire's sinister
agents, Princess Gawain races 
home aboard her starship, 
custodian of the stolen plans 
that can save her people 
and restore stability to the 
[codebase....](http://www.starwars.com/games-apps/star-wars-crawl-creator/?55d5e39ee4b0be1e479fdbf2)

star-wars-wallpaper-a-new-hope-8

PR #5150, opened 2016-04-09
[Travis] Composer install testing

GawainLynch:
This move the Travis Composer install test set-up to use a "real" Composer project, that includes updated vendor directories and theme.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,9 @@ before_install:
   - sudo service nginx restart
   # Start postfix
   - sudo service postfix restart
+  # Environment set up
+  - export COMPOSER_PR_REPO=../.bolt-git-install/
+  - export COMPOSER_INSTALL=../.composer-install/
 
 before_script:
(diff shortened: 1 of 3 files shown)


GawainLynch:
YES! The only damn number I wanted and scored it…

bobdenotter:
> YES! The only damn number I wanted and scored it…\

So, you having #1000, #4000 and #5000 was sheer coincidence? ;-)

GawainLynch:
Yes, that's what I am trying to say… 

I was thinking about this while having a lay in bed earlier, but I probably PR more than anyone in the history of *this* project, so statistically I am just as likely to get a thousand as any other number when compared to other people.

So, yes, all of them were :smile:

SahAssar:
Teach the controversy! @GawainLynch uses his connections to score special PR numbers while leaving nothing but the crumbs to rest of us! We are the 99% percent!

unnamed

PR #5151, opened 2016-04-09
Don't re-fetch record on preview

### LINKED ISSUE #5146: [BUG] Preview page with not publish status catch Fatal error
When i want preview page and status now not publish (Timed publish, Not published, Draft) - Controller generate error: 
```
PHP Fatal error: Bolt core
Error: Call to a member function setFromPost() on a non-object
/src/Controller/Frontend.php
Line: 203
````

Details
-------
 - Relevant Bolt Version: master
 - Install type: Github checkout
 - PHP version: 5.5

GawainLynch:
Fixes #5146

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -193,11 +193,11 @@ class Frontend extends ConfigurableBase
     {
         $contenttype = $this->getContentType($contenttypeslug);
 
-        // First, get the preview from Post.
-        $content = $this->storage()->getContentObject($contenttypeslug);
-
-        if ($id = $request->get('id')) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
I seem to recall there was a very specific reason that we _did_ re-fetch the record on preview. I'mma think 'bout this.

joshuan:
@bobdenotter blame say that added because #4044

SahAssar:
I'm pretty sure that the reason was specifically for inbound relations, since outbound relations will be present in the post data anyway.

bobdenotter:
Looks good to me (after last commit!)

@joshuan Yes, that's the one! :-)

@SahAssar Yes, correct. :-)

PR #5154, opened 2016-04-11
Set session.save_path to a hidden directory, to persist after flush

GawainLynch:
This is currently causing a few problems out-of-the box.


bobdenotter:
:+1:

PR #5156, opened 2016-04-11
Messages language updated with new about-us string

horus68:
Details: 'master' for 3.0 beta 2

Included new about-us string and removed previous one.
Cleared duplicated string(with typo) :
`"Search results for <b> %search% </b>."`

Note: not removed but...

#--- 5 unused messages --------------------------------------------------------
```
"Content for home not found!": "Content for home not found!"
"If there was a Block with 'about-us' for a slug, it would've been shown here. But there isn't one, so that's why you're seeing this placeholder. In the Bolt backend, go to 'check database' and 'Add some sample Records', to create these automatically.": "If there was a Block with 'about-us' for a slug, it would've been shown here. But there isn't one, so that's why you're seeing this placeholder. In the Bolt backend, go to 'check database' and 'Add some sample Records', to create these automatically."
"Latest entries": "Latest entries"
"Related content:": "Related content:"
"You should set 'homepage:' in your 'app/config/config.yml' to an existing record. Make sure that the content is published, otherwise it will not be shown.": "You should set 'homepage:' in your 'app/config/config.yml' to an existing record. Make sure that the content is published, otherwise it will not be shown."
```


bobdenotter:
Looks good to me. Thanks @horus68 :+1:

PR #5159, opened 2016-04-11
Fix issue with imagelists not updating due to event checking

SahAssar:
Fixes issue with imagelists not updating due to event checking (introduced in #4925).
.


bobdenotter:
The failure is on PHP 5.3, which is to be expected. 

Looking good! :+1:

Issue #5164, opened 2016-04-12
[RFC] Renaming/reshuffling/rethinking some configs


SahAssar:
I'm proposing that we rename/reshuffle a few configs to make them more predictable/usable (preferrably for deprecation in 3.1 and removal in 4.0):
### config.yml
1. Consolidate the cookie settings under one array like we do with mailoptions/wysiwyg/branding and most other things.
2. Same as above but for debug settings (and perhaps combine with `debuglog`)
3. Same as above but for frontend settings (`listing_template`, `search_results_template`, `taxonomy_sort` and so on).
4. ~~Remove the grunt/livereload settings from config.yml. They are only for developing the backend (if they even still work), and having them in config.yml confuses people (**this should be done for 3.0** PR: #5166 ).~~
5. Put backend news and stack enable/disable under the branding setting array (where we also already have `news_source` and `news_variable`). Also not including them in the default config.
6. Rename the wysiwyg key to something more logical like `html_field_settings`. Or rename the field to `wysiwyg`, either works, but when looking at the configs one should get that they are connected.
7. Rename recordsperpage and listing_records to make it more clear what is frontend and what is backend.
### contenttypes.yml
1. Rename the `show_in_menu` option under a contenttype to `group` to match the the way we do with fields and also remove the option to set it to `false`. Instead of false one could just do `show_in_menu: "Other content"` and get the same effect (or add the false option to `group`).
2. ~~Not including documentation on YAML repeating nodes in contenttypes.yml, since most people who are new to bolt probably wont be using them until they are more comfortable (and therefore more versed in the docs).~~ See #5168 and #5176
3. deprecate all no-underscore keys like `recordsperpage` and replace them with underscored ones. It might be confusing to have `recordsperpage` and `show_on_dashboard` right next to each other.
4. Have a better example of the settings available for the html area in the `showcases` contenttype.
### taxonomy.yml
1. ~~Rename `chapters` to `groupings` or vice versa. The other taxonomy type examples are named after their `behaves_like` value, that makes this inconsistent.~~ Fixed in #5169
2. Provide some more helpful comments in taxonomy.yml and menu.yml. The other yml files have pretty good comments, but these two are pretty barren.

bobdenotter:
Thums up, I say: :+1:

SahAssar:
I've PR'ed the ones that only change examples or comments (except better comments in menu.yml and taxonomy.yml, those require some thought),

GawainLynch:
[...]

PR #5166, opened 2016-04-12
Remove livereload settings from config.yml and _page.twig

SahAssar:
Remove livereload settings from config.yml and _page.twig per #5164 and irc.


GawainLynch:
ping @rarila

bobdenotter:
I'm pretty sure this won't work at the moment. :gun:

Issue #420, opened 2013-05-14
Content type slug is sent blank


outrunthewolf:
When editing a page url the contenttypeslug can be sent as blank forcing an error (see screenshot).

This isn't a solution, but more of a patch so I could continue using the CMS. 
**line 1803, app/src/Bolt/Storage.php**

<pre>
    public function getUri($title, $id = 0, $contenttypeslug = "", $fulluri = true)
    {

        if(!$contenttypeslug) $contenttypeslug = "entries";

        $contenttype = $this->getContentType($contenttypeslug);
        $tablename = $this->prefix . $contenttype['slug'];
        ....
</pre>

![screenshot from 2013-05-14 21 55 33](https://f.cloud.github.com/assets/814246/503928/745d00ae-bcd9-11e2-9a02-8561a40fd3b2.png)

PR #5167, opened 2016-04-12
Remove number from contenttypes.yml

SahAssar:
It's been deprecated since about 3 years back (I couldn't be bothered to blame all the way through the history, but #420 from May 18, 2013 references it as deprecated).

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -306,5 +306,3 @@ blocks:
 # select - varchar(256) - select with predefined values
 # templateselect - varchar(256) - select with template filenames
 # checkbox - integer - checkbox-field which is 1 (checked) or 0 (unchecked)
-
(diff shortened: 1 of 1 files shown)


bobdenotter:
I say it's about time. :+1:

bacbos:
Coincidentally, I just used that one today when I read that it's deprecated as of 3.0@beta3 😨 

What's the suggested alternative?

bobdenotter:
@bacbos You can use either one of these: 

```
# integer - integer - Input field for integer numbers.
# float - double - Input field for floating numbers, stored as 'Double'.
```

bacbos:
Lazy me, I could've figured that out myself by just looking it up, sorry... but thanks @bobdenotter

bobdenotter:
@bacbos No worries! :-)

PR #5168, opened 2016-04-12
Remove nodes from contenttypes.yml

SahAssar:
Remove documention of YAML nodes from contenttypes.yml per #5164 and since it is documented here: https://docs.bolt.cm/release-2-2/contenttypes-and-records#advanced-yaml-repeated-nodes

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -230,63 +230,6 @@ blocks:
     icon_many: "fa:cubes"
     icon_one: "fa:cube"
 
-
-# Using YAML repeated nodes
-#
-# YAML allows the defining of "repeated nodes". These allow you to define a 'node'
-# with a 'key: &name' and referenced later with '<<: *name'
-#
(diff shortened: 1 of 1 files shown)


GawainLynch:
Just had a water cooler with @bobdenotter about this.

This was added in #3430 with the intent of *also* matching what is in the docs. The discussion at the time was that like the field examples below, it would be useful for the "lazy" type person who doesn't go to the docs.

An example given of the "lazy" type would be people like Bob & I who use the commented routes in `app/config/routing.yml`

That given, @SahAssar, do you still feel there is a valid reason to remove this?

SahAssar:
@GawainLynch well, considering that after config.yml this is the file that is most likely for a newbie to edit I think there is a case that something that the docs puts under the section "Advanced" should not be there. But if we keep it I think we should move the "Possible field types" section above it.

bobdenotter:
> But if we keep it I think we should move the "Possible field types" section above it.

I could live with that. :+1:

SahAssar:
Closing cause #5176

PR #5169, opened 2016-04-12
Rename chapter taxonomy example to group

SahAssar:
Per #5164 to conform with the other examples.


SahAssar:
I just can't figure out how to fix the conflict here... I'll fix it tomorrow.

Issue #5162, opened 2016-04-12
[BUG] CKEditor breaks in Repeater fields when Moving Up / Down


mikenosko:
I have a repeater field setup in contentype.yml:

```
repeater:
            type: repeater
            group: repeater
            limit: 3
            prefix: "<p>This allows you to create multiple sets of fields. Use the add button at the bottom to create a new empty set.</p>"
            fields:
                repeattitle:
                    type: text
                repeatimage:
                    type: image
                    extensions: [ gif, jpg, png ]
                repeatcontent:
                    type: html
```

Now when I fill in the fields and and try to reorder them using Move Up / Move Down buttons. The repeaters reorders ok but the moved CKEditor looses the content and can't be interacted with. Trying to interact with the editor ie. pressing a button on the editor after moving produces a:

`Uncaught TypeError: Cannot read property 'getSelection' of undefined`

The relevant hidden textarea element has the content so save etc works. Reloading the page initialises the editors ok with the right content.

Details
-------

 - Relevant Bolt Version: [ 3.0 | master ]
 - Install type: [ Composer install ]
 - PHP version: [ 7.0 ]
 - Used webserver: [ Apache ] version [ 2.4.16 ]
 - For UX/UI issues: [ Chrome 49.0.2623.110 | Firefox 45.0.1 ]

bobdenotter:
Apparently @jkazimir knows what the issue is, here! :tada:

jkazimir:
Yep! :smile:

Basically, it's the CKE events not surviving the DOM move.  PR incoming in a bit.

bobdenotter:
Fixed by #5171.

PR #5171, opened 2016-04-12
Reset moved ckeditor instances in repeaters

jkazimir:
Reset moved ckeditor instances in repeaters.

Fixes: #5162 

Details
-------

CKEditor doesn't bode too well when moved, due to the way some events are bound in it.  Because of this, sorting a repeater would break any ckeditor moved within.  Simply destroying and re-adding the instance after a move keeps things working.

### DIFF
--- a/app/src/js/widgets/fields/fieldRepeater.js
+++ b/app/src/js/widgets/fields/fieldRepeater.js
@@ -1,8 +1,9 @@
 /**
  * @param {Object} $    - Global jQuery object
  * @param {Object} bolt - The Bolt module
+ * @param {Object} cke  - CKEDITOR global or undefined
  */
-(function ($, bolt) {
+(function ($, bolt, cke) {
     'use strict';
 
     /**
@@ -108,6 +109,7 @@
 
                 setToMove.insertBefore(setToMove.prev('.repeater-group'));
                 self._renumber();
+                self._resetEditors(setToMove);
             });
 
             self.element.on('click', '.move-down', function () {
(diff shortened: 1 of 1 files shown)


app/src/js/widgets/fields/fieldRepeater.js
```diff
@@ -231,4 +252,4 @@
             }
         }
     });
-})(jQuery, Bolt);
+})(jQuery, Bolt, typeof CKEDITOR !== 'undefined' ? CKEDITOR : undefined);
```
bobdenotter: Call my dumb, but isn't that _literally_ the same as just doing this?

    })(jQuery, Bolt, CKEDITOR);

jkazimir: It… very well may be!  I just went with what the ckeditor.js module had, assuming it was done for a reason. ¯\_(ツ)_/¯

app/src/js/widgets/fields/fieldRepeater.js
```diff
@@ -206,6 +209,24 @@
...
+            var editors = container.find('.ckeditor');
+
+            editors.each(function (i, editor) {
+                cke.instances[editor.id].destroy();
+                cke.replace(editor.id);
+            });
+        },
+
+        /**
          * Adds a vlaue to the group counter and adjust button states according to it.
```
bobdenotter: "vlaue"

app/src/js/widgets/fields/fieldRepeater.js
bobdenotter: Here, you dropped this: `\`

jkazimir: Isn't string escaping fun! 😄

bobdenotter:
Whooop! :shipit:

PR #5175, opened 2016-04-13
Updating base-2016 dependencies

bobdenotter:


GawainLynch:
I did wonder where `base-2015` came from for a moment…

PR #5187, opened 2016-04-15
Fix default value for select contenttype

Boorj:


GawainLynch:
Whomever merges, note this one needs to be done manually as it should be targeting `release/3.0`

Boorj:
i made the same commit a while ago, by the way.. also that included an ability to set default value for geolocation and some others..

GawainLynch:
Manually merged into `release/3.0`.

Thanks @Boorj :+1:

PR #5202, opened 2016-04-16
Temporary hack to get the permission name associated with the route

GawainLynch:
We'll address this better in 3.1-dev


SahAssar:
:+1: Thanks!

Issue #5191, opened 2016-04-15
[BUG] Omnisearch gives template error


Sjord:
When using the Omnisearch in the backend, it gives the following error:

 Twig_Error_Syntax
Arguments must be separated by a comma. Unexpected token "name" of value "with" ("punctuation" expected with value ",") in "@bolt/omnisearch/omnisearch.twig" at [line 42](https://github.com/bolt/bolt/blob/release/3.0/app/view/twig/omnisearch/omnisearch.twig#L42).

![bolt - fatal error - mozilla firefox_007](https://cloud.githubusercontent.com/assets/113030/14564425/ca8e55c2-0326-11e6-9b2e-cd1550b1a22c.png)

This bug was introduced with [this commit](https://github.com/bolt/bolt/commit/dd1c83d4062010b905534fa76ec508a93c448e75).

Details
-------

 - Relevant Bolt Version: v3.0.0-beta2-49-gce44208
 - Install type: Github checkout
 - PHP version: 7.0
 - Used webserver: Apache
 - For UX/UI issues: Firefox 45.0.2

Reproduction
------------

1. Log into the backend
2. In the header "Find" input, type something and click the top item.
3. An error is shown. I expect search results to be shown.

GawainLynch:
You're doing a magnificent job of turning up bugs, thank you :+1:

Sjord:
After I fixed the error on omnisearch.twig:42, I get another error on [_aside.twig:7](https://github.com/bolt/bolt/blob/release/3.0/app/view/twig/omnisearch/_aside.twig#L7):

Unknown "widget" function. Did you mean "widgets", "getwidgets", "haswidgets", "form_widget", "countwidgets" in "@bolt/omnisearch/_aside.twig" at line 7?

SahAssar:
Sjord: `{{ getwidgets('dashboard', 'right_first') }}` is the fix there

PR #5203, opened 2016-04-16
Add omnisearch test

SahAssar:
To catch things like #5202 and #5191


GawainLynch:
One way to win the favour of a :princess: :koala: is to write tests… You my friend are a legend! :+1:

PR #5205, opened 2016-04-16
.control-label for image, imagelist and textarea (part 2)

Boorj:
please, delete this PR if i did something wrong...

### DIFF
--- a/app/view/twig/editcontent/fields/_image.twig
+++ b/app/view/twig/editcontent/fields/_image.twig
@@ -89,7 +89,7 @@
 
 {% block fieldset_label_text  labelkey %}
 {% block fieldset_label_info  option.info %}
-{% block fieldset_label_class 'control-label col-xs-12' %}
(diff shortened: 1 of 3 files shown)


GawainLynch:
Merged on rebase

rarila:
Here's something going wrong here. There was a reason why the control-label class was not set (can't remember what it was atm) but looking at thIs [code](https://github.com/Boorj/bolt/blob/release/3.0/app/view/twig/_base/_fieldset.twig#L21). So there are two possibilities. Either the code mentioned above has to be changed to always add it or it shouldn't be added manually as whatever reason to just add it to those cases when it is used as inline label is valid.

Boorj:
agree 100%. Let's add that permanently. And I also had that question, asked @GawainLynch in slack.

rarila:
Question is, why you want to add it. Can't check at the moment as I can't access the desktop. But the reason could be that one is a label for a single control and the one liner is morevof a headline for a collection of controls.

Did you run the bootstrap test?

PR #5208, opened 2016-04-16
Only force enabling user entity on first user creation

### LINKED ISSUE #5190: [BUG] Newly created users are always enabled
When creating a new user, the user is always set to enabled, even if the "User account is enabled" dropdown is set to "no".

Details
-------

 - Relevant Bolt Version: v3.0.0-beta2-43-gb78bd7d
 - Install type: Github checkout
 - PHP version: 7.0.4
 - Used webserver: Apache
 - For UX/UI issues: Chromium Version 49.0.2623.108 Ubuntu 16.04 (64-bit)

Reproduction
------------

1. Under Configuration, Users and permissions, click the button "Add a new user"
2. Fill out the form. Leave the field "User account is enabled" on "no".
3. Click the button "Save user".
4. Notice that the created user is enabled (it has no strike-through), even though we set "user account is enabled" to "no".

GawainLynch:
Fixes #5190 … and apparently give the :koala: some special :leaves: according to @rarila

### DIFF
--- a/src/Controller/Backend/Users.php
+++ b/src/Controller/Backend/Users.php
@@ -109,7 +109,6 @@ class Users extends BackendBase
         // Check if the form was POST-ed, and valid. If so, store the user.
         if ($request->isMethod('POST')) {
             $userEntity = $this->validateUserForm($request, $form, false);
-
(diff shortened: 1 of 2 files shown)


rarila:
:leaves::cake: ?

GawainLynch:
:laughing:

PR #5209, opened 2016-04-16
Tests for issue #5190

AlphaRecon19:
Added some tests for creating new users. Also modified the addNewUser function to enable/disable new users.

Related to #5190


GawainLynch:
…also married to #5208

PR #5217, opened 2016-04-18
Look for template files in deeper folders.

### LINKED ISSUE #5212: [BUG] Twig templates not showing up for templateselect
I've been using the latest version of Bolt, starting off with a clean install, and I'm having issues getting my twig templates to show up for templateselect. I can't get all my twig templates in there from my theme, not by messing with the `filter` value in `contenttypes.yml` and not by explicitly defining them in `theme.yml`. Attached is my directory structure and theme config.

screen shot 2016-04-17 at 18 45 41

What ends up in the templateselect dropdown in the admin panel is the one and only twig file on the second level of nesting in my theme folder, and if I add twig files there I will see them on the dropdown. This is pretty weird however and I can't really find the root cause of my templateselect config not taking effect. I'm trying my best to keep my theme well organized and reusable and being confined to just one level of nesting is quite limiting, if this is indeed the intended behavior.

screen shot 2016-04-17 at 18 46 38

If I dump my theme config in the frontend, I can see that the `templateselect` values are what I set them to be in the config, so that can't be the issue.

I feel like I'm missing something obvious with my file paths or something, but I can't really find anything wrong and I'm working on a fresh install. I think there might be something wrong with how Bolt looks for the twig files in the theme folder recursively.

bobdenotter:
Fixes #5212

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -188,7 +188,7 @@ class RecordHandler
             ->notPath('node_modules')
             ->notPath('bower_components')
             ->notPath('.sass-cache')
-            ->depth('<2')
+            ->depth('<4')
             ->path($name)
             ->sortByName()
         ;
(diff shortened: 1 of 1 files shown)


GawainLynch:
Why do I have a picture of Bruce Willis in my head from that branch name?!?

PR #5207, opened 2016-04-16
Database contraints

GawainLynch:
Fixes #5165

PR #5219, opened 2016-04-19
Have json_array fields use an empty JSON array string as a default

GawainLynch:
Follow up to #5207… as… well… DBAL.


bobdenotter:
Yay!

PR #5221, opened 2016-04-19
Fix #5218 – Bad default values for list types

### LINKED ISSUE #5218: Bad default values for list types
Not sure if this should read "Help, @rossriley!" or "Help, @rarila!"

On new showcases (that have virtually nothing in them) these warnings are getting triggered on save: 
image

Looking at the database entries most other JSON fields end up defaulting to an empty array `[]` but imagelist and filelist are being getting double quotes `" "` (space added for effect)
image

rarila:
This is a little bit hacky. ``context.content.get(key)`` should always be set and and return a working default. At the moment it returns an empty string instead of an empty array. So the correct fix would be to change that on the part and remove those ``|default([])`` afterwards. But this should work meanwhile.

:bell: Ping @rossriley 

Fixes #5218

### DIFF
--- a/app/view/twig/editcontent/fields/_filelist.twig
+++ b/app/view/twig/editcontent/fields/_filelist.twig
@@ -21,11 +21,7 @@
     }
 } %}
 
-{% if context.content.get(contentkey) is defined %}
-    {% set list = context.content.get(contentkey) %}
-{% else %}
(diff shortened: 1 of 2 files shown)


bobdenotter:
A hack a day keeps the doctor away. 🍎

PR #5223, opened 2016-04-19
Don't throw a fatal exception on missing table

GawainLynch:
Corner case, but happens when you drop tables with cookies still set.


bobdenotter:
Whoop whoop!

PR #5234, opened 2016-04-21
Sync configuration defaults with config.yml

SahAssar:
And remove debug_compressjs, debug_compresscss and 'caching/rendering'
since those are not used anymore


SahAssar:
@bobdenotter Since `wysiwyg/ck/viewSourceRoles` seems to be completely unimplemented I was considering removing that too (from config.yml), but wanted to check with you first... Thoughts? It was added in https://github.com/bolt/bolt/commit/35fa25853efe8d3dbc8c64780a5ebb3d2fc8b438

GawainLynch:
Also would be worth getting some input from @CarsonF on this as we're about to tear apart that whole class.

SahAssar:
@GawainLynch Agreed, but so far this is just removing keys or changing values, so I'm assuming that it won't change the refactor much.

GawainLynch:
No stress… was more just tagging him for the notification and associated "Look here, Carson"

bobdenotter:
Yes, imho we can merge this in now, regardless of upcoming changes.

bobdenotter:
Just to confirm: `allowedContent` and `viewSourceRoles` both have no effect. 

\*insert emoji of dodo here\*

CarsonF:
👍

bobdenotter:
Whoooop!

PR #5236, opened 2016-04-21
Remove the viewSourceRoles option that does not work

### LINKED ISSUE #3998: allowedContent and viewSourceRoles
Changing of setting items viewSourceRoles and allowedContent does not take effect.
The source button is allowed for all roles despite of viewSourceRoles setting.
Change allowedContent to 'p strong', but wysiwyg still allows all other tags.

SahAssar:
Removes the never implemented `viewSourceRoles` option.

I'm also waiting for confirmation on if the `allowedContent` option actually does something in it's current iteration, otherwise there might be reason to remove that from `config.yml` too.

EDIT: allowedContent was confirmed to do nothing either, so it's offed. This closes #3998


bobdenotter:
👍

Issue #2365, opened 2014-12-27
Bolt Extensions don't work if they have composer.json dependencies 


jacobtolar:
I built a couple of extensions for Bolt that pull in some dependencies via composer. When I try to use autoloading in my extension:

```php
require 'vendor/autoload.php';
```

it breaks :) Extensions are [required to include 'init.php' inside the autoload section](https://docs.bolt.cm/extensions/config#json-file) - this seems to cause some sort of cyclic dependency when it hits the require statement.

I brought this up in IRC a few days ago and someone suggested a workaround I can use for now, but I thought I'd post the issue here so it could be tracked.

bobdenotter:
Bump @jacobtolar

jacobtolar:
Sorry for the delay. 

Can't remember the IRC name of who suggested this, but the idea was to download the Bolt source, add my dependency to the bolt core composer.json ( https://github.com/bolt/bolt/blob/master/composer.json ) and install using composer. 

What I did instead was download the dependency using Composer and manually 'require' php files I need in extension.php.

I plan to pull this out into its own repo and submit it as an extension but for now, here's what I'm doing:

composer.json:
https://github.com/jacobtolar/mafmc-web/blob/master/bolt/extensions/local/jacobtolar/GoogleCalendar/composer.json

extension.php:
https://github.com/jacobtolar/mafmc-web/blob/master/bolt/extensions/local/jacobtolar/GoogleCalendar/Extension.php#L8-L16

Line 8 crashes Bolt completely - can't load a single page.

caseyamcl:
Here is another **workaround** that allows uses an autoloader instead of `require` statements:

First, require the Symfony Classloader (`~2.6`) in your extension's `composer.json` file, and update dependencies (`composer update`)

Then, your `Extension.php` class can include a `setupAutoloader()` method:

```php
use Symfony\Component\ClassLoader\ClassLoader as SymfonyClassLoader;
use Symfony\Component\ClassLoader\Psr4ClassLoader;

class Extension extends BaseExtension
{
    public function initialize()
    {
        $this->setupAutoloader();
        
        // The rest of your Bolt code here.
    }

    public function getName()
    {
        return "your_ext_name";
    }

    // ---------------------------------------------------------------

    protected function setupAutoloader()
    {
        require_once(__DIR__ . '/vendor/symfony/class-loader/Symfony/Component/ClassLoader/ClassLoader.php');
[...]

PR #5241, opened 2016-04-23
[WIP] Add support for extension dependancies

jnvsor:
I'm not sure what would be the best way to read the composer lockfile, but this automatically installs extension dependencies while preventing duplicate installation of root bolt dependencies.

If you don't have any extensions with dependancies to test this on try jnvsor/boltkint

Something similar can be done for 3.0 in `src/Composer/JsonManager.php`

Fixes: #2365

Can anyone think of anything that would break this? I would also like to make a test case to demonstrate, but there don't seem to be any with mocked composer packages to base it off

### DIFF
--- a/src/Composer/Action/BoltExtendJson.php
+++ b/src/Composer/Action/BoltExtendJson.php
@@ -91,7 +91,6 @@ final class BoltExtendJson
         $pathToWeb = $app['resources']->findRelativePath($app['resources']->getPath('extensions'), $app['resources']->getPath('web'));
 
         // Enforce standard settings
-        $json['repositories']['packagist'] = false;
         $json['repositories']['bolt'] = array(
             'type' => 'composer',
             'url'  => $app['extend.site'] . 'satis/'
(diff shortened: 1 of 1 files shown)


GawainLynch:
OK, sorry I missed where you were going with this in IRC. 

Basically this is a no-go for Bolt v2 as we're long past feature lock and the branch is in Long Term Support mode.

We intend to tackle *something* during the v3.x life cycle, and currently there are some skunkworks options for adding certain dependencies to the extension site's Satis, but they don't take into account the root package.

What we can/should be doing is extending the Composer libraries to read/insert the `installed.json` data (similar to how you have done here) into the runtime configuration… 

I am unsure how other core team members feel, but I tried something similar once and the `extensions/composer.json` because quickly unusable/unreadable taking this approach (I still managed to get the entire Laravel framework to install as a Bolt extension :stuck_out_tongue_winking_eye:) 

An ugly and hackish option that we've often used to this point is the commit the (yeah, I know) libraries into somewhere like `lib/` in the extension itself and add that to the PSR-4 loader in `composer.json`

jnvsor:
So this would be better implemented in 3.x?

Then in 3.x we'd do something similar to this - take extensions/composer.json and merge it with the lock data and then use the new file for the composer update instead of the original one?

GawainLynch:
> So this would be better implemented in 3.x?

It can't be merged into v2, so whatever is done will be targeting 3.something (3.0 is due out 1 May, so after that either way).

> Then in 3.x we'd do something similar to this

Maaaaaaaaaaaaybe :laughing: 

Both @rossriley and I have separately looked at this, and all current approaches (including this) break spectacularly in differing ways (see the Travis tests on this one for an example), so the solution is going to come down to extending parts of Composer itself much in the same was at the local extension helper does in v3.

Issue #5199, opened 2016-04-16
[Upgrade] Fatal error in dashboard from getLatest()


GawainLynch:
Database taken from a `release/2.2` install with a default pre-fill.

![image](https://cloud.githubusercontent.com/assets/1427081/14579787/6a573098-03b7-11e6-9897-1afe9782e6b8.png)

@rossriley & @CarsonF any thoughts or opinions on approach with these?

GawainLynch:
Ping @rossriley for ideas on approach

PR #5253, opened 2016-04-28
Handle exception when table is missing

rossriley:
This prevents the bug whereby users upgrading to 3.0 with an old db were getting a fatal error on the dashboard. If the table is missing now it will be silently caught and the normal 'Database needs to be repaired' error will show.

Fixes: #5199


GawainLynch:
Awesome, thanks mate!

Issue #5117, opened 2016-04-04
Database entities won't hydrate if columns use captials


GawainLynch:
If a table name has columns like `customerID` they will create and save OK, but the `FieldTypeBase::hydrate` chokes as the `$data` array is all keyed with lower case column names

jadwigo:
ouch, optional capitalization in MySQL strikes again

rossriley:
@GawainLynch any idea how these are getting through, according to this line: https://github.com/bolt/bolt/blob/master/src/Config.php#L604

All the field names should be lower-cased, and hyphens replaced with underscores, as far as I can see the metadata driver uses these parsed values to pass to the hydrate method.

Is there any other places they can slip through without being parsed here?

GawainLynch:
@rossriley Yeah it was a custom extension table. The table creates fine with column names CamelCased, the pdo calls work fine, but the hydration is skipped for those columns.

PR #5258, opened 2016-04-28
Fix case error in entity hydration

rossriley:
This takes care of the error where database hydration was happening in a case-insensitive manner meaning that only lowercase column names were correctly hydrated.

This defers to the original column name rather than the key generated by doctrine dbal.

Fixes: #5117


GawainLynch:
:+1:

bobdenotter:
:shipit:

PR #5262, opened 2016-04-29
Theme overrides (homepage and taxonomy_sort)

SahAssar:
Allow theme to override homepage and taxonomy_sort and add examples to theme.yml for taxonomy_sort, homepage, listing_sort, listing_records and search_results_records. Also fix some wrapping in theme.yml.

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -95,8 +95,9 @@ class Frontend extends ConfigurableBase
      */
     public function homepage(Request $request)
     {
-        $listingparameters = $this->getListingParameters($request, $this->getOption('general/homepage'));
-        $content = $this->getContent($this->getOption('general/homepage'), $listingparameters);
+        $homepage = $this->getOption('theme/homepage') ?: $this->getOption('general/homepage');
+        $listingparameters = $this->getListingParameters($request, $homepage);
+        $content = $this->getContent($homepage, $listingparameters);
 
         $template = $this->templateChooser()->homepage($content);
         $globals = [];
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -1086,7 +1086,7 @@ class Storage
         // Sort on either 'ascending' or 'descending'
         // Make sure we set the order.
         $order = 'ASC';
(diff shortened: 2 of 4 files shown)


bobdenotter:
Haven't tested this yet, but I wonder if this PR is complete.. You've added: 

```
 # taxonomy_sort: DESC
 # homepage: page/1
 # listing_sort: datepublish DESC
 # listing_records: 6
 # search_results_records: 10
```

In the changed code, I see no changes related to `listing_sort`, `listing_records`, `search_results_records`. Did those already work (undocumented), or do they need to be added?

SahAssar:
@bobdenotter yep, they are implemented here:

* listing_sort: https://github.com/bolt/bolt/blob/master/src/Controller/Frontend.php#L304
* listing_records: https://github.com/bolt/bolt/blob/master/src/Controller/Frontend.php#L297
* search_results_records: https://github.com/bolt/bolt/blob/master/src/Controller/Frontend.php#L389

I can't reproduce the test failure locally though, It only seems to happen on the composer install... Last time that was the case @GawainLynch said it was due to other things, is that also true here?

bobdenotter:
Passing locally here as well, and the code looks good with mokeytesting. 

Is this safe to merge, @GawainLynch ?

GawainLynch:
> Is this safe to merge, @GawainLynch ?

No… it is against master. I have to go out again (busier day than expected) but will be home in a couple of hours and can test/merge it into the correct branch.

SahAssar:
@GawainLynch I purposely PR'ed against master, I thought that this was supposed to go there... I wouldn't object to putting it in 3.0 though, and if so I can open a PR against release/3.0.

GawainLynch:
Oh, maybe it is… haven't looked at (this) code today… looked and agreed.

Let me merge-down `release/3.0` and then I might get you to rebase against master… I just want to catch the "why" of that test failure… I doubt it is related to this PR, but it'd be good to know if it is "gone"

GawainLynch:
Discussed with @bobdenotter and we decided to sent this into 3.0… so attached the branch to `release/3.0` and merged manually to there :+1:

PR #5266, opened 2016-04-30
[RFC] Should we disable the "live editor" by default?

bobdenotter:
The base-2016 theme doesn't really support it, and I don't recall ever having seen a third-party theme in the wild that does.


GawainLynch:
:+1:

bobdenotter:
No objections, merging in! :shipit:

PR #5273, opened 2016-05-02
Update messages.cs_CZ.yml

lexislav:
updated czech translation of message file for upcoming bolt 3

### DIFF
--- a/app/resources/translations/cs_CZ/messages.cs_CZ.yml
+++ b/app/resources/translations/cs_CZ/messages.cs_CZ.yml
@@ -1,161 +1,29 @@
-# app/resources/translations/cs_CZ/messages.cs_CZ.yml – generated on 2016-03-15 17:19:33 Europe/Berlin
+# app/resources/translations/cs_CZ/messages.cs_CZ.yml – generated on 2016-05-02 16:01:20 Europe/Berlin
 
 # Warning: Translations are in the process of being moved to a new keyword-based translation
-#          at the moment. This is an ongoing process. Translations currently in the
-#          repository are automatically mapped to the new scheme. Be aware that there
-#          can be a race condition between that process and your PR so that it's
-#          eventually necessary to remap your translations. If you're planning on
+#          at the moment. This is an ongoing process. Translations currently in the 
+#          repository are automatically mapped to the new scheme. Be aware that there 
+#          can be a race condition between that process and your PR so that it's 
+#          eventually necessary to remap your translations. If you're planning on 
 #          updating your translations, it's best to ask on IRC to time your contribution
 #          in order to prevent merge conflicts.
 
-#  37 unused messages
-#  65 untranslated messages
-#  22 untranslated keyword based messages
-# 340 translations
-# 185 keyword based translations
+#  no unused messages
+#  no untranslated messages
+#  no untranslated keyword based messages
+# 390 translations
+# 215 keyword based translations
 
-#--- 37 unused messages -------------------------------------------------------
-
-"An error occurred.": "Došlo k nějaké chybě."
-"Are you sure you wish to delete these records? There is no undo.": "Určitě chcete odstranit tyto záznamy? Změny není možné vrátit zpět."
-"Bolt Extension could not be found. Please check at %site% to ensure correct name.": "Rozšíření Boltu se nepodařilo nalézt. Zkuste na webu %site% ověřit správnost názvu."
-"Content '%title%' could not be deleted.": "Obsah '%title%' nelze odstranit."
-"Content '%title%' has been changed to '%newStatus%'": "Obsah '%title%' byl pozměněn: '%newStatus%'"
-"Content '%title%' has been deleted.": "Obsah '%title%' byl odstraněn."
-"Delete selected": "Smazat vybrané"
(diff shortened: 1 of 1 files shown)


GawainLynch:
Very much appreciated, thank you!

horus68:
You know you can translate via Crowdin, right? Your translations where also included.
https://crowdin.com/project/bolt

GawainLynch:
@horus68 That has very much *not* been agreed upon.

lexislav:
Well, such service may be helpful. As it finds the way into the official bolt docs, I will start to use it.
However, czech translation files in the crowding have wrong lang code - en-GB instead of cs_CZ

GawainLynch:
Yes, and we are moving in that direction… but we are going to do it as a planned, coordinated, effort and adjust the code base as required.

We have informed @horus68 of this several times, but he seems to think of himself as not having to follow process or considerate of the impact on those trying to run the project :wink:

horus68:
@GawainLynch what is your claim? User contribute to translations and submite them to Bolt via Github. This is the plan now, right? the way people is preparing the contributions to git can be via notpade or whateaver they want. Or is not so? I just informed above user that he can also translate via Crowdin and then PR.
So the contributes to translations are still on GIT, and GIT is the actual place to find the lateste translations. Once again, what is the problem here? Please do not create issues to translators that are simply waiting for devs to agree when (in what version) they will change things. Until then we need to work on translations, some directly on GIT, others using the tools they can use.
In the end you will end up with translations on GIT so there is no change on the way BOLT code is working.
Or is it?

bobdenotter:
@horus68 The main issue is that things need to follow _process_. Bolt is a project that has many people with a vested interest. Nobody (not even me or Gawain) can oversee the entirety of the things that are going on. 
As such, we simply cannot change things 'on a whim', because it will undoubtedly cause issues in another part of the project. To prevent that, we need to [RFC] significant changes like this. 

As such, we hold off on directing people to crowdin or whatever different service for translations. Doing that would suggest that decisions were made, which they are not. We might move to crowdin, we might move to transifex, we might keep translations on github or roll our own solution. We simply do not know yet. 

Your telling people to "you can use crowdin" also suggests that a decision was made. Which is not the case. We need to weigh the options, before we do that.

horus68:
I'm a plain user so nobody will ever consider my say as a "project approved"
Need to remember to not reply to anyone with a recommendation for "you can use notepad to translate files".
Translators should be able to choose where they want to translate as long they bring here to GIT the final work. Os is it just as long nobody says that to them?
Why is people always arguing that this is a disruptive work? If Git is the official place to have translations why do you care on where are people **preparing** those translations? 
Why some of you consider this difficult to understand by users?
Also: I already stated that i could maintain the workflow to update translations into GIT so there is no extra work to devs. When you decide to make it in another way it will be done. 

So I will just wait for Bolt version v99 when someone finally end up making a decision.
Whatever! Feel free to ignore translators needs I will rest in silence and wait for the decisions that will never come!

GawainLynch:
> I'm a plain user so nobody will ever consider my say as a "project approved"

But people don't know who you are… and yes, they **have** assumed that you are stating what is "approved". That's where, and why, we're reacting… people start making assumptions and then we get stuck having to explain to people; a) how to do something that we don't ourselves understand; b) why we're not following "approved" procedures ourselves when "@horus68 told me to do it this way"… etc :smile: 

> Translators should be able to choose where they want to translate as long they bring here to GIT the final work.

Yes… But see above. Also to add to that, a lot of people struggle with git… even some programmers, and if a programmer is struggling then it is fair to say that it may well be more difficult for someone who is not a programmer but just wants to help you and the rest of the Bolt team translate some strings and feel included in the community. Trust me when I say that accounts for a lot of people.

> Why is people always arguing that this is a disruptive work? 

See above… see discussions you have had with many of the core team. We've all been doing this for a *very* long time, and we are cautious due to having had to "clean up the mess" when honestly well intentioned people behave like this. 

> Also: I already stated that i could maintain the workflow to update translations into GIT so there is no extra work to devs.

Yes, and please believe me when I say, your effort is appreciated… But please please please understand that we don't want to start down the Yellow Brick Road only to have to turn around and go a different direction and be left dealing with angry emails from people who are upset that we changed things after "@horus68 told me to do it this way". Seriously… it happen so many time… if you could step away from your own ego for a few seconds and listen to what we've all been telling you for months now, it might appear obvious.

> So I will just wait for Bolt version v99 when someone finally end up making a decision.

Honestly, grow up! You're 48 years of age, act like it!

We are about to close off an 18 month development cycle, with a raw delta of over 170,000 lines of code… I can't even begin to imagine how many lines of code that **actually** is if you were to add them all up! I can't tell you how many days that I personally haven't even gone to bed and have worked a full 24 hours of a day to make this release happen! I am not the only one on the core team who has done this!

As we have repeated to you **endlessly**, we just want to get 3.0.0 out the door, take a little bit of vacation, and then get back onto things. Then it will happen… You just need to look at the project's history if you don't believe me… it is all in public record!

horus68:
Ok, I'm not discussion bit by bit as there is no point on that.
I will stay silent waiting for your time of decisions.

GawainLynch:
Thanks mate, it is also appreciated :+1: 

We just want to get 3.0.0 out and move forward.

PR #5281, opened 2016-05-03
Fixed bug with FR translation

nnussbaum:
On the page : bolt/extend
Unescaped quote in a translation hurt the javascript.

### DIFF
--- a/app/resources/translations/fr/messages.fr.yml
+++ b/app/resources/translations/fr/messages.fr.yml
@@ -231,7 +231,7 @@
 "Long messages": "Messages longs"
 "Main configuration": "Configuration principale"
 "Maintenance": "Maintenance"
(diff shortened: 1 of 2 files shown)


bobdenotter:
Thanks for bringing this to our attention. I think we had better fix this by escaping the string in the source code though. That way it'll be 'fixed' for other languages too.

nnussbaum:
That would be a better idea :)

nnussbaum:
As translations are enclosed in double quotes, enclosing the translations results in double quotes too makes better sense to me

bobdenotter:
@nnussbaum Thanks for this fix! 👍 

I'll merge this in, and add a ninja-tweak to add an escape filter (`|e`) to it!

PR #5290, opened 2016-05-09
Change YAMLupdater regex to be non-greedy

SahAssar:
Change the YAMLupdater regex to be non-greedy. Example below:

If you have `canonical: 'https://example.com'` in a config the current regex matches `canonical: 'https` as the key as it is is greedy and will therefore create invalid YAML (that is not written to file, but instead exits nut with an error message).

The same issue seems to apply for `example: { test: test }` as it will there read `example: { test` as the key.


GawainLynch:
Works for me :+1:

PR #5292, opened 2016-05-09
[Feature] Automatic Translation Inclusion for Extensions

Raistlfiren:
At the current time translations are not automatically included by Bolt extensions. With this PR it allows for automatic inclusion of all translations included in the `translations` directory. They will need to follow standard filenames, otherwise it will default to ''. For example, the translations filename must be called `en.yml` or `sv.yml`,

Details
-------

- Bolt Version 3.1+

### DIFF
--- a/src/Extension/TranslationTrait.php
+++ b/src/Extension/TranslationTrait.php
@@ -6,31 +6,64 @@ use Bolt\Filesystem\Handler\DirectoryInterface;
 use Pimple as Container;
 
 /**
- * Twig function/filter addition and interface functions for an extension.
+ * Automatic translation inclusion for an extension based upon three factors.
+ *  - All translations are in the translations dub-directory of the extension
+ *  - Translations are named as en.yml, en_GB.yml, or etc... based upon the locale
  *
- * @author Carson Full <carsonfull@gmail.com>
- * @author Gawain Lynch <gawain.lynch@gmail.com>
+ * @author Aaron Valandra <amvalandra@gmail.com>
  */
 trait TranslationTrait
 {
+
+    /** @var array $translations */
+    private $translations = [];
 
     /**
      * Call this in register method.
      *
      * @internal
      */
-    final protected function addTranslations()
+    final protected function extendTranslatorService()
+    {
+        $app = $this->getContainer();
+
+        $app['translator'] = $app->share(
+            $app->extend(
+                'translator',
+                function ($translator) {
+                    $this->loadTranslationsFromDefaultPath();
+
+                    if ($this->translations) {
+                        foreach ($this->translations as $translation) {
+                            $translator->addResource($translation[0], $translation[1], $translation[2]);
+                        }
+                    }
+
+                    return $translator;
+                }
+            )
+        );
+
+    }
+
+    /**
+     * Load translations from every extensions translations directory.
+     * Filenames must follow common naming conventions - en.yml, en_GB.yml, otherwise
+     *  function will need to be modified.
+     */
+    private function loadTranslationsFromDefaultPath()
     {
         $app = $this->getContainer();
 
         $translationDirectory = $this->getBaseDirectory()->getDir('translations');
         if ($translationDirectory->exists()) {
             foreach ($translationDirectory->getContents(true) as $fileInfo) {
-                $filename = explode('.', $fileInfo->getFilename());
                 if ($fileInfo->isFile()) {
-                    $translation = isset($filename[0]) ? $filename[0] : '';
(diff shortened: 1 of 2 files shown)


GawainLynch:
Awesome mate! Thank you!

OK… just as an FYI… this (correctly) targeted master, so when ready, this can go in for 3.1.  There are a couple of things that need updating & fixing, but we're gearing up for release of 3.0.0 tomorrow so if you could give us a bit of time to flag stuff I personally would really appreciated it :smile_cat:

Raistlfiren:
Cool, sounds good! Thanks @GawainLynch

SahAssar:
Does this enforce filenames and locales like `sv.yml` instead of `sv_SE.yml`? Cause we have had issues when making that assumption before... For example pt_BR and pt_PT are very different.

Raistlfiren:
@SahAssar, no it does not. Basically all the method does is look for the `translations` directory(per extension), and loops through all the files while adding them to the translator service. It adds the translation based upon the filename. So `sv.yml` would be `sv` and `sv_SE.yml` would be `sv_SE`.

SahAssar:
@Raistlfiren Ah, great, thanks!

src/Extension/TranslationTrait.php
```diff
@@ -0,0 +1,48 @@
...
+        $app = $this->getContainer();
+
+        $translationDirectory = $this->getBaseDirectory()->getDir('translations');
+        if ($translationDirectory->exists()) {
+            foreach ($translationDirectory->getContents(true) as $fileInfo) {
+                $filename = explode('.', $fileInfo->getFilename());
+                if ($fileInfo->isFile()) {
+                    $translation = isset($filename[0]) ? $filename[0] : '';
+
+                    $app['translator']->addResource('yml', $app['paths']['extensionspath'] . DIRECTORY_SEPARATOR . $fileInfo->getPath(), $translation);
```
GawainLynch: First one… 

`$app['paths']` is Bolt v1 and we were going to remove in 3.0 but ran out of time.

From v2 onwards you should be using either `$app['filesystem']` (recommended for core stuff) or `$app['resources']` until the former takes over. e.g.
```
$app['resources']->getPath('extensions/ . $fileInfo->getPath());
```

src/Extension/TranslationTrait.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Bolt\Filesystem\Handler\DirectoryInterface;
+use Pimple as Container;
+
+/**
+ * Twig function/filter addition and interface functions for an extension.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
```
GawainLynch: Might want to update the docblock :wink:

src/Extension/TranslationTrait.php
GawainLynch: With `$app['translator']` you should be extending the service, not simply calling it here.

```
        $app['translator'] = $app->share(
            $app->extend(
                'translator',
                function ($translator) {
                    // $translator is $app['translator'] … do logic

                    return $translator;
                }
            )
        );
```

src/Extension/TranslationTrait.php
```diff
@@ -0,0 +1,48 @@
...
+trait TranslationTrait
+{
+
+    /**
+     * Call this in register method.
+     *
+     * @internal
+     */
+    final protected function addTranslations()
```
GawainLynch: In line with the other functions, the should be named something more like `extendTranslatorService`

src/Extension/TranslationTrait.php
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+    }
+
+    /** @return Container */
+    abstract protected function getContainer();
+
+    /** @return string */
+    abstract public function getName();
```
GawainLynch: This trait doesn't need `getName()` so can be safely dropped as-is

GawainLynch:
Sorry for the long delay on feedback, @Raistlfiren 

OK, two things… the tests can't run due to something that we've long fixed in master, so if you could rebase against current master that'd be awesome … in the event you need help, just call out on Slack or IRC.

Only thing that needs to change, and I can get it if you want/need, is this:
```
$path = $app['resources']->getPath('extensions' . DIRECTORY_SEPARATOR . $fileInfo->getPath());
```

…it should be using `$app['filesystem']`. There are a few or more examples in the extension traits if you feel like taking that on.

Otherwise, we're good to go!

Raistlfiren:
Thanks @GawainLynch. I have updated the code as you requested.

GawainLynch:
:+1:

PR #5294, opened 2016-05-10
Only regex match from beginning of multi-line

GawainLynch:
More than a slight performance gain on this one :-/


bobdenotter:
Yay, teamwork! \o/

bobdenotter:
Whoop! merging in!

SahAssar:
I think that this was intentional... See #4368

GawainLynch:
With 2 added JS files, it will raise backend load times from ~200ms to ~16000ms :-/

SahAssar:
o_0

Alright, that's a bit more than I expected...

PR #5295, opened 2016-05-10
Set list id to last id + 1 instead of list.length in file/imagelists

SahAssar:
Bug recreation:

* Have a imagelist with two images. (they will have the id's 0,1)
* Remove the first image. (the remaining image still has the id 1)
* Add one image. (bolt will try to assign it the id 1 since the list length is 1)
* Bolt will silently fail to add the image since one with the id already exists.

The fix is to sort the list, get the last image and increment it's id by one.

This bug only seems to be in 2.2, not 3.0 and master

### DIFF
--- a/app/src/js/upload-files.js
+++ b/app/src/js/upload-files.js
@@ -101,11 +101,16 @@ var FilelistHolder = Backbone.View.extend({
     },
 
     add: function (filename, title) {
+        this.list.sort();
+        var nextid = 0;
+        if (this.list.models.length) {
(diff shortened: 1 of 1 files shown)


SahAssar:
ping @rarila or if anyone else has a better solution.

SahAssar:
The tests failed because Travis couldn't restart nginx :laughing:

GawainLynch:
> The tests failed because Travis couldn't restart nginx

But I can … and did.

rarila:
@SahAssar I think I'm on your side here. If this patch works our dev time is better spent in 3.0+

Issue #5270, opened 2016-05-01
[BUG] Contenttypes with slugs in snake_case aren't getting handled correctly


robbert-vdh:
Contenttypes with slugs containing underscores (e.g. `foo_bars`) won't hydrate correctly. If the contenttype doesn't have an explicitly set slug Bolt will hydrate the entity incorrectly and throw an error when trying to save (Bolt will somehow convert the snake_case to kebab-case).

````yaml
foo_bars:
#    slug: foo_bars
    name: foobars
    singular_slug: foo_bar
    singular_name: foobar
    fields:
        thing:
            type: text
        otherthing:
            type: integer
````
````
Content {#2054 ▼
  #contenttype: ContentType {#2038 ▼
    #boltname: "foo_bars"
    #contentType: array:14 [▼
      "name" => "foobars"
      "singular_slug" => "foo_bar"
      "singular_name" => "foobar"
      "fields" => array:2 [▼
        "thing" => array:6 [▶]
        "otherthing" => array:6 [▶]
      ]
      "slug" => "foo-bars"
      "show_on_dashboard" => true
      "show_in_menu" => true
      "sort" => false
      "default_status" => "draft"
      "viewless" => false
      "liveeditor" => false
      "tablename" => "foo_bars"
      "allow_numeric_slugs" => false
      "groups" => []
    ]
  }
  #_legacy: ContentLegacyService {#2069 ▶}
  #id: 1
  #slug: "thing"
  #datecreated: Carbon {#2016 ▶}
  #datechanged: Carbon {#2015 ▶}
  #datepublish: Carbon {#2014 ▶}
  #datedepublish: Carbon {#2013 ▶}
  #ownerid: null
  #status: ""
  #relation: null
  #taxonomy: null
  #group: null
  #sortorder: null
  +_fields: array:1 [▼
    "username" => ""
  ]
  #_internal: array:1 [▶]
  +"templatefields": []
}
````
````
Bolt\Exception\StorageException thrown with message "Attempted to load mapping data for unmapped class foo-bars"
````

If the contenttype does have an explicitly set slug, then `Bolt\Legacy\Storage::getContentType($slug)` will return false and break every query.

````
[...]

PR #5298, opened 2016-05-11
Fix contenttype slugs with underscores not loading correctly

robbert-vdh:
This fixes `\Bolt\Legacy\Storage::getContentType` so that it works correctly with contenttype slugs containing underscores. `getContentType()` would always use Slugify to create a slug version of the queried `$contenttype` even if there was an exact match, causing some contenttypes to not load. I've also added a test covering this specific case.

Fixes: #5270


GawainLynch:
Closing this one, as #5299 will be merged down to master

PR #5299, opened 2016-05-11
Fix contenttype slugs with underscores not loading correctly

robbert-vdh:
This is a port of #5298 for the release/3.0 branch.

Fixes: #5298

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2326,15 +2326,12 @@ class Storage
         // See if we've either given the correct contenttype, or try to find it by name or singular_name.
         if (!$contenttype) {
             foreach ($this->app['config']->get('contenttypes') as $key => $ct) {
-                if (isset($ct['slug']) && ($contenttypeslug === $ct['slug'])) {
-                    $contenttype = $ct;
-                    break;
-                }
-                if (isset($ct['singular_slug']) && ($contenttypeslug === $ct['singular_slug'])) {
-                    $contenttype = $ct;
-                    break;
-                }
-                if ($contenttypeslug === $this->app['slugify']->slugify($ct['singular_name']) || $contenttypeslug == $this->app['slugify']->slugify($ct['name'])) {
+                if (
+                    (isset($ct['slug']) && ($contenttypeslug === $ct['slug'])) ||
+                    (isset($ct['singular_slug']) && ($contenttypeslug === $ct['singular_slug'])) ||
+                    $contenttypeslug === $this->app['slugify']->slugify($ct['name']) ||
+                    $contenttypeslug === $this->app['slugify']->slugify($ct['singular_name'])
+                ) {
                     $contenttype = $ct;
                     break;
                 }
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2314,8 +2314,6 @@ class Storage
      */
     public function getContentType($contenttypeslug)
     {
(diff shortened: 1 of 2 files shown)


robbert-vdh:
It seems I have included 969b7d73dc1fe4e87000a4c8935faec979f6583a by accident, I'll rebase again to sort that out.

GawainLynch:
@robbert-vdh are you comfortable to rebase this properly onto `release/3.0` or would you like me to do at merge time?

GawainLynch:
Beat me by a second :laughing:

robbert-vdh:
@GawainLynch Done, silly me didn't notice I was including those two commits. I guess I should make some more coffee.

bobdenotter:
Tests included! 💟

src/Legacy/Storage.php
```diff
@@ -2314,29 +2314,34 @@ private function parseWhereParameter($key, $value, $fieldtype = false)
...
+                        (isset($ct['singular_slug']) &&
+                            (($contenttypeslug === $ct['singular_slug']) || ($slugifiedContentType === $ct['singular_slug']))
+                        ) ||
+                        $slugifiedContentType === $this->app['slugify']->slugify($ct['name']) ||
+                        $slugifiedContentType === $this->app['slugify']->slugify($ct['singular_name'])
+                    ) {
+                        $contenttype = $ct;
+                        break;
+                    }
                 }
```
GawainLynch: So… this *is* legacy, but I can't help but think given everything we're dealing with here it would be easier to:

```
    public function getContentType($contentTypeSlug)
    {
        // Return false if empty, can't find it.
        if (empty($contentTypeSlug)) {
            return false;
        }

        $contentType = $this->app['config']->get('contenttypes/' . $contentTypeSlug);

        // See if we've either given the correct contenttype, or try to find it by name or singular_name.
        if (!$contentType) {
            // Also check for the slugified version of the content type
            $slugifiedContentType = $this->app['slugify']->slugify($contentTypeSlug);
            $contentType = $this->app['config']->get('contenttypes/' . $slugifiedContentType);

            if (!$contentType) {
                foreach ($this->app['config']->get('contenttypes') as $key => $ct) {
                    $isCtPlural = isset($ct['slug']) && ($contentTypeSlug === $ct['slug']) || ($slugifiedContentType === $ct['slug']);
                    $isCtSingle = isset($ct['singular_slug']) && ($contentTypeSlug === $ct['singular_slug']) || ($slugifiedContentType === $ct['singular_slug']);
                    $isCtPluralMod = $slugifiedContentType === $this->app['slugify']->slugify($ct['name']);
                    $isCtSingleMod = $slugifiedContentType === $this->app['slugify']->slugify($ct['singular_name']);

                    if ($isCtPlural || $isCtSingle || $isCtPluralMod || $isCtSingleMod) {
                        $contentType = $ct;
                        break;
                    }
                }
            }
        }

        if (!empty($contentType)) {
            return $contentType;
        } else {
            return false;
        }
    }
```

On one hand we lose OoE on the condition (and it is covering edge-case), but on the other it becomes a lot more readable.

robbert-vdh: I decided to go with the one liner because it's a tiny bit more efficient, but since we're only dealing with a maximum of three additional checks and four times the number of content types variable assignments I guess it won't matter much. If you guys find this easier to read I'll change it.

GawainLynch:
Thanks @robbert-vdh :+1:

PR #5305, opened 2016-05-12
Extend Page Exceptions

GawainLynch:
This should remove the Extend page dependency on Whoops being enabled, as well as preventing random useless garbage like:
image


bobdenotter:
Cool!

PR #5317, opened 2016-05-13
Set TCP/IP address columns to a max length of 45 characters 

### LINKED ISSUE #5316: [BUG] IPv6 adresses can be longer than the used column
Currently IP columns are up to 32 characters in length, but IPv6 can be up to [39 or 45 characters long](http://stackoverflow.com/a/7477384)

Details
-------

 - Relevant Bolt Version: 3.0
 - PHP version: 5.6.20
 - Used webserver: Apache

Reproduction
------------

Trying to log on with a connection with an IPv6 address that is longer than 32 characters returns an error.

Relevant files:

https://github.com/bolt/bolt/blob/release/3.0/src/Storage/Database/Schema/Table/AuthToken.php#L22  
https://github.com/bolt/bolt/blob/release/3.0/src/Storage/Database/Schema/Table/LogSystem.php#L24  
https://github.com/bolt/bolt/blob/release/3.0/src/Storage/Database/Schema/Table/Users.php#L22

GawainLynch:
Fixes #5316

:koala: :gun:


bobdenotter:
👍

PR #5318, opened 2016-05-13
Fix/upload limit calculation

### LINKED ISSUE #5170: [RFC] Limit on pictures?
There seems to be a limit of 2MB uploading a picture when creating a new entry. Most digital pictures are much bigger and this is rather inconvenient. This limit was not present in Bolt 2.2.

bobdenotter:
Fixes #5170


bobdenotter:
Thank you, @lenlorijn! Your suggestion was indeed key to the solution!

PR #5322, opened 2016-05-15
WIP Add # BEGIN/END CONTENTTYPE NAME

evertalbers:
I started doing this in my own templates to quickly recognize the contenttype separations with 'big' contenttypes. 

Admittedly, there's not much added value when scrolling down (you will see the contenttype name in the first line anyway) but it helps to know where you are when scrolling up.

See if you like the idea. If so, it could be added as a Geheimtip to the Docs, too.

### DIFF
--- a/app/config/contenttypes.yml.dist
+++ b/app/config/contenttypes.yml.dist
@@ -11,6 +11,8 @@
 # list the different types of services a company provides, you can group these
 # together.
 
+# BEGIN PAGES
+
 pages:
     name: Pages
     singular_name: Page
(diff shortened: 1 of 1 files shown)


evertalbers:
Or does the proper use of codefolding make this idea obsolete?

SahAssar:
Personally I'm against this, I think these things are best left to the individual dev, and are very much a matter of taste.

bobdenotter:
Since we all use editors with proper code highlighting / coloring (we do, right?), I agree that this change doesn't have much benefit. It's easy enough to see the different parts in one glance, I think.

GawainLynch:
Yeah, I have wavered backward and forward on this, but I think in the end it might hurt more than hinder in this case… Agreed with @bobdenotter

evertalbers:
I don't see how highlighting/coloring could help you see the name of the contenttype that is only mentioned 50 lines above, but folding could. 

The extensive Bolt user survey above shows that 67% of Bolt users does not see benefits, and 33% is not sure, so I'll close this one.

PR #5337, opened 2016-05-19
Before removing existing taxonomies, check the slugified value too.

### LINKED ISSUE #5283: [BUG] - Taxonomy lost after saving
Details
-------

 - Relevant Bolt Version: [ 2.2.* ]
 - Install type: [ Github checkout ]
 - PHP version: [ unknown ]
 - Used webserver: [ Apache ]
 - For UX/UI issues: [ Chrome / Safari ]

Reproduction
------------

Link for video with screencast: https://www.dropbox.com/s/ah3acd2ygrtplm0/Taxonomy_4_mei%202016%2009.01.04.mov?dl=0 

1. **Bug summary**

After saving the existing taxonomy is lost from the content-type.
When a taxonomy is declared, the user changes something in (for example) a text-field, saves, then the taxonomy is lost.

 2. **Specifics**: 

- Happens on bunkerdag.nl (also on testserver bunkerdag.twokings.eu).
- Happens in Bolt 2.2.20 but was also observed in 2.2.18.
- Method of installing Bolt? Unknown.
- Tested in Safari (9.*), Chrome (50.0.2661.94)
- See video for example.

 3. **Steps to reproduce**: 

Add a taxonomy. SAVE.
Change a text-field or something else . SAVE
Reload page with taxonomy.

 4. **Expected result**: 

I expected to see the taxonomy as saved

 5. **Actual result**: 

Result was an undefined taxonomy

bobdenotter:
Fixes #5283

Due to inconsistent use of `-` vs `_` in taxonomies, the values weren’t checked correctly before deleting. 

This fix isn’t pretty, but it should fix the problem for Bolt 2.2. 

I’ll check to see if this is also a problem in Bolt 3.

### DIFF
--- a/src/Storage.php
+++ b/src/Storage.php
@@ -2514,6 +2514,7 @@ class Storage
                 $slugkey = '/' . $configTaxonomies[$taxonomytype]['slug'] . '/' . $slug;
 
                 if (!in_array($slug, $newSlugsNormalised)
+                    && !in_array($this->app['slugify']->slugify($slug), $newSlugsNormalised)
                     && !in_array($valuewithorder, $newSlugsNormalised)
                     && !array_key_exists($slugkey, $newSlugsNormalised)) {
                     $this->app['db']->delete($tablename, array('id' => $id));


bobdenotter:
Confirmed that this is not an issue in 3.0.

GawainLynch:
Meh… let's just blame the :koala: and while we're at it #1555

PR #5342, opened 2016-05-20
Allow extra plugin for Ckeditor

WedgeSama:
For now, the `config.extraPlugin` of Ckeditor is set during the `initcke` without taking the default value.
The PR just check if there is a value in `config.extraPlugin` and if it does, it will not be ignore.

This allow you to set the default global `CKEDITOR.config.extraPlugin` without loosing it. (for example to add table advanced plugin).

If necessary, I can PR for 3.0 too.

### DIFF
--- a/app/src/js/modules/ckeditor.js
+++ b/app/src/js/modules/ckeditor.js
@@ -70,7 +70,7 @@
             config.resize_enabled = true;
             config.entities = false;
             config.fillEmptyBlocks = false;
-            config.extraPlugins = 'codemirror';
+            config.extraPlugins += (config.extraPlugins?',':'')+'codemirror'; // Allow to add extra plugin from default CKEDITOR.config.
 
             config.toolbar = list(
                 [                 { name: 'styles',      items: list( [                 'Format'        ],


bobdenotter:
Will this change mean you can add extra CKeditor plugins via `config.yml`? If so, could you update `config.yml.dist` to say so?

> If necessary, I can PR for 3.0 too.

Yes please. Bolt 2.2 is in "bugfix only mode". This means we're not adding new features to Bolt 2.2.x anymore.

SahAssar:
Also, perhaps some docs besides the config.yml.dist comment/key would be nice.

GawainLynch:
@SahAssar assigning you as sponsor for this.

@WedgeSama as mentioned above, 2.2 is in bug fix mode, so we're going to re-target this against current master (which will land soon in 3.1).

Over to you, @SahAssar :wink:

SahAssar:
@WedgeSama If you could retarget this to master it would be awesome, after it is merged we/I can tinker with getting a hook for extensions to be able to add ckeditor plugins too :)

GawainLynch:
Merged into master via c4a2efef889ddc6bafa7248d9319dfce5f39a561

PR #5344, opened 2016-05-20
Fix to working Bolt with PostgreSQL database.

mosiyash:
See diff.


GawainLynch:
Better approach in #5345

mosiyash:
Yes, it is right way.

GawainLynch:
No stress… Thank you for your effort :+1:

Issue #5357, opened 2016-05-22
[BUG] bolt_log_system collecting "garbage". (AssetTrait.php)


jan-t3000:
For example: 
`Couldn't add file asset '//fonts.googleapis.com/css?family=Roboto:400': File does not exist in either web://extensions/local/my/extension or /theme/my-theme/ directories.`

Or:
`Couldn't add file asset '/bolt-public/view/css/my.css': File does not exist in either web://extensions/local/my/extension or /theme/my-theme/ directories.`

I'm loading these assets in an extension via registerAssets() using setZone(Zone::BACKEND),
however, on each 'Frontend' visit these errors are getting logged.

jan-t3000:
concerning #5358 

To stop the bolt_log_system table from collecting these 'false' issues, for this instance, I see it as the following:
If a user decides to add an external resource asset (http://, https://, ... //) there is no need to check the filesystem for existence of this asset.

PR #5358, opened 2016-05-22
Update AssetTrait.php

jan-t3000:
Quick proposal...

Fixes parts of : #5357


bobdenotter:
Shouldn't we modify it to behave the same for other non-local assets.. So, starting with `http://` and `https://` as well?

GawainLynch:
Adjusted slightly and merged manually into the 3.0 branch

PR #5362, opened 2016-05-24
Handle json_array for Sqlite separately

### LINKED ISSUE #5251: [BUG] Can't update db - getting exception.
#### v3.00b3

I was adding some fields for SEO and meta generation to my table.

when i'm trying to perform db_check, it gives an exception:

text: 

>    An error occured while updating `bolt_articles`. The error is An exception occurred while executing 'ALTER TABLE bolt_articles ADD COLUMN share_image CLOB NOT NULL': SQLSTATE[HY000]: General error: 1 Cannot add a NOT NULL column with default value NULL

What am I doing wrong? **Cannot add a NOT NULL column with default value NULL** — why is it so?

GawainLynch:
Fixes #5251

OK… Not really a fan of this approach, but DBAL currently `\Doctrine\DBAL\Platforms\SqlitePlatform` still uses `\Doctrine\DBAL\Platforms\AbstractPlatform::getJsonTypeDeclarationSQL` which just returns `CLOB`… 

@SahAssar & @rossriley if you could sanity check on available platforms. Adding any array field type to an **existing** table (with records) will trigger the problem.


SahAssar:
WFM :)

bobdenotter:
Would you look at that! 🐑

Issue #4161, opened 2015-09-29
Default 'pages' content type name change has no effect on backend UI


sbonardt:
When I change the 'pages' content type name and singular_name fields to have them set in another language the change has no effect on the UI.

So when setting the values to this: (Dutch/NL names)

```
pages:
    name: Paginas
    singular_name: Pagina
    fields:
        title:
```

gives me a UI that still has the linktext in the backend sidebar as "Pages", Bekijk 'Pages', Nieuwe 'page'. 

But the slugs do get updated to `/pagina/` and `/paginas/`!  (see screenshot)

I expect the translated names to appear in the UI as well.

Note:
If I create a new content type "articles" and give theme the dutch name "artikelen"  and singular_name "artikel" these get printed properly in the backend UI. "Artikelen", Bekijk Artikelen, Nieuwe Artikel.

![screen shot 2015-09-29 at 10 01 06](https://cloud.githubusercontent.com/assets/5597447/10158190/a736367c-6690-11e5-810a-667d8178c6b4.png)

rarila:
All backend contenttype localisation is done using menu ``Settings: Translations | Contenttypes``. (Ends up in ``bolt/app/resources/translations/nl_NL/contenttypes.nl_NL.yml`` in your case)

sbonardt:
Exactly, that's the weird thing about it. "Artikelen" does get used in the UI. How is that possible if it's _not_ being used from the translation file "bolt/app/resources/translations/nl_NL/contenttypes.nl_NL.yml" as thats all commented out?

bobdenotter:
I still think this is an issue. @rarila: while technically you are correct, this issue is confusing and hard to understand for users. So in that regard it _is_ broken.

rarila:
Was looking that up at the moment. In [latest version](https://github.com/rarila/bolt/blob/master/app/resources/translations/nl_NL/contenttypes.nl_NL.yml#L77) is "Nieuwe pagina" in contenttypes.nl_NL.yml. So, @sbonardt What do you have there?

@bobdenotter I always told you it is a broken design to have localisation data in two places ;-) This gotta be fixed in 3.0

bobdenotter:
The thing is that sbonardt _explicitly_ defines the `name: ` for the contenttypes, and for _one_ of them the name is ignored, while for the _other_ the name is used as defined in `contenttypes.yml`. This is not something that can be explained to people.

sbonardt:
jep. that's it. :)

[...]

PR #5365, opened 2016-05-24
translate taxonomy labels

psychonaut:
Trivial patch, translate taxonomy labels and display them in record footer.

### DIFF
--- a/app/theme_defaults/_sub_taxonomylinks.twig
+++ b/app/theme_defaults/_sub_taxonomylinks.twig
@@ -2,9 +2,9 @@
     {% for type, values in record.taxonomy %}
         <em>
         {% if values|length < 2 %}
-            {{ config.get('taxonomy')[type].singular_name }}:
+            {{ __(config.get('taxonomy')[type].singular_name) }}:
         {% else %}
-            {{ config.get('taxonomy')[type].name }}:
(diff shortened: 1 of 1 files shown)


bobdenotter:
I'm not sure this is something that should be handled in this way.

A while ago we've had a _loooooong_ discussion about whether or not we should 'magically' translate labels that are defined in contenttypes and taxonomies. My opinion is that they shouldn't. If I say `label: Pineapple`, and in the frontend it'll turn into `Ananas`, without being able to control that, I'm going to get really annoyed. 

This (probably) needs to be handled somewhere. When we do, it needs to be done thouroughly and consistently. And most of all, there needs to be consensus. 

For reference: #4161

psychonaut:
From my point of view this is frontend interface part. Everything else is translated apart from these two words 'Categories' and 'Tags'. So you have for example 'More entries' translated to 'Więcej wpisów' and just couple of lines below you have 'Categories' and 'Tags'.

But I can see you point here. If this would be part of template then it wouldn't be a problem. But this is a part of code/application so it's more hassle with keeping patched version.

bobdenotter:
We've discussed this, and while we _do_ agree that this is a problem that needs fixing, we also agree that this is not the correct way to do it. 

Going forward, we'll start working on a more substantial fix. Feel free to join in on #4161, to participate. :-)

PR #5370, opened 2016-05-26
Unfinished sentence

FlorentATo:
The sentence was cut in the middle.
Side note: this header is obsoleted by Content Security Policy level 2 (cf. https://www.w3.org/TR/CSP11/#frame-ancestors-and-frame-options)


bobdenotter:
Thank you, good catch! 👍

PR #5377, opened 2016-05-27
Setting 'provided_link', allowing for more flexibility in "branding"

### LINKED ISSUE #5343: [RFC] Change how provided_by works by allowing to put an http url instead of forcing an e-mail address
Right now, the branding section in config.yml looks like this : 

    branding:
        name: Administration
        path: /admin
        provided_by: [ "contact@company.com", "Company" ]

However in many backends, the footer will redirect to a website and not an e-mail address
Why is mailto: automatically prepended there ?

For example I want to be able to do something like this : 

    branding:
        name: Administration
        path: /admin
        provided_by: [ "http://www.company.com", "Company" ]

And if I need a mail, I guess we could still do something like this:

    branding:
        name: Administration
        path: /admin
        provided_by: [ "mailto:contact@company.com", "Company" ]

Doing like so could have backward compatibility issues, so we could check if there is an @ in first argument and automatically append mailto there.

I think it would be a nice little quality of life change, as redirecting to an URL is common (and cool :smile: ) and it seems easy to implement.

bobdenotter:
~~WIP: Don't merge yet, needs some tests for the new bit of code.~~ 

Allow for using URLs as well as email addresses in the `branding` part of `config.yml`. Fixes #5343

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -228,15 +228,14 @@ accept_file_types: [ twig, html, js, css, scss, gif, jpg, jpeg, png, ico, zip, t
 
 # If you want to 'brand' the Bolt backend for a client, you can change some key
 # variables here, that determine the name of the backend, and adds a primary
-# support/contact link to the footer.
+# support/contact link to the footer.  Add a scheme, like `mailto:` or
+# `https://` to the email or URL.
 #
-# Additionally you can change the mount
-# point for the backend, either for convenience or to obscure it from prying
-# eyes.
+# Additionally you can change the mount point for the backend, either for
+# convenience or to obscure it from prying eyes.
 #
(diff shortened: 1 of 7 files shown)


bobdenotter:
Should be GTG now!

GawainLynch:
`BoltDataCollectorTest` needs updating… it make Travis go boom boom :stuck_out_tongue_winking_eye:

bobdenotter:
Ha, if they are set in the test, `parseGeneral()` in `Config.php` isn't run, so it won't be set correctly anymore: 

```
        // Set the link in branding, if provided_by is set.
        $general['branding']['provided_link'] = Html::providerLink(
            $general['branding']['provided_by']
        );
```

How should I fix this?

PR #5382, opened 2016-05-28
Cast to string in thumbnail handling, prevent "substr() expects parameter 1 to be string"

### LINKED ISSUE #5329: [BUG] Thumbnail must return a string value
Details
-------

 - Relevant Bolt Version: 3.0
 - Install type: Zip install
 - PHP version: 5.5
 - Used webserver: Apache

Reproduction
------------

1) Installed locally, inside the container Docker
2) Gave permission `chmod -R 777 .`
3) Created a root user
4) Logged
5) Displayed a page "Files"
6) Thumbnails do not show up
7) I went to the picture http://bolt.dev/thumbs/1000x1000r/agriculture-cereals-field-621.jpg
8) Displayed a following error:

> Catchable fatal error: Method Bolt\Thumbs\Thumbnail::__toString() must return a string value in /www/bolt.dev/vendor/symfony/http-foundation/Response.php on line 400

bobdenotter:
Fixes #5329

### DIFF
--- a/src/Helpers/Image/Thumbnail.php
+++ b/src/Helpers/Image/Thumbnail.php
@@ -212,7 +212,7 @@ class Thumbnail
     public function setScale($scale)
     {
         $valid = ['b', 'c', 'f', 'r'];
-        $scale = substr($scale, 0, 1);
+        $scale = substr((string) $scale, 0, 1);
         $scale = in_array($scale, $valid)
             ? $scale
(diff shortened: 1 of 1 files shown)


rarila:
I wonder if that shouldn't be fixed near the location where this non string value is generated. That way IMHO it looks a bit hackish as it doesm't fix the real cause of the problem.

The function documentation clearly states that a ``string`` is expected. It also could happen, that this cast later gets removed in a review as the cast looks superfluous.

*Greetings from the department of O.C.D.*. :innocent:

PR #5389, opened 2016-05-30
Use path that includes subdirectories

### LINKED ISSUE #5336: [BUG] Extend-page in backend not working if Bolt is installed in subdirectory.
For my local development setup, I'm using a vhost, setup as `http://bolt.local` and have my Bolt-related projects organised in sub-folders like `http://bolt.local/project1` and so on.

But it seems, when Bolt is installed in a subdirectory, opening the `Extend`-Page (`http://bolt.local/project1/public/admin/extend`) in backend results in the following error:

```
An unknown error occurred. This was the error message:
JSON.parse: unexpected character at line 1 column 1 of the JSON data
```
From what I can tell, the ajax calls are trying to fetch `http://bolt.local/admin/extend/installed` instead of `http://bolt.local/project1/public/admin/extend/installed`. The same goes for `http://bolt.local/admin/extend/installInfo?package=...`.

I know, with the move towards having a `public` folder, Bolt is not intended to be used this way, but as everything else seems to work fine, i thought, i'd report it.

Details
-------

 - Relevant Bolt Version: 3.0.1
 - PHP version: 7.0
 - Used webserver: Apache version 2.4.20

### LINKED ISSUE #5359: [BUG] Extension Manager does not use correct URL to fetch extension list json
When accessing the extension manager in the backend, I'm getting a JSON.parse Error. 
Bolt does not seem to take into account whether the Site is installed in the domain root or in a subfolder. The bolt admin is accessible under http://domain.tld/subfolder/admin/ but Bolt wants to fetch the extension list from http://domain.tld/admin/. 

 - Relevant Bolt Version: 3.0.1
 - PHP version: 5.5 

------------
bolt_extension_error

GawainLynch:
Fixes #5336
Fixes #5359

### DIFF
--- a/app/view/twig/extend/extend.twig
+++ b/app/view/twig/extend/extend.twig
@@ -197,7 +197,7 @@
     {{ data('extend.text.updated',        __('page.extend.message.updated')) }}
 
     {{ data('extend.siteurl',             context.site) }}
-    {{ data('extend.baseurl',             '/' ~ app['controller.backend.extend.mount_prefix']|trim('/') ~ '/') }}
+    {{ data('extend.baseurl',             app['resources'].url('bolt') ~ 'extend'|trim('/') ~ '/') }}
     {{ data('extend.rootpath',            paths.rootpath) }}
 
 {% endblock page_main %}


bobdenotter:
I'm sure a few people will appreciate this! 👍

PR #5393, opened 2016-06-01
Handle site root directory moves on Sqlite

GawainLynch:
We're going to re-do exception handling soon… but this will stop the corner-case when someone sets up a site locally and moves it, from biting.


bobdenotter:
Is this correct? It seems to me that the  `LowLevelDatabaseException` on line 235 / 247 will still be thrown.

GawainLynch:
The one in the `else {}`?

bobdenotter:
I didn't see anything! 😉

Issue #5395, opened 2016-06-01
[Discussion] New Storage Engine Fetches Whole Resultset when Querying pages/2


Raistlfiren:
As discussed with @rossriley on slack, the new storage engine fetches the whole result set instead of an individual item based upon the id. For example: `$page = $app['query']->getContent('pages/2');` returns all of the pages. Link to slack discussion - https://boltcms.slack.com/archives/general/p1464787773012213

Details
-------

 - Relevant Bolt Version: [ 3.0 + ]
 - Install type: [ All ]

Reproduction
------------
`$page = $app['query']->getContent('pages/2');`

As an alternative this can be used.

`$page = $app['query']->getContent('pages', ['returnsingle' => true, 'id' => 2]);`

PR #5401, opened 2016-06-02
Refinements for content fetching 

rossriley:
This implements the missing named record select in the new storage engine.

It includes the intelligence to drop into single record mode where appropriate and thus means the following queries will all work correctly.

`$page = $app['query']->getContent('pages/2'); // returns a single page with an id of 2`

`$page = $app['query']->getContent('pages/lorem-ipsum'); // returns a single page with aslug of 'lorem-ipsum'`

`$pages = $app['query']->getContent('pages,entries/2'); // returns multiple records where id is 2`

Fixes: #5395 

There's also corresponding test updates to reflect these changes, which should have no BC implications since it was functionality that was not correctly implemented before.


GawainLynch:
> There's also corresponding test updates to reflect these changes

^ this :+1: :smile:

PR #5408, opened 2016-06-06
Fix #5226: Placing the delay parameter accordingly to reduce the amount of cance…

### LINKED ISSUE #5226: [BUG] backend search function starts new searches on every key press -> dead server
The small search box in the header of bolt's backend executes a search request on every keypress. On systems with more than a couple of pages, a search request may take some seconds. Now when a user starts to type in a longer search text with eg 10 characters, this will execute 10 different search requests. The javascript cancels all requests befor sending a new one, but the PHP scripts will of course continue to run, leading to 10 concurrent search requests on your database.
This will kill many servers or at least slow down the websites for minutes.

The solution i implemented in a similar case was to block all further search requests until the currently running one really finished. In addition, not every key press should start a search but only a pause of maybe 800ms after the last keypress.

Details
-------

 - Relevant Bolt Version: 2.2.19-pl1
 - PHP version: 5.4
 - Used webserver: Apache 2.4
 - For UX/UI issues: Chrome 49 on OSX

Reproduction
------------

- Open the CMS and login
- Open a network proxy or the developer tools > tab network
- type away into the search field

Actual Result:
a list of search requests will popup, all canceled but the last one. And the server will slow down. A list of running PHP threads will show all search requests as still being processed and your Mysql db will show all queries still running.

Expected:
Only one process is running. No further request will be made until this one is finished.

rossgardt:
Placing the delay parameter accordingly to reduce the amount of cancelled XHRs (fixes #5226)

**Problem:**
Current implementation issues a server-request on every keystroke and is cancelling already running requests. This results an a overload of requests not being processed.

**Solution:**
The `delay` parameter is being placed within the `ajax` object to take affect.

**Reminder**:
[ ] Run `grunt updateBolt` and compile assets to make the fix take affect

Best regards,
Benedikt

### DIFF
--- a/app/src/js/modules/omnisearch.js
+++ b/app/src/js/modules/omnisearch.js
@@ -30,11 +30,11 @@
 
         $('.omnisearch select').select2({
             width: '100%',
-            delay: 250,
             placeholder: bolt.data('omnisearch.placeholder'),
             minimumInputLength: 3,
             multiple: true, // this is for better styling …
             ajax: {
+                delay: 600,
                 url: bolt.conf('paths.async') + 'omnisearch',
                 dataType: 'json',
                 data: function (params) {


GawainLynch:
:shipit:

GawainLynch:
@bobdenotter Can you do the rebuild… build chain is broken here right now… 'cause NPM :frowning:

Issue #5419, opened 2016-06-07
[BUG][NOT SURE]Bolt V3 Storage -> addHandler NOT calling addOperation


Raistlfiren:
When adding handlers to the new Bolt Storage level, it isn't calling addOperation() as well. I am not sure if this is intended or not.

Details
-------

 - Relevant Bolt Version: 3.0 | master
 - Install type: [ Github checkout | Zip/tarball install | Composer install ]
 - PHP version: [ 5.4 | 5.5 | 5.6 | 7.0 ]
 - Used webserver: [ PHP ]

Reproduction
------------
In a service provider I add a new handler:  

`$app['query.parser']->addHandler('pager', new PagingHandler());`
The new handler is not recognized when being calling `$query->getContent('pages/pager');`.

In order to get it working, it must be added to the operations list as so - 
`$app['query.parser']->addOperation('pager');`

Then it is recognized as an operation and executes it correctly. 

Proposed Solution
------------
```php
    /**
     * Adds a handler for the named operation.
     *
     * @param string   $operation
     * @param callable $callback
     */
    public function addHandler($operation, callable $callback)
    {
        $this->handlers[$operation] = $callback;
        $this->addOperation($operation);
    }
```

P.S.
-------
Should addOperation() throw an Exception instead of silently not adding the operation to the list? It would be easier to debug if it did IMHO.

SahAssar:
Fixed in #5423, no?

PR #5423, opened 2016-06-08
[FEEDBACK] Call addOperation in addHandler

Raistlfiren:
Information can be viewed about this PR in #5419.

Fixes: #5419

### DIFF
--- a/src/Storage/Query/ContentQueryParser.php
+++ b/src/Storage/Query/ContentQueryParser.php
@@ -320,7 +320,7 @@ class ContentQueryParser
     }
 
     /**
-     * Adds a handler for the named operation.
+     * Adds a handler AND operation for the named operation.
      *
      * @param string   $operation
      * @param callable $callback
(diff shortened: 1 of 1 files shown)


rossriley:
While you're pushing this, I've just noticed an anomaly in that the addOperation has a chance of not being added if the operation already exists, so just to be safe it's also best to call `removeOperation($operation)` too just before adding since the handler will get updated no matter what.

Raistlfiren:
I wonder if it isn't better to throw an exception instead of removing the operation. Otherwise, an extension may try to overwrite the functionality of a handler.

rossriley:
Maybe, but then is that such a bad thing, it means someone can write extensions that replace the built in functionality.

We could perhaps make it explicit and have a separate method for `replaceHandler` and then throw an exception in the `addHandler` method.

Raistlfiren:
Possibly. The biggest issue I see is that you can't really expand on the current handlers. If they could be inherited/expanded, then maybe. The biggest issue I see is that there might be new functionality added or bugs fixed in the current handlers (to core). Then the extension handlers are completely outdated and may not be able to handle the new functionality. 

I guess I could see a few rare cases where it would be nice to overwrite a default handler. I feel as though a new handler would be the best solution, and calling the new handler instead of using a "core" one.

GawainLynch:
:+1: from all

PR #5434, opened 2016-06-10
Fixed invalid ExtensionInterface namespace in Controller Resolver

maciejmiara:
There is serious issue in Bolt v3.0, resulting in 

> Reflection Exception: Interface \Bolt\Extensions\ExtensionInterface does not exist

This happens, when someone tries to register controller in extension.
Namespace for this interface has been changed for Bolt v3.0, but no one updated interface path in \Bolt\Routing\ControllerResolver on line 25.

Correct interface path is: \Bolt\Extension\ExtensionInterface.

For future I'd recommend usage of ::class constant, rather than typing paths to classes/interfaces by hand. Thanks to that, if someone forgets to update paths in one file, fatal error during composer execution will occur, which will make programmer aware of the problem.

### DIFF
--- a/src/Routing/ControllerResolver.php
+++ b/src/Routing/ControllerResolver.php
@@ -22,7 +22,7 @@ class ControllerResolver extends Silex\ControllerResolver
     protected function instantiateController($class)
     {
         $refCls = new \ReflectionClass($class);
-        if ($refCls->implementsInterface('\Bolt\Extensions\ExtensionInterface')) {
+        if ($refCls->implementsInterface('\Bolt\Extension\ExtensionInterface')) {
             /** @var \Bolt\Extensions\ExtensionInterface[] $extensions */
             $extensions = $this->app['extensions']->getEnabled();
 


src/Routing/ControllerResolver.php
```diff
@@ -22,7 +22,7 @@ class ControllerResolver extends Silex\ControllerResolver
     protected function instantiateController($class)
     {
         $refCls = new \ReflectionClass($class);
-        if ($refCls->implementsInterface('\Bolt\Extensions\ExtensionInterface')) {
+        if ($refCls->implementsInterface('\Bolt\Extension\ExtensionInterface')) {
             /** @var \Bolt\Extensions\ExtensionInterface[] $extensions */
             $extensions = $this->app['extensions']->getEnabled();
```
GawainLynch: `$app['extensions']` no longer implements `getEnabled()`

GawainLynch:
> This happens, when someone tries to register controller in extension.

I am curious as to how you're hitting this?

That name space in indeed wrong, but the use of the `['extensions']` service is also wrong

maciejmiara:
Please look at
https://github.com/bolt/bolt-extension-starter-extended/blob/master/src/ExtensionNameExtension.php
and 
https://github.com/bolt/bolt-extension-starter-extended/blob/master/src/Controller/ExampleController.php

I didn't find official documentation for creating Controllers in extension, so I based on Your example.

Method from my extension class:
`protected function registerFrontendControllers()
    {
        $config = $this->getConfig();
        return [
            '/user/login' => new Auth($config),
        ];
    }`

And whole Auth controller:
`<?php

namespace Bolt\Extension\XXX\YYY\Controller\User;

use Bolt\Controller\Frontend;
use Bolt\Translation\Translator;
use Silex\Application;
use Silex\ControllerProviderInterface;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;

class Auth implements ControllerProviderInterface
{
    /** @var array The extension's configuration parameters */
    private $config;
    
    /**
     * Initiate the controller with Bolt Application instance and extension config.
     *
     * @param array $config
     */
    public function __construct($config)
    {
        $this->config = $config;
    }

    /**
     * Specify which method handles which route.
     *
     * Base route/path is '/example/url'
     *
     * @param Application $app An Application instance
     *
     * @return ControllerCollection A ControllerCollection instance
     */
    public function connect(Application $app)
    {
        /** @var $ctr \Silex\ControllerCollection */
        $ctr = $app['controllers_factory'];
        // /example/url/template
        $ctr->get('/user/login', [$this, 'login'])
            ->bind('profile_login_pl');
        return $ctr;
    }

    /**
     * The default before filter for the controllers in this file.
     *
     * Refer to the routing.yml config file for overridding.
     *
     * @param Request $request The Symfony Request
     *
     * @return null|BoltResponse|RedirectResponse
     */
    public function before(Request $request)
    {
        return null;
    }

    public function after(Request $request)
    {
        return null;
    }

    public function login(Application $app, Request $request)
    {
        return $app['twig']->render("templates/user/auth/login.twig");
    }
}`

As You can see, I'm also forced to create before() and after() methods.
Extending \Bolt\Controller\Frontend is not possible due to 
> Call to a member function start() on a non-object
in \Bolt\Controller\Frontend line 53.

Maybe I'm doing sth wrong, then please tell me, what should I change.

GawainLynch:
The documentation [is here](https://docs.bolt.cm/3.0/extensions/building/controllers-routes), and the extension starter is also a valid example to follow.

Extending `Frontend` is probably not what you want to do, the problem you're hitting is likely that you're overriding the `connect()` method in that case without calling the parent (neither of which is shown in your example).

I use these methods in a lot of extensions without problem, so my guess is there is another bug in your code.

GawainLynch:
Merged into `release/3.0` and cleaned up additional use of the SP.

PR #5438, opened 2016-06-10
Updated Russian Translations

panda-madness:
Just some translations. A couple might need reworking, since the context of the message is not always obvious.

Details
-------

Not sure what to put here? I just edited some text files and made sure it was valid YAML.


panda-madness:
I feel like I did something wrong, I just wanted to suggest my own single commit...

GawainLynch:
Yes, it should have targeted the `release/3.0` branch, but that's not obvious … so my bad :disappointed: 

I'll take care of it this end.

GawainLynch:
Merged manually

Issue #4865, opened 2016-02-18
[RFC] Adding sortable multiselect and relations to core in v3.1


SahAssar:
At Intendit we have been making heavy use of @rossriley's awesome sortable relations extension and have found it very, very useful, however since #4340 the old way of doing sortable relations is no longer possible since the extension cannot get the original order of the relations in the save event.

I propose that we make it part of bolt proper. Both extensions are quite light and I think this would be a useable feature for many users of bolt.

GawainLynch:
Sounds like a job for… Ninja Ross! ;-)

:+1:

PhillippOhlandt:
:+1: just do it! :P

GawainLynch:
@PhillippOhlandt, just so we're on the same page, it's **late** freeze right now for 3.0… but for 3.1 absolutely.

PhillippOhlandt:
Sure, I meant 3.1:) I was just too excited xD

bobdenotter:
Fixed in #5441

SahAssar:
Just to be clear then, we are not going to implement sortable relations then? I'm totally okay with that and was going to suggest it anyways, but that's the idea, right?

bobdenotter:
I closed this during cleanup, but I might've been a bit too hastily. I assumed this was fixed by #5441, but it was only mentioned in there. 

I know @rossriley was planning to overhaul relations some day, so this will probably get included with that then, right?

SahAssar:
Well, sortable relations really seems more like a hack the more I've tried it out, and goes against the idea of relations (being bidirectional), so IMO it's better to leave relations without sortability, and focus on them being more like other fields, so this being closed makes sense.

PR #5441, opened 2016-06-11
Sortable Select Fields

rossriley:
This is Part 1 of the process discussed in #4865 

This PR adds the option to declare a select field as sortable, eg:

```
    selectentry:
        type: select
        values: entries/id,title
        autocomplete: true
        multiple: true            
        sortable: true
```

It also includes the select2 plugin and css that will be used to add sortable relations in a following PR.


bobdenotter:
Oooh, shiny! 🌟

rarila:
Shouldn’t we pull ``select2.sortable`` from github instead of adding it to our repo?

bobdenotter:
> Shouldn’t we pull select2.sortable from github instead of adding it to our repo?

If that's a possibility, yes, I think we should.

rossriley:
Well it's not an official plugin, one I half adapted and modified from one that  was compatible with the old version of select2.

bobdenotter:
Ah, right. Then let's merge this one in, right? :-)

rarila:
ah, ok

rarila:
Hmm, a tiny comment [here](https://github.com/bolt/bolt/blob/release/3.0/app/src/lib/Versions.md) where it was from and that it is patched would probably be nice for future reference

PR #5442, opened 2016-06-11
Fix duplicate button functionality in repeater groups

### LINKED ISSUE #5406: [Bug] Duplicate Set button not working
Clicking the Duplicate Set button has no effect.

Details
-------

 - Relevant Bolt Version: 3.03
 - Install type: Zip install
 - PHP version: 5.5.12
 - Used webserver: Apache 2.4.9
 - For UX/UI issues: Firefox 46.0.1 and Chrome 51 (win 7) in both normal and safe mode

Reproduction
------------

1. Clean install from zip. No dummy content, no extensions.
2. Add a new showcase in admin
3. Go to the repeater tab, fill in the first repeater and save the record
4. Click the  Duplicate Set button

**Expected Result**:
The repeater set should be duplicated.

**Actual Result:**

1. Nothing happens
2. Clicking a second time the button greys out, and hovering shows a not-allowed cursor
3. No console errors
4. Incidentally, on second click the Add New Repeater button at the bottom also greys out and gets a not-allowed cursor, but does work if you click it

rossriley:
This one fixes #5406 which was not fully implemented. Adds a new group after with the values copied across.


bobdenotter:
Noice!

PR #5444, opened 2016-06-11
Fix #5371 – Add an ability to delete a record to mobile version of backend

### LINKED ISSUE #5371: [RFC] Add an ability to delete a record to mobile version of backend
For me it's still not obvious - how to delete a record from narrow screen device. 

Also no buttons inside a record.
Am i missing something? Or it's a strategy?

rarila:
The action column is now visible on mobile with a stripped down edit button. (no rounded edges on the left, but ok …)

Not much screenspace lost but much functionality gained.

Fixes #5371


bobdenotter:
The lack of rounded edges is only _slightly_ annoying. We'll get to that when we re-do the backend. For now, this looks fine! 👍

bobdenotter:
For those interested, it looks like this: 

screen shot 2016-06-13 at 09 21 32

Boorj:
thanks, that was needed

Issue #5449, opened 2016-06-14
[BUG] Namespace not set in Filesystem Plugins


nikgo:
Namespace not set in Filesystem Plugins. This causes e.g. permissions in FileManager not work for different namespaces.

Details
-------

 - Relevant Bolt Version:  3.0.5 | master 
 - Install type:  Github checkout 
 - PHP version: 5.5 
 - Used webserver: Apache version 2.4.7 

Reproduction
------------

Use bolt master or release/3.0 branch:

 1. Go to Bolt File Manager `/bolt/files`
 2. Append a filesystem mount point to `/bolt/files` e.g `/bolt/files/cache` or `/bolt/files/app`
 3. You can view files from not allowed mount points via FileManager.

Only allowed prefixes (files,theme,config) in `\Bolt\Filesystem\FilePermissions` should authorized. 

Technical
-----------
This is caused because namespace not set in constructor of Filesystem Plugins which inherits from `\Bolt\Filesystem\Plugin\AdapterPlugin`. Default namespace is "files". So other namespaces are still authorized.

Solution
----------
PR is in work.

GawainLynch:
Nice find… Awaiting the PR :+1:

PR #5450, opened 2016-06-14
Hotfix filesystem plugins

nikgo:
This PR set correct mount point / namespace in AdapterPlugin before plugin methods executes.
Fixes: #5449 

Details
-------
- Add test to check correct namespace authorization over Authorized Plugin.

Questions / Remarks
--------
- If plugin method from an filesystem object is invoked, then `setFilesystem` is called. I think its right place to fix this here. 
- Parameter `namespace` in AdapterPlugin constructor is not used anymore (insofar as I can evaluate all code :wink: ).
- Suggestion: Filesystem manager should mount a filesystem object only once, otherwise mount point is overwrite and end in a wrong behaviour, too.

### DIFF
--- a/src/Controller/Backend/FileManager.php
+++ b/src/Controller/Backend/FileManager.php
@@ -150,8 +150,12 @@ class FileManager extends BackendBase
         $filesystem = $this->filesystem()->getFilesystem($namespace);
 
         if (!$filesystem->authorized($path)) {
-            $error = Trans::__('general.phrase.access-denied-permissions-view-file-directory', ['%s' => $path]);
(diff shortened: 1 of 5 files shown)


GawainLynch:
> Add test to check correct namespace authorization over Authorized Plugin.

Legendary :+1:

nikgo:
@GawainLynch Only a simple test, but effective :sunglasses:

GawainLynch:
@nikgo the acceptance tests are failing as are normal use-cases, e.g. `/bolt/files/themes` will 403.

GawainLynch:
:shipit:

PR #5453, opened 2016-06-15
I have a feeling we're now required to bump the version to 4.0.0…

bobdenotter:
See also #5312.


GawainLynch:
Took me a minutes… so obligatory #1555 callout

PR #5457, opened 2016-06-15
Don't evaluate an empty constraint

GawainLynch:
If an (local) extension's `composer.json` is missing `"require": { "bolt/bolt": "x.y" }` a cryptic exception will be thrown.


bobdenotter:
Green lights! 👍

PR #5461, opened 2016-06-17
More Z-index shuffling.

bobdenotter:


bobdenotter:
Don't merge yet.. Not working as it should.

bobdenotter:
Should be GTG now!

PR #5481, opened 2016-06-24
Do not catch exception when adding user with Nut

dantleech:
Currently, if you try and register a user with a short name, you get the following error:

```
./app/nut user:add admin admin admin@example.com admin admin
Error creating user: admin
```

Which is not very helpful. By letting the exception get thrown:

```
  [Bolt\Exception\AccessControlException]                           
  Can not save a password with a length shorter than 6 characters!
```

We could show the exception message, but I personally prefer to let exceptions be thrown, especially here as you can get their stack trace with `--verbose`. But happy to change.

### DIFF
--- a/tests/phpunit/unit/EventListener/StorageEventListenerTest.php
+++ b/tests/phpunit/unit/EventListener/StorageEventListenerTest.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace Bolt\Tests\EventListener;
+
+use Bolt\Storage\Entity\Users;
+use Bolt\Storage\EntityManager;
+use Bolt\Storage\Database\Schema\Manager;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Bolt\Logger\FlashLoggerInterface;
+use PasswordLib\Password\Factory as PasswordFactory;
+use Bolt\EventListener\StorageEventListener;
+use Bolt\Events\StorageEvent;
+use Bolt\Config;
+use Prophecy\Argument;
+use PasswordLib\Password\Implementation\Blowfish;
+
+class StorageEventListenerTest extends \PHPUnit_Framework_TestCase
+{
+    private $user;
+    private $entityManager;
+    private $config;
+    private $manager;
+    private $urlGenerator;
+    private $flashLogger;
+    private $passwordFactory;
+    private $listener;
+    private $storageEvent;
+
+    public function setUp()
+    {
+        $this->user = $this->prophesize(Users::class);
+        $this->entityManager = $this->prophesize(EntityManager::class);
+        $this->config = $this->prophesize(Config::class);
+        $this->manager = $this->prophesize(Manager::class);
+        $this->urlGenerator = $this->prophesize(UrlGeneratorInterface::class);
+        $this->flashLogger = $this->prophesize(FlashLoggerInterface::class);
+        $this->passwordFactory = $this->prophesize(PasswordFactory::class);
+
+        $this->listener = new StorageEventListener(
+            $this->entityManager->reveal(),
+            $this->config->reveal(),
+            $this->manager->reveal(),
+            $this->urlGenerator->reveal(),
+            $this->flashLogger->reveal(),
+            $this->passwordFactory->reveal(),
+            5
+        );
+
+        $this->storageEvent = $this->prophesize(StorageEvent::class);
+    }
+
+    /**
+     * It should throw an exception if the password is shorter than 6 characters.
+     *
+     * @expectedException Bolt\Exception\AccessControlException
+     * @expectedExceptionMessage Can not save a password with a length shorter than 6 characters!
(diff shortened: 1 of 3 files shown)


GawainLynch:
Yeah, works for me. I don't suppose you could fix the one test that this breaks too?

dantleech:
Updated. I removed the test which tested the exception catching.

Tests were failing on 5.5 for an unrelated dependency problem:

```
  Problem 1
    - Installation request for lstrojny/phpunit-function-mocker dev-master@dev -> satisfiable by lstrojny/phpunit-function-mocker[dev-master].
    - lstrojny/phpunit-function-mocker dev-master requires php >=5.6.0 -> your PHP version (5.5.33) does not satisfy that requirement.
```

GawainLynch:
> Tests were failing on 5.5 for an unrelated dependency problem:

Yep, see #5484 for the details on that.

Otherwise, thank you!

GawainLynch:
Just a slight tweak/process request. That test is still valid and shouldn't be removed, but can be simply:

```php
    public function testFailure()
    {
        $app = $this->getApp();
        $command = new UserAdd($app);
        $tester = new CommandTester($command);

        $this->setExpectedException('Bolt\Exception\AccessControlException', 'Can not save a password with a length shorter than 6 characters!');

        $tester->execute(
            [
                'username'    => 'koala',
                'displayname' => '',
                'email'       => '',
                'password'    => '',
                'role'        => '',
            ]
        );
    }
```

dantleech:
> That test is still valid and shouldn't be removed

I thought about that, but imo this is the wrong level to test it, we are just testing the domain model by proxy, throwing that exception is not a responsiblity of this class. I would move the test back to the model if it does not already exist. wdyt?

GawainLynch:
OK, I don't know what I did with my last reply… but here goes again :laughing: 

Agree that our tests need a lot of work, and if you're willing to move that test to a better location as part of this PR please be my guest. But that test was added to catch a real bug that was reported so needs to exists… somewhere.

dantleech:
Would it be enough to add a unit test for the `Bolt\EventListenerBolt\EventListener\StorageEventListener` class which emmites this error?

Going further, this could this be better handled by a [validator](http://symfony.com/components/Validator) on the model?

GawainLynch:
> Would it be enough to add a unit test for the `Bolt\EventListenerBolt\EventListener\StorageEventListener` class which emmitts this exception?

Absolutely! :+1: 

> Going further, could this be better handled by a validator on the model?

Potentially, please feel very welcome to RFC a proposed approach… Gives everyone visibility :smile:

tests/phpunit/unit/EventListener/StorageEventListenerTest.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->passwordFactory = $this->prophesize(PasswordFactory::class);
+
+        $this->listener = new StorageEventListener(
+            $this->entityManager->reveal(),
+            $this->config->reveal(),
+            $this->manager->reveal(),
+            $this->urlGenerator->reveal(),
+            $this->flashLogger->reveal(),
+            $this->passwordFactory->reveal(),
+            5
```
dantleech: The number of unused dependencies for this test probably makes a good case for breaking the logic out of here.

GawainLynch: Yes, we still have much to do… 3.0 was over a year to get this far, and more iterations are needed. All continued help most welcome.

dantleech:
Added a unit test -- I made no attempt here to copy existing style - as wanted to show how I would do this, can change as required.

Notable I am using Prophecy for mocking, and I do not extend the BoltTestCase - as that test case would make this a *functional* test.

tests/phpunit/unit/EventListener/StorageEventListenerTest.php
GawainLynch: Just updating my above point … before I get jumped on :laugh: … please also note that, process-wise, direction changes should first be discussed in an RFC. 

We're pretty open to improving things generally, just not without direction and oversight.

dantleech: ok cool. I wil try adapt this to the standard form from the other tests and use this code in an RFC

GawainLynch:
> Notable I am using Prophecy for mocking, and I do not extend the BoltTestCase - as that test case would make this a functional test.

Our tests need love, but we must keep the level of testing intact (regardless of mixed approaches currently).

For this I have no problem, and also prefer the approach but if you would like to expand on this approach, a *short* RFC on the subject would get it moving. Also if you're keen to jump on some of this (and you're help is wanted, have I mentioned that yet :wink: ), please feel free to join the dev meetings on IRC on Tuesday nights.

GawainLynch:
Failure just only due to #5484 … otherwise GTG.

PR #5319, opened 2016-05-14
Don't emit autoloader warnings from class_exists()

GawainLynch:
I know we all hate using the `@` silencer, but when removing an extension manually things get annoying and the warnings are note useful in this context… 'cause PHP.

```
PHP Warning:  include([snip]extensions/vendor/composer/../clippy/clippy/src/ClippyExtension.php): 
failed to open stream: No such file or directory in [snip]vendor/composer/ClassLoader.php on line 412
PHP Stack trace:
PHP   1. {main}() [snip]app/nut:0
PHP   2. require() [snip]app/nut:14
PHP   3. call_user_func:{[snip]app/bootstrap.php:131}() [snip]app/bootstrap.php:131
PHP   4. Bolt\{closure}() [snip]app/bootstrap.php:131
PHP   5. Bolt\Application->initialize() [snip]app/bootstrap.php:127
PHP   6. Bolt\Application->initExtensions() [snip]src/Application.php:116
PHP   7. Bolt\Extension\Manager->addManagedExtensions() [snip]src/Application.php:256
PHP   8. Bolt\Extension\Manager->addManagedExtension() [snip]src/Extension/Manager.php:170
PHP   9. class_exists() [snip]src/Extension/Manager.php:235
PHP  10. spl_autoload_call() [snip]src/Extension/Manager.php:235
PHP  11. Composer\Autoload\ClassLoader->loadClass() [snip]src/Extension/Manager.php:235
PHP  12. Composer\Autoload\includeFile() [snip]vendor/composer/ClassLoader.php:301
PHP Warning:  include(): Failed opening '[snip]extensions/vendor/composer/../gawain/clippy/src/ClippyExtension.php' 
for inclusion (include_path='[snip]vendor/phpunit/dbunit:[snip]vendor/symfony/yaml:.:/usr/share/pear:/usr/share/php') 
in [snip]vendor/composer/ClassLoader.php on line 412
PHP Stack trace:
PHP   1. {main}() [snip]app/nut:0
PHP   2. require() [snip]app/nut:14
PHP   3. call_user_func:{[snip]app/bootstrap.php:131}() [snip]app/bootstrap.php:131
PHP   4. Bolt\{closure}() [snip]app/bootstrap.php:131
PHP   5. Bolt\Application->initialize() [snip]app/bootstrap.php:127
PHP   6. Bolt\Application->initExtensions() [snip]src/Application.php:116
PHP   7. Bolt\Extension\Manager->addManagedExtensions() [snip]src/Application.php:256
[...]

PR #5485, opened 2016-06-25
Do not hide error

dantleech:
I ran into problems when writing an extension, where there was no error message. It turns out it was repressed. This PR removes the error hiding from `class_exists` so that I can now see my error:

```
PHP Fatal error:  Declaration of DTL\Bolt\Extension\Fixtures\DtlBoltFixturesExtension::register(Bolt\Application $app) must be compatible with Silex\ServiceProviderInterface::register(Silex\Application $app) in /home/daniel/www/fnair_asso_sf/extensions/local/dtl/bolt-fixtures/src/DtlBoltFixturesExtension.php on line 13
```

(which is good).

I do not know why that `@` was there?

### DIFF
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -232,7 +232,7 @@ class Manager
     private function addManagedExtension(PackageDescriptor $descriptor)
     {
         $className = $descriptor->getClass();
-        if (@class_exists($className) === false) {
+        if (class_exists($className) === false) {
             if ($descriptor->getType() === 'local' && @class_exists('Wikimedia\Composer\MergePlugin') === false) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
Very specifically no. This breaks the three groups principle from the manifesto.

All for handling and reporting the error though, but it shouldn't bring down a site.

dantleech:
Is it not worse that the site shows a whitescreen of death and no error? What non-fatal error would be caught here?

GawainLynch:
Tagging #5319 for reference

GawainLynch:
@dantleech I am still having no luck in reproducing your problem … have you got a simple test case I can use to explore covering both bases?

GawainLynch:
Handled in #5565 :+1:

PR #5492, opened 2016-06-29
Add option for google maps api key

SahAssar:
See https://github.com/bolt/bolt/wiki/Dev-meeting-2016-06-28 for more info.

Issue #5484, opened 2016-06-25
Unit tests failing on PHP 5.5


GawainLynch:
Our `require-dev` component `lstrojny/phpunit-function-mocker` just bumped their required version to PHP >=5.6.0 causing Travis to break on PHP 5.5 tests.

Upstream issue c.f https://github.com/lstrojny/phpunit-function-mocker/issues/6

GawainLynch:
Fixed upstream, with thanks to @lstrojny

PR #5493, opened 2016-06-29
Lock lstrojny/phpunit-function-mocker to 0.3.0 for PHP 5.5 support

GawainLynch:
See #5484 
See #5495
See #5492 fails


SahAssar:
Please see #5495 for countdown/details.

PR #5497, opened 2016-07-01
[2.2] Enforce relative schema on Twig {{ url() }} calls

GawainLynch:
If a site's admin area is served over strict HTTP2 certain URLs are generated with `http://`. This forces the `{{ url() }}` call to return a relative schme to work around this

See [upstream](https://github.com/symfony/twig-bridge/blob/2.8/Extension/RoutingExtension.php#L48-L51)


bobdenotter:
👍

PR #5500, opened 2016-07-02
Hyphen vs underscore … :koala: vs :koala: … there can only :leaves: one

### LINKED ISSUE #5363: [BUG] Contenttype vs. Table name "_" and "-"
Hey guys,

I'm back working with bolt for the v3 upgrade.
At first: Thank you very much for the nice peace of software! You are doing a good job.

My Problem is that the contenttypes `name` seem to be transformed from `_` to `-` which causes a *undefined index* warning.

My questions are: 
Do I need to rename the contenttype or is there a fix? 
Is this known or even a feature instead of a bug? 
What's the best way to solve it?

Details
-------

- Bolt version: 3.0.2

Reproduction
------------

1) Create a content type like: `form_feedbacks`

2) Use bolt forms and setup the write to database through the contenttype:
```yml
  database:
    contenttype: form_offer_checks
```
3) Take a look in the logs:
> PHP message: PHP Notice:  Undefined index: form_feedbacks in /var/www/vhosts/bolt/vendor/bolt/bolt/src/Storage/Database/Schema/Builder/ContentTables.php on line 33"

4) Dump info:
*Bolt\Storage\Database\Schema\Builder\ContentTables::getSchemaTables()*
```php
    public function getSchemaTables(Schema $schema, Config $config)
    {
        $contentTypes = $config->get('contenttypes');
        dump($this->tables->keys(), $contentTypes);exit;
```
You will see that the table key is `form_feedbacks` and the contenttype key is `form-feedbacks`.

GawainLynch:
Fixes #5363

### DIFF
--- a/src/Storage/Database/Schema/Builder/ContentTables.php
+++ b/src/Storage/Database/Schema/Builder/ContentTables.php
@@ -15,6 +15,9 @@ use Doctrine\DBAL\Schema\Table;
  */
 class ContentTables extends BaseBuilder
 {
+    /** @var array */
+    protected $tableSchemas;
+
     /**
      * Build the schema for Bolt ContentType tables.
      *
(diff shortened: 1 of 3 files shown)


bobdenotter:
Looks good to me, disregarding Travis.

PR #5503, opened 2016-07-03
Initialize variable

dantleech:
When developing with a high level this uninitialized var causes lots of errors on the search page.


GawainLynch:
Merged into 3.0 via 9acadc9077d1d4f0f0a37243d0937435bd409977

Thanks again, @dantleech :+1:

PR #5518, opened 2016-07-07
Refactor the hydration event to use an ArrayObject

rossriley:
Apologies, this is a replacement for #5510 the approache used there wasn't successful since the GenericEvent in symfony was dereferencing the subject.

This is a better approach, since it ensures that the object will always be passed by reference down the line.


GawainLynch:
Ha! I did think the last one wouldn't work, but erred on the side of "Ross has it in hand" :wink: 

Nice stuff mate!

rossriley:
Obviously coded with too little beer.

next time I'll drink harder.

PR #5520, opened 2016-07-08
Updating dependencies for base-2016.

bobdenotter:


GawainLynch:
Here's hoping this won't be a #1555

PR #5522, opened 2016-07-08
[3.1] Extension repo button

GawainLynch:
image


PhillippOhlandt:
👍

bobdenotter:
I like the idea to be be able to do that, but I'm not too fond of adding another spaced button. It's becoming a bit too much, I think. How about we make this into a split/pull down button, like this: 

screen shot 2016-07-08 at 15 17 12

Then we add the pull-down under what is now 'README', with two options "Go to Repository" and "View on extensions.bolt.cm". That way we have the extra functionality, but it doesn't clutter the screen as much.

GawainLynch:
I totally agree @bobdenotter, and meant to tag this as RFC due to the visuals … and we know how far Gawain stays away from UI :wink: 

I'll bug people after work hours for some HTML help

GawainLynch:
@bobdenotter in the "stay away from UI, Gawain" department … what do you think about this ASCII-frame for the button layout:

```
Details v
    View "read me"
    View on marketplace
    Source repository
```

GawainLynch:
BTW… my thinking for having a "Details" top-level is that quite often there is no README, and the button is suppressed.

GawainLynch:
Thinking about it more … see, this is why we keep me away from UI … the config button is the same, it isn't guaranteed to be there in all extensions…

So … thoughts on something more like this then:
```
Extension Details v
    View README
    Edit configuration
    View on marketplace
    View source repository
```

bobdenotter:
I very much like the 'details' button, that can just show what's underneath it! 👍

bobdenotter:
Meging in!

Issue #5506, opened 2016-07-03
[Storage] Custom repository for Content


dantleech:
Currently it is not possible to use a custom repository for a contenttype AND have the legacy functionality of the content records.

For example, if I use a custom repository for `pages` then `record.link` does not work.

This is due to the in-between state of the storage layer (see comment [here](https://github.com/bolt/bolt/blob/release/3.0/src/Storage/EntityManager.php#L282)).

I realize this is a temporary state, but just thought I would bring it up as it is somewhat of an obstacle when it comes to keeping code organized.

@rossriley what is the plan here? does it depend on other refactorings? (e.g. routing) or could it be fixed in a minor release?

GawainLynch:
Closed via #5523

PR #5523, opened 2016-07-08
Allow callables to be registered as repository classes

rossriley:
This allows a callable factory to be registered as a Repository. It should hopefully allow a workaround for scenarios such as in #5506 where you may want to return a different repository instance based on the contenttype.


GawainLynch:
Should this go into 3.1?

PR #5525, opened 2016-07-09
Translation cache option

nikgo:
If you develop extensions and work with translation files, it's tiring to clear cache all the time.

Now you can set `general/caching/translations` to `false` if you need it. Recommend only for extension development. Cache is enabled by default.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -155,6 +155,10 @@ recordsperpage: 10
 #                  cases you should *NOT* enable this, because it will cause
 #                  side-effects if the website shows different content to
 #                  authenticated users.
+# - thumbnails:    Caches thumbnail generation.  
+# - translations:  Caches translation files. It is recommend to leave this 
+#                  enabled. Only if you develop extensions and work with 
+#                  translation files you should turn this off.   
 caching:
     config: true
     templates: true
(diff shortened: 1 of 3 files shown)


GawainLynch:
I am guessing … and I literally mean guessing … that the test fails here have something to do with the call to `$app['translator.caching'] = $app['config']->get('general/caching/translations');` not setting a default, but the result is strictly tested against a `false` where as an unset value will return a `null` meaning that `if ($app['translator.caching'] === false) {` will only pass the check if explicitly set.

nikgo:
Thanks to look at this. For reference, how the cache options are implemented, I have checked the `ThumbnailsServiceProvider` and `TwigServiceProvider`:

That the check passes only `if ($app['translator.caching'] === false)` is by design. Any other value (or option is unset) should leave cache enabled. In the failing test is translation cache disabled via provider registration with `'translator.cache_dir'`. This is no longer possible with my changes, because `'translator.cache_dir'` is overwriten by a closure. 

Now I have set a default value and would change the test function `registerTranslationServiceWithCachingDisabled` (similiar like in `TwigServiceProviderTest`) to pass the test and it work. 

Or there are other / better solutions?

GawainLynch:
> Or there are other / better solutions?

I think this is good enough, and @CarsonF agreed, so… 5… 4… 3…

nikgo:
2... 1... merged in. All right! :sunglasses:

Issue #5392, opened 2016-06-01
[BUG] viewless setting in contenttype also affects backend omnisearch


annejan:
According to related issue #2149 setting `viewless: true` should cause:
- the contenttype is not associated with any template
- the contenttype does not produce any frontend routes
- there is no 'Preview' button in the backend for this contenttype
- the contenttype does not appear in any search results or listings on the frontend

Unfortunately in practice this also means that the content of that type is hidden from the (backend) omnisearch.
## Details
- Relevant Bolt Version: 2.2
- Install type: Composer install
- PHP version: 5.6
- Used webserver: Apache
## Reproduction

Setting `viewless: true` on a contenttype hides it from backend omnisearch.
This is unwanted behaviour in my opinion.

I have currently patched this on my instance with the following crude patch.

```
--- vendor/bolt/bolt/src/Storage.php    2016-05-31 17:35:44.000000000 +0200
+++ vendor/bolt/bolt/src/Storage.php    2016-06-01 11:21:28.000000000 +0200
@@ -783,9 +783,7 @@
             $contenttypes = array_filter(
                 $contenttypes,
                 function ($ct) use ($appCt) {
-                    if ((isset($appCt[$ct]['searchable']) && $appCt[$ct]['searchable'] === false) ||
-                        (isset($appCt[$ct]['viewless']) && $appCt[$ct]['viewless'] === true)
-                    ) {
+                    if (isset($appCt[$ct]['searchable']) && $appCt[$ct]['searchable'] === false) {
                         return false;
                     }

```

SahAssar:
@annejan You want to PR that patch? :smile:

SahAssar:
Closing this, see #5531 for more info.

Issue #2149, opened 2014-11-11
[RFE] Viewless content types


tobias2k:
Proposal: add a property `viewless` to `contenttypes.yml` (or, alternatively, allow explicitly setting the `template` properties to `null` - discuss!), which indicates that:
- the contenttype is not associated with any template
- the contenttype does not produce any frontend routes
- there is no 'Preview' button in the backend for this contenttype
- the contenttype does not appear in any search results or listings on the frontend

The purpose of such contenttypes is to provide manageable "auxiliary" content that can be referenced from other content records and templates (either through `getContent` / `{% setcontent %}`, or by means of Relationships), but doesn't make sense on its own. Popular examples include snippets of information that you may want to share between proper content records, but that don't deserve frontend pages of their own, such as contact addresses, content blocks for headers and footers, etc.

Technically, all that is required to do this is add the property, and check it in a few strategic places to block the things listed above.

Regarding naming, I propose the term **viewless contenttypes**; suggestions for a better name are extremely welcome.

PR #5531, opened 2016-07-11
Don't hide viewless from backend omnisearch

annejan:
Setting viewless: true on a contenttype hides it from backend omnisearch.
This is unwanted behaviour in my opinion.

Fixes: #5392 

According to related issue #2149 setting viewless: true should cause:

the contenttype is not associated with any template
the contenttype does not produce any frontend routes
there is no 'Preview' button in the backend for this contenttype
the contenttype does not appear in any search results or listings on the frontend
Unfortunately in practice this also means that the content of that type is hidden from the (backend) omnisearch.

### DIFF
--- a/src/Storage.php
+++ b/src/Storage.php
@@ -783,9 +783,7 @@ class Storage
             $contenttypes = array_filter(
                 $contenttypes,
                 function ($ct) use ($appCt) {
-                    if ((isset($appCt[$ct]['searchable']) && $appCt[$ct]['searchable'] === false) ||
-                        (isset($appCt[$ct]['viewless']) && $appCt[$ct]['viewless'] === true)
-                    ) {
+                    if (isset($appCt[$ct]['searchable']) && $appCt[$ct]['searchable'] === false) {
                         return false;
                     }
 


GawainLynch:
Test failures are unrelated, and seem to be from Travis builds pulling in phpunit 5 despite having the constraint `~4.4`

SahAssar:
Did some monkeytesting and this also seems to allow viewless records to be searched on the frontend...

@annejan could you have a look at that? IIRC you need to do a zone-check to see if viewless should or should not be allowed.

annejan:
Will have to check that out later today, realised we don't have front-end search in project I created this patch for so never noticed.

bobdenotter:
@annejan Gentle nudge, could you look into this?

GawainLynch:
@annejan another gentle nudge, it would be good to get this happening soon if you're able.

SahAssar:
Closing this one as it has been half a year since it was opened. If the fix for the frontend/backend search is fixed please either reopen, rebase and retarget this or submit a new PR :)

Issue #5534, opened 2016-07-11
[BUG] unable to choose different session storage methods, filesystem hard-coded


dwolfhub:
In the SessionServiceProvider, when gathering options for the Session classes in `registerOptions(Application $app)`, the method ends by adding `$app['session.options']`, effectively hard-coding these values and not allowing me to override them in any way. By switching this line and the ones adding `$app['session.storage.options']` (lines 168-170), we can overwrite the option values in our `.bolt.php` files and customize the way sessions are handled. (PR incoming)

dwolfhub:
unfortunately this still doesn't solve my original issue, which was that i wanted to use native php redis sessions... it seems like there could be some work done to this provider to add more flexibility based on different hosting situations.

GawainLynch:
Added a comment on #5535 regards that.

As per our conversation on Slack yesterday, this ended up having to be finished too quickly to push 3.0.0 out, so if you have time to dig deeper that would be really useful.

CarsonF:
The real problem is bigger than this. Silex/Pimple was designed to define services and parameters _first_, then override default values in a configuration phase. Bolt does/tries to do the opposite, configure everything first, then add services/parameters. It's a huge nightmare on a lot of fronts.

Unfortunately, the right way and our way had to be bridged somewhere, which is where `configure` comes in. Either way though I would expect `session.options` to be overridden after the service provider is registered if needed.

---

`.bolt.php` is still in a new concept. I want to explore it more and add more features, but as @GawainLynch said it was decided to release 3.0. I want to keep that file as small as possible though. It should only be used to bridge your paths/app to the common web endpoint (`index.php`) and `nut`. Features should be added to this file very cautiously. It shouldn't replace a config file.

CarsonF:
I'm not sure where `session` gets invoked and therefore freezing the config parameters.
What would work for now is subclassing `Bolt\Application`, and overriding `initSession`. You can update `session.options` before the `session` service gets invoked.

The end goal is clearly defined loading phases where you can override services and parameters in one spot. But we are in transition still from a Service Locator pattern to Dependency Injection, which causes pain for this exact problem.

ToBe998:
I just stumbled on this too. Our Bolt2 instances use a clustered memcache session handler. In Bolt2, you could just put `session_use_storage_handler: false` into your config.yml and it worked.
Now in Bolt3, this option is no longer used.

[...]

PR #5535, opened 2016-07-11
Change order that options are added in SessionServiceProvider

dwolfhub:
Change order that options are added in SessionServiceProvider to allow custom options

Fixes: #5534

### DIFF
--- a/src/Provider/SessionServiceProvider.php
+++ b/src/Provider/SessionServiceProvider.php
@@ -165,11 +165,11 @@ class SessionServiceProvider implements ServiceProviderInterface
                     }
                 }
 
+                $options->add($app['session.options']);
+
                 if (isset($app['session.storage.options'])) {
                     $options->add($app['session.storage.options']);
                 }
-
-                $options->add($app['session.options']);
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
@dwolfhub I am not sure this is going to do anything, as `$app['session.options']` is (currently) hard coded as an empty array, so the `->add()` is just going to do an `array_replace()` affecting matching values

@CarsonF can you remember the intention here, i.e. is `$app['session.options']` supposed to be a catch-all bucket, i.e. ultimately being able to override previously set properties from `$app['session.storage.options']` & friends?

CarsonF:
Commented on #5534 

```php
if (isset($app['session.storage.options'])) {
  $options->add($app['session.storage.options']);
}
```
This was an attempt to maintain BC, it really shouldn't be used going forward.

GawainLynch:
@CarsonF assigning you, but I'll co-sponsor this one with you if you need.

dwolfhub:
I'm gonna close this one and open another one for feedback.

Thanks.

PR #5538, opened 2016-07-13
[WIP] [FEEDBACK] Session Provider

dwolfhub:
Before I jump into writing tests and setting up environments to test all these different storage methods, I wanted to get some feedback and see if I'm headed down the right path.

I know the provider was in a state of flux and so if I'm missing the point of the rebuild please let me know, I can pivot.

Overall, though, the idea behind this is to allow someone to configure sessions to be stored in any way they choose, without needing to hack in a `.bolt.php` file.

Cheers

Fixes: #5534

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -85,7 +85,7 @@ class Application extends Silex\Application
         $this->register(new Provider\SessionServiceProvider());
 
         $this->extend('session.bag.options', function (OptionsBag $optionsBag, Silex\Application $app) {
-            $optionsBag->set('filesystem', $app['config']->get('general/session/handler', 'filesystem'));
+            $optionsBag->set('handler', $app['config']->get('general/session/handler', 'filesystem'));
 
             return $optionsBag;
         });
--- a/src/Application.php
+++ b/src/Application.php
@@ -85,10 +85,22 @@ class Application extends Silex\Application
         $this->register(new Provider\SessionServiceProvider());
 
         $this->extend('session.bag.options', function (OptionsBag $optionsBag, Silex\Application $app) {
-            $optionsBag->set('handler', $app['config']->get('general/session/handler', 'filesystem'));
+            // Backwards compatibility
+            // @todo remove?
+            if ($app['config']->get('general/session_use_storage_handler', false)) {
+                $optionsBag->set('handler', 'null');
+
+                return $optionsBag;
+            }
+
+            // Plug in settings from config
+            foreach ($app['config']->get('general/session', []) as $key => $value) {
+                $optionsBag->set($key, $value);
+            }
 
             return $optionsBag;
         });
+
     }
 
     public function initialize()
--- a/src/Application.php
+++ b/src/Application.php
@@ -7,6 +7,7 @@ use Bolt\Events\MountEvent;
 use Bolt\Provider\LoggerServiceProvider;
 use Bolt\Provider\PathServiceProvider;
 use Bolt\Provider\WhoopsServiceProvider;
+use Bolt\Session\OptionsBag;
 use Cocur\Slugify\Bridge\Silex\SlugifyServiceProvider;
 use Silex;
 use Symfony\Component\Stopwatch;
@@ -81,9 +82,13 @@ class Application extends Silex\Application
 
     protected function initSession()
     {
-        $this
-            ->register(new Provider\SessionServiceProvider())
-        ;
+        $this->register(new Provider\SessionServiceProvider());
+
+        $this->extend('session.bag.options', function (OptionsBag $optionsBag, Silex\Application $app) {
+            $optionsBag->set('filesystem', $app['config']->get('general/session/handler', 'filesystem'));
+
+            return $optionsBag;
+        });
     }
 
     public function initialize()
--- a/src/Provider/SessionServiceProvider.php
+++ b/src/Provider/SessionServiceProvider.php
@@ -12,6 +12,7 @@ use Bolt\Session\SessionListener;
 use Bolt\Session\SessionStorage;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
+use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBag;
 use Symfony\Component\HttpFoundation\Session\Flash\FlashBag;
 use Symfony\Component\HttpFoundation\Session\Session;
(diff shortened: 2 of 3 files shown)


GawainLynch:
@CarsonF temporarily assigning you, if you could have a look over as this was your baby … assign back to me when done if you like

GawainLynch:
One comment for when this is closer to merge.

We're about to try to move as much "commented" functionality out of the YAML files as we can soon, so probably better married with a [PR to docs](https://github.com/bolt/docs) when ready to go… more than happy to help on that one.

dwolfhub:
Good to know. I had assumed there would need to be some good docs around this as well.

src/Provider/SessionServiceProvider.php
```diff
@@ -3,373 +3,193 @@
...
+                $app['monolog'],
+                new MetadataBag()
+            );
+        });
+
+        if ($app->offsetExists('session.storage.handler') === false) {
+            $app['session.storage.handler'] = $app->share(function (Application $app) {
+                $storageConfigOption = $app['config']->get('general/session/handler', 'filesystem');
+
+                if ($app['session.test'] === true or $storageConfigOption === 'null') {
```
CarsonF: These handlers could all be broken out into their own DI keys

CarsonF:
@dwolfhub These services shouldn't be relying on the `config` service. That's the whole point of `session.options`. `config` should only be used to provide values to DI parameters (simple keys, not services). With these changes so far it's harder to break out into a separate repo because of this coupling.

CarsonF:
> I know the provider was in a state of flux and so if I'm missing the point of the rebuild please let me know, I can pivot.

Respectfully, I think you are missing the point. Really the only thing I wasn't happy with was `parseConnections` and how the handlers were tied to the options. 

I see what you are going for with having each handler have it's config section. I'm fine with that and debated it myself. I was also attempting to maintain BC, since this was written when we were targeting v2.3.

Can the `session.handler_factory.*` be modified to accept these config values (decoupled through a DI parameter)?

CarsonF:
Maybe it's confusing, and I know there are no docs for this. Currently you can do this:

```php
$app['session.options']['save_handler'] = 'redis' // any handler name

// Choosing "redis" invokes this factory
$app['session.handler_factory.redis']($options) // $options is $app['session.options']
```

So really `session.options` could be hooked up directly to a section of the config.

dwolfhub:
Thanks for the feedback.

Decoupling the configuration makes perfect sense. I think I had trouble understanding why it was necessary to add so any new elements to the DI container, when most of this logic, including the handler factory, was strictly used in the provider itself.

I'm going to take another crack at it, now that we're more aligned.

CarsonF:
@dwolfhub Good question. It allows the individual handlers to be overridden/added without having to modify the "routing" logic. For example, with your changes, if wanted to change something about `FilesystemHandler` I would have to override the entire $app['session.storage.handler']. This may not be a big deal if you are an end user/dev but say an extension or some other library wants to make more of a discrete change this is impossible (without assuming service is the way it's defined in core and copying it completely).

Feel free to ask away since I seem to be the only one with knowledge in this area, happy to have others on board! You can also ping me (`@carson`) on slack if you want.

dwolfhub:
updates:
- add all the things to container
- decoupling config / options bag

todo:
- initSession
- config
- comments
- tests
- docs

am i getting closer?

src/Provider/SessionServiceProvider.php
```diff
@@ -10,366 +10,253 @@
...
+                    [
+                        'host' => '127.0.0.1',
+                        'port' => false
+                    ]
+                ],
+                'database' => 0,
+                'dir' => false,
+                'gc_divisor' => 100,
+                'gc_maxlifetime' => 1440,
+                'gc_propability' => 1,
```
GawainLynch: Typo here, should be `gc_probability`

src/Provider/SessionServiceProvider.php
```diff
@@ -10,366 +10,253 @@
...
+                'handler' => 'filesystem',
+                'lazy_write' => false,
```
GawainLynch: Do we really want to disable this as a default?

src/Provider/SessionServiceProvider.php
```diff
@@ -10,366 +10,253 @@
...
+                return new SessionListener($app['session'], $app['session.bag.options']);
             }
         );
 
-        $this->registerOptions($app);
+        $app['session.storage'] = $app->share(function (Application $app) {
+            return new SessionStorage(
+                $app['session.bag.options'],
+                $app['session.storage.handler'],
+                $app['session.storage.random_generator'],
```
GawainLynch: Typo?

GawainLynch:
@dwolfhub I haven't got time to dig much today, but this kills a few things like request sessions for non-predis installs.

@CarsonF if you have some time. :wink:

src/Provider/SessionServiceProvider.php
dwolfhub: good catch

src/Provider/SessionServiceProvider.php
dwolfhub: this was the default where it's used, in `Bolt\Session\SessionStorage`. @CarsonF may be able to chime in.

src/Provider/SessionServiceProvider.php
dwolfhub: sorry where?

src/Provider/SessionServiceProvider.php
GawainLynch: Ah, no stress… running on 2 hours sleep here :laughing:

dwolfhub:
@GawainLynch thanks for the once over!
> this kills a few things like request sessions for non-predis installs.  

can you explain this a bit further? i'm probably missing something simple

src/Provider/SessionServiceProvider.php
GawainLynch: `'session.storage.random_generator'` isn't defined

dwolfhub: oh right. thx

GawainLynch:
Yeah, I quickly booted Bolt from this branch and had to fiddle with a bit to get past exceptions being thrown, but somehow it was preventing the session object being associated with the request, e.g. [this returns a null](https://github.com/bolt/bolt/blob/release/3.0/src/EventListener/StorageEventListener.php#L132);

dwolfhub:
ah, forgot to add the session listener to the dispatcher, thanks.

src/Application.php
```diff
@@ -81,9 +82,13 @@ protected function initConfig()
...
-        $this
-            ->register(new Provider\SessionServiceProvider())
-        ;
+        $this->register(new Provider\SessionServiceProvider());
+
+        $this->extend('session.bag.options', function (OptionsBag $optionsBag, Silex\Application $app) {
+            $optionsBag->set('handler', $app['config']->get('general/session/handler', 'filesystem'));
+
+            return $optionsBag;
+        });
```
dwolfhub: @CarsonF before i continue with more options, is this what you had in mind?

PR #5547, opened 2016-07-15
Hide that we know the future in userlist (fix null in lastseen on user displaying as `now()`)

### LINKED ISSUE #5546: [BUG] User last seen value continues to increment for never logged in users
Ideally the value should display "Never".

Details
-------

 - Relevant Bolt Version: 3.0.10
 - For UX/UI issues: Chrome

Reproduction
------------

1: Create a new user, and never log into it.
2: Observe the menu under `Configuration` >> `Users & Permissions`
3: Refresh the page and observe the "Last seen" value being `in a few seconds`.

SahAssar:
Fixes another issue reported in slack. We can't tell people that we know the future...

The issue was that a null in the lastseen column gave the current date when run through the date filter, which moment.js then turned into "in a few seconds".

Oh, and this fixes #5546

### DIFF
--- a/app/view/twig/users/_userlist.twig
+++ b/app/view/twig/users/_userlist.twig
@@ -27,7 +27,7 @@
                 </td>
 
                 <td class="hidden-xs">
-                    {{ user.lastseen|date('Y') > '1901' ? buic_moment(user.lastseen) : '-' }}
+                    {{ user.lastseen ? buic_moment(user.lastseen) : '-' }}
                 </td>
 
                 <td class="contenttypes">


Maelstrome26:
Doctor Who has been suppressed! Rabble rabble rabble!

SahAssar:
The failures I'm getting are the same I'm getting in a fresh branch (without this commit), so I'd assume it's not related but if it is I'll fix it... Just lemme know!

GawainLynch:
Test failures unrelated … and this can never have happened …

PR #5550, opened 2016-07-17
Remove the Foreign Key Constraint properties from diffs when supportsForeignKeyConstraints() is false

GawainLynch:
See:
  *  https://github.com/doctrine/dbal/pull/242
  *  https://www.sqlite.org/foreignkeys.html


SahAssar:
:train: :train:

PR #5552, opened 2016-07-18
Add scripts to run grunt without global

SahAssar:
This makes it so that you don't need to have grunt installed globally to be able to run the backend JS/CSS build. Instead of `grunt` one can (optionally) run `npm run grunt` or `npm start`. You can also run the subscripts by doing `npm start -- prepareCkeditor` or `npm run grunt -- prepareCkeditor`.

Of course it will still work if one wants to use the global version, but this makes it easier if someone has multiple projects that require different versions of grunt.

This is how it's setup in base-2016, see #5189

ping @rarila

### DIFF
--- a/app/src/package.json
+++ b/app/src/package.json
@@ -46,6 +46,8 @@
         "select2": "~4.0.3"
     },
     "scripts": {
+        "start": "grunt",
+        "grunt": "grunt",
         "install": "napa"
     },
     "napa": {


GawainLynch:
You would do this after I had to tear out the system Node.js packages to be able to compile Bolt's stuff :laughing:

rarila:
@SahAssar Jo!

GawainLynch:
:de: has spoken!

rarila:
@GawainLynch This was just a personal comment and was not meant to be an official statement reflecting the opinion of all Germans!

PR #5554, opened 2016-07-18
fix #5346 [Bug] Using Html 'base' tag breaks pagination links

### LINKED ISSUE #5346: [Bug] Using Html 'base' tag breaks pagination links
The way Paging works is not compatible with the use of html 'base' tag.

they are many use case where we must define the html base tag  within the html head tag.: 
example : 
`<base href="{{ paths.rooturl }}">`

The problem is : 
given a content type "entries", 

```
entries:
    name: Entries
    singular_name: Entry
    fields:
    ...
    listing_records: 10
```

we can list 10 of them with this url http://site.com/entries, which match the default route 

```
contentlisting:
    path: '/{contenttypeslug}'
    defaults:
        _controller: 'Bolt\Controllers\Frontend::listing'
    requirements:
        contenttypeslug: 'Bolt\Controllers\Routing::getPluralContentTypeRequirement'
```

We just have to use the {{ pager() }} twig function in our template to generate the pagination.

By default the html generated by the app/theme_defaults/sub_pager.twig template is something like that : 

```
<div class="pagination pagination-centered ">
  <ul class="pagination">
    <li class="active"><a href="?&page_entries=1">1</a></li>
    <li><a href="?&page_entries=2" rel="noindex, follow">2</a></li>
    <li><a href="?&page_entries=2" rel="noindex, follow">›</a></li>
  </ul>
</div>
```

you can notice that the generated link href attributes does not contain the full url ( it's a relative link ).
So how works a relative link  ?
if the url is : 
http://site.com/entries , the relatives links should target to : http://site.com/entries?page_entries=XXX
EXCEPT if we have defined an html base tag, in that case, the relative link use the base tag url ... as base.

So it's quite usual to define the html base tag to the root url, as i write abose : 
`<base href="{{ paths.rooturl }}">` 
==> which generate 
`<base href="http://site.com">` 

SO in that case the pagination links target to http://site.com/?page_entries=XXX and ... that's not the good url. ( it does not contains the content type part (entries), so the link do not paginate anything... and redirect to the root, in that case. )

We can fix that behavior in the Bolt\Pager class, within the makeLink() method.
I think that method should retrieve the current "script" url and use it as prefix before adding the pagination parameters.

Just to be quick, this is the actual Pager::makeLink() implementation: 

```
/**
     * Used for calling from template to build up right paginated URL link.
     *
     * @return string
     */
    public function makelink()
    {
        /*
         * If link set directly that forces using it rather than build
         */
        if ($this->link) {
            return $this->link;
        }

        $pageid = static::makeParameterId($this->for);
        $parameters = $this->app['request']->query->all();
        if (array_key_exists($pageid, $parameters)) {
            unset($parameters[$pageid]);
        } else {
            unset($parameters['page']);
        }

        $parameters[$pageid] = '';
        $link = '?' . http_build_query($parameters);

        return $link;
    }
```

I think we can just add something like that : 

```
...
        $parameters[$pageid] = '';
        $link = $this->app['paths']['current'] . '?' . http_build_query($parameters);

        return $link;
}
```

where $this->app['paths']['current'] contain the '/entries' part.
That way pagination links are no longer fully relative, but relative to the root ( so that  links are not relying on the base tag and still works. )

Any thoughts on that ? 

Details
-------
 - Relevant Bolt Version: [ >= 2.2 & 3.0 ]

dadaxr:
fix #5346 [Bug] Using Html 'base' tag breaks pagination links

### DIFF
--- a/src/Pager.php
+++ b/src/Pager.php
@@ -54,7 +54,8 @@ class Pager extends \ArrayObject
         }
 
         $parameters[$pageid] = '';
-        $link = $this->app['paths']['current'] . '?' . http_build_query($parameters);
+        $currentUrl = $this->app['resources']->getUrl('current');
+        $link = $currentUrl . '?' . http_build_query($parameters);
 
         return $link;
     }
--- a/src/Pager.php
+++ b/src/Pager.php
@@ -54,7 +54,7 @@ class Pager extends \ArrayObject
         }
 
         $parameters[$pageid] = '';
-        $link = '?' . http_build_query($parameters);
+        $link = $this->app['paths']['current'] . '?' . http_build_query($parameters);
 
         return $link;
     }


SahAssar:
IMO this is good for release/2.2 since it fixes a bug with <base>. @dadaxr Does this bug also exist in 3.0?

dadaxr:
nop only in bolt 2 :)

SahAssar:
Btw, those test failures look unrelated to me, but I've falsely made that assumption before... :D

@GawainLynch ?

src/Pager.php
```diff
@@ -54,7 +54,7 @@ public function makelink()
         }
 
         $parameters[$pageid] = '';
-        $link = '?' . http_build_query($parameters);
+        $link = $this->app['paths']['current'] . '?' . http_build_query($parameters);
```
GawainLynch: `$app['paths']` is a hangover from 0.x days and should not be used.

See `$app['resources']->getPath()`

GawainLynch:
> Btw, those test failures look unrelated to me, but I've falsely made that assumption before

Yeah, for a reason that I haven't gotten to the bottom of yet, Travis is pulling in v5 of PHPUnit when we  `"require": "~4.4"`

src/Pager.php
dadaxr: :+1: done :)

GawainLynch:
:+1:

Issue #5487, opened 2016-06-28
[RFC] Testing


dantleech:
In this RFC I want to propose a clear distinction between Unit and Functional tests and advocate the use of Prophecy.
## Unit and Functional testing
- [Unit tests](https://en.wikipedia.org/wiki/Unit_testing): Tests where only the test unit is instantiated - all dependencies are mocked. In general they always extend `PHPUnit_Framework_TestCase` directly.
- Functional/integration tests: Tests which are executed "from the container/application", **no mocks** are used (definition of functional testing is subjective).

Currently (afaict) bolt uses a Hybrid approach, it has the `BoltTestCase` which creates some mocks AND boots the container/application.

I propose that there be a clear separation such that the tests directory looks like:

```
tests/
    /unit
    /functional
    /...
```

(we do this in the CMF, [here for example](https://github.com/symfony-cmf/routing-auto-bundle/tree/master/Tests))
## Prophecy

[Propehcy](https://github.com/phpspec/prophecy) is a (now native) alternative to PHPUnit's mocking system. The following is an example of a unit test:

``` php
class ResourceEnhancerTest extends \PHPUnit_Framework_TestCAse
{
    public function setUp()
    {
        $this->registry = $this->prophesize(RegistryInterface::class);
        $this->urlGenerator = $this->prophesize(UrlGeneratorInterface::class);
        $this->requestConfigurationFactory = $this->prophesize(RequestConfigurationFactory::class);
        $this->requestStack = $this->prophesize(RequestStack::class);
        $this->enhancer = new ResourceEnhancer(
            $this->registry->reveal(),
            $this->requestStack->reveal(),
            $this->requestConfigurationFactory->reveal(),
            $this->urlGenerator->reveal()
        );

        $this->cmfResource = $this->prophesize(CmfResource::class);
        $this->metadata = $this->prophesize(Metadata::class);
    }

    /**
     * It should return true if the payload is supported by Sylius.
     */
    public function testSupportedBySylius()
    {
        $this->cmfResource->getPayloadType()->willReturn('stdClass');
        $this->registry->getByClass('stdClass')->willReturn($this->metadata->reveal());
        $this->assertTrue(
            $this->enhancer->supports($this->cmfResource->reveal())
        );
    }
}
```

With prophecy you describe (or "prophesize") the behavior of objects.

I personally find it far easier to work with than the old PHPUnit mock framework.

GawainLynch:
[...]

PR #5559, opened 2016-07-20
[WIP] Test reorganization

dantleech:
Fixes https://github.com/bolt/bolt/issues/5487 and as discussed breifly at an IRC meeting.

This PR aims to reorganize the tests, with the specific aim of making a diferentiation between `functional` and `unit` tests.

In this case defined as:

- **functional**: Requiring the bolt application to be booted.
- **unit**: Extends `\PHPUnit_Framework_TestCase`; generally no non-mocked dependencies.

Currently all tests are functional tests under a folder named `unit`.

Practical benefits of this reorganization will be allowing and encourging unit tests to be written - encouraging because people may be less likely to write functional tests when a unit test would be more suitable (sometimes it may be desirable to have both).

Unit tests are many times faster than functional tests, and as they do not depend on application state, are much more able to fully test all code paths in any given class.


dantleech:
Doh. Wrong branch.

PR #5560, opened 2016-07-20
[WIP] Test Reorganization

dantleech:
Fixes https://github.com/bolt/bolt/issues/5487 and as discussed breifly at an IRC meeting.

This PR aims to reorganize the tests, with the specific aim of making a diferentiation between `functional` and `unit` tests.

In this case defined as:
- **functional**: Requiring the bolt application to be booted.
- **unit**: Extends `\PHPUnit_Framework_TestCase`; generally no non-mocked dependencies.

Currently all tests are functional tests under a folder named `unit`.

Practical benefits of this reorganization will be allowing and encourging unit tests to be written - encouraging because people may be less likely to write functional tests when a unit test would be more suitable (sometimes it may be desirable to have both).

Unit tests are many times faster than functional tests, and as they do not depend on application state, are much more able to fully test all code paths in any given class.

### DIFF
--- a/tests/phpunit/nutty
+++ b/tests/phpunit/nutty
@@ -3,9 +3,11 @@
 
 namespace Bolt\Tests;
 
+use Bolt\Tests\BoltFunctionalTestCase;
+
 require_once __DIR__ . '/bootstrap.php';
 
-class Nutty extends BoltUnitTest
+class Nutty extends BoltFunctionalTestCase
 {
     public function getApp($boot = true)
     {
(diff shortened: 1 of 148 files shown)


GawainLynch:
@dantleech are you still planning on working on this?

If so, there is a small but growing number of tests that are starting to switch to Prophecy so the approach might need a bit of tweaking … not to mention a rebase :wink: 

If not, feel free to close this PR but leave the issue open as we're in agreement that this needs to transition but without a dedicated person driving it, it is going ot take a bit of time :smile:

dantleech:
Hey @GawainLynch , yeah, its probably a bit late for this one, will close!

GawainLynch:
Thanks for you effort so far though mate, I have really fallen in love with Prophecy as a result :+1:

PR #5561, opened 2016-07-20
Session Hotfixes

dwolfhub:
Hotfixes to SessionProvider->parseConnections and redis session handler.

Setting up Redis for sessions and i had to extend the SessionProvider and the RedisHandler to make hotfixes. Figured I would submit a patch as well.

### DIFF
--- a/src/Provider/SessionServiceProvider.php
+++ b/src/Provider/SessionServiceProvider.php
@@ -20,6 +20,7 @@ use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\HttpFoundation\Session\Storage\Handler\MemcachedSessionHandler;
 use Symfony\Component\HttpFoundation\Session\Storage\Handler\MemcacheSessionHandler;
 use Symfony\Component\HttpFoundation\Session\Storage\MetadataBag;
+use GuzzleHttp\Psr7\parse_query;
 
 /**
  * Because screw PHP core.
@@ -347,9 +348,13 @@ class SessionServiceProvider implements ServiceProviderInterface
             }
         } elseif (isset($options['save_path'])) {
             foreach (explode(',', $options['save_path']) as $conn) {
-                $conn = new ParameterBag($conn);
-                $conn->set('uri', new Uri($conn));
-                $toParse[] = $conn;
+                $uri = new Uri($conn);
+
+                $connBag = new ParameterBag();
+                $connBag->set('uri', $uri);
+                $connBag->add(parse_query($uri->getQuery()));
+
+                $toParse[] = $connBag;
             }
         }
 
--- a/src/Session/Handler/RedisHandler.php
+++ b/src/Session/Handler/RedisHandler.php
@@ -25,7 +25,7 @@ class RedisHandler implements \SessionHandlerInterface, LazyWriteHandlerInterfac
      */
     public function __construct($redis, $maxlifetime)
     {
-        if (!$redis instanceof Redis || !$redis instanceof Predis) {
+        if (!$redis instanceof Redis && !$redis instanceof Predis) {
             throw new \InvalidArgumentException('Argument must be an instance of Redis or Predis\ClientInterface');
         }
         $this->redis = $redis;


GawainLynch:
Test failures are unrelated and fixed in #5557

src/Session/Handler/RedisHandler.php
```diff
@@ -25,7 +25,7 @@ class RedisHandler implements \SessionHandlerInterface, LazyWriteHandlerInterfac
      */
     public function __construct($redis, $maxlifetime)
     {
-        if (!$redis instanceof Redis || !$redis instanceof Predis) {
+        if (!$redis instanceof Redis and !$redis instanceof Predis) {
             throw new \InvalidArgumentException('Argument must be an instance of Redis or Predis\ClientInterface');
```
GawainLynch: Style nitpick, but could you flip the `and` to `&&` for code base consistency.

GawainLynch:
@dwolfhub does this achieve what you were needing for #5538 / #5534 ?

dwolfhub:
It's a step in the right direction. You still can't change session options without extending Bolt\Application. I have another PR coming for that.

src/Session/Handler/RedisHandler.php
dwolfhub: sure thing, made that change

CarsonF:
It's been awhile..what does this change for redis handler?
I'm pretty sure this breaks several options: `persistant`, `database`, `prefix`, etc.

src/Provider/SessionServiceProvider.php
```diff
@@ -341,9 +341,9 @@ protected function parseConnections($options, $defaultHost, $defaultPort)
             }
         } elseif (isset($options['save_path'])) {
             foreach (explode(',', $options['save_path']) as $conn) {
-                $conn = new ParameterBag($conn);
-                $conn->set('uri', new Uri($conn));
-                $toParse[] = $conn;
+                $connBag = new ParameterBag();
+                $connBag->set('uri', new Uri($conn));
+                $toParse[] = $connBag;
             }
```
dwolfhub: @CarsonF an explanation: `$conn` is a string, but the constructor for Parameter bag expects an array. Also, you were passing the parameter bag into the `Uri` constructor and then setting that to 'uri' in the same ParameterBag. It didn't make sense to me (and threw an exception), but maybe I'm missing something.

dwolfhub:
@CarsonF it shouldn't, this is only when parsing save_paths for things like redis.

CarsonF:
Yeah ok maybe it was a derp. I think I've seen in php.ini where those other other options will be query string parameters on the url/save_path. So maybe they need to be pulled out after parsing `Uri` and set on `$connBag`?

```
localhost:6379?prefix=bolt_&database=2
```

dwolfhub:
will look into that

src/Provider/SessionServiceProvider.php
```diff
@@ -341,8 +342,12 @@ protected function parseConnections($options, $defaultHost, $defaultPort)
...
         } elseif (isset($options['save_path'])) {
             foreach (explode(',', $options['save_path']) as $conn) {
+                $uri = new Uri($conn);
+
                 $connBag = new ParameterBag();
-                $connBag->set('uri', new Uri($conn));
+                $connBag->set('uri', $uri);
+                $connBag->add(parse_query($uri->getQuery()));
```
dwolfhub: @CarsonF is this what you had in mind?

src/Session/Handler/RedisHandler.php
```diff
@@ -25,7 +25,7 @@ class RedisHandler implements \SessionHandlerInterface, LazyWriteHandlerInterfac
      */
     public function __construct($redis, $maxlifetime)
     {
-        if (!$redis instanceof Redis || !$redis instanceof Predis) {
+        if (!$redis instanceof Redis && !$redis instanceof Predis) {
```
CarsonF: Who did that smh

GawainLynch: I know, right :wink:

dwolfhub:
@CarsonF rebased and ready!

CarsonF:
@GawainLynch Good for me :shipit:

GawainLynch:
Merged into `release/3.1` via https://github.com/bolt/bolt/commit/dcd15150d07699e1f31b862bcd624619ba3ee7b8

PR #5563, opened 2016-07-21
Session settings

### LINKED ISSUE #5534: [BUG] unable to choose different session storage methods, filesystem hard-coded

dwolfhub:
First, as an implementer of Bolt, I would expect Bolt would use my default session settings set in my `php.ini` file. If for some reason I was using a crappy hosting service or was held back by a grumpy sysadmin, this PR would also let me override those session settings in my `config.yml` file.

It fixes #5534 for me because I wouldn't need to extend `Bolt/Application` to switch away from storing all sessions in `app/cache/.sessions/`. I could just make changes in my `php.ini` file or in my `config.yml` file.

### DIFF
--- a/src/Provider/SessionServiceProvider.php
+++ b/src/Provider/SessionServiceProvider.php
@@ -105,11 +105,9 @@ class SessionServiceProvider implements ServiceProviderInterface
      */
     public function configure(Application $app)
     {
-        $app['session.options'] = [
+        $app['session.options'] = $app['config']->get('general/session', []) + [
             'name'            => 'bolt_session_',
             'restrict_realm'  => true,
-            'save_handler'    => 'filesystem',
-            'save_path'       => 'cache://.sessions',
             'cookie_lifetime' => $app['config']->get('general/cookies_lifetime'),
             'cookie_path'     => $app['resources']->getUrl('root'),
             'cookie_domain'   => $app['config']->get('general/cookies_domain'),
@@ -165,8 +163,16 @@ class SessionServiceProvider implements ServiceProviderInterface
                     }
                 }
 
+                // @deprecated backwards compatibility:
                 if (isset($app['session.storage.options'])) {
                     $options->add($app['session.storage.options']);
+                }
+
+                // PHP's native C code accesses filesystem with different permissions than userland code.
+                // If php.ini is using the default (files) handler, use ours instead to prevent this problem.
+                if ($options->get('save_handler') === 'files') {
+                    $options->set('save_handler', 'filesystem');
+                    $options->set('save_path', 'cache://.sessions');
                 }
 
                 $options->add($app['session.options']);


CarsonF:
I get what you are going for, but this is the wrong way. We already pull ini settings [here](https://github.com/bolt/bolt/blob/master/src/Provider/SessionServiceProvider.php#L164).

Instead we could add a config parameter for that. And `null` would use default value.

dwolfhub:
@CarsonF you pull the ini settings on line 162, but then immediately override them with defaults from 'session.storage.options' on line 164. There is currently no way to use your default ini settings. In fact, there is currently no way to use anything other than `filesystem` with `cache://.sesssions`. It's all getting overridden. I'll take another crack at a PR.

CarsonF:
Right so maybe we need to filter null values out of the `session.options` before it's `->add` to the defaults.

dwolfhub:
@CarsonF by removing the configuration from the SessionProvider, i was able to not only decouple it like you wanted, but also fix my issue of not being able to configure the sessions without extending the bolt application. Please have a look. I think we're close.

src/Application.php
```diff
@@ -84,6 +84,15 @@ protected function initSession()
...
+
+        $this['session.options'] = array_merge(
+            $this['session.options'],
+            $this['config']->get('general/session', [])
+        );
+        $this['session.options']['cookie_lifetime'] = $this['config']->get('general/cookies_lifetime');
+        $this['session.options']['cookie_path']     = $this['resources']->getUrl('root');
+        $this['session.options']['cookie_domain']   = $this['config']->get('general/cookies_domain');
+        $this['session.options']['cookie_secure']   = $this['config']->get('general/enforce_ssl');
     }
```
GawainLynch: These are throwing problems in Travis:

```
PHP Notice:  Indirect modification of overloaded element of Bolt\Application has no effect in /home/travis/build/bolt/bolt/src/Application.php on line 92
PHP Stack trace:
PHP   1. {main}() /home/travis/build/bolt/bolt/tests/phpunit/nutty:0
PHP   2. Bolt\Tests\Nutty->getApp() /home/travis/build/bolt/bolt/tests/phpunit/nutty:21
PHP   3. Bolt\Tests\BoltUnitTest->getApp() /home/travis/build/bolt/bolt/tests/phpunit/nutty:12
PHP   4. Bolt\Application->initialize() /home/travis/build/bolt/bolt/tests/phpunit/unit/BoltUnitTest.php:44
PHP   5. Bolt\Application->initSession() /home/travis/build/bolt/bolt/src/Application.php:101
```

dwolfhub: right, sorry. fixed.

CarsonF:
@dwolfhub Yeah if you move it back into SessionSP I'll be good. I worked very hard in 3.0 dev cycle to shrink the Application class.

dwolfhub:
@CarsonF Yea, I can do that, if you prefer. You had mentioned that you wanted to decouple the config from this service provider so i thought it made sense to make that bridge in the Application. Will patch.

CarsonF:
Yeah sorry about flipping on you. Hard to know each step to take with a far away target.

GawainLynch:
@CarsonF — if you're happy when @dwolfhub updates this, I am going to send it into `release/3.1`

CarsonF:
There's still session options that will be missing now in most user's config.
I would just revert to what was there before and then add your line 90 after the defaults are set.

CarsonF:
```php
$defaults = [
...
];
$app['session.options'] = $app['config']->get('general/session', []) + $defaults;
```

CarsonF:
That still overwrites native php.ini options though...

dwolfhub:
So the order in initConfig now goes:
1. defaults outlined on line 136
2. overwritten with values from php.ini
3. overwritten with values from session.options (if set either explicitly in a .bolt.php or by config.yml)

If we hard code anything in session.options it would overwrite php.ini values...

src/Provider/SessionServiceProvider.php
```diff
@@ -105,17 +105,17 @@ public function boot(Application $app)
...
-            'name'            => 'bolt_session_',
-            'restrict_realm'  => true,
-            'save_handler'    => 'filesystem',
-            'save_path'       => 'cache://.sessions',
-            'cookie_lifetime' => $app['config']->get('general/cookies_lifetime'),
-            'cookie_path'     => $app['resources']->getUrl('root'),
-            'cookie_domain'   => $app['config']->get('general/cookies_domain'),
-            'cookie_secure'   => $app['config']->get('general/enforce_ssl'),
-            'cookie_httponly' => true,
-        ];
```
CarsonF:
```php
        $app['session.options'] = $app['config']->get('general/session', []) + [
            'name'            => 'bolt_session_',
            'restrict_realm'  => true,
            'cookie_lifetime' => $app['config']->get('general/cookies_lifetime'),
            'cookie_path'     => $app['resources']->getUrl('root'),
            'cookie_domain'   => $app['config']->get('general/cookies_domain'),
            'cookie_secure'   => $app['config']->get('general/enforce_ssl'),
            'cookie_httponly' => true,
        ];
```

dwolfhub:
@CarsonF rebased and ready!

CarsonF:
Looks good 👍

GawainLynch:
Merged into `release/3.1` at https://github.com/bolt/bolt/commit/3e7cb6e904e4b1ac9b4d939d051f053568038cd7

PR #5567, opened 2016-07-22
fix #5398

### LINKED ISSUE #5398: [BUG] Html snippets are not loaded because of an unexpected cache-control setting
As discussed in Slack, I spent the past few days debugging a problem with my bolt install where the assets of all extensions would not load. The htmlsnippets were just not included in the html. 

Assets are not loaded because of this line in Application.php which was last edited for release 2.2.20 (see PR  https://github.com/bolt/bolt/pull/5121 by @SahAssar)

```
if (isset($this['htmlsnippets']) && ($this['htmlsnippets'] === true) && ($response->headers->get('Cache-Control') === "no-cache")) {
```

I deployed the site locally, I deployed it to another server, and the behaviour is not present. I only see this behaviour on one of our servers. I have confirmed the server is the only variable, and changing the line of code below to what it was in 2.2.19-pl1 actually 'fixes' it.

I will describe the specifics of this server below.

Details
-------

 - Relevant Bolt Version: 2.2.21
 - PHP version: 7.0.6
 - Used webserver: Apache version 2.4.6 (CentOS)

Environment information
----------------------
Please rename these txt files to html to view the detailed phpinfo() output in your browser
[phpinfo_not_working.txt](https://github.com/bolt/bolt/files/294360/phpinfo_not_working.txt)
[phpinfo_working.txt](https://github.com/bolt/bolt/files/294361/phpinfo_working.txt)

Reproduction
------------

 1. Do a clean install of bolt 2.2.21 on a similar environment
 2. Install any extension that uses html snippets to load assets
 3. Watch as the html snippets are not loaded into the html.

I tested this with the bolt/labels extension.

dadaxr:
there is case where the header cache-control can be either 
"no-cache" either "private, must-revalidate" 

see : /vendor/symfony/http-foundation/Symfony/Component/HttpFoundation/ResponseHeaderBag.php::computeCacheControlValue() method for more information

we must handle that possibility to correctly insert some html snippet.

( This bug is present in bolt 2.2.23 ( and previous i think ) but i dont know for bolt 3 because the application.php file does not work the same !

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -562,7 +562,10 @@ class Application extends Silex\Application
 
         // true if we need to consider adding html snippets
         // note we exclude cached requests where no additions should be made to the HTML
-        if (isset($this['htmlsnippets']) && ($this['htmlsnippets'] === true) && ($response->headers->get('Cache-Control') === "no-cache")) {
+        if (isset($this['htmlsnippets'])
+            && ($this['htmlsnippets'] === true)
+            && (in_array($response->headers->get('Cache-Control'), ['no-cache','private, must-revalidate']))
+        ) {
             // only add when content-type is text/html
             if (strpos($response->headers->get('Content-Type'), 'text/html') !== false) {
                 // Add our meta generator tag.


SahAssar:
hey @dadaxr, first of all thanks for fixing a bug that I introduced :koala: !

 2.2 is still supporting 5.3 so you have to use the old array() syntax so, if you can fix that it'd be awesome (and also rebase out that merge perhaps :) ).

GawainLynch:
Manually merged (and array syntax fixed) in c027192b4f0753554695ad4c8cf997498a747da0

dadaxr:
Thanks =)

PR #5568, opened 2016-07-23
Create custom exception for invalid repo, and throw this when accessed

rossriley:
Small patch for issue discussed in #5537. 
Gives a more descriptive error message when an invalid repository is loaded.

PR #5569, opened 2016-07-23
Set a flash and redirect to to dashboard if ContentType doesn't exist fetching repository

### LINKED ISSUE #5537: [RFC] Improve error handling of accessing non-existing contenttypes in dashboard
Currently if you visit, for example `/bolt/editcontent/thisdoesntexist`, you get the error:
`Attempted to load mapping data for unmapped class thisdoesntexist`

This error doesn't say that the contenttype doesn't exist, and me (and at least one other user) have run into this problem and had trouble figuring out what resulted into this error.

I propose a better handling of this situation so users get a clear 404 error and see a clear error.
For example: `Contenttype thisdoesntexist doesn't exist.`.

Details
-------

 - Relevant Bolt Version: 3.0 | master
 - Install type: All
 - PHP version: All

Reproduction
------------

If you have a running Bolt installation, simply go to `/bolt/editcontent/thisdoesntexist` to reproduce the error. Any string (except an existing contenttype) instead of 'thisdoesntexist' will result in the unclear error.

GawainLynch:
Closes #5537
Follows-up #5568

### DIFF
--- a/src/Controller/Backend/Records.php
+++ b/src/Controller/Backend/Records.php
@@ -1,6 +1,7 @@
 <?php
 namespace Bolt\Controller\Backend;
 
+use Bolt\Exception\InvalidRepositoryException;
 use Bolt\Storage\ContentRequest\Listing;
 use Bolt\Storage\ContentRequest\ListingOptions;
 use Bolt\Translation\Translator as Trans;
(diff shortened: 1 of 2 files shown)


bobdenotter:
Looks good to me!

PR #5571, opened 2016-07-24
Setting Extensions Composer options

### LINKED ISSUE #5499: [extension] Allow prefer-source option
Frequently when I install an extension of my own I need to make changes to it within my project. It therefore needs to be a source checkout, not the dist checkout.

With composer `dev` dependencies are git clones, and there is also a `--prefer-source` option which will force depdencies to be git cloned rather than downloaded.

With bolt it seems that extensions are always downloaded and never cloned - would it be possible to support cloning? (maybe a `--prefer-source` option?)

GawainLynch:
Closes #5499

@SahAssar: I was back-and-forth on what to do with `app/config/config.yml.dist`, but opted to add some of the useful bits here now, and it can get caught up in a bigger sweep later.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -155,10 +155,10 @@ recordsperpage: 10
 #                  cases you should *NOT* enable this, because it will cause
 #                  side-effects if the website shows different content to
 #                  authenticated users.
-# - thumbnails:    Caches thumbnail generation.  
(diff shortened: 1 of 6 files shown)


SahAssar:
@GawainLynch Sounds good, I need to write up docs for the configs before removing stuff anyway.

PR #5572, opened 2016-07-24
Four fixes for the extend screen

SahAssar:
* Add checks for config and readme
* Don't add button classes on dropdowns
* Have max three buttons (readme, config and copy theme), with the rest in a dropdown
* Don't output a manage dropdown on composer-plugin's

Also adds href's to the \<a\> tags that did not have any so that it's valid html and automatically gets the correct cursor.

Before:
image

After:
image


GawainLynch:
McFreakingAwesome! :shipit:

Issue #5391, opened 2016-05-31
[BUG] Search in the new storage classes does not reset parameters between queries


rossriley:
Because of the way the objects are shared in the QueryServiceProvider searches in the new storage classes do not reset parameters between queries this means that multiple searches or searches against multiple content types will not have the compiled queries reset and thus will error.
## Details
- Relevant Bolt Version: [ 3.0 + ]
- Install type: [ Alll ]
## Reproduction

`$pages = $app['query']->getContent('(pages,entries)/search', ['filter' => 'lorem']);`

Will fail, whereas the first single run will work. eg this:

`$pages = $app['query']->getContent('(pages)/search', ['filter' => 'lorem']);`

Thanks to @ratty for spotting.

PR #5575, opened 2016-07-24
Fixed issue when searching multiple contentypes in Bolt V3 Storage Layer

Raistlfiren:
When searching multiple content types in the Bolt V3 storage layer, it is being overwritten by the previous query. This should correct the issue by cloning a fresh instance of `SearchQuery` to be used each time in the `foreach` loop.

Fixes: #5391


GawainLynch:
I don't suppose, :koala: to :penguin:, I could interest you in adding/refining a unit test that catches this?

PR #5582, opened 2016-07-25
Folder handling doesn't need parent value concatenated

### LINKED ISSUE #5478: [BUG] sub-folder rename with Umlaut-names does not work
Details
-------

 - Relevant Bolt Version: 3.0.4 - 3.0.8
 - Install type: Zip/tarball install
 - PHP version: 7.0.6 FPM/FastCGI
 - Used webserver: Apache 2.4.7
 - For UX/UI issues: Chromium 50.0.2661.102; Firefox 47.0

Reproduction
------------

 1. **Bug summary**: It is not possible to rename sub-folders whose names contain Umlaute (German diacritical characters - ä ö ü ß)
 2. **Specifics**: In the admin backend, folders whose names contain one or multiple Umlauts cannot be renamed (either from or to the Umlaut-version), but they can be created. This only applies if attempted in a sub-folder of the files/ directory. I.e. it *does* work as expected directly below files/, but not on a deeper level.
 3. **Steps to reproduce**:
* create a folder of any name below files/ via the admin backend's file manager
* in this sub-folder, create another folder called "testü"
* attempt to rename this folder; this does not work
* instead of steps 2 & 3, you can also create a folder called "test" and attempt to rename it to "testü"; this does not work either
 4. **Expected result**: Folders should be able to be renamed to/from a name containing Umlauts *not* only if they're directly below files/
 5. **Actual result**: A javascript alert() message pops up with the following localized message:
"Ordner konnte nicht umbenannt werden: events\/test\u00fc" [Folder could not be renamed: events\/test\u00fc]. Additionally, a console.log() message gets written after the first message is closed,
with the text "Something went wrong renaming this folder!"
In this case, event/ is a folder directly below files/, and the folder testü/ was attempted to be renamed to test/

bolt rename bug
(the long execution time visible in the debug bar is only due to a large number of directories and files)

GawainLynch:
Fixes #5478

### DIFF
--- a/src/Controller/Async/FilesystemManager.php
+++ b/src/Controller/Async/FilesystemManager.php
@@ -295,13 +295,12 @@ class FilesystemManager extends AsyncBase
     public function removeFolder(Request $request)
     {
         $namespace = $request->request->get('namespace');
-        $parentPath = $request->request->get('parent');
         $folderName = $request->request->get('foldername');
 
         try {
-            $this->filesystem()->deleteDir("$namespace://$parentPath$folderName");
+            $this->filesystem()->deleteDir(sprintf('%s://%s', $namespace, $folderName));
 
-            return $this->json("$parentPath$folderName", Response::HTTP_OK);
+            return $this->json($folderName, Response::HTTP_OK);
         } catch (ExceptionInterface $e) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
Looks good to me!

PR #5584, opened 2016-07-25
[FEEDBACK] Separated `Directives` from `Handlers`.

Raistlfiren:
This isn't an issue, just a refactoring. This could break V3 extensions if they extend upon any of the current directives.

The purpose of this is to separate the logic from `Handlers` and `Directives`. I think the two should be split up since they do different things. Here are my definitions of `Handlers` and `Directives`:

*Handlers* - Short segments of code that execute queries AND returns data from the db
*Directives* - Short segments of code that manipulate queries

Thoughts?

### DIFF
--- a/src/Storage/Query/ContentQueryParser.php
+++ b/src/Storage/Query/ContentQueryParser.php
@@ -3,18 +3,18 @@
 namespace Bolt\Storage\Query;
 
 use Bolt\Storage\EntityManager;
+use Bolt\Storage\Query\Directive\HydrateDirective;
+use Bolt\Storage\Query\Directive\OrderDirective;
+use Bolt\Storage\Query\Directive\PagingDirective;
+use Bolt\Storage\Query\Directive\PrintQueryDirective;
+use Bolt\Storage\Query\Directive\ReturnSingleDirective;
 use Bolt\Storage\Query\Handler\FirstQueryHandler;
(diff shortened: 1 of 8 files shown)


GawainLynch:
I like the concept… my worry is the BC.

Worth putting these in with the legacy class aliases?

Raistlfiren:
I have always liked just breaking things ;)

rossriley:
I'll give you the rationale for the split, but I'm certainly not opposed to changing it, especially when we properly document it.

To try and help with tokenising and parsing, I divided the existing query syntax into `{% setcontent var = '<contenttype(s)>/<operation>/<args> where {xxxxx} <directive1> <directive2> %}` 

So `'pages/latest/5'` would be past to the latest query handler, `'entries/random/10'` would be passed to the random query handler etc. 

Directives were the name for the single commands that appear after the where parameters in Bolt syntax eg. `printquery`, `hydrate`, `returnsingle` etc. 

For convenience I grouped limit and orderby here too although they are a little different because they aren't strictly a directive, supporting arguments too.

So the short explanation is that the split was more determined on where they appeared to the query parser, rather than what their functionality entailed.

GawainLynch:
@rossriley (pre-:coffee: warning) — do I take that as a :+1: from you?

Matching docs https://github.com/bolt/docs/pull/453

rossriley:
yes :+1: from me

PR #5587, opened 2016-07-25
(In select fields with content values) the valuefield is last and defaultception

### LINKED ISSUE #5517: Select by records does not work with multiple fields.
f.e. this from the default bolt setup:

```yaml
        selectentry:
            type: select
            values: entries/id,title
            postfix: "Select an entry"
            autocomplete: true
            sort: title
```

bug

Both entries have a title:

SahAssar:
Fixes #5517.

The valuefield is the last one, not the first one, and we can't assume that `lookupfieldlist` is set so, unfortunately we need some defaultception until this code is rewritten.

Also prevents certain twig errors in some cases because of that defaultception.


GawainLynch:
I …

PR #5597, opened 2016-07-27
add index on slug in taxonomy table

SahAssar:
Fixes #4545 

Was assigned to ross, but might aswell take this myself since it's so simple.

PR #5602, opened 2016-07-28
Set relations indexes to be multi column indexes

SahAssar:
Because any query with from_contenttype will include from_id and the same with to_*. Gives a very nice performance boost when dealing with sites that have a large amount of relations.

Before:
```
SELECT * FROM bolt_relations WHERE from_contenttype=? AND from_id IN (?) ORDER BY id 
Parameters: ["kommentarer",{"681786":"681786"}] 
Time: 371.70 ms
```
After:
```
SELECT * FROM bolt_relations WHERE from_contenttype=? AND from_id IN (?) ORDER BY id 
Parameters: ["kommentarer",{"681786":"681786"}] 
Time: 0.33 ms
```


GawainLynch:
Seriously … who the heck set up those indexes … :koala: :gun:

SahAssar:
Oh, and just like #5597 this will need an update to the tests db, so it will fail the tests (but lets let them run so we see that is the only failure it gets).

GawainLynch:
:shipit:

PR #5604, opened 2016-07-28
Implement a new option in nut's database:check

brewneaux:
This relates to Issue 5995 - implemented an option for database:check
which will cause the database statements to be output on the console
with nut.

Changes include:
 - Adding jdorn/sql-formatter to composer, so that queries are more
   legible
 - Making changes to DatabaseCheck, to add the new option and outputs.

Future enhancement could include using the SqlFormatter class within the
web backend to achieve formatting and colourized output in
/dbcheck?debug=1.  Unfortunately, I'm not good enough with the way Twig is used in the backend to get that integrated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,8 @@
         "symfony/yaml" : "^2.8",
         "tdammers/htmlmaid" : "~0.7",
         "twig/twig" : "^1.18",
-        "ua-parser/uap-php" : "~3.4"
+        "ua-parser/uap-php" : "~3.4",
+        "jdorn/sql-formatter": "^1.2"
     },
     "require-dev" : {
         "bolt/codingstyle" : "~1.0",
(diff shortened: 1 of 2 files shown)


GawainLynch:
Personally, I am not overly comfortable with `jdorn/sql-formatter` … its last tag was 27 Nov 2013. The risk being getting locked into a package that holds us in place.

Otherwise :+1:

brewneaux:
@GawainLynch Normally, I would agree on the extremely old package.  But, in this case, I would think its ok, and heres why:

* It is a completely self-contained class
* It is very lightly used
* There isn't much that would have to be added...ever.  SQL, in general, doesn't change very much.

If you have another suggestion on a different package to use, I can look into it, but there isn't really much out there.

bobdenotter:
I've looked over the code in https://github.com/jdorn/sql-formatter/blob/master/lib/SqlFormatter.php , and I don't see much harm in it. In general I agree with Gawain's statement that we should be wary to include new packages that seem to be discontinued, but in this case we could argue that it's "finished", so there's not much to maintain there. 

I'm 👍

GawainLynch:
Good enough for me :smile:

Issue #5468, opened 2016-06-21
[BUG][UI] When deleting one or N records in the listing, all records disappear


PhillippOhlandt:
I think this gif shows it really good:

![](http://g.recordit.co/YaSaZQ1cXG.gif)

It's just an UI issue, the not selected record is still there after a reload.

Details
-------

 - Relevant Bolt Version: 3.0.7
 - Install type: composer
 - PHP version: 7.0
 - Used webserver: nginx
 - For UX/UI issues: chrome, latest stable

GawainLynch:
Yeah, and I think recent (/me looks at z-index PRs) as I was doing this on a 3.0.4 site last week and it worked.

rarila:
Hmm, can't reproduce (Chrome, Apache, 3.0.8), works as expected …
Any errors in the console?

PhillippOhlandt:
@rarila just tested it again on 3.0.7, no errors in the console.

rarila:
Ok, I don’t think 3.0.7/3.0.8 makes a difference.

Next step, could you please check if:

- a XHR request is sent to ``/async/content/action`` with formdata ``bolt_csrf_token``, ``contenttype`` and something looking like ``actions[pages][9][delete]``

- a response with status code ``200`` and a html fragment like ``<table class=" dashboardlisting listing">…</table>`` as data is returned

bobdenotter:
*BUMP*

Is this still an issue, @PhillippOhlandt ? I can't reproduce it on 3.0.11.

PhillippOhlandt:
@bobdenotter, @rarila, can't reproduce on 3.0.10 anymore. seems to be fine now :)

SahAssar:
actually, It's still an issue, and one of the ones that #5610 will fix once cleaned up enough to go in.

SahAssar:
The problem was/is that ajaxy requests don't show unpublished records, while non-ajaxy backend requests did. So, you'd only see this issue with unpublished records.

PhillippOhlandt:
Oh yes, @SahAssar is right. Still a bug with two unpublished records. Should this be reopened?

SahAssar:
Reopening and adding to #5610

bobdenotter:
Fixed by #5610

PR #5610, opened 2016-07-31
Fix backend publish process quirks (#5085)

### LINKED ISSUE #5085: [BUG][UX] back-end publish process quirks
The process of publishing from the content listing has some weird quirks.

The first two steps are normal
image

image

But the third step tells me "there is no undo" which is just FUD for this action.
image

The real quirk comes after I confirm - only the published records are displayed and I have to refresh the page to get the other records back. The _pager_ is hidden too.

image

Details
-------

 - Relevant Bolt Version: 3.0 / master
 - PHP version: 5.6
 - Used webserver: Apache
 - For UX/UI issues: Chrome Version 49.0.2623.87 (64-bit) Linux

SahAssar:
Fixes #5085 and #5468

This does two things:
1. check if the request is safe (status change) or unsafe (delete) in listing actions and only ask the user to confirm if it is unsafe.
2. Allow unpublished records in the async zone, since all async routes are also backend routes (right?) and should therefore show unpublished too.

Since it's a bugfix I targeted release/3.0, lemme know if this should go in 3.1 instead.

### DIFF
--- a/app/src/js/widgets/buic/buicListing.js
+++ b/app/src/js/widgets/buic/buicListing.js
@@ -31,13 +31,13 @@
         /**
          * Execute commands on triggered button.
          *
-         * @param {string} action - Triggered action (Allowed: 'delete').
+         * @param {string} action - Triggered action.
          * @param {array} ids - Array of ids to perform the action on.
          * @param {string} buttonText - Button text to be displayed on ok button.
          */
         modifyRecords: function (action, ids, buttonText) {
             var self = this,
-                safeQuery = true,
+                modifications = {},
                 actions = {
                     'delete': {
                         'safe': false,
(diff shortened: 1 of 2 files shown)


rarila:
A few things:

- I would prefer to feed ``modifications`` as function parameter into ``sendQuery()`` as it is of local nature.

- I'm not quite happy with extracting ``sendQuery()`` as what it does is more of ``sendModifyRecordsQuery()``. buicListing (if I remember correctly, though I may be wrong) is more to be seen as general listing object with just one specialized functionality added yet. To show that connection… naming, inlining, …? At least it should be a private method (underscore).

- I think there's no need for ``safeQuery``, just ``if (!actions[action].safe) {`` will do as there's only one action in that function call. You could even simplify it to ``if (action === 'delete') {`` and save some more lines.

- Please add some spaces, eg. ``if(`` => ``if (``, ``)else(`` => ``) else (``, …. Otherwise one could think the OCD is weak in you :smiley:  And a doc block for ``sendQuery()`` would be awesome, too.

SahAssar:
My OCD has failed you, master. My work shall double tomorrow to make up for this! ;)

rarila:
@SahAssar  As discipline you now write 1000 spaces on that blackboard! :black_large_square:

SahAssar:
@rarila Should be fixed now, and my blackboard is filled with spaces!

rarila:
:rotating_light: We have a nice PR here, ready to merge! Well done! :+1: :smile:

bobdenotter:
Let's go, then! 👍

PR #5611, opened 2016-07-31
Feature: Better sanitization of content on save.

### LINKED ISSUE #5339: [RFC] Allow or disallow `<script>` in content.
Until now, we've always allowed `<script>` tags in content. 

In fact, more than a few times, people have submitted "serious security issues", because a registered and logged on Editor is able to put `<script>alert('xss');</script>` in a title field. 

Similarly, we allow `<iframe>` in wysiwyg fields, because otherwise you couldn't insert Youtube and Vimeo embeds. However, this means it's also possible to (unknowingly to an editor) insert remote javascripts in your content, because the editor copy/pasted it from a different website. Because CKEditor does _not_ execute these scripts, this can happen without the editor noticing it. 

We will also filter `text`, `textarea` and `html` fields using the `htmlcleaner:    allowed_tags` setting, expanded with the `wysiwyg` settings like `embed`. This would prevent inadvertently adding script to the wysiwyg fields by an editor. 

Items to do: 

- [ ] Use our sanitation for all text-like fields, not just Markdown.
- [ ] Make order consistent: first sanitize, and _then_ parse snippets as twig.
- [ ] When explicitly allowing things like 'embed' in Wysiwig, temporarily add these to whitelist for sanitizing.
- [ ] Document these changes, and in extensions documentation mention that extensions are able to insert stuff _after_ sanitization took place. 
- [ ] Always disallow `<script>` and `onX='..'` attributes in content, because that shit's whack, yo!

bobdenotter:
Fixes #5339.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -208,7 +208,7 @@ thumbnails:
 # left in the content that is shown to users. For example, tags like `<script>`
 # or `onclick`-attributes.
 htmlcleaner:
-    allowed_tags: [ div, p, br, hr, s, u, strong, em, i, b, li, ul, ol, blockquote, pre, code, tt, h1, h2, h3, h4, h5, h6, dd, dl, dh, table, tbody, thead, tfoot, th, td, tr, a, img ]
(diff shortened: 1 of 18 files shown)


bobdenotter:
Looks good! 👍

PR #5620, opened 2016-08-02
Get correct path to compare in 'current' filter, when Bolt is in a subfolder/

### LINKED ISSUE #5585: [BUG] current() twig filter always returns false when running bolt in a subfolder
Currently active menu items are not styled any differently.
Dumping `item|current` and `item.record` from within the for loop in  `base-2016/partials/_sub_menu.twig` shows that the `current()` twig function appears to only return false, and [`item.record`](https://docs.bolt.cm/3.0/content/menus) is `null`.

**Added after debugging:** This is because the current() twig filter always returns false when running bolt in a subfolder.

Details
-------

 - Relevant Bolt Version: 3.0.11 (worked in bolt 2)
 - Install type: Zip/tarball
 - PHP version: 7.0.8
 - Used webserver: Apache 2.4.23

Reproduction
------------

1. Create a new bolt installation (In a subfolder)
2. View the sample site, and note that menu items referring to the current record are not getting the "active" class they should be
3. Dump `item.record` from inside the for loop in `base-2016/partials/_sub_menu.twig`. It returns null, suggesting that `current()` returning false is just a symptom of the real issue.

On my installation:
* `base-2016/partials/_sub_menu.twig` does indeed test `item|current` on line 23
* `config/menu.yml` defines paths, not links

No visible error messages are produced.

bobdenotter:
Fixes #5585

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -40,7 +40,7 @@ class RecordHandler
     {
         /** @var Request $request */
         $request = $this->app['request'];
-        $requestUri = $request->getPathInfo();
+        $requestUri = $request->getRequestUri();
         $routeParams = $request->get('_route_params');
 
         // If passed a string, and it is in the route.


bobdenotter:
Tests will pass after #5622 has been merged in.

PR #5621, opened 2016-08-02
Replacing deprecated localdate with localedatetime.

### LINKED ISSUE #5591: [BUG] Listing template uses deprecated Twig filter localdate
The template **@bolt/_base/_listing.twig** uses the deprecated filter `localdate` which now causes a deprecated exception within Twig. The template is used on the dashboard for example.
The strange thing is, that the exception seems to be suppressed in bolt itself. Whe you use the same template to display a list of records in your own code (a widget in my case) it throws the error as show below.

```
An exception has been thrown during the rendering of a template ("Twig Filter "localdate" is
 deprecated. Use "localedatetime" instead in @bolt/_base/_listing.twig at line 196.") in
 "@bolt/dashboard/dashboard.twig" at line 3.
```

Details
-------

 - Relevant Bolt Version: 3.0.11
 - PHP version: 5.5
 - Used webserver: Apache
 - For UX/UI issues: Chrome 51.0.2704.106 (64-bit) on Mac

Reproduction
------------
As told above, Bolt seems to suppress the error. I have a widget that calls same template.
I could remove the exception by changing the template as specified in the error message.

bobdenotter:
Fixes #5591.

### DIFF
--- a/app/view/twig/_base/_listing.twig
+++ b/app/view/twig/_base/_listing.twig
@@ -193,7 +193,7 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
                 {% if content.status == 'timed' %}
                     <i class="fa fa-clock-o status-timed fa-fw"></i> {{ buic_moment(content.datepublish) }}<br>
                 {% else %}
-                    <i class="fa fa-circle status-{{ content.status }} fa-fw"></i> {{ content.datepublish|localdate('%x') }}<br>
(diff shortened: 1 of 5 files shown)


bobdenotter:
Tests will pass after #5622 has been merged in.

evertalbers:
Will this have any consequences for the docs? Just asking...

SahAssar:
evertalbers: yep, it will be changed in #467

PR #5626, opened 2016-08-03
Adjust probability formula for session garbage collection

### LINKED ISSUE #5625: [BUG] Garbage collection of session files doesn't work
Hi.

After commit https://github.com/bolt/bolt/commit/2d2634a5fe111613aa60eaed56bb7ab61b9b5a3e
Change cache://session to cache://.sessions functionality of cleaning session files has stopped working.

In bolt\filesystem\src\Iterator\PathFilterIterator.php
In accept method we have problem in:

       foreach ($this->noMatchRegexps as $regex) {
            if (preg_match($regex, $filename)) { //here is the problem with dot files - .session/

In my opinion we have to register new mount point in Bolt\Configuration\ResourceManager eg. $this->setPath('session', 'app/sessions');

and that mount point should be used to store session files, it can not be dot directory.

Details
-------

 - Relevant Bolt Version: 3.0 | master 
 - PHP version: 5.6
 - Used webserver: Apache 2.4
 - For UX/UI issues: not relevant

GawainLynch:
Closes #5625

Fixes a number of problems we were seeing, such as on the extension store, with session directories getting too large.

### DIFF
--- a/src/Session/SessionStorage.php
+++ b/src/Session/SessionStorage.php
@@ -353,9 +353,10 @@ class SessionStorage implements SessionStorageInterface
             return;
         }
         $divisor = $this->options->getInt('gc_divisor');
+        $probabilityPercentage = 100 * $probability / $divisor;
+        $chance = mt_rand(1, 100);
 
-        $rand = mt_rand(0, $divisor);
-        if ($rand < $probability) {
+        if ($chance <= $probabilityPercentage) {
             $this->handler->gc($this->options->getInt('gc_maxlifetime'));
         }
     }


bobdenotter:
Whooop!

src/Session/SessionStorage.php
```diff
@@ -353,9 +353,10 @@ protected function collectGarbage()
             return;
         }
         $divisor = $this->options->getInt('gc_divisor');
+        $probabilityPercentage = 100 * $probability / $divisor;
+        $chance = mt_rand(1, 100);
 
-        $rand = mt_rand(0, $divisor);
-        if ($rand < $probability) {
+        if ($chance <= $probabilityPercentage) {
```
CarsonF:
```
[1-100] <= 0.1
```
That will never be true...

GawainLynch: Hmm… OK, but previously it would/could never be true.

`$rand` will always be an integer, and by default `$probability` is 1 … so 1 is never going to be less than 1.

Seems this is one to break both our brains :smile:

CarsonF: You're forgetting zero

GawainLynch: Also, I miust have copy/pasta-ed something … and this was firing given enough refreshes

Issue #5555, opened 2016-07-19
Repository::findWith return type


dantleech:
Find with returns `false` in the case that no records are found.

It should return an empty array:

- You cannot iterate over `false`
- The consumer has to always condsider that `false` may be returned.
- If the user wants to handle the case that no records are returned it is fine to use `count() === 0`
- Mixed return types are arguably always to be avoided.

For example, in Bolt Members:

```php
    public function getAccountMetaAll($guid)
    {
        // ...

        $meta = $this->findWith($query);

        /** @var Entity\AccountMeta $value */
        foreach ($meta as $key => $value) {
            $meta[$value->getMeta()] = $value;
            unset($meta[$key]);
        }

        return $meta;
    }
```

Will cause an error because `$meta` is false.

GawainLynch:
Yes, I agree, and I know, and understand, all the downsides. The reason I have gone that route is that it mirrors what Bolt core does, which mirrors what Doctrine does, and I intended this to be consumed without having to shift that mindset … plus it was having to be architected and built in amazingly short time frames and resources.

That aside, I am not against it. But need to have a look/think about what the effect is to be consistent across the library.

GawainLynch:
Haha… replied in the wrong issue, thought this was the Members one referencing this. Big day :disappointed:

PR #5629, opened 2016-08-03
Return an empty array rather than false from findWith

rossriley:
Minor fix to return an iterable empty array rather than false from the result of a `findWith` query.

Fixes: #5555


GawainLynch:
Ah… awesome, I have been meaning to ask you about this one for weeks!

Issue #5636, opened 2016-08-04
Relations seem to affect each other in odd ways


SahAssar:
Okay this one is a bit odd, so hang with me.  Tested on 3.0.11 and master.

To reproduce:

1. Setup a new bolt install.
2. Create user and log in.
3. Prefill the db with a bunch of records.
4. In the entries contenttype add a relation to entries, the relations block should look something like this:

    ```
        relations:
            pages:
              multiple: false
              order: title
              label: Select a page
            entries:
              multiple: false
              order: title
              label: Select an entry
    ```
5. Edit one of the entries with a higher id, like 4 or 5.
6. Set the entry relation to one of the lower id's like 2 and save: ![image](https://cloud.githubusercontent.com/assets/5096632/17394659/a3630a3c-5a2a-11e6-9c1b-fa270684e21d.png)

7. Reload the page, notice that while the entry relation is saved just fine, under the "This record is related to the following Entries:" it will show that the entry is related to itself. That is oddity no. 1.

8. Now add a relation to pages **with a higher id than the relation you added for entries, but an id that also exists as an entry**. For example, lets set the relation to page no. 4 and then save.

9. Reload the page and see that the entry->entry relation has been set to no. 4 too, even though we did not change it.

10. If you go to the frontend and dump the record the relations are correct (entries->2, pages->4), but since the preselected relation in the backend is incorrect as soon as one saves the record again it sets both to no. 4 even though entries should have been set to no. 2.

Possibly related to #5541

GawainLynch:
Ah … I think I know what this is!

PR #5637, opened 2016-08-04
Incoming relations must be checked with both contenttype and id.

SahAssar:
Fixes part of #5636, where intra-contenttype relations will show as related to themselves.

### DIFF
--- a/src/Storage/Collection/Relations.php
+++ b/src/Storage/Collection/Relations.php
@@ -190,7 +190,7 @@ class Relations extends ArrayCollection
     public function incoming($entity)
     {
         return $this->filter(function ($el) use ($entity) {
-            return $el->getTo_contenttype() == (string) $entity->getContenttype();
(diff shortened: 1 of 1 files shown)


src/Storage/Collection/Relations.php
```diff
@@ -190,7 +190,7 @@ public function getField($fieldname, $biDirectional = false)
     public function incoming($entity)
     {
         return $this->filter(function ($el) use ($entity) {
-            return $el->getTo_contenttype() == (string) $entity->getContenttype();
+            return ($el->getTo_contenttype() == (string) $entity->getContenttype() && $el->getTo_id() == $entity->getId());
```
GawainLynch: Are we sure we need loose comparison?

Also the outer `()` isn't needed.

SahAssar: You're right, in this case strict should be fine. Also removed the outer parens.

PR #5638, opened 2016-08-04
Fix ImageHandler imageInfo method for some rare cases.

WedgeSama:
In some cases, the `Reader` return `false` and of course, the rest of the method cant work cause it try to use `$exif` as an object but it is a `boolean`.

Only an error on `2.2` cause the method is changed on `3.0`.


bobdenotter:
The failing test is only for PHP 5.3. It does that on our Bolt 2.2 branch.. 

Looks good, merging in. 👍

Issue #5596, opened 2016-07-26
HTML field in repeater doesn't work if not the last in the field set.


dantleech:
I thought I had already created a bug report for this, but I don't think I did.

Works:

```
examples:
    name: examples
    singular_name: example
    fields:
        text:
            type: textarea
        sections:
            type: repeater
            fields:
                image:
                    type: image
                afficher:
                    type: select
                    values: [ "Gauche", "Doite" ]
                text:
                    type: html
```

Doesn't work:

```
examples:
    name: examples
    singular_name: example
    fields:
        sections:
            type: repeater
            fields:
                image:
                    type: image
                afficher:
                    type: select
                    values: [ "Gauche", "Doite" ]
                text:
                    type: html
        text:
            type: textarea
```

![brok](https://cloud.githubusercontent.com/assets/530801/17153228/8d4024c8-5373-11e6-87e0-9c0780f70be2.png)

Tested with bolt v3.0.10

(super helpful) Javascript error:

```
TypeError: c is undefined bolt.js:8:8921
```

PR #5639, opened 2016-08-04
Bugfix for HTML fields inside repeaters

rossriley:
This addrresses a bug whereby HTML fields in repeaters would fail if they appear early in the order before ckeditor initialisation has happened.

Fixes: #5596


bobdenotter:
Haven't tried this out yey, but is this correct? `{% if 'html' not in context.fieldtypes %}`

We include it, if 'html' is _*not*_ in `context.fieldtypes` ?

Issue #5541, opened 2016-07-14
[BUG] Relationships with Multiple: True not displaying in editor correctly


Jovinsson:
Details
-------

 - Relevant Bolt Version: 3.0.9
 - Install type: Zip/tarball install
 - PHP version: 5.5.35
 - Used webserver: Apache (XAMPP 5.5.35)
 - Host OS: Windows Server 2012

Reproduction
------------

**Bug Summary**
Contenttypes with relations that have the property of `multiple: true` do not correctly show the multiple records when viewing it after saving, instead, they only show 1 related record.

The relationships are still stored but not shown. When I go to the front end of the CMS (user side) it shows the related records correctly as per my .twig structure. It seems it is only impacting the "Relationships" tab of the editing screen.

**Steps to Reproduce**
1. Create multiple contenttypes with a relation to each other and have the `multiple: true` property set
2. Relate more than two records of a single contenttype to another and save the record
3. Refresh the page
4. View the relations again to find only 1 related record is shown per contenttype

**Expected Result**
It should show all related records for each contenttype.

**Actual Result**
Only a single record is displayed.

**Screenshots**
Note: I have removed the names of the records in paint for privacy reasons
 1. I have selected all related records ('Sites') and clicked save
![before saving](https://cloud.githubusercontent.com/assets/5206571/16855150/a586c392-4a46-11e6-806d-31599e069a7d.PNG)
 2. I have refreshed and can only see one record listed
![after saving and refreshing](https://cloud.githubusercontent.com/assets/5206571/16855151/a587db74-4a46-11e6-91e4-1894b89734da.PNG)

Please advise if any further information is required!

Jovinsson:
Any ideas?

GawainLynch:
Potentially just a case of real life overload hitting everyone's time. Currently triaging for 3.1 so let's see if we can reproduce and fix for that, or tag it for further look :smile:

Jovinsson:
Figured your team would be onto it! Thanks @GawainLynch

GawainLynch:
OK, just tried on 3.0.11 and 3.1-dev and I can't reproduce this on either, I don't suppose you have a dev install you could try 3.0.11 on do you?

SahAssar:
@Jovinsson Do you perhaps have an incoming relation from another record to that record? Or another outbound relation from it? I'm tinkering with bidirectional relations right now, and there seems to have been some bugs about relations affecting each other in odd ways.

Jovinsson:
[...]

PR #5641, opened 2016-08-05
Fix bidirectional relations and relations affecting each other

### LINKED ISSUE #5636: Relations seem to affect each other in odd ways

SahAssar:
Fixes #5636 and (in my testing at least) #5541.

This makes bidirectional relations work as they did in 2.2 (one could show/delete relations from either record if they both had the relation defined), fixes intra-contenttype relations, and also introduces a number of checks to make sure that multiple relations on the same contenttype don't affect one another.

### DIFF
--- a/app/view/twig/editcontent/fields/_relationship.twig
+++ b/app/view/twig/editcontent/fields/_relationship.twig
@@ -11,7 +11,7 @@
 {# Currently selected relationship values #}
 {% set relselect = [] %}
 {% for item in values %}
-    {% set relselect =  relselect|merge([item.getToId()]) %}
+    {% set relselect =  relselect|merge([item.toid]) %}
 {% endfor %}
 
 {# Build the select options array #}
--- a/app/view/twig/editcontent/fields/_relationship.twig
+++ b/app/view/twig/editcontent/fields/_relationship.twig
@@ -6,9 +6,8 @@
 {# Use option groups #}
 {% set groupby = relation.groupby|default(false) %}
 
-{% set values = context.content.relation.getField(relcontenttype, true, context.contenttype.slug, context.content.id)|default([]) %}
-
 {# Currently selected relationship values #}
+{% set values = context.content.relation.getField(relcontenttype, true, context.contenttype.slug, context.content.id)|default([]) %}
 {% set relselect = [] %}
 {% for item in values %}
     {# We use item.toid instead of item.to_id here to make sure that inversed relations are picked up correctly #}
--- a/app/view/twig/editcontent/fields/_relationship.twig
+++ b/app/view/twig/editcontent/fields/_relationship.twig
@@ -11,6 +11,7 @@
 {# Currently selected relationship values #}
 {% set relselect = [] %}
 {% for item in values %}
+    {# We use item.toid instead of item.to_id here to make sure that inversed relations are picked up correctly #}
     {% set relselect =  relselect|merge([item.toid]) %}
 {% endfor %}
 
--- a/app/view/twig/editcontent/fields/_relationship.twig
+++ b/app/view/twig/editcontent/fields/_relationship.twig
@@ -6,10 +6,12 @@
 {# Use option groups #}
 {% set groupby = relation.groupby|default(false) %}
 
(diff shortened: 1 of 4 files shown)


app/view/twig/editcontent/fields/_relationship.twig
```diff
@@ -6,10 +6,12 @@
...
 
+{% set values = context.content.relation.getField(relcontenttype, true, context.contenttype.slug, context.content.id)|default([]) %}
+
 {# Currently selected relationship values #}
 {% set relselect = [] %}
-{% for item in context.content.relation.getField(relcontenttype, true)|default([]) %}
-    {% set relselect =  relselect|merge([item.to_id]) %}
+{% for item in values %}
+    {% set relselect =  relselect|merge([item.getToId()]) %}
 {% endfor %}
```
GawainLynch: In Twig, you don't need the `get` part … Twig will always internally check for a function name that matches either , `is*()`, `has*()`, and `get*()`, so `item.toid` would be fine.

SahAssar: Well, the problem is that since this can be an inversed relation that will get the actual to_id, instead of the inversed to_id if i use twigs getters. see https://github.com/bolt/bolt/blob/858d652e49c3b61fef7e0f66f47750a4fe97d766/src/Storage/Entity/Relations.php#L81-L88

GawainLynch: Not relevant. The object property is `to_id`, the function name is `getToId()` :wink:

SahAssar: :tired_face: Fixed now.

app/view/twig/editcontent/fields/_relationship.twig
```diff
@@ -6,10 +6,12 @@
...
 
+{% set values = context.content.relation.getField(relcontenttype, true, context.contenttype.slug, context.content.id)|default([]) %}
+
 {# Currently selected relationship values #}
 {% set relselect = [] %}
-{% for item in context.content.relation.getField(relcontenttype, true)|default([]) %}
-    {% set relselect =  relselect|merge([item.to_id]) %}
+{% for item in values %}
+    {% set relselect =  relselect|merge([item.toid]) %}
 {% endfor %}
```
GawainLynch: OK… now going into COD mode … and would just ninja this but figure it's better to make it part of the dialogue.

We probably should put a `{{# comment #}}` explaining the "why" one vs the other … will happily do if your dance card is full though :+1:

SahAssar: Done :)

SahAssar:
On bidirectional relations that last commit will flip them on save since the correct state is the updated state where the saved record is now the `from` whereas in the old state it was the `to`. It also fixes the issue of deleting a bidirectional relation from the `to` record.

GawainLynch:
Awesome mate, thanks so much for this work!

PR #5643, opened 2016-08-05
Fix two issues with deferred widgets.

bobdenotter:
This PR fixes two issues related to deferred widgets:

 - It makes deferred widgets work in the backend
 - It prevents deferred widgets from being rendered twice (once during the initial request, and again when fetched ajaxy)

### DIFF
--- a/src/Asset/Widget/Queue.php
+++ b/src/Asset/Widget/Queue.php
@@ -152,8 +152,9 @@ class Queue implements QueueInterface
     /**
      * Render a location's widget.
      *
-     * @param string $location Location (e.g. 'dashboard_aside_top')
-     * @param string $zone     Either Zone::FRONTEND or Zone::BACKEND
+     * @param string $location        Location (e.g. 'dashboard_aside_top')
+     * @param string $zone            Either Zone::FRONTEND or Zone::BACKEND
+     * @param string $wrapperTemplate Template file for wrapper
      *
      * @return string|null
      */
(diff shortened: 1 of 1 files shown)


src/Asset/Widget/Queue.php
```diff
@@ -99,7 +99,7 @@ public function process(Request $request, Response $response)
     {
         /** @var WidgetAssetInterface $widget */
         foreach ($this->queue as $widget) {
-            if ($widget->getZone() === Zone::FRONTEND && $widget->isDeferred()) {
+            if ($widget->isDeferred()) {
                 $this->addDeferredJavaScript($widget, $response);
```
GawainLynch: Sanity check only… but there was a reason you didn't want deferred widgets in the backend, can you remember what that was?

src/Asset/Widget/Queue.php
```diff
@@ -164,7 +164,12 @@ public function render($location, $zone = Zone::FRONTEND, $wrapperTemplate = 'wi
         /** @var WidgetAssetInterface $widget */
         foreach ($this->sort($this->queue) as $widget) {
             if ($widget->getZone() === $zone && $widget->getLocation() === $location) {
-                $widgets[] = [ 'object' => $widget, 'html' => $this->getHtml($widget) ];
+                if ($widget->isDeferred()) {
+                    $html = '';
```
GawainLynch: Should be null to indicate that there wasn't deliberately a string there

src/Asset/Widget/Queue.php
bobdenotter: I _guess_ because during the refactor they didn't actually work as deferred in the backend. I did some monkey-testing, and at the moment it seems to work fine in my tests.

GawainLynch:
:+1:

PR #5657, opened 2016-08-08
Replace JSHint with ESLint

rarila:
This PR replaces JSHint with ESLint. This is because with this tool we have a superior fine grained control over what and how things get checked and are now able to also check coding style in one go, as JSCS got merged into ESLint.

So far only the old settings were converted and a few new ones added. Errors found were fixed. Expect more to be come.

Note: No js files were generated in this PR (sofar)


bobdenotter:
Looks good to me. 👍 

Note that when Gawain merges `release/3.1` into master, it'll update `package.json` again, but that should give no conflicts.

rarila:
As long as JSHint doesn’t get in again… :smile:

PR #5667, opened 2016-08-10
More exceptions work

### LINKED ISSUE #4921: "White screens of death" in 404 handler.
Details
-------

 - Relevant Bolt Version: [ master ]
 - PHP version: [ 5.5 | 5.6 | 7.0 ]

Reproduction
------------

 - Edit 'config.yml', set `strict_variables: true`.
 - Edit 'record.twig', and add to the body somewhere `Doesn't exist: {{ foo }}`
 - In the browser, trigger a 404 by going to `/ggggg` or whatever. 
 - Boom. Nothing! 

screenshot 2016-02-27 09 53 33

### LINKED ISSUE #5579: Migration of extension to Bolt3 crashes if getname() is used
If you migrate any extension from Bolt 2 to Bolt 3, you will most probably have a **getName()** method in your old extension, that is returning a string. In Bolt 3 however, this method is already defined in the base classe AbstractExtension. And it is defined as *final*.

Since you are now extending a method defined as **final**, your Bolt will crash. The problem with it is, Bolt's Error handling will brak to. You will have **no error**, and **no log entry** leading to a long hunt for invisible bugs.

**Suggestion**
1) I heard the error handling is rewritten. Please take this case into account and allways return something ro at least use the basic error reporting and logging. There is nothing worse than having to develop blind.

2) If there is no good reason for declaring the method **final** please consider keeping it consitant with the old method definiton or maybe even rename it. This creates problems with migration of extensions.

3) Update the migration guide to include this detail please.

Details
-------

 - Relevant Bolt Version: 3.0.11
 - PHP version:5.5
 - Used webserver: Apache

Reproduction
------------

Just create a getName() method in one of your extensions (local works too) and return some string there.

GawainLynch:
Fixes #4921
Fixes #5579

### DIFF
--- a/src/EventListener/BootInitListener.php
+++ b/src/EventListener/BootInitListener.php
@@ -94,7 +94,7 @@ class BootInitListener implements EventSubscriberInterface
         return [
             KernelEvents::REQUEST => [
                 ['onBootEarly', Application::EARLY_EVENT],
-                ['onBoot', 0],
+                ['onBoot', 33], // After routes determined. @see RouterListener::getSubscribedEvents
             ],
         ];
     }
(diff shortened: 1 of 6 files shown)


bobdenotter:
Whoooop! 💯

PR #5670, opened 2016-08-11
Hydrate repeater in templatefields

SahAssar:
This is done very much like https://github.com/bolt/bolt/commit/e28c253ad5d07f07c75f896a41c8634099071e8d in #5542, so hopefully this is mostly correct.

I've tested that it works with multiple repeaters in the same templatefields and that normal repeaters are uneffected.

This also removes the `populateTemplateFieldsContenttype` call since that method was removed in 	2f28f16 in the original template specific fields PR.

### DIFF
--- a/src/Storage/Entity/ContentValuesTrait.php
+++ b/src/Storage/Entity/ContentValuesTrait.php
@@ -297,10 +297,8 @@ trait ContentValuesTrait
                 }
 
                 if (is_array($unserdata)) {
-                    $templateContent = new \Bolt\Legacy\Content($this->app, $this->getTemplateFieldsContentType(), [], false);
-                    $value = $templateContent;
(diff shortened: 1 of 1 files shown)


src/Storage/Entity/ContentValuesTrait.php
```diff
@@ -397,6 +397,22 @@ public function setValues(array $values)
                 $this->values[$key] = $video;
             }
 
+            if ($this->fieldtype($key) == 'repeater' && is_array($this->values[$key]) && !$this->isRootType) {
```
GawainLynch: CDO nitpick: The comparisons here don't need to be loose

SahAssar: Ah, I just copied from L371. did a pass over that file, and they should all be strict now.

PR #5675, opened 2016-08-11
Fix slash in taxonomies

### LINKED ISSUE #5673: [BUG] Adding a tag with a slash crashes the content in backend and frontend
It is currently possible to add a taxonomy of type tag that includes the `/` character to a contenttype in the editor. If the contentobject is then safed, bolt will crash whenever this contentobject would be displayed.
This includes the contentobject's view on the frontend, aggregations and the backend edit view.

The crash can only be resolved by deleting the value containing that evil character from the bolt_taxonomy table manually.

**Suggested Solution**
Add a Whitelist for characters allowed in freetext taxonomies like taks.

Details
-------

 - Relevant Bolt Version: [ 3.0 ]
 - PHP version: [ 5.5 ]
 - Used web server: [ Apache ] version [ 2.4 ] Mysql
 - For UX/UI issues: [ Chrome ]

Reproduction
------------
- Open the edit view of any contenttype with a taxonomy field of type tag
- enter a value with a`/`character
- Press safe
- Reload page and/or open it on the frontend

SahAssar:
Fixes #5673

In taxonomy slugs we basically allow any character (since we do not slugify them). One way to test this is to add a tag with a bunch of special chars or with a taxonomy like this:

```yaml
categories:
    name: Categories
    slug: categories
    singular_name: Category
    singular_slug: category
    behaves_like: categories
    options: {'test&%¤': test}
```

The only character that we cannot use or have is a slash since that denotes a separation of url portions.

This PR removes any slash found in a taxonomy slug (since that's how we do it in record slugs) and shows a warning flash message if a slash is put in a slug in taxonomy.yml.

PRing into 3.0 since this is a bug fix, let me know if I should retarget to 3.1.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -859,6 +859,26 @@ class Config
     }
 
     /**
+     * Sanity check for slashes in in taxonomy slugs.
+     */
+    public function checkTaxonomy()
+    {
+        foreach ($this->data['taxonomy'] as $key => $taxonomy) {
+            if (!empty($taxonomy['options']) && is_array($taxonomy['options'])) {
+                foreach ($taxonomy['options'] as $optionkey => $optionvalue) {
+                    if (strpos($optionkey, '/') !== false) {
+                        $error = Trans::__(
+                            'general.phrase.invalid-taxonomy-slug',
+                            ['%taxonomy%' => $key, '%option%' => $optionvalue]
+                        );
+                        $this->app['logger.flash']->error($error);
+                    }
+                }
+            }
+        }
+    }
+
+    /**
      * Sanity checks for doubles in in contenttypes.
      */
     public function checkConfig()
--- a/src/Config.php
+++ b/src/Config.php
@@ -861,7 +861,7 @@ class Config
     /**
      * Sanity check for slashes in in taxonomy slugs.
      */
-    public function checkTaxonomy()
+    private function checkTaxonomy()
     {
         foreach ($this->data['taxonomy'] as $key => $taxonomy) {
             if (empty($taxonomy['options']) || !is_array($taxonomy['options'])) {
@@ -1005,6 +1005,7 @@ class Config
                 }
             }
         }
+        $this->checkTaxonomy();
     }
 
     /**
--- a/src/Config.php
+++ b/src/Config.php
@@ -864,20 +864,23 @@ class Config
     public function checkTaxonomy()
     {
         foreach ($this->data['taxonomy'] as $key => $taxonomy) {
-            if (!empty($taxonomy['options']) && is_array($taxonomy['options'])) {
-                foreach ($taxonomy['options'] as $optionkey => $optionvalue) {
-                    if (strpos($optionkey, '/') !== false) {
-                        $error = Trans::__(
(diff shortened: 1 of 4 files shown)


src/Config.php
```diff
@@ -859,6 +859,26 @@ protected function parseConnectionParams($params, $defaults = [])
...
+                        $error = Trans::__(
+                            'general.phrase.invalid-taxonomy-slug',
+                            ['%taxonomy%' => $key, '%option%' => $optionvalue]
+                        );
+                        $this->app['logger.flash']->error($error);
+                    }
+                }
+            }
+        }
+    }
```
GawainLynch: PSR-2 on the variable name case.

Also the nesting is a tad aggressive… Does this read better to you on review?
```php
    public function checkTaxonomy()
    {
        foreach ($this->data['taxonomy'] as $key => $taxonomy) {
            if (empty($taxonomy['options']) || !is_array($taxonomy['options'])) {
                continue;
            }
    
            foreach ($taxonomy['options'] as $optionKey => $optionValue) {
                if (strpos($optionKey, '/') === false) {
                    continue;
                }
    
                $error = Trans::__(
                    'general.phrase.invalid-taxonomy-slug',
                    ['%taxonomy%' => $key, '%option%' => $optionValue]
                );
                $this->app['logger.flash']->error($error);
            }
        }
    }
```

src/Controller/Backend/BackendBase.php
```diff
@@ -71,6 +71,7 @@ public function before(Request $request, Application $app, $roleRoute = null)
         // Sanity checks for doubles in in contenttypes. This has to be done
         // here, because the 'translator' classes need to be initialised.
         $app['config']->checkConfig();
+        $app['config']->checkTaxonomy();
```
GawainLynch: Just flagging for the a.m. 

This is not the place to be doing/firing this …

SahAssar: Well, where is the right place? I didn't really know where to put it, so I followed the checkConfig way of doing things, and assumed that following established code was the way to go until we actually get to the config refactor. That way it's more predictable, no?

GawainLynch: Not sure yet, was mid-wife-night and just noticed this walking past my screen … let me look post- :coffee:

GawainLynch: OK … soo looking at it, and thinking about it … probably best for now to put that at the end of `checkConfig()` as that is one of the things on my :fire: list for the 3.2 exception/request work.

Short version is that we're checking on every request the configuration of the configuration  … ConfigCeption … that is already cached … if it is cached, then we know it's already passed checks.

GawainLynch: Also moving the call means we can make that function private and not expose another API

GawainLynch: c.f. https://github.com/bolt/bolt/compare/master...GawainLynch:hotfix/config-check-request

GawainLynch:
Green it is going, it is … :shipit:

PR #5679, opened 2016-08-12
Fix for uploading to `themes/` folder in backend.

### LINKED ISSUE #5678: Uploading a files in `themes/` seems to be broken.
Uploading a file here: 

screen shot 2016-08-12 at 16 55 31

Gives this error: 

screen shot 2016-08-12 at 16 55 24

bobdenotter:
Fixes #5678 

This fix is a bit hackish. Open to better suggestions. :-)

### DIFF
--- a/app/view/twig/_macro/_macro.twig
+++ b/app/view/twig/_macro/_macro.twig
@@ -77,7 +77,10 @@
 {% macro typepopover(types) %}
     {% set types = types|default(app.config.get('general/accept_file_types')) %}
     {% set text = __("field.general.allowed-filetypes") ~ "<code>" ~ types|join("</code>, <code>") ~ "</code>." %}
-    <span class="label info-pop" data-title="" data-html="true" data-content="{{ text }}">{{ __('general.phrase.types') }}</span>
+    <span class="label label-default info-pop" data-title="" data-html="true" data-content="{{ text }}">
+        <i class="fa fa-info-circle"></i>
+        {{ __('general.phrase.types') }}
+    </span>
 {% endmacro %}
 
 {##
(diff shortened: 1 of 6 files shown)


rossriley:
Looks fine to me. I did make a note somewhere that if you mount a new filesystem name then there needs to be a corresponding entry in Resource Manager. 

One thing to check do we set in Configuration/Composer too if so we'll need to add the alias in that class too.

GawainLynch:
This is mounted in `bolt/filesystem` not here, deliberately.

CarsonF:
> Looks fine to me. I did make a note somewhere that if you mount a new filesystem name then there needs to be a corresponding entry in Resource Manager.

Do you mean the opposite?

bobdenotter:
Sorry guys, you've lost me. :-/

rossriley:
Sorry, was on mobile so I couldn't check. Was referring to this: https://github.com/bolt/bolt/blob/release/3.0/src/Provider/UploadServiceProvider.php#L65

No @bobdenotter no problem, just needed to check that we didn't replace `themebase` in the composer configuration but we don't so no need to worry.

CarsonF:
@rossriley Ah gotcha.

@bobdenotter I think what @GawainLynch means is we already have `themes` defined in [FilesystemServiceProvider](https://github.com/bolt/bolt/blob/release/3.0/src/Provider/FilesystemServiceProvider.php#L30).

bobdenotter:
> @bobdenotter I think what @GawainLynch means is we already have themes defined in FilesystemServiceProvider.

Yes, but uploading to the themes folder didn't work.

rossriley:
@bobdenotter sorry for confusing things, this PR is exactly what we need. No need to change anything.

My comment was just a note that in future if we add a new filesystem (themes was added to replace themebase) then we need to add a corresponding `getPath` in ResourceManager which is exactly what you've done in this PR.

GawainLynch:
@bobdenotter see 1af24d0705f2fc3c88583870d462ca9febafdd07

GawainLynch:
Teamwork FTW :+1:

PR #5680, opened 2016-08-13
Timed publish tweak

GawainLynch:
This makes the parameters for timed publishing checks independent of caching.

**[RFC]**
This PR introduces the `performance` configuration key, as a grouping for these types of parameters, with a sub key of `dispatchers` … thoughts on naming?

These should be considered advanced user territory.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -176,6 +176,13 @@ caching:
     thumbnails: true
     translations: true
 
+# Performance specific settings
+#
+# performance:
+#     timed_records:
+#         interval: 3600    # Number of seconds in between timed record checks
+#         use_cron: false   # If set to `true`, timed record checks will only be done via the hourly cron job
+
 # Set 'enabled' to 'true' to log all content changes in the database.
 #
 # Unless you need to rigorously monitor every change to your site's content, it
--- a/src/Config.php
+++ b/src/Config.php
@@ -1194,6 +1194,12 @@ class Config
                 'allowed_tags'       => explode(',', 'div,span,p,br,hr,s,u,strong,em,i,b,li,ul,ol,mark,blockquote,pre,code,tt,h1,h2,h3,h4,h5,h6,dd,dl,dh,table,tbody,thead,tfoot,th,td,tr,a,img,address,abbr,iframe'),
                 'allowed_attributes' => explode(',', 'id,class,style,name,value,href,src,alt,title,width,height,frameborder,allowfullscreen,scrolling'),
             ],
+            'performance'                 => [
+                'timed_records' => [
+                    'interval' => 3660,
+                    'use_cron' => false,
+                ],
+            ],
         ];
     }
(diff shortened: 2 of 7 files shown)


bobdenotter:
I'm not too enthusiastic about the name, but I don't have a better alternative yet. 

We should probably also ass some defaults to https://github.com/bolt/bolt/blob/master/src/Config.php#L1034 .

GawainLynch:
> We should probably also add some defaults to

Hmm … interesting point, architecturally. Right his very second the defaults are in the `get()`, i.e.

```
$this->config->get('performance/dispatchers/timed_records', 60)
```

So shift them into `Config`?

SahAssar:
Imo shifting them to config would be better, personally I don't like that we have config defaults spread out so much.

bobdenotter:
I agree with @SahAssar

GawainLynch:
WFM

bobdenotter:
Let's see if we come up with a better name, otherwise we'll merge this in! :-)

rossriley:
By way of an alternative suggestion, I quite like what Drupal does with these and allows performance conscious users to move them out of the request cycle, and install a cron job instead that runs every minute.

I'm not suggesting we do that immediately, but it might make for good naming consistency in the future if we do something like:

```
performance:
    cronjobs:
        timed_records: 60
```

Then in a future version when we offer cron versions of these, the config can be set to false either globally or individually:

```
performance:
    cronjobs:
        timed_records: false
        another_operation: 60
```
or
```
performance:
    cronjobs: false
```

GawainLynch:
At the risk of over thinking this … I can see a genuine need to, and not, use a cron job … ¿porque no los dos?

Would it make sense to have two, e.g.
```
performance:
    dispatchers:
        timed_records: false
    cronjobs:
        timed_records: 30
```

bobdenotter:
Since we already have a [Cron handler](https://github.com/bolt/bolt/blob/release/3.0/src/Cron.php), it makes sense to use that. 

Perhaps flip it around: 

```
performance:
    timed_records: 
        interval: 30
        use_cronjob: true
    foo: 
        interval: 60
        use_cronjob: false
```

bobdenotter:
What should we do with this one? Merge as-is, or update the code to my latest suggestion first? :-)

GawainLynch:
Ha! Yeah, I was looking at this PR last night.

I am tagging this for 3.3 as I want to land the configuration load/validate stuff first and take up the work that will come out of this week's hackathon from @PhillippOhlandt, @SahAssar and myself.

GawainLynch:
> Perhaps flip it around:

I like the overall idea, but something I am trying to work though in my head right now is the `interval` … for a request, as it is caching we're touching, the value of `interval` is and integer of seconds. With cron it is a string: `hourly`, `daily`, 'weekly`, etc …

GawainLynch:
Should be GTG now

PR #5683, opened 2016-08-14
Upload exceptions

GawainLynch:
@CarsonF pointed out that doing this check during the service construction creates a two+ requests for the filesystem … which on local is no big deal, but on remotes like S3 it's a different story.


bobdenotter:
👍

PR #5684, opened 2016-08-14
Improve Hydration Process in Repeaters

### LINKED ISSUE #5403: custom fields not working in repeater field, bug?
1. **Bug summary**: I am using the "https://github.com/bolt/bolt-extension-starter-extended" example, where a custom field called "examplefield", is created but I can not work with a field of type
"repeater", to save does not return any error (the field is ignored), and if you added the "name" attribute with the name of `repeat-field-name[index][fieldname-in-repeat],` in this case" `repeater [0] [bikeken] `"by submitting the form, it returns an error" unknown column "

**Expected result**: Save field :smile: 

 **Actual result**:
 By Default: 
To save does not return any error (the field is ignored)
Forcing repeat-field-name[index][fieldname-in-repeat]
If you added the "name" attribute with the name of `repeat-field-name[index][fieldname-in-repeat],` in this case" `repeater [0] [bikeken] `"by submitting the form, it returns an error" unknown column "
**Screenshots**: 
captura de pantalla_2016-06-03_00-17-01
captura de pantalla_2016-06-03_00-16-49

Reproduction
------------
**Steps to reproduce**: 
 - Create an extension that adds a custom field, based on the package "https://github.com/bolt/bolt-extension-starter-extended" start, make sure it is installed from the "extensions" on the backend of Bolt.
-  Add the field to a content type, section repeater (eg using the content type showcase).
-  In our content type, add a new entry, and fill our field in the repeater field.
-  Click on "save" button.

-------

 - Relevant Bolt Version:  3.0.3
 - Install type:Composer install 
 - PHP version:  5.6?
 - Used webserver: php built-in

rossriley:
This PR improves Repeater fields to make sure that both custom and built-in Bolt fieldtypes are correctly hydrated.

Fixes #5403

### DIFF
--- a/src/Storage/Field/Collection/FieldCollection.php
+++ b/src/Storage/Field/Collection/FieldCollection.php
@@ -112,8 +112,8 @@ class FieldCollection extends AbstractLazyCollection
             $instances = $repo->findBy(['id' => $this->references]);
 
             foreach ((array) $instances as $val) {
-                $boltType = $this->em->getMapper()->getFieldTypeFor($val->getContenttype(), $val->getName(), $val->getFieldname());
-                $fieldInstance = $this->em->getFieldManager()->get($boltType, ['fieldname' => $val->getFieldname() ]);
+                $fieldMapping = $this->em->getMapper()->getFieldMetadata($val->getContenttype(), $val->getName(), $val->getFieldname());
+                $fieldInstance = $this->em->getFieldManager()->get($fieldMapping['type'], $fieldMapping);
 
 
                 $fieldtype = $val->getFieldtype();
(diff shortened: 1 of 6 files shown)


src/Storage/Field/Collection/FieldCollection.php
```diff
@@ -115,7 +116,23 @@ protected function doInitialize()
                 $field = $this->em->getFieldManager()->getFieldFor($fieldtype);
                 $type = $field->getStorageType();
                 $typeCol = 'value_' . $type->getName();
-                $val->setValue($val->$typeCol);
+
+                // Because there's a potential for custom fields that use json storage to 'double hydrate' this causes
+                // json_decode to throw a warning. Here we prevent that by replacing the error handler.
+                set_error_handler(
+                    function ($errNo, $errStr, $errFile) {},
```
GawainLynch: @rossriley do you have an easily reproducible test case for the double hydration?  The Symfony debug handlers in 3.2 can convert the error into an exception and we should get ready to catch that.

PR #5691, opened 2016-08-17
More session save_path bugfixes

CarsonF:
- Fixes parsing url (`save_path`) when scheme is omitted, by defaulting it.
- Prevent `save_path` for a different `save_handler` from bleeding into options.
- Use default connection if not specified with `save_path` or `connection(s)`.


GawainLynch:
LGTM :+1:

PR #5696, opened 2016-08-20
Case-insensitive username lookup at login

### LINKED ISSUE #5695: Login seems to be case-sensitive
Details
-------

 - Relevant Bolt Version: [ 3.0.12 ]

Reproduction
------------

 - Use an admin user to create a new user 'test', with a straightforward password.
 - Open an incognito window, go to the log in page
 - Try loggin on with user 'Test' (note the capital 'T') and the chosen password.
 - See the `Your account is disabled. Sorry about that.` notice. 

See screenshot: 

screen_shot_2016-08-19_at_14_20_10

GawainLynch:
For the record, I am 1,000% **against** this being done in core.

Nonetheless this implements the database check for the username case-insensitively.

Closes #5695

### DIFF
--- a/src/Storage/Repository/UsersRepository.php
+++ b/src/Storage/Repository/UsersRepository.php
@@ -94,7 +94,10 @@ class UsersRepository extends Repository
         if (is_numeric($userId)) {
             $qb->where('id = :userId');
         } else {
-            $qb->where('username = :userId')->orWhere('email = :userId');
+            $qb
+                ->where($qb->expr()->like('username', ':userId'))
+                ->orWhere('email = :userId')
+            ;
         }
         $qb->setParameter('userId', $userId);
 
--- a/tests/phpunit/unit/Storage/Repository/UsersRepositoryTest.php
+++ b/tests/phpunit/unit/Storage/Repository/UsersRepositoryTest.php
@@ -27,7 +27,7 @@ class UsersRepositoryTest extends BoltUnitTest
         $this->assertEquals('SELECT * FROM bolt_users users WHERE id = :userId', $queryGetUserByID->getSql());
 
         $queryGetUserByName = $repo->getUserQuery('user');
-        $this->assertEquals('SELECT * FROM bolt_users users WHERE (username = :userId) OR (email = :userId)', $queryGetUserByName->getSql());
+        $this->assertEquals('SELECT * FROM bolt_users users WHERE (username LIKE :userId) OR (email = :userId)', $queryGetUserByName->getSql());
 
         $queryHasUsers = $repo->hasUsersQuery();
         $this->assertEquals('SELECT COUNT(id) as count FROM bolt_users users', $queryHasUsers->getSql());


rarila:
[RFC] I suggest bringing the login process to an absolute new level of convenience by adding AJAX driven suggestion boxes to the username and password fields. That unique feature would generate a real painless login experiance and makes an end to endless login attempts when you once again have forgotten how exactly your one hundred and eighty four character long password was spelled.

Don’t forget, Bolt stands for Simplicity!

loginng

SahAssar:
Wouldn't an easier/more correct solution be to throw `pattern="[a-z0-9]+"` on the input field so that it is simply validated on the frontend as lowercase?

bobdenotter:
I think the main reason why we need this is the following: Usernames are already stored lowercase. Consider the following example: 

 - Add a new user, name: `Mike`, password `hunter42`. 
 - Email Mike, telling him hist credentials: Mike / hunter42
 - Mike tries to log on, using `Mike` and hi password. 

This will not work, because on the log in, only `mike` in lowercase will be accepted. This PR does not in any way decrease security, it only lessens  confusion in certain situations.

SahAssar:
Well, the way I see it is that we have a requirement that all usernames are lowercase. Why not just add that requirement to the frontend validation for all username fields (including "create user" and "first user" forms)? Allowing incorrect input seems odd when we can just make the requirement clear for all users/admins when they create the user.

bobdenotter:
@SahAssar The field accepts the username or email, and email _is_ case sensitive, and allows for other characters that are hard to regex correctly. :-)

GawainLynch:
> This PR does not in any way decrease security, it only lessens confusion in certain situations.

Actually, it does… significantly.

bobdenotter:
> Actually, it does… significantly.

How? not trying to be facetious here, but I don't see that. Remember, the usernames are stored in lowercase anyways, so there's no decrease in the number of possibilities here.

GawainLynch:
Well, the "bug" is also that we're storing the usernames in lowercase only.

So when brute-forcing, the number of variants on a given n-Character string is one:
  * aaa

As opposed to (using a 3 character string as an example):
  * aaa
  * aaA
  * aAa
  * Aaa
  * aAA
  * AaA
  * AAa
  * AAA

i.e. we have a larger problem, so my comment is really meant to be taken about the approach as a whole, not just what is done with this PR

bobdenotter:
> Well, the "bug" is also that we're storing the usernames in lowercase only.

Yes, but that's a whole different discussion. This PR only lessens the confusion, in the situation that we happen to have right now. :-)

GawainLynch:
> Yes, but that's a whole different discussion. 

Actually no. Your comment probably indicates that I am doing a bad job of explaining myself (again) though.

> This PR only lessens the confusion, in the situation that we happen to have right now. :-)

It still reduces the attempts needed during blind-brute-force by orders of magnitude. 

Personally, I am not trying to block this change … but I am putting on public record that we're handling this the wrong way :disappointed:

rarila:
Is there some db versioning built in? That way it could be handled the “correct” way for newly created databases?

GawainLynch:
> Is there some db versioning built in? 

Nope.

rarila:
Adding a field ``UsernameCaseSensitive`` and use that case sensitive for newly created users? Or some boolean field ``UserNameWasCreatedCaseSensitive``…

bobdenotter:
> It still reduces the attempts needed during blind-brute-force by orders of magnitude.

Let's discuss this in person. I understand what you mean, I just disagree. The 'extra entropy' for case variations was not there, so 'lowercasing' it doesn't take away what wasn't there to begin with. :-)

BFDL says: Merging in!

bobdenotter:
@rarila We could do something like this, but that's also out of the scope of this PR. It should be taken into account with the 'larger problem' gawain mentioned.

rarila:
@bobdenotter Didn’t target this PR, just wanted to collect some ideas for a future solution…

GawainLynch:
I'm still laughing hard at that graphic you posted, @rarila

rarila:
@GawainLynch History tells that revolutionary people are always laughed at in the beginning  :fist: :stuck_out_tongue:

PR #5697, opened 2016-08-20
Show correct error message.

bobdenotter:
.. because at this point we don’t know yet if the account is enabled, only that the user wasn’t fetched. To keep the notice vage, we should return the generic “username or password incorrect”.

### DIFF
--- a/src/AccessControl/Login.php
+++ b/src/AccessControl/Login.php
@@ -186,7 +186,7 @@ class Login extends AccessChecker
     protected function getUserEntity($userName)
     {
         if (!$userEntity = $this->repositoryUsers->getUser($userName)) {
-            $this->flashLogger->error(Trans::__('general.phrase.login-account-disabled'));
+            $this->flashLogger->error(Trans::__('general.phrase.error-user-name-password-incorrect'));
 
             return null;
         }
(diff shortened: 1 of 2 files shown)


GawainLynch:
Then this one should be changed too https://github.com/bolt/bolt/blob/c902bdab3db674dd3ed9ed97e97d8c6c1c02f599/src/AccessControl/Login.php#L196

bobdenotter:
Lemme check something.. back in 10 mins.

GawainLynch:
WFM :+1: 

… should we tweak the "human" message, or is that already GTG?

bobdenotter:
Ok, modified PR.. Currently, it works like this: 

 - non-existing username (regardless of password) -> `username or pass incorrect`
 - Enabled user, wrong pass -> `username or pass incorrect`
 - Disabled user, wrong pass -> `username or pass incorrect`
 - Disabled user, correct pass -> `sorry, account disabled` 
- Correct user + pass -> Login success

So, the change in this PR will 'leak' less information about disabled users.

bobdenotter:
GTG, i think!

GawainLynch:
> So, the change in this PR will 'leak' less information about disabled users.

Love your work :-)

PR #5593, opened 2016-07-26
Fix attempts to access values via $this->values in storage

SahAssar:
Fixes #5466

I took out the taxonomy based routing because as far as I can tell it hasn't worked in 2.2 or 3.0 (tried with tags and groups, seems to not have worked for quite a while), but if there is some quick way to get it to work or if I'm wrong then please let me know.

PR #5698, opened 2016-08-21
Allow taxonomies in contentlinks again

SahAssar:
This reintroduces taxonomies in contentlinks which was ripped out in #5593 because it seemed to not work. It contains one case for legacy and one for the new storage as opposed to the old one removed in 5593 which only worked in legacy. The issue was reported on slack by techwilk.

Example route:

```
category:
    path: /{groups}/{slug}
    defaults:
        contenttypeslug: pages
        groups: none
    requirements:
        groups: [ controller.requirement:singleTaxonomy, [groups] ]
    contenttype: pages
```


GawainLynch:
:+1:

PR #5700, opened 2016-08-23
Default profiler exception redirection to false

GawainLynch:
While nice for a small amount of people, the redirection will change the route, disabling `F5`'s practicality when stepping though a problem.


bobdenotter:
👍

PR #5710, opened 2016-08-26
Fix .bolt being allowed to specify application as string

CarsonF:
.bolt files should be allowed to specify application as string. This fixes the check to ensure the string is a class name that extends Silex\Application.


GawainLynch:
I think I just spent 2 minutes staring at this … Fridays & coding should be avoided.

PR #5711, opened 2016-08-26
Fix loading controllers without before/after middlewares

CarsonF:
When loading `routing.yml` if a controller is referenced that doesn't have `before` and `after` methods, an exception would be thrown. This fixes that to just do nothing in those cases.

### DIFF
--- a/src/Routing/CallbackResolver.php
+++ b/src/Routing/CallbackResolver.php
@@ -58,7 +58,7 @@ class CallbackResolver extends \Silex\CallbackResolver
         if (isset($this->classmap[$cls])) {
             return true; // Will use service definition
         }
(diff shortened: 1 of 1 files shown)


src/Routing/CallbackResolver.php
```diff
@@ -58,7 +58,7 @@ public function isValid($name)
         if (isset($this->classmap[$cls])) {
             return true; // Will use service definition
         }
-        if (!class_exists($cls)) {
+        if (!class_exists($cls) || !method_exists($cls, $method)) {
             return false; // Can't handle this
```
GawainLynch: "Do do do do… do do do do… You can't handle this!"

PR #5712, opened 2016-08-26
contenttype -> ContentType (uppercasing the C and the T in contenttype)

SahAssar:
Use ContentType instead of contenttype, since the latter is not correct English and we have also moved to using ContentType in the docs. See bolt/docs/pull/482 for the docs commit.


SahAssar:
Alright, I'll need to fix the tests on this one...

GawainLynch:
The good news is … :beers:

SahAssar:
This now seems to fail because of the backend news feed which would output a log message saying "Using cached data", but since it can't fetch the news it's not logging that message. Let's just rerun travis when the news is back up, there is no rush to get this one in.

GawainLynch:
Oh look, Travis changed its mind … Just don't ask what the bribe was :wink:

PR #5722, opened 2016-08-31
fix link to available locales

Crunch09:
Hi,

while creating my first bolt project i noticed that the link in the config.yml which should list the available locales is 404-ing.

PS: For what it's worth, i don't like brown M&M's 😂


bobdenotter:
Hi, @Crunch09, thank you! 👍

GawainLynch:
Merged into 3.1 via 284441ec2f9f3ef4151c4bcaf4bba73868def744

PR #5723, opened 2016-08-31
Remove content edit link from toolbar.

CarsonF:
The toolbar is not the right place for it. It's available to the templates from the record object, which is even already used in the base-2016 theme.

### DIFF
--- a/app/view/toolbar/bolt.html.twig
+++ b/app/view/toolbar/bolt.html.twig
@@ -57,14 +57,6 @@
         </div>
         {% endif %}
 
-        {% if collector.editlink %}
-        <div class="sf-toolbar-info-piece">
-            <hr style="margin: 8px 0 12px; padding: 0; border: 0; border-bottom: 1px solid #777;">
-            <span> Edit: <a href="{{ collector.editlink }}">{{ collector.edittitle|excerpt(40)|default("(no title)") }}</a> </span>
-        </div>
-        {% endif %}
-
-
     {% endset %}
     {% include '@WebProfiler/Profiler/toolbar_item.html.twig' with { 'link': profiler_url } %}
 {% endblock %}
--- a/src/Application.php
+++ b/src/Application.php
@@ -69,10 +69,6 @@ class Application extends Silex\Application
         $locales = (array) $this['config']->get('general/locale');
         $this['locale'] = reset($locales);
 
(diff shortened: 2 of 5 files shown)


bobdenotter:
Very much 👎 . I (and at least a few people i know) use that all the time.

bobdenotter:
To expand on that, I'm sure the feature can be improved upon. I'll gladly merge in an _improved_ version of this feature, bur we shouldn't _remove_ features and leave them to someone else to reimplement later.

GawainLynch:
I am on the fence here … but what **must** go is things like this in `Application`
```
        $this['editlink'] = '';
        $this['edittitle'] = '';
```

CarsonF:
What @GawainLynch said. @bobdenotter I didn't think this was a feature _removal_ as it's in the current theme.

I still don't think this is the right place for it though. The toolbar is for developers, not editors. The toolbar is not even enabled in production, which it shouldn't be for performance reasons. The editors at GMO never even see it.

bobdenotter:
> I still don't think this is the right place for it though. The toolbar is for developers, not editors. 

I'm inclined to agree, but the current feature is in use, so we shouldn't rip it out. 

> The editors at GMO never even see it.

People tend to grow "blind spots" very quickly. Same thing happens with the news in the backend, or the warnings that you're about to install using SQLite. That's no reason to remove it still. :-)

CarsonF:
What I mean is it's never enabled in production, so they never could access it. Not that they are "missing it". It never shows for them.

The fact that your editors are using it raises questions as to why it's enabled for them in production. That's a big performance hit.

GawainLynch:
@CarsonF we need a way to move this forward when you have time.

Everyone is in agreement that the way data is getting to the toolbar isn't correct — darn annoying, but I digress — but that feature is used a lot during site development 

… and during that period most people who want/know to use that will have written their own templates that probably won't have the edit link in use there (I wouldn't).

So I'll leave this open to update, but current approach of this PR just won't fly :disappointed:

CarsonF:
Ok I'll see if I can find another way to just remove it from Application then.

GawainLynch:
@CarsonF are you going to tackle this in this branch, or can we clear the PR off?

CarsonF:
Yeah very low priority now.

PR #5726, opened 2016-09-01
Allow use of string class names in app bootstrap

### LINKED ISSUE #5725: [BUG] Can't extends Bolt\Application using .bolt.yml file
As a developer, extending the basic class is required in some cases, especially when the application has certain way to overwrite the class.

In my case, I'm trying to change the Bolt session to write it to memcached instead of filesystem.
Bolt offer the way to overwrite the Application class by `.bolt.yml` or `.bolt.php` file. 

Application is working when the `.bolt.php` is returning `$config['application']` with an initialized app instance, but if `$config['application']` only contain the Application classname it's not working.
After some debugging, I found out the bootstrap is checking the `$config['application']` classname using an instance checking (`is_a`) instead of classname checking (`is_subclass_of`).

https://github.com/bolt/bolt/blob/release/3.1/app/bootstrap.php#L121
```php
    // Create the 'Bolt application'
    $appClass = Application::class;
    if ($config['application'] !== null && is_a($config['application'], Silex\Application::class, true)) {
        $appClass = $config['application'];
    }
    $app = new $appClass(['resources' => $resources]);
```

The Fix will be on line 121
```php
    // Create the 'Bolt application'
    $appClass = Application::class;
    if ($config['application'] !== null && is_subclass_of($config['application'], Silex\Application::class, true)) {
        $appClass = $config['application'];
    }
    $app = new $appClass(['resources' => $resources]);
```

Details
-------

 - Relevant Bolt Version: master
 - Install type: GitHub checkout
 - PHP version: 5.6
 - Used web server: php buildin server

Reproduction
------------

1. create a file call `.bolt.yml` with `application: My\Application`
2. create a folder `my` and `application.php` file in it
```php
<?php

namespace My;

class Application extends \Bolt\Application
{

}
```
3. update `composer.json` autoload custom namespace `My` and run `composer update`
```json
    ...
    "autoload" : {
        "psr-4" : {
            "Bolt\\" : "src",
            "My\\" : "my"
        },
        ...
    }
    ...
```
​

rossriley:
Credit to @helman for this one.

Fixes #5725

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -118,7 +118,7 @@ return call_user_func(function () {
 
     // Create the 'Bolt application'
     $appClass = Application::class;
-    if ($config['application'] !== null && is_a($config['application'], Silex\Application::class, true)) {
+    if ($config['application'] !== null && is_subclass_of($config['application'], Silex\Application::class)) {
         $appClass = $config['application'];
     }
     $app = new $appClass(['resources' => $resources]);


app/bootstrap.php
```diff
@@ -118,7 +118,7 @@
 
     // Create the 'Bolt application'
     $appClass = Application::class;
-    if ($config['application'] !== null && is_a($config['application'], Silex\Application::class, true)) {
+    if ($config['application'] !== null && is_subclass_of($config['application'], Silex\Application::class)) {
         $appClass = $config['application'];
```
GawainLynch: ¿porque no los dos? As in, if it is a concrete `Silex\Application`, then `is_subclass_of()` will return `false`, so an `instanceof` check should follow (also to avoid deprecation notice on `is_a()`).

rossriley: I think if it gets to this block it's already not a concrete `Silex\Application` if you look up to line 95 then it returns.

I guess there's an argument for allowing `Silex\Application` as a string but I'm not sure how that would work out in terms of interface.

GawainLynch: You raise good points kind sir. :+1:

PR #5736, opened 2016-09-06
fix too generic selector in filescreenuploader

### LINKED ISSUE #5715: [UI] File types arrow's position is messed up
On 3.1
screen shot 2016-08-29 at 4 24 01 pm

SahAssar:
Fixes #5715

This selector was catching the tooltip arrow too.

In the end we'd probably want to not use any of these generic selectors, but this should be "good enough for now™".


bobdenotter:
Ugh, I completely missed that! Herp derp. 

Thanks, @SahAssar

PR #5744, opened 2016-09-09
Add FR translations

nbehier:
Add FR translation for new meta tabs.


GawainLynch:
So sorry for the delay in getting to this … one of those months :laughing: 

Thank you, as always :+1:

Issue #5729, opened 2016-09-04
[RFC] Don't use hidden files .bolt.yml and .bolt.php


rossriley:
Since we've moved to this setup we've had quite a few issues with users' production deploys breaking because many ftp/sftp applications and indeed operating systems ignore hidden files.

Historically dotfiles are primarily used for storing user preferences or extra metadata that ideally don't disable the program in question by being wiped / not there whereas in our case these files specify a directory structure that is co-dependent on the hidden files being present.

How do we feel about 

  1. Having them as plain filesystem files `bolt.yml` and `bolt.php` or maybe
  2. Having a new top level `boot` folder and keeping them in there for some extra structure.

Obviously in the intervening period we could support both but for new distributions we can update to the new method.

PhillippOhlandt:
I like the hidden file but I also see the problem. I suggest moving to `bolt.yml` and `bolt.php` but still fallback to `.bolt.*` for the people who like that.

GawainLynch:
> Having a new top level `boot` folder

:-1: 

>  keeping them in there for some extra structure.

That might work, but the place for them is probably `app/` IMHO (follow upstream's behaviour). However, the problem is two fold:
  1. This is where we define paths
  2. It needs to know where its paths are to find the path of the file that has the location of the files that knows where the files are to find the file that has the location of the …

The file itself is a compromise to allow path configuration, that needs to be loaded early enough that it can be used prior to the configuration cycle. We're almost there on fully abstracted filesystems :smile: 

Transition…  What is valid and override the other? A lot of installs are going to have edited/adapted a `.bolt.XXX` file now (and committed to git, etc) and we'd expect every one of those sites to reconfigure if they were installing from tarball as that would drop a new (non-hidden) file.

Moreover, a lot of people will want to stay with the hidden file approach as, to us, the non-hidden approach doesn't make sense for a file that should only ever be edited once.

bobdenotter:
This is the same issue as #5720

bobdenotter:
> Having them as plain filesystem files bolt.yml and bolt.php

I'd be in favor of this. With a 'fallback' to `.bolt.yml` for BC.

GawainLynch:
> I'd be in favor of this. With a 'fallback' to `.bolt.yml` for BC.

Yes, but that will break installs for people who use tarball *and* have modified the file's contents

bobdenotter:
> Yes, but that will break installs for people who use tarball and have modified the file's contents

[...]

PR #5748, opened 2016-09-11
Update bootstrap to check non-hidden config files

rossriley:
As discussed in #5729 Check for alternative non-dot versions of `.bolt.yml` and `.bolt.php`.

There's an additional minor improvement here and that allows a `bolt.php` file to just return a pre-constructed application rather than array. This is optional but simplifies the flow of php bootstrap files that will normally take care of creating an instance of `Bolt\Application`.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -95,9 +95,9 @@ return call_user_func(function () {
 
     // An extra handler if a PHP bootstrap is provided, allow the bootstrap file to return
     // a pre-initialized Bolt Application rather than the config array.
-    if ($php !== null && is_array($php)) {
+    if (isset($php) && is_array($php)) {
         $config = array_replace_recursive($config, $php);
-    } elseif ($php !== null && $php instanceof Silex\Application) {
+    } elseif (isset($php) && $php instanceof Silex\Application) {
         return $php;
     }
 
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -95,9 +95,9 @@ return call_user_func(function () {
 
     // An extra handler if a PHP bootstrap is provided, allow the bootstrap file to return
     // a pre-initialized Bolt Application rather than the config array.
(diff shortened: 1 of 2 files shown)


bobdenotter:
Huzzah! 🎊🎉

app/bootstrap.php
```diff
@@ -84,11 +84,21 @@
...
-        if (is_array($php)) {
-            $config = array_replace_recursive($config, $php);
-        }
+    } elseif (file_exists($rootPath . '/bolt.php')) {
+        $php = include $rootPath . '/bolt.php';
+    }
+
+    // An extra handler if a PHP bootstrap is provided, allow the bootstrap file to return
+    // a pre-initialized Bolt Application rather than the config array.
+    if (is_array($php)) {
```
GawainLynch: PHP isn't defined for all paths … Travis go bye bye

rossriley: Travis the fascist.

app/bootstrap.php
```diff
@@ -84,11 +84,21 @@
...
         $php = include $rootPath . '/.bolt.php';
-        if (is_array($php)) {
-            $config = array_replace_recursive($config, $php);
-        }
+    } elseif (file_exists($rootPath . '/bolt.php')) {
+        $php = include $rootPath . '/bolt.php';
+    }
+
+    // An extra handler if a PHP bootstrap is provided, allow the bootstrap file to return
+    // a pre-initialized Bolt Application rather than the config array.
```
CarsonF: This is already allowed below though?

CarsonF:
Should bolt/composer-install be updated as well?

bobdenotter:
As discussed in the dev-meeting: Merging in for 3.2, so we can test and bugfix before release! See https://github.com/bolt/bolt/wiki/Dev-meeting-2016-09-20

PR #5751, opened 2016-09-12
Updating default `debug_error_level` from `6135` to `8181`.

bobdenotter:


GawainLynch:
Well, on 3.2 if you have it turned up that high, Symfony converts errors to exceptions … so do this with caution :smile:

bobdenotter:
@GawainLynch it's `8181` in `config.yml.dist as well, so it makes sense that they're consistent, at least. https://github.com/bolt/bolt/blob/release/3.1/app/config/config.yml.dist#L261

GawainLynch:
Yeah, but we still might want to tone down the default, and advise people tracing issues to turn it up.

bobdenotter:
Actually, this is "toning it down" compared to the previous version, right? 

previous: "all, but not notices and deprecations" 
new: "all, but not notices, deprecations and also not warnings"

GawainLynch:
Possibly… But I've been having to wind my setting down at certain times lately on `release/3.2`+.

bobdenotter:
I'm not at all against reviewing the setting for 3.2, but I think this PR _does_ make it more consistent for 3.1.. And when we update for 3.2, it'll be easier to find the two same values to update them to whatever we choose to set it to. ;-)

GawainLynch:
There you go with the logic again… :stuck_out_tongue_winking_eye:

Issue #5759, opened 2016-09-13
[BUG] Filebrowser breaks when folder has files with certain filenames


xiaohutai:
With thanks to **suther** for reporting this on Slack.

Related thread on forums (no solution): https://discuss.bolt.cm/d/427-error-when-opened-theme-folder

Details
-------

 - Tested on Bolt Versions: 3.0.10 and 3.1.1
 - Install type: Composer install
 - PHP version: 5.6
 - Used web server: Apache

Reproduction
------------

 * When logged in the back-end, navigate to **File Management** -> **View/edit Templates** (i.e. `/bolt/files/themes/`)
 * Create a file with one of the breaking options below:
   * a file without an extension (e.g. `LICENSE`)
   * a file with a name and the extension `.save` (e.g. `foo.save`)
   * a file like `config.yml` with something appended (e.g. `config.yml.foo` or `config.ymlx`, etc.)
   * and more?

The following exception will be thrown:
```
An exception has been thrown during the rendering of a template ("Requested url themes is not available") in "@bolt/files/_files.twig" at line 53.
```
This refers to https://github.com/bolt/bolt/blob/fde4e50fbbca13d90fea34f502477b4e26796593/app/view/twig/files/_files.twig#L53

It might be the magic stuff that's related to:
https://github.com/bolt/bolt/blob/release/3.1/src/Filesystem/Plugin/AdapterPlugin.php#L46

CarsonF:
The problem is ResourceManager doesn't have a `themes` url.

Even though it was referenced (indirectly) like this:
```php
$fs->url('themes://foo.jpg')
```

The `url` method calls `$app['resources']->getUrl('themes')`, but nothing was there.

PR #5760, opened 2016-09-13
Fixed: Files without recognised extensions don't break the file browser. 

bobdenotter:
Fixes: #5759, until a better solution lands.

### DIFF
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -90,6 +90,7 @@ class ResourceManager
         $this->setUrl('upload', '/upload/');
         $this->setUrl('bolt', '/bolt/');
         $this->setUrl('theme', '/theme/');
+        $this->setUrl('themes', '/theme/'); // Deprecated, here for BC. See #5759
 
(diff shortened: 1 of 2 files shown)


CarsonF:
One thing though, it's not really deprecated anymore than the class itself is.

bobdenotter:
Should I then just omit the comment there? It might be good to keep a reference to the issue number in, right?

CarsonF:
Sure, if you want. I don't think it's necessary. It was really just an oversight before.

bobdenotter:
Updated the comment.

CarsonF:
Added explaination to that issue too

CarsonF:
Really the line above it for `setUrl('theme'` could be removed. It isn't needed there as it's set in `setThemePath`. If it looks confusing to you to have both there, otherwise :shipit:

bobdenotter:
I'm a bit wary to remove that one. Afraid it'll come back to haunt us in unexpected ways. :-)

CarsonF:
Ok no worries.

PR #5761, opened 2016-09-13
[WIP] Update FS's url plugin to use Symfony's Asset Packages

CarsonF:
Currently the Filesystem Url Plugin references `ResourceManager::getUrl` which is a deprecated class. This PR replaces that plugin with two plugins that use Symfony's Asset component. This is a good fit because the Asset component is already used for creating urls in our templates, and manages different base paths/urls and versioning strategies.

The `Plugin\Url` class handles a single package for a single filesystem, and `Plugin\ComposeUrls` class handles multiple packages. We'll only use the latter one right now, but the former one could be used to override for a single filesystem.

As [`filesystem`](https://github.com/bolt/bolt/blob/master/src/Provider/FilesystemServiceProvider.php) replaces `ResourceManager::paths`, now [`asset.packages`](https://github.com/bolt/bolt/blob/master/src/Provider/AssetServiceProvider.php#L22-L34) will replace `ResourceManager::urls`.

There are already some packages there, we can add more but I would only like to add what is needed. Some filesystems don't make sense to have public urls, like _cache_.


CarsonF:
Moved to a bigger PR incoming

GawainLynch:
@CarsonF Removing 2 lines doesn't make it bigger…  :stuck_out_tongue_winking_eye:

PR #5768, opened 2016-09-14
Bolt changed to make image aliases work #3703

ToBe998:
Changes to the default configuration and to the twig filters image and thumbnail to allow aliases for image resolutions as discussed in #3703. 

The new config value `restrict_alias` is set in all the old style `$app["myconfig param name"] things in the providers to keep it in line with current code. **But** the code will only use the $app->["config"]->get()` methods to access them like the new code does.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -196,7 +196,7 @@ thumbnails:
     save_files: false
     allow_upscale: false
     exif_orientation: true
-    restrict_alias: false
+    only_aliases: false
 #    browser_cache_time: 2592000
 
 # Define the HTML tags and attributes that are allowed in 'cleaned' HTML. This
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -196,6 +196,7 @@ thumbnails:
     save_files: false
     allow_upscale: false
     exif_orientation: true
+    restrict_alias: false
 #    browser_cache_time: 2592000
 
 # Define the HTML tags and attributes that are allowed in 'cleaned' HTML. This
--- a/src/Config.php
+++ b/src/Config.php
@@ -1080,6 +1080,7 @@ class Config
                 'cropping'          => 'crop',
                 'notfound_image'    => 'view/img/default_notfound.png',
                 'error_image'       => 'view/img/default_error.png',
+                'restrict_alias'    => false,
             ],
             'accept_file_types'           => explode(',', 'twig,html,js,css,scss,gif,jpg,jpeg,png,ico,zip,tgz,txt,md,doc,docx,pdf,epub,xls,xlsx,csv,ppt,pptx,mp3,ogg,wav,m4a,mp4,m4v,ogv,wmv,avi,webm,svg'),
             'hash_strength'               => 10,
--- a/src/Config.php
+++ b/src/Config.php
@@ -1080,7 +1080,7 @@ class Config
                 'cropping'          => 'crop',
                 'notfound_image'    => 'view/img/default_notfound.png',
                 'error_image'       => 'view/img/default_error.png',
-                'restrict_alias'    => false,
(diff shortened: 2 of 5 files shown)


bobdenotter:
Liking the look of this at first glance! 👍 

(You've made the PR against release/3.0, but this will probably land in Bolt 3.2, since that's the first upcoming version with new features)

CarsonF:
> (You've made the PR against release/3.0, but this will probably land in Bolt 3.2, since that's the first upcoming version with new features)

And you can change that with the edit button next to the title.

bobdenotter:
> And you can change that with the edit button next to the title.

Theoretically. ;-)

CarsonF:
Can `aliases` go under `thumbnails` instead of another root key, `image_aliases`?

bobdenotter:
> Can aliases go under thumbnails instead of another root key, image_aliases?

There is no `thumbnails` in `theme.yml`. I think it makes sense the way it is, personally.

CarsonF:
Also why do we need a different route for this? Why can the alias just generate the same route with the parameters it's configured with?

ToBe998:
Thanks for the feedback so far! :)

@bobdenotter I'll try to pick a fight with github!

@CarsonF I actually like the idea of **thumbnails > aliases**. It would then have same structure as the block in config.yml. THis would allow us to merge both configs or move to either file in the future. Im very sure the future will have a $app["config"]->get("thumbnails/aliases") which would decide for itself if it's in "general" or "theme". This would make all kinds of overriding possible. Like it! Objections to the change?

@CarsonF Afaik each route in silex needs it's own unique route id. If you dont specify one with bind, it will do so itself internally. THis does however **not** affect the url. The new function of course uses the same url as the old thumbnails function, just with different parameters. So these two work now:
```
/thumbnails/myimageformat/myimage.jpg
/thumbnails/100x200c/myimage.jpg
```

CarsonF:
> @CarsonF I actually like the idea of thumbnails > aliases.

👍 


> /thumbnails/myimageformat/myimage.jpg

I'm saying why have this route? Why not just read the alias, and use it's values to generate the same route as without an alias.
```
myalias:
  size: [400, 300]
  cropping: crop

{{ thumbnail('foo.jpg, 'myalias') }}

/thumbs/400x300c/foo.jpg
```

ToBe998:
@CarsonF the whole point of this change is to protect against calls with random resolutions. If we make the twig filter just generate the old style url from the given alias (with resolutions and stuff) and then serve all resolutions as before, we would not win much. We would just introduce aliases to template designers but not protect the thumbnail generator in any way and still be open to DoS attacks. That's why we need to generate urls with the alias and make the generator resolve them on it's own.

CarsonF:
So the concern is abusing our thumbnail route, and our solution has nothing to do with locking that route down? 💀 

Don't get me wrong, I think named aliases are a great way to dedup thumbnail config. But I think that needs to be a separate argument/feature from the security concern.

ToBe998:
@CarsonF Did you check the PR to bolt-thumbs? I think the route is locked down now. Did i miss some security hole?

CarsonF:
👊  my b, only glanced through. 

I'll review it properly...with GitHub's **NEW** PR reviews :metal:

GawainLynch:
@CarsonF any chance to have a look and update @ToBe998 … I'm happy with your call regards merge.

bobdenotter:
Bumping this for @CarsonF and @ToBe998 

3.2 is drawing closer. Let's get this one over the line. ;-)

ToBe998:
afaik all feedback has been integrated.

GawainLynch:
@CarsonF GTG then?

GawainLynch:
Thanks @ToBe998 … sorry for the delay on this!

PR #5773, opened 2016-09-15
Fix TemplateChooser templates existence checks

CarsonF:
Updated TemplateChooser to check if templates exist from twig environment. This is better as there can be additional loaders, that are in different locations than standard. Only checking the path breaks that abstraction anyways.


GawainLynch:
Merged into `release/3.2` via eaaff6b17a3b01ed543ebe2bccca5f149acf5e50

PR #5774, opened 2016-09-15
add compatibility with strict_variables params

dadaxr:
add "is_defined" check to be compatible with strict_variables: true
else it breaks badly !


rarila:
Ok, I'm a bit late… :smile: 

``{% if t.class is defined and t.class is not empty %}`` could be shortened to ``{% if t.class.default() %}`` and ``t.class.default()…`` in the next line could be shortened to ``t.class…`` as we make shure it is defined and not empty in the if clause anyway.

PR #5787, opened 2016-09-19
Fix the underscore hyphen debacle

rossriley:
OK, so this is an interim fix that's designed to workaround the existing issues with hyphens / underscores without changing too much of the code that drives config itself.

The underlying issue is that currently the main config check doesn't happen until boot, whereas most of the new storage system assumes that the config is already read and processed on register.

In reality there's probably only a small sub-set of functionality that needs to wait until boot but in the current state, nothing is processed in time.

So this PR moves config to a two stage process that hopefully gives us the functionality that we need in most use cases, at the expense of consistency between register and boot states.

Edit to add:

This also adds a soft warning when a user puts a hyphen in their contenttypes definition.

Under the hood it is translated to an underscore, but with the warning that they should update.

screen shot 2016-09-19 at 11 10 05


GawainLynch:
> In reality there's probably only a small sub-set of functionality that needs to wait until boot but in the current state, nothing is processed in time.

Just to add context for those reading along … the configuration loading/validation is being replaced in 3.3, and will address this. 

The problem currently is the validation & loading needs services that shouldn't be started until boot.

src/Configuration/ConfigurationValueProxy.php
```diff
@@ -0,0 +1,114 @@
+<?php
+/**
+ * Class ConfigurationProxy a simple wrapper that allows passing a pointer to the eventual
+ * compiled and validated configuration.
+ * @package Bolt\Configuration
+ * @author Ross Riley <riley.ross@gmail.com>
+ */
```
GawainLynch: G-pick … could you push that down on to of the class declaration and drop the `@package`

GawainLynch: Given it is uncertain the life time of the one, could you also tag it `@internal`

bobdenotter:
Interim fix or not, this makes me quite happy! 👍

bobdenotter:
As discussed in the dev-meeting: Merging in for 3.2, so we can test and bugfix before release! See https://github.com/bolt/bolt/wiki/Dev-meeting-2016-09-20

dadaxr:
This fix is not working on my side : 

just tried to add a content type with those 2 fields : 

```
...
image:
            type: image
            attrib: title # Note: retrieve this in your template with {{ record.values.image.title }}
            extensions: [ jpg, png ]
mobile_image:
            type: image
            extensions: [ jpg, png ]
```

Tried to add a new record of that type using the admin panel and got when saving : 

```
Uncaught Fatal Error:

Details

Doctrine\DBAL\Exception\NotNullConstraintViolationException in AbstractMySQLDriver.php line 112

An exception occurred while executing 'INSERT INTO bolt_sliders (slug, datecreated, datechanged, datepublish, datedepublish, username, ownerid, status, templatefields, title, caption, css_class, position, url, target, image, mobile_image) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params ["slug-f25d016fbae3576b0c01e1777e8a0e60", "2016-09-28 09:19:29", "2016-09-28 09:19:29", "2016-09-28 09:19:12", null, "", "1", "published", "[]", "Slider 1", "", "", 0, "#", "_self", "{\"file\":\"2016-09\\\/slider-1.jpg\"}", null]:

SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'mobile_image' cannot be null

```

rossriley:
Ok, I've tracked this down now. The bug is actually at another part of the codebase, so this PR is still ok.

I'll put in another PR shortly to fix.

PR #5793, opened 2016-09-21
Bumping grunt-eol to v1.1.3

rarila:
An upstream fix makes [grunt-eol](https://github.com/psyrendust/grunt-eol) install on windows again.

PS: The two remaining ones with grunt v1.0 problems (``grunt-phpdocumentor`` and ``grunt-remove``) have pull requests in the pipeline but seem somewhat abandoned. As a workaround you can use ``npm`` v3 to turn the install error into a warning.


bobdenotter:
Great, thanks! 

(Yeah, i've seen the warnings as well. :-) )

Issue #5796, opened 2016-09-21
No longer possible to have "no tabs", instead there's always "ungrouped"


bobdenotter:
It is no longerpossible to have no tabs in 'edit content', where all fields are displayed in one 'screen'. Instead, there's always an 'ungrouped' tab, even if no groups were defined. 

Details
-------

 - Relevant Bolt Version: [  3.1 ]

Reproduction
------------

Create a simple contenttype: 

```

dummy:
    name: Dummies
    singular_name: Dummy
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        content:
            type: textarea
            default: pompidom
```

Create a new record, and you'll see the ungrouped tab: 

![screen shot 2016-09-21 at 13 06 48](https://cloud.githubusercontent.com/assets/1833361/18708874/5fc5b61e-7ffd-11e6-856d-50924a1873ba.png)

GawainLynch:
#1555

PR #5797, opened 2016-09-21
Fixed: Don't display 'ungrouped' tab, if no groups are defined.

bobdenotter:
Fixes: #5796


GawainLynch:
Given the effort required to back trace on this one, I'm just blaming the :koala:

PR #5801, opened 2016-09-21
Better feedback when Records can't be saved.

### LINKED ISSUE #5788: [RFC] clearer warning when Bolt cannot save
Ohai!
When saving goes wrong, there's a small little warning that appears, that can easily be overlooked. 
It would make sense to give this a more warning - like appearance. 
I think this can be done easily by adding a class 'alert alert-danger'.

top: alert now.
bottom: prererred alert.

screen shot 2016-09-19 at 14 26 45

Details
-------

 - Relevant Bolt Version: [ 3.1 ]

### LINKED ISSUE #5803: [RFC] When a record is saved, show a tick icon to indicate saving has been successful
For a better user experience, when the record has been successfully saved, it should show a tick icon and after a few seconds revert back to the flag.

While there is checking the timestamp as a way of making sure it's saved, often people continue to look at the button they've pressed, so it just adds that little extra bit of UI sugar.

Example:

saved

And after 2/3 seconds or so, revert back to current format

tosaved

Applied `.fa-fw` class in both examples to the icon so that the icon change doesn't "bump" the buttons around.

Details
-------

 - Relevant Bolt Version: 3.0+
 - For UX/UI issues: Chrome 33+

bobdenotter:
Fixes #5788
Fixes #5803

When saving a record fails, you'll get a more visible error message now. See: 

http://recordit.co/eeeYMiT986

Screenshot of an actual Bolt error occurring: 

screen shot 2016-09-21 at 15 54 54


src/Controller/Base.php
```diff
@@ -155,10 +155,16 @@ protected function redirect($url, $status = 302)
...
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse|Symfony\Component\HttpFoundation\JsonResponse
      */
     protected function redirectToRoute($route, array $parameters = [], $status = 302)
     {
+        // Don't redirect on ajaxy requests, but abort with a 500 code instead.
+        if ($this->app['request']->isXmlHttpRequest()) {
+            $response = ['error' => ['message' => 'Redirect detected. Check if you are still logged on.'] ];
+            return new JsonResponse($response, 500);
+        }
```
GawainLynch: Wrong approach mate. This will affect all AJAX requests requiring a redirect and 995 would be valid.

bobdenotter: Currently, all 'ajaxy' requests that will redirect will fail silently. What happens is that it gets redirected, and returns an HTML page of the login screen, with a `200` status. As far as I can tell these just fail silently, and in javascript the `.done()` or `.success()` will be triggered, which is _not_ correct. 

This change will send a 500, so in the requesting javascript it can be caught. At the very least it'll trigger the `.fail()` method, allowing us to give feedback to the user.

bobdenotter:
As mentioned in #5803, I've also updated the icon in the save button, to reflect a successful or failed save: 

Failure: screen shot 2016-09-21 at 17 50 12

Success: http://recordit.co/YBxyHjHC2J

src/Controller/Base.php
Maelstromeous: 👍 for using proper error codes!

GawainLynch:
:tada: merging…

PR #5805, opened 2016-09-21
In fields, only apply the default to an actual null value

### LINKED ISSUE #5795: Fields with a `default: ..` value keep getting that value.
A field with a `default: ..` value should get that value when creating a new record. Currently they also get this value when editing an existing record in the backend.

Details
-------

 - Relevant Bolt Version: [ 3.0 and up ]

Reproduction
------------

Create a simple contenttype, like this: 

```
dummy:
    name: Dummies
    singular_name: Dummy
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        content:
            type: textarea
            default: Pompidom
```

Create a new record, and you'll this, as expected: 

screen shot 2016-09-21 at 13 06 48

Edit the fields, but _empty_ the textarea fields and save: 

screen shot 2016-09-21 at 13 07 18

The database will contain the expected results: 

screen shot 2016-09-21 at 13 09 10

Refresh the page, and the `content` will again contain the `default` value. 

screen shot 2016-09-21 at 13 09 57

rossriley:
Hopefully this fixes #5795, only add a default value on real null values, not if we've replaced with an empty text field.

### DIFF
--- a/app/view/twig/editcontent/fields/_text.twig
+++ b/app/view/twig/editcontent/fields/_text.twig
@@ -29,7 +29,7 @@
         required:        option.required,
         title:           option.title,
         type:            (option.pattern in ['url', 'email']) ? option.pattern : 'text',
-        value:           context.content.get(contentkey)|default(field.default)|default(''),
+        value:           context.content.get(contentkey)|default(''),
     }
 } %}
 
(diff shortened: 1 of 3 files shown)


bobdenotter:
Unfortunately, this doesn't fix the issue.. In the example in 5795, the `pompidom` keeps reappearing.

bobdenotter:
Whoop! Merging in.

PR #5816, opened 2016-09-24
Removal of Slect2

GawainLynch:


SahAssar:
What is it replaced with or planned to be replaced with? We use select2 for quite a bit of functionality, and this'd be a breaking change for some extensions, so this needs to go into a major release, no?

GawainLynch:
Problem is that you can't use it in an extension anyway … it keeps breaking.

This needs to be replaced with something that works.

SahAssar:
The only way the shipped version is broken IIRC is that it does not work on IE (works on edge though).

Just straight up removing it without implementing an alternative is not the way to go. This removes a lot of functionality.

GawainLynch:
> Just straight up removing it without implementing an alternative is not the way to go. This removes a lot of functionality.

Yes, and it will need to be replaced … but I am not a good one to be let near CSS, etc.

GawainLynch:
Also important to note in this, is that some of it works for sure … but you can't reliably implement with it when working on, or extending, the backend … sometimes it will work, sometimes not.

I've wasted this whole week tracking this junk to some sort of resolution, the library is crap, as is NPM & Grunt and they've gotta go

SahAssar:
> Yes, and it will need to be replaced … but I am not a good one to be let near CSS, etc.

Well, then someone needs to take this on and implement an alternative before this is merged, right?

> I've wasted this whole week tracking this junk to some sort of resolution, the library is crap, as is NPM & Grunt and they've gotta go

I appreciate that you are frustrated with it, but we can't just 🔥 stuff without implementing an alternative solution, or if we do we need to do it in a major release.

GawainLynch:
The alternative should be built on top of what we remove here.

I firmly disagree that this is a major version issue … but screw it … we'll just split out the backend

PR #5817, opened 2016-09-24
[Travis] Test npm & grunt install, and run

GawainLynch:
Every other day the running joke that is the JavaScript ecosystem breaks the ability to install, update or debug our components.

Simply, NPM/Grunt has to go as part of our build chain!

As a precursor, this PR attempts to build the JS/CSS assets and libraries and will fail to pass if that dies.


GawainLynch:
No buy-in on this, and NPM is :gun: :fire: :boom: … closing!

PR #5822, opened 2016-09-25
Fixed: Fix time picker regexp: slashes are not needed for regex as string.

bobdenotter:


GawainLynch:
Fails due to remote timeouts

PR #5836, opened 2016-09-29
Filesystem / Assets Phase II

CarsonF:
# Overview
- Rewrote canonical code
- Replaced all `ResourceManager::getUrl` calls with url generator or assets
- Replaced some `ResourceManager::getPath` calls with our filesystem abstraction
- All JS code like `bolt.conf('paths.async')` were replaced with routes generated from Twig.
- Removed visibility checks for remote filesystem compatibility
- Updated UIs for file managing
- Rewrote Stack to work with our filesystem abstraction
- Stack changes now apply to stack lists ("Select from stack" button for content fields)
- Various other Stack fixes

# What you need to know

#### PHP
```php
$app['resources']->getUrl('...')
$app['paths'][...]
```
#### Twig
```twig
{{ paths.foo }}/bar
{{ app.resources.url(...) }}
{{ app.resources.path(...) }}
```

### Do this instead:

This needs to be inside the request cycle.

#### PHP
```php
// For routes, generate them with the url generator
$app['url_generator']->generate('homepage');

// For assets (files to be loaded in browser)
$app['asset.packages']->getUrl('img/foo.jpg', 'packageName')
// Package name is the the grouping similar to url/path namespace
// Valid package names: bolt, files, theme, themes, extensions
```
#### Twig
```twig
//  Route
{{ path('homepage') }}

// Asset
{{ asset('img/foo.jpg', 'bolt')
```

#### PHP
```php
$app['resources']->getUrl('canonical');
$app['resources']->getUrl('canonicalurl');
$app['resources']->setUrl('canonicalurl', '...');
```

#### Twig
```twig
{{ paths.canonical }}
{{ paths.canonicalurl }}
{{ app.resources.canonical }}
{{ app.resources.canonicalurl }}
```

### Do this instead:

This needs to be inside the request cycle.

#### PHP
```php
$app['canonical']->getUrl();
```

#### Twig
```twig
{{ canonical }}
```

## Filesystem Paths

Try to use `$app['filesystem']` over `$app['resources']->getPath()`. I know this is less defined, and there is still work that needs to be done here.


app/web.php
```diff
@@ -14,6 +14,14 @@
     if (is_file($_SERVER['DOCUMENT_ROOT'] . preg_replace('#(\?.*)$#', '', $_SERVER['REQUEST_URI']))) {
         return false;
     }
+
+    // Fix server variables for PHP built-in server so base path is correctly determined.
```
GawainLynch: Shouldn't we wrap this in a check for that?

tests/codeception/acceptance/100-BackendAdmin/BackendAdminCest.php
```diff
@@ -569,10 +569,8 @@ public function logoutAdminUserTest(\AcceptanceTester $I)
         $I->see('Dashboard');
         $I->click('Logout');
 
-        // Removed as we now unset the session cookie at logout
-        //$I->see('You have been logged out');
-
         $I->amOnPage('/bolt');
-        $I->see('Please log on');
+
+        $I->seeInCurrentUrl('/bolt/login');
```
GawainLynch: Is switching to BK killing flashes?

This, and a few of the other checks for flashes, were to catch a recurring problem where a change in the code base "somewhere" would stop some messages being displayed, or in some way interfered with, and get missed during regular "monkey tests".

app/web.php
GawainLynch: Ugh … and I'd had coffee by then too … nothing to see here!

GawainLynch:
Ladies and gentlemen, benevolent ones, princesses, and drop bears, of the class of 2016 … we have a green light!

tests/codeception/acceptance/100-BackendAdmin/BackendAdminCest.php
CarsonF: I don't think so....this is the only case that needed changing. I'm not sure I could reproduce the flash message running the test steps myself - so I changed it

PR #5838, opened 2016-09-30
Spanish translations

nestordedios:
First part of the Spanish translations for the contenttypes, info and messages done.


nestordedios:
¡De nada amigo!

PR #5853, opened 2016-10-03
Fix underscored field serialization

rossriley:
As mentioned in #5787 this is the other bugfix.


bobdenotter:
Happy days! Works like a charm! 

giphy 1

PR #5856, opened 2016-10-03
Less assumptions for configuration

CarsonF:
The idea here is to not assume Bolt is configuring the system, but that it is a part of something bigger. So instead of applying defaults from our config, we don't set them if they aren't provided.

These config values are:
- debug
- error_level
- timezone
- locale
- twig's strict_variables

Tweaked `dump()` (PHP function, not Twigs) so it won't dump anything if debug is false.

Changed most of our event listeners to check if they are set before subscribing them. This way they can be disabled by unsetting them.

If debug was disabled we had a hardcoded error_level we set: https://github.com/CarsonF/bolt/commit/d116895fdfcbd5ebde336e862be997a8e9507359. I made this configurable with a new config key, `production_error_level`. It's disabled by default though, since we want to assume the system is configured correctly.

---

I would also like to comment out `strict_variables` in config by default. Twig internally sets this based on app's debug flag: app debug = strict_variables. This is logical to me because you want to see these errors while you are deving/playing, but don't want production to see them where debug is disabled. Having it commented out by default let's it operate as an override if someone wants to force it one way or the other. 
Thoughts @bolt/backend-developers @bobdenotter?

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -283,7 +283,7 @@ debuglog:
 
 # Use strict variables. This will make Bolt complain if you use {{ foo }},
 # when foo doesn't exist.
-strict_variables: true
+#strict_variables: false
 
 # There are several options for giving editors more options to insert images,
 # video, etc in the WYSIWYG areas. But, as you give them more options, that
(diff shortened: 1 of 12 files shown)


GawainLynch:
> I would also like to comment out strict_variables in config by default.

:+1:

CarsonF:
Forgot to change defaults in `Config` to null; rebased those in.

GawainLynch:
Oh, forgot to add a note to others… I asked @CarsonF to send this to 3.2 as it cleans up more of our exception handling in areas I was already worried about … so for me this is bug-fix.

Happy to negotiate though … tomorrow :smile:

GawainLynch:
:traffic_light:

PR #5859, opened 2016-10-04
Fix memcache session handler closing connection

### LINKED ISSUE #5857: [BUG] Memcache Session close but never open
Hi,

When users try to login on the backoffice, they are redirect to /admin/login with the message  general.phrase.error-session-data-login

In the method handlePostLogin, we try to get the token from Memcache. 
SessionHandler return false because the connection to memcache was closed.

The connection to Memcache  is close by the SessionStorage when it's save  the session and never open it again. Informations retrieve from  session after this aren't correcte.

In fact the session is open but the code of the handler juste return true

`   public function open($savePath, $sessionName)
    {
        return true;
    }
`
A quick fix : Do we really need to close Memcache(d) connection ?

Details
-------

 - Relevant Bolt Version: [  3.1.2 ]
 - Install type:  Composer install multi site
 - PHP version: [ 5.6 ]
 - Used web server: [ Apache ]

Reproduction
------------
Configure Memcache in php.ini
​try to login.

CarsonF:
Fixes #5857

### DIFF
--- a/src/Provider/SessionServiceProvider.php
+++ b/src/Provider/SessionServiceProvider.php
@@ -5,6 +5,7 @@ namespace Bolt\Provider;
 use Bolt\Session\Generator\RandomGenerator;
 use Bolt\Session\Handler\FileHandler;
 use Bolt\Session\Handler\FilesystemHandler;
+use Bolt\Session\Handler\MemcacheHandler;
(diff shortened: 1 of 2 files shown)


GawainLynch:
:+1:

PR #5860, opened 2016-10-04
Codeception fixes backport

GawainLynch:
This is a back port of the Codeception changes from #5836 

These changes do not affect production, only acceptance tests, with the exception of two small changes to _not_ fatal during debug/test if fields are not included in a POST.

…and one small assassination of a super global.

Obviously 99% of the work on this was @CarsonF … so props there :+1:

### DIFF
--- a/src/Storage/ContentRequest/Save.php
+++ b/src/Storage/ContentRequest/Save.php
@@ -168,6 +168,8 @@ class Save
             $content->setOwnerid($user['id']);
         }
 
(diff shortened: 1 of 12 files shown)


GawainLynch:
Oh, and for the record … the intent of the Codeception changes is to allow the correct use of `xdebug` (or `dump()` for that matter) in our Codeception tests when issues come up … and given that 3.2 is significant due to changes in the exception handling it makes sense.

PR #5864, opened 2016-10-04
Deprecate "app" and "view" filesystem

CarsonF:
I deprecated and removed usages of `app` and `view` in favor of a new `bolt` filesystem. This is only for our filesystem abstraction currently, not `ResourceManager::getPath`.

My reasoning is this: `app` and `view` always confused me as composer installation user. I would think those those would be _my_ app folder and _my_ view folder. Instead they are Bolt's core files.

The line was blurred in 2.x days where a lot of people used git installations (or zip installs which matched git). It was hard for users to understand that those files should not be modified as they will be replaced on update.

Now that all are using composer installs 🎉  (not including core development) that line is clear.
So `bolt` represents core files, of any kind, and is not user configurable.

---

@GawainLynch pointed out to me that seeing
```yaml
notfound_image: bolt://app/view/img/default_notfound.png
```
in `config.yml` could be confusing to new users.

I don't disagree, but I think the solution to this is just documentation. I want to move everything else over to our filesystem abstraction in 3.3 and completely deprecate `ResourceManager`. Then it would make a lot of sense to document a) `ResourceManager` -> it's replacement, and b) each filesystem and what it's purpose is.


src/Asset/File/Queue.php
```diff
@@ -51,6 +51,11 @@ public function __construct(Injector $injector, Packages $packages)
      */
     public function add(FileAssetInterface $asset)
     {
+        if (!$asset->getPackageName()) {
+            // Deprecated.
```
GawainLynch: Nit pick … I don't get many chances to make them in your direction :wink: 

Wouldn't a PHPDoc `@deprecated` be a better option here, it makes regex searching a bit easier during major version Rip & Replace sessions, skipping the need to check `git blame` & Co.

tests/phpunit/unit/Extensions/AssetsProviderTest.php
```diff
@@ -188,7 +188,7 @@ class AssetsProviderTest extends AbstractExtensionsUnitTest
     protected function getApp($boot = true)
     {
         $app = parent::getApp(false);
-        $mock = $this->getMock('\Bolt\Asset\BoltVersionStrategy', ['getVersion'], [$app['filesystem']->getFilesystem('extensions'), $app['asset.salt']]);
+        $mock = $this->getMock('\Bolt\Asset\BoltVersionStrategy', ['getVersion'], [$app['filesystem']->getFilesystem('extensions')->getDir(''), $app['asset.salt']]);
```
GawainLynch: Did you do a "Gawain" and forget that we're on PHP 5.5+ now :stuck_out_tongue_winking_eye: 

Zero stress & changes needed here btw… just woke up in a very good, and very cheeky, mood :wink:

app/config/config.yml.dist
```diff
@@ -204,8 +204,8 @@ thumbnails:
     default_image: [ 1000, 750 ]
     quality: 80
     cropping: crop
-    notfound_image: view://img/default_notfound.png
-    error_image: view://img/default_error.png
+    notfound_image: bolt://app/view/img/default_notfound.png
+    error_image: bolt://app/view/img/default_error.png
```
GawainLynch: Just flagging this for others, as per @CarsonF's  comment on the PR itself.

My concern with "public facing" use of `bolt://` is that to a large part of our target audience of implementers, it might appear that `bolt://` is the prefix for all Bolt files, rather than the root of the code.

I personally don't have a merge-blocking concern, but did want to flag it for those advocating for other target user groups … ping @bobdenotter

bobdenotter: If this stuff gets documented, I'm fine with it.

bobdenotter:
Merging in! 👍

PR #5870, opened 2016-10-05
Lock Twig at < 1.26 to preserve BC break on extension driven Twig fun…

### LINKED ISSUE #5868: [BUG] Since Twig 1.26, only the last extension using TwigTrait will get registered by Twig
[Since Twig deprecated the usage of getName() in Twig extensions](https://github.com/twigphp/Twig/commit/f58cce77d773a805e5e4d0ddf5b5e5b17b097d93), multiple Bolt extensions using `TwigTrait` keep overriding the previously loaded extension, resulting in only the last one having its functions, filters, etc. registered. This is because Twig now treats every instance of `Bolt\Twig\DynamicExtension` as the same extension.

Details
-------

 - Relevant Bolt Version: 3.1/all
 - PHP version: 5.6
 - Used web server: Nginx 1.11.3

GawainLynch:
Fixes #5868

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "symfony/web-profiler-bundle" : "^2.8",
         "symfony/yaml" : "^2.8",
         "tdammers/htmlmaid" : "~0.7",
-        "twig/twig" : "^1.18",
+        "twig/twig" : "<1.26",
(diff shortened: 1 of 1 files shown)


bobdenotter:
👍

PR #5883, opened 2016-10-08
Fix the info Popovers in Repeaters

### LINKED ISSUE #5783: [BUG] Info option not working for Repeater fields
Seen by @marcingajda in Bolt 3.0.11
Confirmed by me, still an issue in Bolt 3.1.2

Here's the sample repeater in my ContentType:
`        myreps:
            type: repeater
            fields:
                repeatingitem:
                    type: text
                    info: some info here
`
You will see the (i) icon next to the repeater field, but when you hover over it, it only displays a question mark as opposed as showing a popup with the info defined in the CT.

rossriley:
Popovers are in the legacy `init` javascript object so the init for popovers needs to be called manually whenever a repeater block is added.

Fixes #5783


bobdenotter:
Merging in! I'll rebuild the assets too.

PR #5884, opened 2016-10-08
Improve Look&Feel of the new Exception pages

bobdenotter:
Note: This PR also ensures that this info won't be shown to non-logged in users. 🕵️

See screenshot: 

screen shot 2016-10-08 at 14 30 17

screen shot 2016-10-08 at 14 30 28


GawainLynch:
At least now we can be satisfied that if we do end up with #1555 triggering the end of the known universe, we'll look damn fine!

PR #5890, opened 2016-10-10
fix context for isallowed permissions check

cosmod:
Fixes the display of the aside delete button on editcontent views for 'editor' user roles. `isAllowed` is checking for a slug, legacy Content, or a string.


bobdenotter:
Good catch. Thank you, @cosmod 👍

Issue #5891, opened 2016-10-10
Exception rendering exception template


CarsonF:
![](http://www.breaknenter.org/wp-content/uploads/2011/09/inception_movie_poster_small.png)

`exception` and `request` aren't always defined when rendering exception templates.
It can be triggered with a configuration error, such as, a templatefield without any fields defined.
Maybe an exception could be created for this case, @GawainLynch? It seems weird to me to render an exception view without an exception.

We probably should not use `app` either. Maybe `request` can be passed in from controller.

bobdenotter:
> We probably should not use app either. Maybe request can be passed in from controller.

I would think that if `{{ app.request }}` doesn't exist yet, that we also can't use `$this->app['requests]` in the controller, right before it's rendered, right?

CarsonF:
> I would think that if {{ app.request }} doesn't exist yet, that we also can't use $this->app['requests] in the controller, right before it's rendered, right?

I'm more ok with that. I think using `app` in Twig is less acceptable than in PHP.

I'm not sure when/how that method is being called though. Typically in a controller you just "ask" (pass in) request object, and it's given to you. But if we have to do `$app['request']` here, I'd be ok with that for now - It still abstracts it from Twig.

GawainLynch:
Yeah, this is kinda what I wanted to clean up too… I initially just extended exiting templates, 'cause "get it done".

Not convinced it is blocker, but more than happy to see it resolved before stable release … and 3.3/3.4 should put some nails in some coffins, but we have to get there (safely and stably) first.

CarsonF:
Right I was more so wondering if `{{ app.request }}` could even be called without giving an exception in all cases. Default filter and is defined checks can get this out the door, that's what is blocking in my mind.

GawainLynch:
> It can be triggered with a configuration error, such as, a templatefield without any fields defined.
Maybe an exception could be created for this case, @GawainLynch?

That was the plan, yes.

> We probably should not use app either. Maybe request can be passed in from controller.

:+1:

GawainLynch:
> Default filter and is defined checks can get this out the door, that's what is blocking in my mind.

:+1:

bobdenotter:
I've updated the PR, so it doesn't use `{{ app }}` anymore.

CarsonF:
If @GawainLynch merged it, I guess it's good.

PR #5892, opened 2016-10-10
Fixes and Tweaks to exception page.

bobdenotter:
- Add a 'Google this exception' button. 
 - Shows correct line numbers in snippet
 - Abbreviates the classname of the Exception that was thrown
 - Should fix the failure if `Request` doesn't exist (yet), as reported by @CarsonF (see #5891) 

screen shot 2016-10-10 at 20 57 34

### DIFF
--- a/app/view/twig/exception/exception.twig
+++ b/app/view/twig/exception/exception.twig
@@ -60,7 +60,7 @@
                             <p class='exception'>
                                 {# Based on this classic tweet: https://twitter.com/divineomega/status/695744177557106688 #}
                                 {% set query = 'Bolt ' ~ exception.classbase|default('unknown') ~ ' in ' ~ exception.filebasename|default('unknown') ~ ' line ' ~ exception.object.line|default('unknown') ~ ': ' ~ exception.object.message|default('unknown') %}
-                                <a class='btn btn-default' href='https://www.google.nl/search?q={{ query|url_encode }}' target='_blank'>Google this Exception</a>
+                                <a class='btn btn-default' href='https://www.google.com/search?q={{ query|url_encode }}' target='_blank'>Google this Exception</a>
                             </p>
 
                             {% block request %}
--- a/app/view/twig/exception/exception.twig
+++ b/app/view/twig/exception/exception.twig
@@ -44,8 +44,8 @@
                             {% endblock exception %}
 
                             <p class='exception'>
-                                <tt><abbr title="{{ exception.class|default('unknown') }}">{{ exception.classbase|default('unknown') }}</abbr></tt> in <tt>{% if debug %}<abbr title="{{ exception.filename|default('unknown') }}">{% endif %}
-                                {{- exception.filebasename|default('unknown') }}{% if debug %}</abbr>{% endif %}</tt> line <tt>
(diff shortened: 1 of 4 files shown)


GawainLynch:
Looks good …but need a fresh brain to merge … someone else can go for it, just need sleep before dong this myself :smile:

bobdenotter:
No rush.. Might need some tweaks after looking at #5891 :-)

app/view/twig/exception/exception.twig
```diff
@@ -43,30 +43,28 @@
...
-<pre class='line-numbers' data-start='{{ exception.object.line|default(1) }}' data-line='6'><code class='language-php'>
+<pre class='line-numbers' data-start='{{ max(exception.object.line - 5, 1) }}' data-line='6'><code class='language-php'>
 {{- exception.snippet -}}
 </code></pre>
                             {%- endif %}
 
+                            <p class='exception'>
+                                {# Based on this classic tweet: https://twitter.com/divineomega/status/695744177557106688 #}
+                                {% set query = 'Bolt ' ~ exception.classbase|default('unknown') ~ ' in ' ~ exception.filebasename|default('unknown') ~ ' line ' ~ exception.object.line|default('unknown') ~ ': ' ~ exception.object.message|default('unknown') %}
+                                <a class='btn btn-default' href='https://www.google.nl/search?q={{ query|url_encode }}' target='_blank'>Google this Exception</a>
```
SahAssar: Perhaps this should be google.com to be more locale-agnostic?

GawainLynch: Adding as a required change, 'cause playing with approval system. :smile:

bobdenotter: Ah, yes.. When poking around with how to add a query to the URL I started out with `google.com`, but it must've been "helpful" by changing it to NL. :-)

src/Controller/Exception.php
```diff
@@ -245,11 +246,21 @@ protected function getContextArray(\Exception $exception = null)
             $snippet = false;
         }
 
+        // We might or might not have $this->app['request'] yet, which is used in the
+        // template to show the request variables. Use it, or grab what we can get.
+        if ($this->app['request'] instanceof Request) {
+            $request = $this->app['request'];
```
bobdenotter: But what if `$app['request_stack']` doesn't exist _yet_? 

I believe that was the issue Carson was encountering, when this is triggered too early in the request cycle.

GawainLynch: If one doesn't, the other doesn't. The sole difference being that the request gets added to the stack during the request cycle…

The request stack is added in the `Application` [constructor](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Application.php#L130) so if we don't have it, we've crashed in bootstrap and … all … hope … is … lost! :laughing:

GawainLynch:
Something is kinda b0rked:
```
Twig_Error_Runtime in Template.php line 394:
An exception has been thrown during the rendering of a template ("Catchable Fatal Error: Object of class Bolt\Controller\Backend\General could not be converted to string") in "@bolt/exception/_request.twig" at line 45.
```

bobdenotter:
@GawainLynch I saw that one time as well, which is why I added this check: https://github.com/bolt/bolt/pull/5892/files#diff-c55ad1867f9650f74e2851fbaafcb896R42

Can you reproduce it?

PR #5897, opened 2016-10-11
Better title for exception page

bobdenotter:
As suggested by @CarsonF

### DIFF
--- a/app/view/twig/exception/exception.twig
+++ b/app/view/twig/exception/exception.twig
@@ -5,7 +5,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
I'm on the fence. Opinions, @GawainLynch ?

GawainLynch:
Actually … this might relate to a niggle I struck yesterday when I was trying tout the new [One True Exception Handler](https://twitter.com/divineomega/status/695744177557106688) and the query ended up being for "Bolt exception …"

Where my concern lays is that we're applying branding in the wrong way … most of the time Bolt is working correctly, but something else isn't … but we quite proudly "own" (brand) the exception and tell the world that it's ours, rather than us trying to provide a usable fall back for people who've hit a bug (often extensions), or misconfigured their system … "It's OK, we'll take the blame" style :laughing: 

Now above is a very long way of saying that IMHO our exception page should be less "branded", but still have the look, feel, and functional (where possible) that people love our product for.

bobdenotter:
Ok, changed. :-)

PR #5899, opened 2016-10-12
Fix twig in live editor

CarsonF:
- Don't parse Twig for live editor as it saves the compiled HTML.
- Set Twig syntax as protected source for CKEditor.
- Handle protected source in live editor.
- Fix JS error when switching CKEditor to source view. Credit to @dadaxr.


bobdenotter:
Looks good to me! :

PR #5902, opened 2016-10-13
Revert 94ecd7e as it hides twig in ckeditor view.

CarsonF:
This can still be opt in, but I don't feel like it should be default.

@bobdenotter - as we discussed.


GawainLynch:
c.f. #5899

Issue #5905, opened 2016-10-14
Update base-2016 theme to use `{{ asset() }}` instead of `{{ paths }}`


bobdenotter:

CarsonF:
Could probably just backport my commits on 3.3.

bobdenotter:
Fixed by #5906

PR #5906, opened 2016-10-15
Replace paths string concat with path() route and asset() generation.

GawainLynch:
This is a cherry-pick of @CarsonF's #5836 clean up on `base-2016` for 3.x

I've put the branch on the  org repo, so "go for it" wrt to tweaks as needed, I'll merge branches again on Monday before we tag 3.2-RC


bobdenotter:
Fixes #5905

PR #5912, opened 2016-10-16
Revert hidden file changes

### LINKED ISSUE #5894: Both .bolt.yml and bolt.yml installed in a Composer create-project install
Speaking of testing :wink:

### LINKED ISSUE #5909: [RFC][Urgent] Handling .bolt.yml to bolt.yml migration in 3.2 — Fixing #5894
Ping @bobdenotter @CarsonF @PhillippOhlandt @rossriley @SahAssar et al.

Realistically, this **needs to be addressed today** as we're plan to tag RC tomorrow and is a simple coding problem to fix, we just need to agree on the approach.

### Problem

`bolt/composer-install` and many/most/all existing 3.x installs have the hidden file. The changes in core are currently half-baked, and non-archive installs will end up with both a hidden an non-hidden file … the hidden file has, and should have, usage priority, and therefore create yet one more thing that will needlessly trip up a significant percentage of our implementer base.

In the case of the Composer target, it should stay hidden, as IMHO this is the _correct approach_. Making the non-hidden file option available is part of our ongoing effort to bridge the gap between novice and experienced implementers.

We're about to open up a world of embarrassing pain for all of our implementers, in what we believe is going to be the best 3.x minor release branch yet. Let's not fall on our faces here :wink: 

### Proposal

Modify the `ScriptHandler` `post-create-project-cmd` callback as follows:

  *  Add a question that has the default as the novice friendly option:

>  "Would you like the load configuration file (bolt.yml) to be hidden? (no): "

  * If the non-hidden option is selected, **remove** the `.bolt.yml` file

The problem I foresee with Step 2 is that, you can re-run `composer create-project` in an existing project root and that will skip straight to executing the `post-create-project-cmd` callback … and it is easy to miss the questions, as wall-o-text, and nuke your hidden file,

I really think we've completely taken the wrong approach to this problem so far, so can we all have a think about:

a) What problem are we actually trying to solve
b) The negative impacts that we've already created by doing a half arsed job of approaching wrt to this a)
c) How we can achieve a unified agreement on the end goal to providing a solution to a)
d) What is the solution we're going to effect :laughing: 

… aaaaand … Go! :vertical_traffic_light: 

---------------------

Linking #5894 and https://github.com/bolt/composer-install/pull/16

GawainLynch:
As discussed in Slack #development today.

Fixes #5894 
Closes #5909
Closes bolt/composer-install#16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
   fast_finish: true
   allow_failures:
     - php: hhvm
+    - php: 7.1
 
 services:
   - mysql
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -84,13 +84,8 @@ return call_user_func(function () {
     if (file_exists($rootPath . '/.bolt.yml')) {
         $yaml = Yaml::parse(file_get_contents($rootPath . '/.bolt.yml')) ?: [];
         $config = array_replace_recursive($config, $yaml);
-    } elseif (file_exists($rootPath . '/bolt.yml')) {
-        $yaml = Yaml::parse(file_get_contents($rootPath . '/bolt.yml')) ?: [];
-        $config = array_replace_recursive($config, $yaml);
     } elseif (file_exists($rootPath . '/.bolt.php')) {
         $php = include $rootPath . '/.bolt.php';
-    } elseif (file_exists($rootPath . '/bolt.php')) {
-        $php = include $rootPath . '/bolt.php';
     }
 
     // An extra handler if a PHP bootstrap is provided, allow the bootstrap file to return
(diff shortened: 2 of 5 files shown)


CarsonF:
Don't merge this yet. See what I have to say in development channel.

GawainLynch:
Tests are assassinated in current state anyhoo :laughing:

CarsonF:
Ok nvm. As you were.

bobdenotter:
There's a typo in the error. Suggestion: 

```
If you have uploaded this install via a file manager or FTP, please check that "show hidden files" files is turned on. After doing so, you will be able to see this file and you can upload it to the root of your Bolt installation.
```

GawainLynch:
@CarsonF **After** you have had :coffee: can you sanity check this, it includes the default path updates for `Bolt\Configuration\Composer` (ended up a dependency to get this over the line)

bobdenotter:
Whoop! :shipit:

PR #5914, opened 2016-10-17
[WIP] Refactor of ContentType and Field Mapping

rossriley:
Mainly pinging @CarsonF and @GawainLynch here for some early feedback on approach.

The goal of this refactor is to make ContentTypes and their fields proper objects, rather than nested arrays.

So far there is a ContentType class which represents a single contenttype definition. It exposes a getter for each property.

Then each field is also a field definition class, in some cases this is standard, but there are additional field classes to expose field specific functionality, eg the `slug` field expoes a `getUses` method which is not available on other fields.

Each field, and indeed the outer ContentType class manages its own initialisation and validation so this is all moved out of the main Config class.

Both continue to implement ArrayAccess so most of the old functionality should work, eg: `$contenttypes['pages']['fields']['title']['type']` should still get you the value `text`, although in the future, `$contenttypes->get('pages')->get('title')->getType()` will be the preferred way to navigate, once collection classes are in place.
### Still to do

Lots of doc blocks etc. 
Still need to add tests for the ContentType class
Want to turn the array of contenttypes and the array of fields into a Collection rather than array
Need to do a big scan of the codebase to make sure there aren't any 🐨 🐨 hiding from moving to this approach.


GawainLynch:
I'm going to go over this again with a fresh brain mate, but I think on the surface that our two branches are going to work well together.

What the loader will deliver to you (in current form) is a fully validated, and complete, array of values … so the mapping part of your branch here is key to the next step I was aiming to tackle … and you've saved me a *tonne* of work … team work FTW :+1: 

> Lots of doc blocks etc. 

:+1: 

> Still need to add tests for the ContentType class

:+1::+1::+1::+1:

> Want to turn the array of contenttypes and the array of fields into a Collection rather than array

Yes please!

> Need to do a big scan of the codebase to make sure there aren't any :koala: :koala: hiding from moving to this approach.

We really need a drop bear emoji on GitHub :laughing:

CarsonF:
I think I had a bit of a different direction in mind. Instead of making Field / ContentType classes array wrappers they would be value objects. For example, instead of having a generic `$config` property, it would have properties for each...property. i.e. `$type`, `$group`, `$label`, etc. This allows the current state of the object to be more clear and consistent when debugging.
The value objects would also be serializable, so you don't have to go through normalization for every request.

The other thing I want is the business logic in hydrating the Fields / ContentType separated from them. For two reasons, service injection and extendability. The `Slugify` service is an easy example of this. Currently we just create the default service in the object so it can't be customized at all. Secondly, there's currently no way for this logic to be changed without extending the fields. I see this is kind of done with `ContentType::setupFields` which has the `MappingManager` passed in, but this is not the same thing. We are unnecessarily limiting functionality to that object when creating each field. 

There's a number of resources on why factories with business logic should be split from value objects, I'll see if I can find some.

rossriley:
That's fine, I can move them out of the array to be first class properties, to be honest that was just just done for ease so that array access would still work but I can use a property_exists check instead.

$64,000 dollar question is how do we allow service injection without any autowiring here, because for instance the `Slug` definition needs `$app['slugify']` but it can't have it until the `initialize` step, so constructor injection won't work. We can of course provide a `setSlugify()` method but because we want to open out these `Definition` classes to users and extensions then we'd have to create a `SlugifyAwareInterface` but do that for each service we support.

Would we be better imitating the ServiceProvider strategy and implementing a `connect(Bolt\Application $app)` method that gets called on initialize and then it's up to the definition class itself to get the services, eg for Slug:

```
    public function connect(Bolt\Application $app) 
    {
        $this->setSlugify($app['slugify']);
    }
```

Let me know what you think and I'll get it adjusted.

CarsonF:
I'm working on Config DI right now and I'm almost done. I would just assume config will work like any other service with DI and constructor injection.

rossriley:
Ok, so we write the constructor like this...

```
 public function __construct($name, array $parameters, Slugify $slugify)
    {
        $this->name = $name;
        $this->parameters = $parameters;
        $this->slugify = $slugify;
    }
```

and the container will resolve the correct implementation of Slugify?

CarsonF:
Ah _autowiring_, no. You would have to explicitly define that with the factory service.

```php
$app['storage.mapping.content_type_factory'] = $app->share(function ($app) {
   return new ContentTypeFactory($app['slugify']);
});
```

rossriley:
Yes, but that's the problem, how does an extension writer who wants to make a new Field Definition that needs the `$app['slugify']` or any other service, maybe they want to use  `$app['schema']` etc. If we pass the services into a factory then all field definitions can only consume the services that we pass in.

CarsonF:
Well right exactly. If someone wants to do custom logic they need to extend and replace the factory.

CarsonF:
Maybe there's not an actual class for the factory just a DI key. That would make it easier to replace/add/modify the factory for individual fields.

```php
$app['storage.mapping_factory.image'] = $app->protect(function($arrayDefiniton) use ($app) {
  // do logic
  // return Image Definition object
});
```
So we would look for `"storage.mapping_factory." . $type` key and call that. 

I do the same thing for Session handler creation. A main one [`session.handler_factory`](https://github.com/bolt/bolt/blob/master/src/Provider/SessionServiceProvider.php#L194-L202) that calls individual factories ([filesystem](https://github.com/bolt/bolt/blob/master/src/Provider/SessionServiceProvider.php#L194-L202), [memcache](https://github.com/bolt/bolt/blob/master/src/Provider/SessionServiceProvider.php#L274-L293), etc.).

rossriley:
What about what we do here: https://github.com/bolt/bolt/blob/master/src/Provider/DatabaseSchemaServiceProvider.php#L81-L89 and allow an array of definition handlers to be defined under `$app['mapping.definitions']` that way we still get an iterable from there so users can add as many extras (or replacements) as they want, so adding a new one is almost the same as your suggestion above, except it writes to a sub-key:

```
$app['mapping.definitions']['slug'] = $app->protect(function($arrayDefiniton) use ($app) {
    return new CustomSlugDefinition('slug', $arrayDefinition, $app['slugify'], $app['any.other.service']);
});
``

**Comment by CarsonF** [comment]
Yeah I think I would be ok with that. My only concern is adding complexity to the syntax. Is there a reason we need to iterate?

**Comment by CarsonF** [comment]
I think I actually like that better. Maybe something like this?
```php
$app['mapping.definition_factories'] = $app->share(function ($app) {
    $factories = new Pimple();

    $factories['base'] = $app->protect(function ($data, $obj = null) use ($factories) {
        $obj = $obj ?: new Base();
        $obj->setLabel($data['label']);

        return $obj;
    });

    $factories['slug'] = $app->protect(function ($data, $obj = null) use ($factories) {
        $obj = $obj ?: new Slug();
        $factories['base']($data, $obj);

        $obj->setUses($data['uses']);

        return $obj;
    });

    return $factories;
});
```

rossriley:
Cool, OK, watch this space and I'll start refactoring to this.

tests/phpunit/unit/BoltUnitTest.php
```diff
@@ -63,11 +63,11 @@ protected function getApp($boot = true)
             $this->app->initialize();
 
             $verifier = new Config\Validation\Validator($this->app['controller.exception'], $this->app['config'], $this->app['resources']);
-            $verifier->checks();
 
             if ($boot) {
                 $this->app->boot();
             }
+            $verifier->checks();
```
GawainLynch: I'm still laughing at my own stupidity over that one … thanks for catching it :smile:

src/Storage/Mapping/MappingManager.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        if (!array_key_exists('type', $parameters) || empty($parameters['type'])) {
+            $error = sprintf('Field "%s" has no "type" set.', $name);
+
+            throw new InvalidArgumentException($error);
+        }
+
+        $handler = $this->getHandler($parameters['type']);
+
+        if (isset($this->factories[$handler])) {
```
GawainLynch: Is this condition needed for something?

SahAssar:
Asking since I'm trying to do one of my cleanups again... Is this still going?

rossriley:
Yes, bumped to 3.4 now but set to resume soon.

PR #5916, opened 2016-10-17
We don't have 'legacy.php' anymore, so no use in keeping __FILE__ around.

bobdenotter:


bobdenotter:
If this gets merged, I'll udate https://github.com/bolt/composer-install/edit/master/public/index.php accordingly. :-)

GawainLynch:
@bobdenotter About to Green Button this, give me a few to double check that there are not more :koala: :koala: hiding in the trees and we should be GTG for the final testing for RC.

Issue #5900, opened 2016-10-13
[BUG] Content->getImage() method breaks for images with alt and no path


ToBe998:
The function `Content->getImage()` takes the first field of type *image* and the returns either the image path if we have a *file* key or the whole thing if not (as fallback for bolt 1.5.0).

If a user now edits the alt title of an image but does not provide a image path, this check fails and returns the array [alt: „the alt“]. 
Since every part of bolt (including especially the twig template of the backend) assume it’s a string and not an array, this throws an exception, making some views of the backend unusable. 

**The method in question** - ContentValuesTrait.php Line 503
```
        // Grab the first field of type 'image', and return that.
        foreach ($this->contenttype['fields'] as $key => $field) {
            if ($field['type'] == 'image') {
                // After v1.5.1 we store image data as an array
                if (is_array($this->values[$key]) && isset($this->values[$key]['file'])) {
                    return $this->values[$key]['file'];
                }

                return $this->values[$key];
            }
        }
```

Example of a breaking template with a "Array to String Conversion":
*_base/_listing.twig*
```
    {% set thumb_title = __('Image') ~ ': ' ~ content.getImage %}
```

In my case a widget that uses the @bolt/_base/listing.twig block creating the thumbnails breaks, making the dashboard fail critically (Remember: we can't catch exceptions in twig).

**Proposed Fix** - ContentValuesTrait.php Line 503
```
        // Grab the first field of type 'image', and return that.
        foreach ($this->contenttype['fields'] as $key => $field) {
            if ($field['type'] == 'image') {
                // After v1.5.1 we store image data as an array
                if (is_array($this->values[$key])) {
                    return isset($this->values[$key]['file']) ? $this->values[$key]['file'] : "";
                }

                return $this->values[$key];
            }
        }
```

Details
-------

 - Relevant Bolt Version: 3.0.12
 - Install type: Composer install
 - PHP version: 5.5
 - Used web server: Apache
 - For UX/UI issues: Chrome

Reproduction
------------
- Create a content type with a field of type *image* with alt attribute
- Enter an alt value and save
- Use the $record->getImage() method somewhere and observe returned file path

GawainLynch:
How tired am I … I think I just spent 10 minutes staring at the code trying to figure it out too!

But yeah, there was an intention to fix the getter for 3.0 … buuuuuuuuuut. :wink: 

Want to PR it to get the glory?

ToBe998:
[...]

PR #5919, opened 2016-10-18
fix for images with alt attribute but no path - see #5900

ToBe998:
As discussed here: https://github.com/bolt/bolt/issues/5900


GawainLynch:
Merged via 04e2a6452cc170606b6a787dcfbdc1eb31618327

Thank you kindly, @ToBe998

PR #5920, opened 2016-10-18
Set LegacyService for all Repo extending ContentRepository

dadaxr:
Fix a bug introduced with : [PR 5862](https://github.com/bolt/bolt/pull/5862)

All the Repositories extending the ContentRepository must have (for now) a "LegacyService" configured to handle correctly "Content Entities" ( content entities defined with the contenttype.yml file )

The previous PR was not always  setting that legacy service and it could (not always) end up with a bolt exception page when editing a content with the default backend UI.

This PR just ensure that the legacy service is set when retrieving a repository.

### DIFF
--- a/src/Storage/EntityManager.php
+++ b/src/Storage/EntityManager.php
@@ -8,6 +8,7 @@ use Bolt\Legacy\Storage;
 use Bolt\Storage\Collection\CollectionManager;
 use Bolt\Storage\Mapping\ClassMetadata;
 use Bolt\Storage\Mapping\MetadataDriver;
+use Bolt\Storage\Repository\ContentRepository;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata as ClassMetadataInterface;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
@@ -50,6 +51,8 @@ class EntityManager implements EntityManagerInterface
     protected $legacyStorage;
     /** @var Callable */
     protected $defaultRepositoryFactory;
+    /** @var  ContentLegacyService */
+    protected $legacyService;
 
     /**
      * Creates a new EntityManager that operates on the given database connection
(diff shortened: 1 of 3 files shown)


src/Storage/EntityManager.php
```diff
@@ -262,42 +268,54 @@ public function getRepository($className)
...
+
         /*
          * The metadata driver can also attempt to resolve an alias for us.
          * For now we are hardcoding the link between a content entity and
          * the content repository, but in time this should be a metadata level
          * configuration.
          */
-        if ($this->getMapper()->resolveClassName($className) === 'Bolt\Storage\Entity\Content') {
-            return $this->getDefaultRepositoryFactory($classMetadata);
+        if (is_null($repo) && $this->getMapper()->resolveClassName($className) === 'Bolt\Storage\Entity\Content') {
```
GawainLynch: Small style consistency request… can you use comparison rather than functions:
```
if ($repo === null && $this->getMapper()->resolveClassName($className) === 'Bolt\Storage\Entity\Content') {
```

We're pretty much locked in on PSR-2 + [Symfony Coding Standards](http://symfony.com/doc/current/contributing/code/standards.html) … the main difference is that we still use spaces around concatenation operators.

src/Storage/EntityManager.php
```diff
@@ -262,42 +268,54 @@ public function getRepository($className)
...
+        if (is_null($repo) && in_array($className, $this->getMapper()->getUnmapped())) {
+            $repo = $this->getDefaultRepositoryFactory($classMetadata);
+        }
+
+        if (is_null($repo)) {
+            $repo = new Repository($this, $classMetadata);
         }
 
-        return new Repository($this, $classMetadata);
+        if (is_a($repo, Repository\ContentRepository::class)) {
```
GawainLynch: s/is_a/instanceof/

GawainLynch:
Might also want to look at Travis :wink:

GawainLynch:
Test failure is a known problem in 3.3+ 

@rossriley LGTM, same for you?

PR #5923, opened 2016-10-19
Fix edge case bugs with routing.yml

CarsonF:
- Fix middleware array definitions and strings with `classname::method` and `service:method`.
- Fix default middleware definition, `::before`, with controller that is class name string with an `__invoke` method.


bobdenotter:
👍

PR #5925, opened 2016-10-19
Show custom fields in "Relationships" tab

Techworker:
This fixes a problem where the custom fields cannot be displayed when setting the group of a field to `relationships`.

``` yaml
my:
    fields:
        foobar:
            label: Foo Bar
            type: checkbox
            group: relations
```


GawainLynch:
Nice catch, thank you!

PR #5946, opened 2016-10-24
[3.2][RFC] Switch profiler target option to Symfony's target

GawainLynch:
The profiler page was a nice idea, but it remove the ability to just refresh the page and attempt next step debugging.

Also, for a number of us, having the stack trace at the bottom of the page means we have to modify the bootstrap to disable Bolt's exception handler as it is just too much scrolling every few page refresh and not suitable for the BEs.

So this give the "pretty" crowd what they needs, and the rest of us can just fallback to Symfony.

Also, I don't want to go into RC and lock in BC on this.

Ping @CarsonF

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -1,6 +1,7 @@
 <?php
 namespace Bolt\EventListener;
 
+use Bolt\Config;
 use Bolt\Controller;
 use Bolt\Exception\BootException;
 use Bolt\Request\ProfilerAwareTrait;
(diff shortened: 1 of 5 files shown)


bobdenotter:
I'm OK with this. 
Perhaps we should even sneak in that we change the order of "backtrace" and "request variables listing"?

CarsonF:
I definitely think this is a step in the right direction. The only niggle for me is the config option name. It feels like a double negative. "use symfony" when in code it's more like "disable bolts" since symfony is default. Maybe a "bolt error handler" key could be used instead and have it enabled by default?

To Bob's point there are two reasons I disable the error handler. 
 - The stacktrace is too far to scroll too
 - The security risk for production (not debug mode). 

I would love to see both of those changes make it into 3.2

bobdenotter:
> The stacktrace is too far to scroll too

IMHO this is strictly personal, but changing the order is fine with me.

> The security risk for production (not debug mode).

Didn't we decide that we were not going to do that in 3.2 but shortly thereafter?

CarsonF:
> IMHO this is strictly personal, but changing the order is fine with me.

That's why I was thinking we could have collapsing sections or tabs.

> Didn't we decide that we were not going to do that in 3.2 but shortly thereafter?

It's not my call, but it would be releasing a version with a known security risk.

src/EventListener/ExceptionListener.php
```diff
@@ -97,6 +102,8 @@ public function onKernelException(GetResponseForExceptionEvent $event)
                     'message'   => $message,
                 ]
             );
+        } elseif ($this->config->get('general/debug_error_use_symfony')) {
```
CarsonF: Can you pass in this option instead of the entire class?

bobdenotter:
Mergin in, as discussed on Slack. Minor niggles will be addressed in follow-up PR's this week.

PR #5950, opened 2016-10-24
Remove local extension handling

GawainLynch:
This has become a hot spot and novice users are not trying to use them

Test breakage I'll fix prior to merge … enough is enough


rossriley:
I'm a big 👍 for this, my gut feeling is that the choice for an app developer should be:

1. Is my extension specific to this application - Then include your extension as part of your app source code, autoload it using your root `composer.json`

2. Do I want to re-use this extension but not publish to the marketplace? - Then add a local or private repository to `extensions/composer.json` and include the package explicitly in the same file (we explicitly don't overwrite these settings in Bolt for this purpose)

3. Do I want to re-use and share? - Then publish to the marketplace.

I don't think there's a need to support any other scenarios, the only thing you lose in `1` is the auto asset install, but then arguably since it's your root project this should be the responsibility of your root composer.json anyway and I have a `post-update-cmd` script that handles it.

GawainLynch:
@rossriley local extensions now do do auto-installs of assets … 'cause paths.

That aside, this PR was aimed at making a point to some people that want "magic" and the for you, Carson or I to rewrite Composer.

For everyone else. Local extensions were a **hack** we added in v2 to give a migration path for certain  workflows, we even tried to make that easier in v3, including autoloading and a bunch of other goodies.

Nonetheless, it has always intended to be an migration path, and past that a hack that experienced developers can use. The reason being is that we have to do a lot do make it work, it break the Composer interface principles, and the list goes on with the amount of unsupported stuff we have to do to make this happen.

I am going to move this to `release/3.3` and fix tests, but currently there looks to be a 50% agreement in the core team about merging this, and I am going to suggest we talk though it further, but I am pushing this **hard** for 3.3.

GawainLynch:
Tests passing now :smile:

GawainLynch:
Ongoing discussion on this in meeting minutes https://github.com/bolt/bolt/wiki/Dev-meeting-2016-10-25

PR #5956, opened 2016-10-26
Standardise the Doctrine Types used by custom fields

rossriley:
I'm guessing this bug was us depending on unsupported behaviour, but in recent DBAL versions the  type is no longer standardised to the `getName()` value of the Type.

This was manifesting itself in custom fields where a `getStorageType()` is specified.

Hopefully fixes the issue at: https://github.com/nbehier/bolt-field-geojson/issues/9


GawainLynch:
LGTM

PR #5958, opened 2016-10-26
Handle user entity pre-save events very early to mitigate passwords not being hashed if another event stop propagation

### LINKED ISSUE #5947: [BUG]  Unsupported password hash when trying to create the first user, bolt 3.1.5 (and also 3.2
I created a new bolt project with create-project, changed the db name (MySQL) to an already existing one, to use the DB from an earlier version 2 install. And everything was fine.
Destroyed the db to start afresh but then I got the error while attempting to create the first user:

> "Unsupported Password Hash Supplied"

When I look in the database, the user is created but the password is stored in PLAINTEXT. @bobdenotter can verify to that since he saw.

See the image for a full stack trace.

Update to 3.2@beta with composer, threw away config.yml and composer and app.caches but the problem persisted.
Tried to reproduce the error in a clean checkout in another directory and I could not, everything is hunky dory there.
I have no idea what's going on and am posting this on Bobs request. Not sure if it really is an error, or just me, as per ususal [;-)]

screen shot 2016-10-24 at 13 24 35
https://guides.github.com/features/mastering-markdown
screen shot 2016-10-24 at 13 48 57

Details
-------

 - Relevant Bolt Version: [ 3.1.5 | 3.2.0-beta7 | master ]
 - PHP version: [ 5.5.9 ]
 - Used web server: [ Nginx ]

GawainLynch:
Handle user entity pre-save events very early to mitigate passwords not being hashed if another event stop propagation

Fixes #5947

### DIFF
--- a/src/EventListener/StorageEventListener.php
+++ b/src/EventListener/StorageEventListener.php
@@ -14,6 +14,7 @@ use Bolt\Storage\EventProcessor;
 use Bolt\Translation\Translator as Trans;
 use PasswordLib\Password\Factory as PasswordFactory;
 use PasswordLib\Password\Implementation as Password;
+use Silex\Application;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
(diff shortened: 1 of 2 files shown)


bobdenotter:
Merging in!

PR #5962, opened 2016-10-26
If `name:` or `singular_name:` isn't set in 'contenttype.yml', generate something semi logical from the slug.

### LINKED ISSUE #5961: In contenttypes the 'name' doesn't get generated from slug anymore.
Details
-------

 - Relevant Bolt Version: [ 3.1 | 3.2 | master ]

Reproduction
------------

Add a contenttype like this: 

```
blocks:
    slug: blocks
    singular_name: Block
    tablename: blocks
    fields:
        title:
            type: text
            class: large
            group: "Block"
        slug:
            type: slug
            uses: [ title ]
        content:
            type: html
            height: 150px
```

Note the `slug:` instead of the `name:`. 

Go to the bolt backend, get an error: 

screen shot 2016-10-26 at 16 05 16

bobdenotter:
Fixes #5961 

Note: We don't have to check for the case where _both_ `name` and `slug` aren't set. That's already handled here: 

https://github.com/bolt/bolt/blob/hotfix/set-name-from-slug-if-not-defined/src/Config.php#L545-L554

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -562,8 +562,14 @@ class Config
         if (!isset($contentType['slug'])) {
             $contentType['slug'] = Slugify::create()->slugify($contentType['name']);
         }
+        if (!isset($contentType['name'])) {
+            $contentType['name'] = ucwords(preg_replace('/[^a-z0-9]/', ' ', $contentType['slug']));
+        }
         if (!isset($contentType['singular_slug'])) {
             $contentType['singular_slug'] = Slugify::create()->slugify($contentType['singular_name']);
+        }
+        if (!isset($contentType['singular_name'])) {
+            $contentType['singular_name'] = ucwords(preg_replace('/[^a-z0-9]/i', ' ', $contentType['singular_slug']));
         }
         if (!isset($contentType['show_on_dashboard'])) {
             $contentType['show_on_dashboard'] = true;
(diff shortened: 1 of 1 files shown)


GawainLynch:
Why do we need this all of a sudden? My concern here is the underscore debacle.

bobdenotter:
This is just a bug I discovered, when testing the RC. It doesn't have to do with the underscore thing, right?

GawainLynch:
No, I was referring to fixing in on spot and breaking in another

bobdenotter:
I'm not sure I follow. This is "polyfilling" a missing _name_, if the _slug_ is already present. 

It has always worked like this: You could either set both the `name` and `slug` explicitly, or if one of them was missing, the system would assume one. 

If this PR would 'guess' a _slug_ I could see how that might be tricky, but the _name_ can contain spaces, underscores and other characters, and is used for display purposes only.

GawainLynch:
Yes, hence the question I asked:

> Why do we need this all of a sudden? 

i.e. has it always been a bug? If no, why now?

bobdenotter:
It has been a bug since 3.0 at least, looking back quickly.. Even the hints in the backend suggest that it worked like this PR fixes: 

> In contenttype <code>%s</code>, neither 'name' nor 'slug' is set. Please edit <code>contenttypes.yml</code>, and correct this.

If 'name' was required and 'slug' was the only optional one, that message would be different.

GawainLynch:
OK cool … that's more than enough for me … I was just confused as to why it became a problem now, and that it wasn't going to be a source of pain in 3.2

PR #5969, opened 2016-10-28
Fix Issues Caused By Slug/Key inconsistencies

### LINKED ISSUE #5965: [BUG] Multi select field doesn't save values.
Multi select fields (those tag-like ones) don't keep values. After saving and refreshing the editor the field is empty. 

Details
-------

 - Relevant Bolt Version: 3.2.0 RC1
 - PHP version: 7.0
 - Used web server: Nginx

Reproduction
------------

 1. **Bug summary**: 
    * Saving multi select field doesn't work
    * After save and editor refresh the field is empty
 2. **Specifics**:
    * It happens in the record editor
    * I'm using Bolt 3.2.0 RC1 installed with composer
    * I'm using Nginx, PHP 7.0 and MySQL database
    * Tested in Chrome beta
 3. **Steps to reproduce**:
    * Create a content type with multi select field which use records of another type    
    * Add new record of this type
    * Choose some values in this field
    * Save the record
    * Refresh the page
 4. **Expected result**: 
    * Set values should be shown in the editor after refreshing
 5. **Actual result**: 
    * The multi select field is empty

```
presentation_pl:
    name: Presentations (PL)
    singular_name: Presentation (PL)

    slug: presentations_pl
    singular_slug: presentation_pl
    fields:
        <<: *presentation_fields
        speakers:
            type: select
            multiple: true
            values: speaker_pl/anchor
```

rossriley:
This is a problem that was papered over by the fact we always used to write the slug back to the contenttype key. This behaviour was removed in #5939 and so this nasty little bug revealed itself.

We were using the table alias, rather than the original contenttype name to load in the contenttype structure to the entity, so in cases where they differ this was incorrect.

The main difference here is that we consult the original `contenttypes.yml` data to find the key for a given table alias, no longer assuming they are the same.

Fixes #5965

### DIFF
--- a/src/Configuration/ConfigurationValueProxy.php
+++ b/src/Configuration/ConfigurationValueProxy.php
@@ -52,6 +52,18 @@ class ConfigurationValueProxy implements ArrayAccess, EventSubscriberInterface
         ];
     }
 
+
+    /**
+     * Get the data of the loaded config
+     * @return array
+     */
+    public function getData()
+    {
+        $this->initialize();
+
+        return $this->data;
+    }
+
     /**
      *{@inheritdoc}
      */
--- a/src/Storage/Mapping/MetadataDriver.php
+++ b/src/Storage/Mapping/MetadataDriver.php
@@ -172,7 +172,7 @@ class MetadataDriver implements MappingDriver
             $this->unmapped[] = $tblName;
         }
 
-        $contentKey = $table->getOption('alias');
+        $contentKey = $this->getContenttypeFromAlias($table->getOption('alias'));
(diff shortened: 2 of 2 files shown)


GawainLynch:
I don't think I have any objection to this … certainly nothing jumps out except …

Between underscores, hyphens, booleans, slugs, and :snail:, we've managed to recently sprinkle a lot of normalisation logic over the place … so it would be good to get something mapped out, as when you & I get our config branches in, the normalisation should be more centralised … before untangling it becomes a project in, and of, itself.

bobdenotter:
Merging in! :-)

Issue #5963, opened 2016-10-26
File uploads no error when file_uploads ini setting is disabled


CarsonF:
We should definitely check for this ini setting an render an error if `file_uploads` ini setting is disabled.

GawainLynch:
Fixed in #5973

PR #5973, opened 2016-10-30
Check file_uploads ini directive for allowed uploads

ozdemirburak:
Hoping that it fixes the issue #5963. 

There may be one possible issue with testing where the user is not allowed to change the `file_uploads` ini directive. 

That's why added the code block below to the tests.

``` php
if (ini_set('file_uploads', '0') !== false) {
    ...
}
```

### DIFF
--- a/src/Exception/FilesystemException.php
+++ b/src/Exception/FilesystemException.php
@@ -7,10 +7,9 @@ namespace Bolt\Exception;
  */
 class FilesystemException extends \Exception
 {
-    const FILE_NOT_READABLE            = 1;
-    const FILE_NOT_WRITEABLE           = 2;
-    const FILE_NOT_REMOVEABLE          = 4;
-    const FILE_UPLOADS_ARE_NOT_ALLOWED = 5;
+    const FILE_NOT_READABLE   = 1;
+    const FILE_NOT_WRITEABLE  = 2;
+    const FILE_NOT_REMOVEABLE = 4;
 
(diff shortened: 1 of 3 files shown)


GawainLynch:
@CarsonF can you :+1: / :-1: … I know we're aiming to nuke the permission class, but short term this looks to get us something

bobdenotter:
Bumping @CarsonF 👊

GawainLynch:
@bobdenotter just flagging, if you're near the green button, I want to send this to a lower version branch than `master`

ozdemirburak:
@CarsonF updated, couldn't find any specific exception but IOException.

CarsonF:
@GawainLynch What do you think? I'm thinking a subclass should be created for this. It doesn't really feel like an IO exception. Although the subclass could extend IO.

GawainLynch:
> I'm thinking a subclass should be created for this. It doesn't really feel like an IO exception. Although the subclass could extend IO.

Well right this second it is something of a moot point, as we're not catching the exception, and providing feedback. This is into the editor zone of UX after all.

But yeah, a `ThisIsNotTheIniSettingYouAreLookingForException()` might be a bit more useful in catching and displaying something in a clean way. The question becomes, what?

GawainLynch:
Merged into `release/3.3` via 83fa3db98beac6e2a1ef43c700eb297323b975f6

Thanks for your patience @ozdemirburak :+1:

Issue #5968, opened 2016-10-28
[BUG] ContentRepeater order gets mixed up when using postgres


5in4:
I am using a contentrepeater in a contenttype. If more than 10 elements are created, the 11th is displayed after the first. This only happens when using postgres.
## Details
- Relevant Bolt Version: 3.2.0 RC1
- Install type: Zip/tarball install
- PHP version: 7.0.10
- Used web server: Apache/2.2.31
- postgres: postgres (PostgreSQL) 9.5.4
## Reproduction
1. clean install from zip
2. use this contenttypes file [contenttypes.txt](https://github.com/bolt/bolt/files/557939/contenttypes.txt)
3. setup with postgres
4. insert a contenttype "case" with 11 repeater elements
5. navigate somewhere else and reopen the case. The order is mangled (same happens in frontend).

The data is saved correctly in the db, but something goes wrong when trying to display the fields.

bobdenotter:
I'm guessing the ordering is alphabetical, without having tried it myself..

```
0
1
10
2
3
4
5
6
7
8
9
```

GawainLynch:
Fixed in #5976

PR #5976, opened 2016-11-01
Order field values by grouping in postgres (#5968)

5in4:
Fixes: #5968

- The ordering was mangled, because record ids are casted to string.
- This should only affect postgres, because it uses string_agg instead of GROUP_CONCAT
- DISTINCT was removed, but I don't see why it was there anyways


GawainLynch:
Checks out as valid to me … @bobdenotter I am going to send this to `release/3.2`

GawainLynch:
Merged into `release/3.2` via b281b69ef33085e3394fb18fb6e81368f46d112b

PR #5983, opened 2016-11-02
Fixed: New records without a title set don't throw exceptions now.

### LINKED ISSUE #5982: [bug] Exception when saving a new record without title.
Details
-------

 - Relevant Bolt Version: [ 3.2 | master ]

Reproduction
------------

Have contenttype: 

```
blocks:
    slug: blocks
    singular_slug: block
    fields:
        title:
            type: text
            class: large
            group: "Block"
        slug:
            type: slug
            uses: [ title ]
        content:
            type: html
            height: 150px
```

Create "new block", click "save". See: 

screen shot 2016-11-02 at 17 07 40

Full stack trace: 

```
#1	Symfony\Component\Debug\ErrorHandler::handleError(8, "Undefined property: Bolt\Stor…", "/Users/bob/Sites/bolt-origin/ …", 325, [array])
[root]/src/Storage/Entity/Content.php # line 325
#2	Bolt\Storage\Entity\Content::getTitle() 
[root]/src/Storage/Entity/MagicAttributeTrait.php # line 22
#3	Bolt\Storage\Entity\Entity::__get("title") 
[root]/src/Storage/Entity/Content.php # line 325
#4	Bolt\Storage\Entity\Content::getTitle() 
[root]/src/Storage/Entity/Content.php # line 56
#5	Bolt\Storage\Entity\Content::get("title") 
[root]/src/Storage/Field/Type/TextType.php # line 21
#6	Bolt\Storage\Field\Type\TextType::persist(QuerySet, Content) 
[root]/src/Storage/Repository.php # line 304
#7	Bolt\Storage\Repository::persist(QuerySet, Content, [array]) 
[root]/src/Storage/Repository.php # line 382
#8	Bolt\Storage\Repository::insert(Content) 
[root]/src/Storage/Repository.php # line 359
#9	Bolt\Storage\Repository::save(Content) 
[root]/src/Storage/ContentRequest/Save.php # line 242
#10	Bolt\Storage\ContentRequest\Save::saveContentRecord(Content, [NULL], [array], [true], "", "new", "/bolt/overview/foo") 
[root]/src/Storage/ContentRequest/Save.php # line 120
#11	Bolt\Storage\ContentRequest\Save::action([array], [array], "", [true], "new", "/bolt/overview/foo") 
[root]/src/Controller/Backend/Records.php # line 71
#12	Bolt\Controller\Backend\Records::edit(Request, "blocks", "") 
#13	call_user_func_array([array], [array]) 
[root]/vendor/symfony/http-kernel/HttpKernel.php # line 144
#14	Symfony\Component\HttpKernel\HttpKernel::handleRaw(Request, 1) 
[root]/vendor/symfony/http-kernel/HttpKernel.php # line 64
#15	Symfony\Component\HttpKernel\HttpKernel::handle(Request, 1, [true]) 
[root]/vendor/silex/silex/src/Silex/Application.php # line 586
#16	Silex\Application::handle(Request) 
[root]/vendor/silex/silex/src/Silex/Application.php # line 563
#17	Silex\Application::run(Request) 
[root]/src/Application.php # line 93
#18	Bolt\Application::run() 
[root]/index.php # line 12
```

bobdenotter:
Fixes #5982

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -322,7 +322,13 @@ class Content extends Entity
 
         $fieldName = $this->getTitleColumnName($this->contenttype);
 
-        return $this->$fieldName;
+        try {
+            $title = $this->$fieldName;
+        } catch (\Exception $e) {
+            $title = null;
+        }
+
+        return $title;
     }
 
     public function getContenttype()


GawainLynch:
I've been thinking about this more (while having a very entertaining conversation about SEO "experts" on our Slack channel) …

One reason why this just hit the wrong places for me is that we've tried to get away from swallowing errors/exceptions needlessly. While having Bolt crash is a UX problem, so is just being silent about the problem (and this one will exists in anything after early 2.3-dev) and not letting someone know why their slug is "random" actually (IMHO) breaks UX as much as an exception screen does. It tells the end user _nothing_ about why something *they* did wrong is impacting them.

A good example of this is the early uptake in 3.3 & 3.3 so that a number of us can actually see that we've "messed up", rather than wondering why a page gets rendered, minus the "bit" we're expecting.

What also worries me now is that 3.3 & 3.4 is going to take apart the configuration, um, stuff … and in the process we've got to find 2 years worth of stuff we've scattered around the codebase to try to work around (and create new) problems … we're in for a hard release to make up for these hacks-on-hacks-on-hacks-on-hacks. I've written a number of these hacks too, so no fingers pointed, but can we think more about why/where problems are coming from and if we can address them closer/at the core prior to 3.3 & 3.4.

Respect :+1:

bobdenotter:
I'm not saying we should fix it _like this_. I just bumped into it today, and IMHO it's something we should fix for 3.2, right? 

I agree that a random slug for an end-user is less than optimal, but most of them _by far_ will not see an exception screen as a sign that _they_ did something wrong. They will see it as a failure of the system. Ideally we'd show a flash-warning in this case.

GawainLynch:
Yeah, I suck at explaining.

> I'm not saying we should fix it like this. I just bumped into it today, and IMHO it's something we should fix for 3.2, right?

No, it exists for 2 years … and doing it like this is just as bad if not worse.

> I agree that a random slug for an end-user is less than optimal, but most of them by far will not see an exception screen as a sign that they did something wrong. They will see it as a failure of the system. Ideally we'd show a flash-warning in this case.

OK, I really just at making my point … well at least one of them…

The current approach just fails silently … it is almost impossible to debug unless you're at the expert level, so you're hurting everyone from novice to expert with this approach.

If I must state this here, the problem is our configuration if ####ed and done all over the place … underscore/hyphen debacle as an example.

Some wise advice I was given today in terms of what I am trying to say:

> I'd like to see the root cause determined and tell the user why this happened

What we shouldn't do is swallow this. This approach isn't a fix, it is contribution to an unknown outcome.

GawainLynch:
> This approach isn't a fix, it is contribution to an unknown outcome.

The "this" in that statement I meant as "this PR" — see, I do suck at communication :smile:

bobdenotter:
Yes, and I agree completely, but currently in 3.2 we don't swallow it silently. It breaks in a way that many people will see as a system failure. We're making the same point here, I'm just approaching it from the other end. 

Anyhow, I'm taking suggestions on how to fix it properly. :-)

GawainLynch:
But this approach only works if you have debugging turned on, otherwise they are just errors

GawainLynch:
…and if there are other things causing exceptions, you lose that too 

… and yes, thinking the problem though, but it isn't easy to solve right now

bobdenotter:
> But this approach only works if you have debugging turned on, otherwise they are just errors 

True, so using this to set a flash message is no good, either. Regardless of how/if/why/when we fix it, that might be very useful. A flash message that says "Your new record has a weird slug, because you didn't fill out a title!" 

Problem is, that a _lot_ of editors have the following workflow, because this is what they were used to with word: 

 - Click "new page" 
 - Click "Save" 
 - Start typing a title
 - Save again. 
 - Rinse, repeat.

I'll also think if I can figure out a better approach for this.

GawainLynch:
Zero stress … I have pulled the custom error/exception handling from 3.2+ so errors will be as they were, silent.

PR #5993, opened 2016-11-05
Change order on exception page, and show page before dumps have loaded.

bobdenotter:
I've moved the trace nearer to the top of the page (by request). The actual `dump()`s of the arguments are loaded _after_ the rest of the page, and injected into the trace after loading. The result is a page, that appears on screen much sooner.


GawainLynch:
Sending in early … 'cause these templates are not under test, and it must be time for :wine_glass: … and Travis is being flaky anyway today.

PR #5995, opened 2016-11-06
Timed record tweak

GawainLynch:
This was actually requested by @CarsonF recently (read: two months ago), and it slipped my mind.

### DIFF
--- a/src/Provider/StorageServiceProvider.php
+++ b/src/Provider/StorageServiceProvider.php
@@ -288,14 +288,15 @@ class StorageServiceProvider implements ServiceProviderInterface
         $app['storage.event_processor.timed'] = $app->share(
             function ($app) {
                 $contentTypes = array_keys($app['config']->get('contenttypes', []));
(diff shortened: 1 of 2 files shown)


bobdenotter:
```
$interval = $app['config']->get('general/caching/duration', 10) * 60;
```

And later: 

```
-        $this->cache->save(self::CACHE_KEY_HOLD, true, $this->config->get('general/caching/duration', 10));
+        $this->cache->save(self::CACHE_KEY_HOLD, true, $this->interval);
```

Isn't what we pass into `$this->cache->save` 60 times too large?

GawainLynch:
> Isn't what we pass into $this->cache->save 60 times too large?

Caching interval parameter is in seconds, the [duration in config](https://github.com/bolt/bolt/blob/release/3.1/app/config/config.yml.dist#L150) is minutes

bobdenotter:
Cool, just checking! 👍

PR #5999, opened 2016-11-07
Add missing use statement to HiddenType

rossriley:
Causing exception when using hidden fields.


GawainLynch:
That damn :koala: out of its :evergreen_tree: again?

PR #6005, opened 2016-11-07
[BUG] Do not escape markdown content in backend

Rovak:
Details
---------

When using the markdown field to write code examples some characters would be escaped, for example:

```
test(x => y)
```
is being converted to

```
test(x =&gt; y)
```

Reproduce
--------------

1. Make a new post which contains a markdown field
2. Add a code snippet, like the example given above
3. Save the blog post
4. Press F5
5. The code is now escaped

This PR adds the `raw` filter to the markdown field so HTML tags are no longer being escaped by twig

### DIFF
--- a/app/view/twig/editcontent/fields/_markdown.twig
+++ b/app/view/twig/editcontent/fields/_markdown.twig
@@ -48,6 +48,6 @@
 
 {% block fieldset_controls %}
     <div class="col-xs-12">
-        <textarea{{ macro.attr(attributes.text) }}>{{ context.content.get(contentkey)|default('') }}</textarea>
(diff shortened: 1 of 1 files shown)


bobdenotter:
Looks good, at first glance. Thanks, @Rovak! 👍 

I'll go over this tomorrow, and barring anything unforeseen, we'll merge it in.

bobdenotter:
@Rovak I did some more digging, and found that the issue was "deeper" than what we thought. The fix in #6018 will not only fix it here, but also for `text` and `HTML` fields, both in the backend and in the frontend.. 

Thanks for bringing it to my attention! 💯

PR #6008, opened 2016-11-08
Remove record taxonomy based routes from example

### LINKED ISSUE #5211: Routing: Weird behavior on custom bindings
This is the first time I tinker with the custom route bindings. For testing, I enabled chapter route binding in the `routing.yml`

```
chapterbinding:
    path: /{chapters}/{slug}
    defaults:
        contenttypeslug: showcases
        chapters: showcase
    requirements:
        chapters: [ controller.requirement:singleTaxonomy, [chapters, showcase] ]
    contenttype: showcases
```

I have the 3 standard chapter `main`, `meta` and `other`.

Now I have the record with the id `5` in the main chapter but can access it with this URLs:
- `/main/5`
- `/meta/5`
- `/other/5`
- `/showcase/5`

The direct record link generated from the backend is `/showcase/5` because that's the fallback.

Same with the category-binding. I can access the record via a category which is not applied to the record.

This doesn't seem to be the correct behavior, right?
## Details
- Relevant Bolt Version: 3 Beta 2
- Install type: composer
- PHP version: 7.0.4
- Used webserver: php-dev

SahAssar:
Removes the routing examples that do not work since a while back. fixes #5211 and see #5593 for context.

### DIFF
--- a/app/config/routing.yml.dist
+++ b/app/config/routing.yml.dist
@@ -102,34 +102,6 @@ contentlisting:
 #     recordslug: page/about
 
 
-
-# Example: Use the following to add a route for contenttypes in your URL scheme. In this example we
-# add routing for the 'chapter' taxonomy, so we can link to /meta/lorum-ipsum or /main/foo-bar.
-# note that the 'showcase' in the defaults and requirements determine the fallback, for when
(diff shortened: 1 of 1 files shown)


bobdenotter:
Merging in! 👍

PR #6012, opened 2016-11-09
Prevent 404 on missing ContentType for homepage

### LINKED ISSUE #6011: [BUG] 404 page after updating Bolt
I'm having an issue after upgrading one of my Bolt sites to 3.2.0. This issue has already been reported over on the forum [here](https://discuss.Bolt.cm/d/502-3-1-6-and-homepage) but I'll go into a little more detail here with a test case. After upgrading Bolt to 3.2.0 (or 3.1.6) and navigating to the homepage you might see the 404 page. A quick fix for this would be to set the homepage as the 404 page but this is a hack to get the site working again. Reverting back to 3.1.5 also fixes this issue.

Trying to replicate the issue with fresh copy of Bolt (I used a clone from Github to make switching versions easier) I noticed that I was missing the `homepage` key in the `config.yml` for the site that was breaking. Removing this key also broke the fresh copy of Bolt (when I'm checked out at 3.1.6).

Looking at the changes between 3.1.5 and 3.1.6, I found that reverting [this](https://github.com/bolt/bolt/commit/38ace924a78ebd436a04422e23d1783bcba75ff3) commit fixed this issue and looking at what has changed this would make sense. If this is intended behaviour then it would be good to get something about this on the docs as this was working before the update.

As for a fix I'm still working on that but it could be a quick fix.

Details
------------
 - Relevant Bolt Version: [ 3.1.6 | 3.2 ]
 - Install type: [ GitHub checkout | Composer install ]
 - PHP version: 7.0
 - Used web server: Nginx

Reproduction
------------
1. Setup a new install of Bolt 3.1.5
2. Create first user and add some sample content
3. Click the 'view site' button and you should see the homepage
4. Edit the `config.yml` and remove the `homepage` key
5. Click the 'view site' button again and should still get the normal homepage
6. Update the site to either 3.1.6 or 3.2.0
7. Check the homepage and it should be showing the 404 page
8. Restoring the `homepage` key or reverting the commit fixes the homepage

GawainLynch:
Fixes #6011

Credit: @jkazimir :koala: Spotters Inc.

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -485,8 +485,13 @@ class Frontend extends ConfigurableBase
     {
         $contentType = $this->getContentType(current(explode('/', $contentTypeSlug)));
 
+        // If there is no ContentType, don't get parameters for it
+        if ($contentType === false) {
+            return [];
+        }
+
         // If the ContentType is 'viewless', don't show the listing / record page.
-        if ($contentType === false || $contentType['viewless']) {
+        if ($contentType['viewless']) {
             $this->abort(Response::HTTP_NOT_FOUND, 'Page ' . $contentType['slug'] . ' not found.');
         }
 


bobdenotter:
Whoop! 👍

AlphaRecon19:
This also fixes the issue I was having on 3.1.6. Do you want me to backport it to the `release/3.1` branch?

GawainLynch:
All good, @AlphaRecon19. `release/3.1` was closed on Sunday as `release/3.2` is now the stable branch. Thank you for the offer though :+1:

PR #6018, opened 2016-11-09
Prevent encoding of entities "On save" in text fields.

bobdenotter:
Fixes: #6005

Details
-------

Our sanitiser inserts HTML entities for non-correctly used ones when saving user input to the database. For example, `Foo & Bar` becomes `Foo &amp; Bar`.

With this fix: 

`Foo & Bar` stays `Foo & Bar`.
`<b>Foo &amp; Bar</b>` stays `<b>Foo &amp; Bar</b>`.
`<b>Foo &amp; Bar & Baz</b>` becomes `<b>Foo &amp; Bar &amp; Baz</b>`.

This fix checks if the input contains improperly encoded `<`, `&`, etc., and reverts the cleanup on these. This just applies to sanitising just before fields of type `text`, `html`, `textarea` and `markdown` get inserted into the DB only.

@CarsonF I've gone over this with @GawainLynch. If you have a better solution, i'm all ears. For now, this seems to fix it, and it's pretty much contained.

### DIFF
--- a/src/Storage/Field/Sanitiser/Sanitiser.php
+++ b/src/Storage/Field/Sanitiser/Sanitiser.php
@@ -41,6 +41,10 @@ class Sanitiser implements SanitiserInterface
             ? $this->getWyswigAllowedTags()
             : $this->getAllowedTags();
 
+        // If the input does not contain encoded HTML entities, we'll need to
+        // decode the output later, because the sanitiser will insert entities.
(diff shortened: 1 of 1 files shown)


bobdenotter:
Bumping @CarsonF 👊

bobdenotter:
I tried doing it in Maid, which is when I found out that `DomDocument::saveHTML()` disregards the setting for `substituteEntities`. :-)

CarsonF:
Well either way that logic here could be moved there.

Issue #5893, opened 2016-10-11
[RFC] Replacing RandomLib


GawainLynch:
PHP 7.1 (currently in RC3) deprecates `mcrypt_*` functions, possibly for removal as early as 7.2: 
https://wiki.php.net/rfc/mcrypt-viking-funeral

Nut and server logs get very noisy as a result:

```
PHP Deprecated:  Function mcrypt_module_close() is deprecated in [snip]/vendor/ircmaxell/random-lib/lib/RandomLib/AbstractMcryptMixer.php on line 86
PHP Stack trace:
PHP   1. {main}() [snip]/vendor/bolt/bolt/app/nut:0
PHP   2. Symfony\Component\Console\Application->run() [snip]/vendor/bolt/bolt/app/nut:18
PHP   3. Symfony\Component\Console\Application->doRun() [snip]/vendor/symfony/console/Application.php:124
PHP   4. Symfony\Component\Console\Application->doRunCommand() [snip]/vendor/symfony/console/Application.php:193
PHP   5. Symfony\Component\Console\Command\Command->run() [snip]/vendor/symfony/console/Application.php:849
PHP   6. Bolt\Nut\BaseCommand->initialize() [snip]/vendor/symfony/console/Command/Command.php:231
PHP   7. Silex\Application->boot() [snip]/vendor/bolt/bolt/src/Nut/BaseCommand.php:34
PHP   8. Bolt\Provider\SessionServiceProvider->boot() [snip]/vendor/silex/silex/src/Silex/Application.php:197
PHP   9. Pimple->offsetGet() [snip]/vendor/bolt/bolt/src/Provider/SessionServiceProvider.php:99
PHP  10. Pimple::{closure:[snip]/vendor/pimple/pimple/lib/Pimple.php:122-130}() [snip]/vendor/pimple/pimple/lib/Pimple.php:83
PHP  11. Bolt\Provider\SessionServiceProvider->Bolt\Provider\{closure}() [snip]/vendor/pimple/pimple/lib/Pimple.php:126
PHP  12. Pimple->offsetGet() [snip]/vendor/bolt/bolt/src/Provider/SessionServiceProvider.php:61
PHP  13. Pimple::{closure:[snip]/vendor/pimple/pimple/lib/Pimple.php:122-130}() [snip]/vendor/pimple/pimple/lib/Pimple.php:83
[...]

PR #6021, opened 2016-11-10
[3.3] Replacing RandomLib

GawainLynch:
In an effort to get PHP 7.1 a happening thing, and per RFC #5893 … removal of `RandomLib` use in core.

This PR maintains a service under `$app['randomgenerator']` for BC. 

There is a corner-case BC break with this PR, where someone is type hinting on the service itself, which would be pretty rare. 

### Imports

```
use RandomLib\Generator;
```
to:
```
use Bolt\Security\Random\Generator;
```

### Function signatures

In the case of a fully namespaced function hint:

```
function foo(\RandomLib\Generator $generator)
{
}
```
to
```
use Bolt\Security\Random\Generator;

// …

function foo(Generator $generator)
{
}
```

Would generally suffice.

### DIFF
--- a/src/AccessControl/AccessChecker.php
+++ b/src/AccessControl/AccessChecker.php
@@ -1,17 +1,18 @@
 <?php
+
 namespace Bolt\AccessControl;
 
 use Bolt\Events\AccessControlEvent;
 use Bolt\Events\AccessControlEvents;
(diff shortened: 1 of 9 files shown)


bobdenotter:
Looks good to me!

PR #6023, opened 2016-11-10
Enable HTTP Cache to be passed configuration options

GawainLynch:
I'll add a PR to docs if/when this is agreed upon

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -181,7 +181,10 @@ class Application extends Silex\Application
             ->register(new Provider\TwigServiceProvider())
             ->register(new Provider\RenderServiceProvider())
             ->register(new Silex\Provider\HttpCacheServiceProvider(),
-                ['http_cache.cache_dir' => $this['resources']->getPath('cache/' . $this['environment'] . '/http')]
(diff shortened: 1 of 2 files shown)


bobdenotter:
How do these relate to these? https://github.com/bolt/bolt/blob/release/3.2/app/config/config.yml.dist#L169-L176 

I can see this being slightly confusing to novices (and me) :-)

GawainLynch:
> How do these relate to these? 

```
+# For more information on performance settings see https://docs.bolt.cm/performance/
```

But you will need to enable request caching first … in the middle of the "I'll add a PR to docs if/when this is agreed upon" part

> I can see this being slightly confusing to novices (and me) :-)

There is a reason that it is commented out. It is not a feature for a novice, and am happy to remove the `.dist` part if that is people's wishes … rebases are easy :wink:

GawainLynch:
Docs: https://github.com/bolt/docs/pull/573

GawainLynch:
Note that docs are being pulled as we're not documented advanced-level user features.

PR #6029, opened 2016-11-11
readded missing alias config and default sizes

ToBe998:
Readded two missing config values for the thumb-aliases change already integrated into 3.2.
These got "lost" somewhere down the git train, no clue why.


bobdenotter:
Thanks, I'll test in a bit, and will let you know

bobdenotter:
Thanks, @ToBe998 👍

PR #6031, opened 2016-11-11
Fix for Notice and subsequent warning related to repeater fields

graham73may:
PR fixes a notice and warning caused by all repeater fields.
 
Notice: Undefined index: fields in C:\MAMP\htdocs\nuffieldtrust.org.uk\vendor\bolt\bolt\src\Storage\Field\Type\RepeaterType.php on line 86


GawainLynch:
Confirmed here on debug inspection.

Issue #5979, opened 2016-11-02
[RFC] No way to re-trigger copying web assets for extensions..


bobdenotter:
If I install an extension in Bolt, like `SEO`, it comes with a `web/` folder. On installation, the contents of this `web/`folder get copied to `public/extensions/vendor/bolt/seo/`, so these assets become available in a web browser. 

In the case that I accidentally clear this `public/extensions/` folder, I can't "rebuild" those files, without removing and re-installing the `SEO` extension. 

I'd like it if there was an option to re-trigger this copy action. Preferably both as a `app/nut` command, as well as through the "Update all things" button on the Extend page in the backend. 

Details
-------

 - Relevant Bolt Version: [ 3.2 | master ]
 - Install type: [ Zip/tarball install | Composer install ]

GawainLynch:
For those reading along, the title is slightly misleading.

From the CLI this can be achieved using Composer:
```
cd extensions/
composer run-script post-package-update
```

We do need a more convenient way via both Nut and the admin UI to handle this, without a doubt.

bobdenotter:
> From the CLI this can be achieved using Composer:

Additional note: in recent composers this doesn't work as expected: 

```
$ composer run-script post-package-update

  [InvalidArgumentException]
  Script "post-package-update" cannot be run with this command
```

GawainLynch:
Ping @rossriley & @CarsonF 

This is intended as an initial RFC on approach: https://github.com/bolt/bolt/compare/release/3.3...GawainLynch:feature/extension-events

CarsonF:
It's actually `composer run post-install-cmd`

GawainLynch:
> This is intended as an initial RFC on approach: release/3.3...bolt:feature/extension-events

I have deleted said extension. If someone wants to waste their time on this, they are more than welcome, but my code is gone.

GawainLynch:
This is hard, and despite repeated questions to the team on approach nothing has come up.

Filing in the circular filing cabinet.

PR #6034, opened 2016-11-11
Re-add extension:setup command

GawainLynch:
This re-adds Nut's extension:setup command, minus the local extension ~~mess~~ functionality

### DIFF
--- a/src/Nut/ExtensionsSetup.php
+++ b/src/Nut/ExtensionsSetup.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Bolt\Nut;
+
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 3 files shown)


bobdenotter:
My apologies if we've decided on this already, but were we going to add a CLI-based option to "re-copy the extensions assets to `html/bolt-public`" in here, like it happens when you first install an extension?

GawainLynch:
Yes, but this is Step No. 1 …

I was just looking at #5979 and did this as a first step … I don't know how long that branch will take, and this reverts a BC break. So I went with get something done now, address RFC later :smile:

bobdenotter:
Good enough for me.. Let's get shit done. 

👍

PR #6040, opened 2016-11-12
Upstream deprecations

GawainLynch:
* Switch choices key/values around and set choices_as_values to true
  * When upgrading to Symfony 3, we'll need to remove choices_as_values
* Don't catch **errors** (still catching exceptions) on CLI
* Update Nut extensions command
* Remove use of `Bolt\Render` where a `BoltResponse` is neither needed nor appropriate
* Only use `Twig_Node`s for `setcontent` where parameters
* Remove deprecation from `SwitchNode` and fix a bug


bobdenotter:
Whoop! 👍

PR #6058, opened 2016-11-17
More Symfony deprecations: Replace BoltResponse

GawainLynch:
`BoltResponse` is designed to allow us to internally track template information in the response, one of the big problems was that it shifted the Twig rendering into the response class itself. This meant that exceptions and errors were often swallowed.

Twig is removing the environment getter from the template class, as well, meaning that we'd have to find a way to get a hold of it.

This PR moves the rendering back to the Render class, and replaces `BoltResponse` with a simpler `TemplateResponse` that has fluent setters for the additional data we wish to store.


bobdenotter:
Sweet! 👍

PR #6059, opened 2016-11-18
HTTP cache options

GawainLynch:
The parameters passed to http_cache.options must always be an array, cached config can cause an exception to be thrown


bobdenotter:
Whoop!

PR #6063, opened 2016-11-21
Query ~~OCD~~ tidy

GawainLynch:
* A bunch of PHPDoc 
* Some code readability
* Branding and PSR-2
* `\Bolt\Storage\Query\QueryParameterParser::parseValue()` will throw an exception if invalid parameter is passed … a valid array is currently assumed where it is used, and let's not go down that road again :laughing:


bobdenotter:
Looks good to me, too! 👍

PR #6067, opened 2016-11-21
fix accessing slug property from Entity\Content class

### LINKED ISSUE #6066: [BUG] link() method inconsistences between Entity\Content and Legacy\Content
Bolt version 3.2.2. When content is fetched using the repository methods, the link() method doesn't return the correct URL to the entry as opposed to the legacy content class, which returns the correct link.

```PHP
// wrong links: falling back to id instead of slug
$content = $app['query']->getContent('pages/1');
dump($content->link()); // outputs: "/page/1"

$content = $app['storage']->getRepository('pages')->find(1);
dump($content->link()); // outputs: "/page/1"

// correct link: using slug
$content = $app['storage']->getContent('pages/1');
dump($content->link()); // outputs: "/page/SLUG"
```

Matching routing:
```YML
contentlink:
    path: /{contenttypeslug}/{slug}
    defaults:
    requirements:
        contenttypeslug: controller.requirement:anyContentType
```

maetthu:
Fixes #6066 

When accessed from the Entity\Content class, this trait cannot access its slug property, resulting in an incorrect link.

### DIFF
--- a/src/Storage/Entity/ContentRouteTrait.php
+++ b/src/Storage/Entity/ContentRouteTrait.php
@@ -96,7 +96,8 @@ trait ContentRouteTrait
             return null;
         }
 
-        $slug = $this->values['slug'];
+        // slug is either stored in $this->values['slug'] or in $this->slug
+        $slug = $this->values['slug'] ?: $this->slug;
         if (empty($slug)) {
             $slug = $this->id;
         }


GawainLynch:
Thanks, @maetthu 

I've merged into `release/3.2` via d1095f2bb6394ce89c99617da89c82bc1d1faea4 and added a clean up, from another branch to address more edge cases.

maetthu:
Very nice, thanks :)

Issue #6069, opened 2016-11-22
[BUG] HTML showing in Latest system activity widget


graham73may:
![image](https://cloud.githubusercontent.com/assets/6642498/20516233/7498a8f6-b08d-11e6-8fbd-9b5436113559.png)

<tt> tags showing in activity widget (and maybe other HTML, not seen anything else yet). 

Details
-------
 - Relevant Bolt Version: 3.2.2
 - Install type: Composer install
 - PHP version: 7.0.9 
 - Used web server: Apache

Reproduction
------------
Login and look at the dashboard widget

bobdenotter:
Fixed in #6080

PR #6080, opened 2016-11-24
Output the log.message as raw as the logs contain HTML.

graham73may:
Fix for issue #6069 

There is HTML in the logs that gets displayed.


bobdenotter:
Looks good to me! 👍

PR #6087, opened 2016-11-25
Fixing issues with underscores

### LINKED ISSUE #6051: [BUG] Underscores in Contenttypes cause issues with relations
If a Contenttype uses an underscore (e.g. pages_en) this will cause issues with relations, because in Bolt\Legacy\Storage::getRelation() instead of the tablename the slug is being used. For contenttypes with an underscore this by default gets replaced by a dash.

Bolt Version: 3.2.2

Possible solution:
Change
 `$contenttype = $content[util::array_first_key($content)]->contenttype['slug']`
to
`$contenttype = $content[util::array_first_key($content)]->contenttype['tablename']`

rossriley:
This is the first of a couple of PRs to fix some underscores / slug / key naming problems.

This one fixes #6051

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2217,6 +2217,7 @@ class Storage
                         $slugifiedContentType === $this->app['slugify']->slugify($ct['singular_name'])
                     ) {
                         $contenttype = $ct;
+                        $contenttype['key'] = $key;
                         break;
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/Legacy/Storage.php
```diff
@@ -2557,7 +2558,8 @@ protected function getRelation($content)
         }
 
         // Get the contenttype from first $content
-        $contenttype = $content[util::array_first_key($content)]->contenttype['slug'];
+        $ct = $content[util::array_first_key($content)];
```
GawainLynch: This has changed in 3.3, and this will nuke that … https://github.com/bolt/bolt/blob/release/3.3/src/Legacy/Storage.php#L2383-L2384

GawainLynch:
@bobdenotter I am going to merge (or you can), all of these with @rossriley's others, and my two, as I want to keep a close eye on the cascade into `release/3.3` in a number of places … 

Just easier to do it all at once, with the PRs fresh in my mind.

PR #6089, opened 2016-11-26
Fix metadata lookup in repeaters

### LINKED ISSUE #6046: [BUG] 3.2.2 - Unable to save entries in templatefield repeaters
Confirmed as a bug in 3.2.2 only:

theme.yml:

```
templatefields:
    services/index.twig:
        services:
            type: repeater
            label: Services
            fields:
                title:
                    type: text
                body:
                    type: html
```

Whenever an entry has been added to the templatefield, the whole site breaks, throwing the following exception:

uncaught exception fieldconfigurationexception

Also cleared cache, totally removed the cache folder, recreated the `theme.yml` file just incase it was some weird line ending thing, all sorts.

Above config does work in 3.2.1.

Details
-------

 - Relevant Bolt Version: 3.2.2 ONLY
 - PHP version: 5.6

Reproduction
------------

1. Create a theme.yml using above config
2. Create a page / contenttype and have it use the specified theme
3. Create a new record and add entries into the templatefield
4. Observe exception

rossriley:
Bug introduced by move around of metadata storage.

Fixes #6046

### DIFF
--- a/src/Storage/Field/Type/RepeaterType.php
+++ b/src/Storage/Field/Type/RepeaterType.php
@@ -82,7 +82,7 @@ class RepeaterType extends FieldTypeBase
     public function hydrate($data, $entity)
     {
         $key = $this->mapping['fieldname'];
-        if ($this->isJson($data[$key])) {
+        if (isset($data[$key]) && $this->isJson($data[$key])) {
             $originalMapping[$key]['fields'] = $this->mapping['fields'];
             $originalMapping[$key]['type'] = 'repeater';
             $mapping = $this->em->getMapper()->getRepeaterMapping($originalMapping);
--- a/src/Storage/Field/Type/RepeaterType.php
+++ b/src/Storage/Field/Type/RepeaterType.php
@@ -83,7 +83,7 @@ class RepeaterType extends FieldTypeBase
     {
         $key = $this->mapping['fieldname'];
         if ($this->isJson($data[$key])) {
-            $originalMapping[$key]['fields'] = $this->mapping['data']['fields'];
+            $originalMapping[$key]['fields'] = $this->mapping['fields'];
             $originalMapping[$key]['type'] = 'repeater';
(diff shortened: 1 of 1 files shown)


GawainLynch:
So the moving around of said storage will prevent what occurs in #6031 where that extra parameter was added?

rossriley:
Eek. I'm not sure, I've just debugged a save and the error was that there was no `['data']` parameter coming through on the mapping but there was a `['fields']` @graham73may any ideas, has your fix been superceded elsewhere?

GawainLynch:
Obviously not an option for 3.2, but wondering if we can't move that mapping data to an object for 3.3 as `<insert array rant here>`

rossriley:
Yes, that's already covered by the Mapping Refactor PR.

GawainLynch:
> Yes, that's already covered by the Mapping Refactor PR.

Which I think you've told me before … I need more :coffee:

rossriley:
@GawainLynch @graham73may when you guys have time to check this over.

Here's what I see from debugging this both against standard repeaters and repeaters in templatefields.

First the configs:

```
#contenttypes.yml
         .....
         template:
            type: templateselect
            filter: '*.twig'
        imagerepeater:
            type: repeater
            fields:
                title:
                    type: text
                featured:
                    type: checkbox
                image:
                   type: image
                   attrib: [title, alt]
```

```
# theme.yml

templatefields:
    extrafields.twig:
        services:
            type: repeater
            label: Services
            fields:
                title:
                    type: text
                body:
                    type: html
                image:
                    type: image
                    attrib: [title, alt]
```

I'm seeing both save and load working fine here. Here's the debug of `$this->mapping['fields']` in the case of isJson being true which it only is for templatefields.

https://infinit.io/_/3nzuDc2

bobdenotter:
Is this one GTG? Pinging @graham73may, @GawainLynch

GawainLynch:
It triggers an exception for me.

PR #6092, opened 2016-11-26
Backport "canonical" Twig function

GawainLynch:
Not sure what to call this, beyond :koala: 

This is another part of the missing backport of #5836 that "someone" was supposed to take care of, and didn't do a good job of … we're looking at **you**, Kenny :koala: !!!

**Edit:** Obviously credit goes to @CarsonF for this work!

PR #6102, opened 2016-11-28
Now Carson tells me! ;-)

GawainLynch:
So from #5836 there was a part that was missed in #6092 that would cause a #1555 … 

For that mater a #1, #2, #1000, #1024, #2000, #2048, #3000, #3072, #4000, #4096, #5000, #5128, #6000 … luckily #6144 hasn't happened yet so …

More credit to @CarsonF for both the original work, and relentless facepalming at me for not paying enough attention.


bobdenotter:
Aaaight!

PR #6108, opened 2016-11-30
Performance issue fixing in ContentRelationTrait

Yanlaz:
Change function setRelation where arrray_unique and sort on every call can be very slow with big array.
sort has been moved to getRelation.


GawainLynch:
Moved target branch to `release/3.2`, as 3.1 is long closed.

GawainLynch:
Per conversation in #development on slack, we're going to move this to 3.3 as we're not entirely sure that this wont have unintended consequences, and a hotfix release is not a good place to try that. :smile_cat:

PR #6110, opened 2016-11-30
Fixes #6109 - Make content search case insensitive on every platform

### LINKED ISSUE #6109: [BUG] Search not working properly using case-sensitive DBMS
I am using PostgreSQL - searching content doesn't work properly due to Postgres' _LIKE_ operator being case-sensitive unlike most database systems. This is also true for SQLite.

As per [this comment](https://github.com/doctrine/doctrine2/pull/654#issuecomment-17112097), Doctrine chose to let the platform decide how to manage their case sensitivity, this is why I'm opening this ticket on bolt and not on doctrine's repository.

I suggest to change the query and lower the searched fields and the input string with the _LOWER()_ function. That would cover any database system.

Details
-------

 - Relevant Bolt Version: 3.2.1
 - PHP version: 7.0
 - DBMS: PostgreSQL

PR is coming

artggd:
This PR fixes #6109 

I put the searched fields and the input in a _LOWER()_ sql function. That way, the search is now case insensitive on every platform.

Lowering the input string might be seen as double job because the string is already lowered in PHP. I see it more like a double security. I let you to remove one the php or sql part if you think it'd be better.

Note that I worked in the Legacy Storage class, not sure if the non-legacy search is yet in use and how to fix it too.

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -708,7 +708,8 @@ class Storage
         foreach ($fields as $field => $fieldconfig) {
             if (in_array($fieldconfig['type'], $searchableTypes)) {
                 foreach ($query['words'] as $word) {
-                    $fieldsWhere[] = sprintf('%s.%s LIKE %s', $table, $field, $this->app['db']->quote('%' . $word . '%'));
+                    // Build the LIKE, lowering the searched field to cover case-sensitive database systems
+                    $fieldsWhere[] = sprintf('LOWER(%s.%s) LIKE LOWER(%s)', $table, $field, $this->app['db']->quote('%' . $word . '%'));
                 }
             }
         }


GawainLynch:
Note that `master` is 4.0-dev so this would be about a year away from seeing a stable release, in which time this class will be completely removed.

artggd:
Woops! I was about to merge on the 3.x then I saw the contributing guide that says "master is for 3.0". It would be good to update that as well ;)

artggd:
BTW, I created my branch from 3.x, the merge should go smooth (if merge there is)

GawainLynch:
Yes, I have meant to update the guide for a few days … I only make the change late last week before jumping on a plane from the USA back home … 

> BTW, I created my branch from 3.x, the merge should go smooth (if merge there is)

Awesome, I'll move the target then.

GawainLynch:
@bobdenotter I can catch this on merge

GawainLynch:
…and like magic :wink:

bobdenotter:
Looks good to me! 👍

PR #6112, opened 2016-11-30
Fixed: Don't stretch tags to full width

bobdenotter:
Before: 

screen shot 2016-11-30 at 16 29 34

After: 

screen shot 2016-11-30 at 16 37 30


GawainLynch:
Rejecting on the grounds of #1555

Issue #6107, opened 2016-11-30
[RFC] UX improvements to systemlog


jadwigo:
The systemlog in the bolt back-end has some impractical things in the interface.

There is a bug icon that has a lot of usefull information on mouseover - but that info is put into a small popup that is partially off screen.

![image](https://cloud.githubusercontent.com/assets/1488214/20749289/00a1950a-b6f2-11e6-9b30-f60f9d346956.png)

From a UX POV it is like you are hiding the actual thing you want to tell.

A click on the row or icon to expand a a textbox that can be selected might be better.

![image](https://cloud.githubusercontent.com/assets/1488214/20749563/1d0e27fc-b6f3-11e6-8b95-470a29c6e8b6.png)

It might also improve scannability if not all columns that are present are always shown, so something like the IP can be moved to the hidden row.

Details
-------

 - Relevant Bolt Version: [ 2.2 & 3.0 & master ]
 - Install type: [ all ]
 - PHP version: [ all ]
 - Used web server: [ all ]
 - For UX/UI issues: [ all ]

GawainLynch:
This totally has my vote!

bobdenotter:
Looks good to me too.

GawainLynch:
Do you think you can handle this one @bobdenotter ? That probably requires some JS magic … and we all know how dangerous it is to unleash :koala:s  on JS :laughing:

bobdenotter:
Sure, lemme handle it.

PR #6113, opened 2016-11-30
Nicer listing overview for System Log.

bobdenotter:
Before: 

screen shot 2016-11-30 at 16 50 15

After: 

screen shot 2016-11-30 at 17 18 56

Fixes: #6107


GawainLynch:
We need a poster for the office wall: "There is no place like `127.0.0.1` … Except `::1`"

GawainLynch:
PHP 7 took forever to start, and I am impatient tonight … so  … Oh, look … now it finished!

PR #6117, opened 2016-12-01
[BUG] Incoming relation not display when relation is empty

Yanlaz:


bobdenotter:
Tests passing locally. Thanks, @Yanlaz

PR #6121, opened 2016-12-02
Fix for #6115: Added a check on $valueSelect to see if it is a string. 

### LINKED ISSUE #6115: [BUG]  Changes to SelectQueryHandler breaking the JSONAPI extension
Since upgrading to version 3.2.3, the jsonapi extension is breaking. I realize this might not be the place to file a bug about that, however the jsonapi extension is an important one and you may be interested in case the fix belongs on the Bolt side. 

I filed a bug for the extension also:
https://github.com/xiaohutai/jsonapi/issues/34

Details
-------
 - Bolt Version: 3.2.3
 - PHP version: 7.0
 - Used web server: Apache

Reproduction
------------
 1. **Bug summary**: 
    Using the jsonapi extension does not work due to a PHP exception.

 2. **Steps to reproduce**:
    * Install the jsonapi extension.
    * Add some content, for instance entries.
    * Go to .../json/entries

 4. **Expected result**: 
    You should be seeing json output, with a list of the entries.

 5. **Actual result**: 
    Instead you get an exception stemming from the new whitelisting code in SelectQueryHandler.php: 

Uncaught Exception: ContextErrorException
ContextErrorException in SelectQueryHandler.php line 74: 
Warning: preg_split() expects parameter 2 to be string, object given

```php
$metadata = $repo->getClassMetadata();
        $allowedParams = array_keys($metadata->getFieldMappings());
        $cleanParams = [];
        foreach ($queryParams as $fieldSelect => $valueSelect) {
            $stack = preg_split('/ *(\|\|\|) */', $fieldSelect);
            $valueStack = preg_split('/ *(\|\|\|) */', $valueSelect);
```

While debugging I found that "$valueSelect" is coming in as a "Bolt\Extension\Bolt\JsonApi\Converter\Parameter\Type\Page Object" instead of as a string.

skoop:
This PR should fix #6115.

We've added a check to `$valueSelect` to make sure it is a string. If not, we set the `$stack` empty. This is to prevent calling `preg_split()` on `$valueSelect` in the situations where it is an object.

### DIFF
--- a/src/Storage/Query/Handler/SelectQueryHandler.php
+++ b/src/Storage/Query/Handler/SelectQueryHandler.php
@@ -70,8 +70,12 @@ class SelectQueryHandler
         $allowedParams = array_keys($metadata->getFieldMappings());
         $cleanParams = [];
         foreach ($queryParams as $fieldSelect => $valueSelect) {
-            $stack = preg_split('/ *(\|\|\|) */', $fieldSelect);
-            $valueStack = preg_split('/ *(\|\|\|) */', $valueSelect);
+            $stack = [];
+
+            if (is_string($valueSelect)) {
+                $stack = preg_split('/ *(\|\|\|) */', $fieldSelect);
+                $valueStack = preg_split('/ *(\|\|\|) */', $valueSelect);
+            }
 
             if (count($stack) > 1) {
                 $allowedKeys = [];


GawainLynch:
Thanks @skoop :+1:

PR #6124, opened 2016-12-02
Fix twig ismobileclient() to _actually_ not use super global

CarsonF:


GawainLynch:
This is where our OCD collides :wink:

CarsonF:
Which part?

PR #6128, opened 2016-12-04
Configuration check helpers

bobdenotter:
Adds a configuration notice when running "debug" in PROD. Comments welcome. 

screen shot 2016-12-04 at 14 28 52

### DIFF
--- a/app/view/twig/_sub/_configuration_notices.twig
+++ b/app/view/twig/_sub/_configuration_notices.twig
@@ -8,7 +8,7 @@
 {% block panel_class 'panel-configuration' %}
 
 {% block panel_alert %}
-    class="progress-bar progress-bar-info progress-bar-striped active"
+    class="progress-bar progress-bar-warning progress-bar-striped active"
     style="width: 100%; text-align: left; color: #333; padding: 10px 14px; margin-bottom: 15px;"
 {% endblock panel_alert %}
 
@@ -17,6 +17,7 @@
 {% block panel_head %}<b>Configuration Notices</b>{% endblock %}
 
 {% block panel_body %}
+
     {% set shown = [] %}
     <p>
         {{ __('general.bolt-configuration-issues-detected') }}
@@ -24,8 +25,19 @@
     <ul>
     {% for msg in messages %}
         {% if msg not in shown %}
-            <li>{{ msg|ymllink }}</li>
             {% set shown = shown|merge([msg]) %}
+
+            {% if msg is json %}
+                {% set msg = msg|json_decode %}
+                <li>
+                    {{ msg.notice|ymllink }}
+                    {% if msg.info|default() %}
+                        <br><small>{{ msg.info|raw }}</small>
+                    {% endif %}
+                </li>
+            {% else %}
+                <li>{{ msg|ymllink }}</li>
+            {% endif %}
         {% endif %}
     {% endfor %}
     </ul>
--- a/app/view/twig/dashboard/_aside.twig
+++ b/app/view/twig/dashboard/_aside.twig
@@ -3,18 +3,6 @@
 {# Configuration notices #}
 {{ include('@bolt/_sub/_configuration_notices.twig') }}
 
-{# If we're running a development version, show annoying message. #}
-{% if not bolt_stable %}
-<div class="panel panel-default panel-news">
-    <div class="progress-bar progress-bar-warning progress-bar-striped active" style="width: 100%; text-align: left; color: #333; padding: 10px 14px; margin-bottom: 15px;">
-        <b><i class="fa fa-fw fa-bullhorn"></i> Warning: Development version </b>
-    </div>
-    <div class="panel-body">
-        <p>This is a development version of Bolt, so it might contain bugs and unfinished features. Use at your own risk! For 'production' websites, we advise you to stick with the official stable releases.</p>
-    </div>
-</div>
-{% endif %}
-
 {# Dashboard Widget #}
 {{ widgets('dashboard_aside_top', 'backend') }}
 
--- a/src/Application.php
+++ b/src/Application.php
@@ -267,6 +267,11 @@ class Application extends Silex\Application
                 return new Stopwatch\Stopwatch();
             }
         );
+
+        // Initialize our friendly helpers, if available.
+        if (class_exists('\Bolt\Starter\Provider\StarterProvider')) {
+            $this->register(new \Bolt\Starter\Provider\StarterProvider());
+        }
     }
 
     public function initExtensions()
--- a/src/EventListener/GeneralListener.php
+++ b/src/EventListener/GeneralListener.php
@@ -32,19 +32,6 @@ class GeneralListener implements EventSubscriberInterface
     }
 
     /**
-     * Kernel request listener callback.
-     *
-     * @param GetResponseEvent $event
-     */
-    public function onKernelRequest(GetResponseEvent $event)
-    {
-        $request = $event->getRequest();
-
-        $this->mailConfigCheck($request);
-        $this->gdCheck($request);
-    }
-
-    /**
      * Kernel response listener callback.
      *
      * @param FilterResponseEvent $event
@@ -55,37 +42,6 @@ class GeneralListener implements EventSubscriberInterface
         $response = $event->getResponse();
 
         $this->setFrameOptions($request, $response);
-    }
-
-    /**
-     * No mail transport has been set. We should gently nudge the user to set
-     * the mail configuration.
-     *
-     * @see https://github.com/bolt/bolt/issues/2908
-     *
-     * @param Request $request
-     */
-    protected function mailConfigCheck(Request $request)
-    {
-        if (!$request->hasPreviousSession()) {
-            return;
-        }
-
-        if (!$this->app['config']->get('general/mailoptions') && $this->app['users']->getCurrentuser() && $this->app['users']->isAllowed('files:config')) {
-            $notice = "The mail configuration parameters have not been set up. This may interfere with password resets, and extension functionality. Please set up the 'mailoptions' in config.yml.";
-            $this->app['logger.flash']->configuration(Trans::__($notice));
-        }
-    }
-
-    /**
-     * Check whether or not the GD-library can be used in PHP. Needed for making thumbnails.
-     */
-    protected function gdCheck()
-    {
-        if (!function_exists('imagecreatetruecolor')) {
-            $notice = "The current version of PHP doesn't have the GD library enabled. Without this, Bolt will not be able to generate thumbnails. Please enable <tt>php-gd</tt>, or ask your system-administrator to do so.";
-            $this->app['logger.flash']->configuration(Trans::__($notice));
-        }
     }
 
     /**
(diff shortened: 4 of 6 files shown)


src/Config.php
```diff
@@ -371,6 +371,12 @@ protected function parseGeneral()
             $general['branding']['provided_by']
         );
 
+        // Add some common 'development' hosts to the `debug_local_domains` setting:
+        $general['debug_local_domains'] = array_unique(array_merge(
+            (array) $general['debug_local_domains'],
+            ['.dev', 'dev.', 'devel.', 'development.', 'test.', '.test', 'new.', '.new', 'localhost', '.local', 'local.']
+        ));
+
```
GawainLynch: And if you're production domain legitimately uses one of these?

src/Config.php
```diff
@@ -1128,6 +1134,7 @@ protected function getDefaults()
             'debug_enable_whoops'         => false, /** @deprecated. Deprecated since 3.2, to be removed in 4.0 */
             'debug_error_use_symfony'     => false,
             'debug_permission_audit_mode' => false,
+            'debug_local_domains'         => [],
```
GawainLynch: So the people that can't remember to run it off, are going to remember to configure this? Yeah, nah!

src/Config.php
bobdenotter: Doesn't matter. In the off chance your domain is `locahost.org`, you just won't see the notice, no harm done.

src/Config.php
bobdenotter: No, the opposite of that. If you want to get rid of the message, add your dev domain here.

bobdenotter:
I consolidated the 'configuration notices' and the 'development version' warning into one. 

screen shot 2016-12-05 at 14 34 52

bobdenotter:
Rebased and all. Should be GTG now.

bobdenotter:
Green lights. 🚦

CarsonF:
I'm on the fence about this. But the first thing that comes to mind is that this is a check for novices, and novices aren't going to add their "dev domains" to the config. Thus it doesn't even need to be configurable. 

I would rather have these things as a separate composer package, that's included by default with `composer-install`. But it's easy to remove for those that don't want it.

If we insist on keeping these things in core, I would like a "I know what I'm doing" config option.

GawainLynch:
> I would rather have these things as a separate composer package

Ha! You took the words right out of my mouth. I was just thinking though this and came to the same conclusion. 

An even better option would be to have said Composer package install the archive version, that way people doing a default `composer-install` for their base builds can not worry either.

bobdenotter:
> But the first thing that comes to mind is that this is a check for novices, and novices aren't going to add their "dev domains" to the config. Thus it doesn't even need to be configurable.

I don't think that's the point of this, at all. This is something for novices, agreed, but they don't need to add their "dev domains" to config. They _only_ need to do this if they've set up a config name that does _not_ have `.local` or `.dev` or one of the others in it. If they did do that, I wouldn't consider them novices per se. 

In my mind this check _does_ have edge cases where it doesn't "trigger" on production sites. Sure, but the worst that can happen is that it won't show the notice for the dev domains, and it'll act like it does now. In the other 99% of the cases it _will_ work, and it'll help those novices by urging them to turn off debug in prod.  

> I would rather have these things as a separate composer package

Sure, we can add them as composer packages later. Gawain mentioned today that he'd like to move all of these checks out of `GeneralListener.php`. But, that'll take work, and this PR - as it stands - is an improvement over that we have right now.

CarsonF:
Yeah I'm fine with that being the next step. Either way we both agree to remove the config option.

bobdenotter:
> Either way we both agree to remove the config option.

What? no. 

If we do that, Gawain will _always_ see the notice, when he's developing. I'm sure he'll get annoyed by that real quickly.

CarsonF:
Oh I see. The config option is there to disable the check. Why not just have a `check_dev_domain` boolean option instead then?

src/EventListener/ConfigListener.php
```diff
@@ -161,6 +161,8 @@ private function setGlobals()
             $usersCur = null;
         }
 
+        $twig->addGlobal('bolt_name', Version::name());
+        $twig->addGlobal('bolt_version', Version::VERSION);
```
bobdenotter: Those were already there.

bobdenotter: Ah, I see.. They snuck in from here.. https://github.com/bolt/bolt/blob/release/3.2/src/EventListener/ConfigListener.php#L165-L167

I'll remove them.

bobdenotter:
> Why not just have a check_dev_domain boolean option instead then?

Let's say your dev environment is `foo.bar`, and your site will be hosted on `baz.qux`. You start devving, and you get annoyed about the message popping up, as you should.. You then set `debug_local_domains:  [.bar ]`, so it stops bugging you. 

You move the site to PROD at http://baz.qux, forget to turn off debug, and hand over the keys to your editor. The editor will then ask you "So, what is this annoying message about debug in the backend?". You then turn off debug in `config_local.yml`, and all is well. 

president-george-w -bush-mission-accomplished

CarsonF:
> forget to turn off debug

I was assuming that if you disabled the check that doesn't happen.

`debug` is based on an environment variable for me, I don't even use config_local.yml.

CarsonF:
Also `bolt_stable` twig global was there as a way to disable that notice. So if we are removing that then we need to add a config option to disable that check as well.

bobdenotter:
> I was assuming that if you disabled the check that doesn't happen. And if it does that's on you. 

That is the entire point. This is not for you. This is for (mainly) developers who don't have an ingrained mental connect between "debug is on" and "my cache will get out of control in prod if i don't fix that". 

It's a small nudge for those devs to turn it off, should they forget to do so. Or for those who don't even realise that they should do it.

bobdenotter:
> Also bolt_stable twig global was there as a way to disable that notice.

That is something entirely different, mate. I put that in this PR as suggested by Gawain. Basically this PR _consolidates_ the different configuration checks (mail, the 'check for debug on prod' and the 'running unstable versions') into _one_ sidebar widget, instead of spreading them out into multiple.

CarsonF:
How about this, since we can't move it to separate package yet, can you move those to a `NoviceListener`? That way we can just disable it that way. No need to go through the config hassle.

bobdenotter:
> those

Which "those", exactly? all of them?

bobdenotter:
Also, why would you want to disable them?

GawainLynch:
So … 5 minutes https://github.com/GawainLynch/starter-kit

Next 5 minutes the rest of it …

CarsonF:
all of them. I don't need them and I don't want them. I'm running 3.3 on a production site right now. I don't want my editors to see that notice. Is that advised? Of course not. But it's my call as the system owner, not Bolt's.

bobdenotter:
> I'm running 3.3 on a production site right now. I don't want my editors to see that notice.

You really, really are missing the point.. _**YOU**_ won't see this message, at all! If you have your stuff set up correctly, you, nor your editors will see it. EVER.

CarsonF:
The domain check, yes. But there's no way to disable the dev version check now.

bobdenotter:
> But there's no way to disable the dev version check now.

No, but there was no way to turn that off either before this PR.

CarsonF:
```php
// Disable development warning.
$this->on(KernelEvents::REQUEST, function () {
    $this['twig']->addGlobal('bolt_stable', true);
}, 30);
```

CarsonF:
It's hacky, but doable. I would appreciate a better way to do it.

bobdenotter:
I just moved it _out_ of twig and into code, because I thought we all agreed there should be less logic in code. :'-(

CarsonF:
> I just moved it out of twig and into code, because I thought we all agreed there should be less logic in code. :'-(

Definitely. The way I was doing it is wrong. But it needs to be configurable.

I am just looking at the rest of `GeneralListener`. I don't need any of it, so I can just disable that class and I'll be fine.

bobdenotter:
Well, if we can make it a separate package that's included by default (like `bolt/thumbs`), i can live with that. @GawainLynch's repo looks like a good approach to me.

GawainLynch:
OK, Done … If we agree, we'll just register `bolt/starter-kit` on Packagist.

… and I have moved the repo to the org https://github.com/bolt/starter-kit

CarsonF:
As long as bolt doesn't depend on that package it looks good to me.

GawainLynch:
We can then just have something like:
```
        if (class_exists('\Bolt\Starter\Provider\StarterProvider')) {
            $this->register(new \Bolt\Starter\Provider\StarterProvider());
        }
```

GawainLynch:
> As long as bolt doesn't depend on that package it looks good to me.

Archive distribution installation addition.

bobdenotter:
> $this->register(new \Bolt\Starter\Provider\StarterProvider()); 

In `Application.php` or somesuch, right? 

> Archive distribution installation addition.

So, we'll add `bolt/starter-kit` to the composer.json here: https://github.com/bolt/composer-install/blob/release/3.2/composer.json ?

This looks like a good solution to me. I'll be able to add some more "end user friendly" checks there, then. :-)

GawainLynch:
> So, we'll add bolt/starter-kit to the composer.json here: https://github.com/bolt/composer-install/blob/release/3.2/composer.json ?

No, just ask part of the script that builds the distribution. it can do a `composer require` and will just add it to the root project, making it removable afterwards.

GawainLynch:
Packagist + webhooks all set up, so`composer require bolt/starter-kit:^1.0@dev` should now work

>> $this->register(new \Bolt\Starter\Provider\StarterProvider());

>In Application.php or somesuch, right?

Yeah, bottom of the providers section I would guess to be best for now.

GawainLynch:
> This looks like a good solution to me. I'll be able to add some more "end user friendly" checks there, then. :-)

Also what I was thinking. I honestly have no objections to having installs that are easier for certain groups, the easier the better. Just as long as we have choice :+1:

bobdenotter:
> No, just ask part of the script that builds the distribution. it can do a composer require and will just add it to the root project,

I'm not seeing how to implement that properly yet. 

> Yeah, bottom of the providers section I would guess to be best for now.

Ok.

GawainLynch:
After this https://github.com/bolt/bolt-distribution/blob/master/package.sh#L36

You add:
```
composer require bolt/starter-kit:^1.0@dev --working-dir=$COMPILE_DIR
```

bobdenotter:
One leftover tidbit.. What are we gonna do with `debug_local_domains: [ ] ` in `config.yml.dist` ?

GawainLynch:
> One leftover tidbit.. What are we gonna do with debug_local_domains: [ ] in `config.yml.dist` ?

My thought is to add it to the docs only, and have sensible defaults in the package itself. The plan is to reduce what is in  `config.yml.dist` by default, and it is already an intimidating wall to the novice user 

… even this not so novice user gets annoyed at the default size :laughing:

bobdenotter:
I'm having trouble getting this to work. How can I get a setup with this running?

bobdenotter:
It looks like this now: 

screen shot 2016-12-06 at 17 57 16

this depends on the PR to `bolt/starter-kit` being merged, oc.

bobdenotter:
See https://github.com/bolt/starter-kit/pull/1/files

GawainLynch:
I am really liking where we went with this … it bridges gaps :+1:

PR #6130, opened 2016-12-04
Add limit to trace, show less data when not logged in.

bobdenotter:
This PR fixes two things: 

 - It makes the amont of 'dumped' trace data configurable. 
 - It makes the exception page show less sensitive data, when people aren't logged in. 

screen shot 2016-12-04 at 15 11 01

### DIFF
--- a/src/Controller/Exception.php
+++ b/src/Controller/Exception.php
@@ -201,6 +201,10 @@ class Exception extends Base implements ExceptionControllerInterface
         $loggedOnUser = (bool) $this->app['users']->getCurrentUser() ?: false;
         $showLoggedOff = (bool) $this->app['config']->get('general/debug_show_loggedoff', false);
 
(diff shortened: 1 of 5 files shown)


src/Controller/Exception.php
```diff
@@ -201,6 +201,10 @@ protected function getContextArray(\Exception $exception = null)
         $loggedOnUser = (bool) $this->app['users']->getCurrentUser() ?: false;
         $showLoggedOff = (bool) $this->app['config']->get('general/debug_show_loggedoff', false);
 
+        // Note: We set this to a high value deliberately. If 'config' is not yet available, the
+        // user can't influence this, so it shouldn't be too low.
+        $traceLimit = (int) $this->app['config']->get('general/debug_trace_argument_limit', 40);
```
bobdenotter: Same as it does two lines above. I assumed this would work, because i couldn't find a case to trigger it.

PR #6131, opened 2016-12-05
Twig Sandbox

CarsonF:
This replaces `safe_twig` with Twig's Sandbox extension (BC should be maintained).

It can be used in Twig with the [`sandbox` tag](http://twig.sensiolabs.org/doc/tags/sandbox.html).
The `twig` function/filter has also been deprecated for the [`template_from_string` function](http://twig.sensiolabs.org/doc/functions/template_from_string.html).


GawainLynch:
Another one of those long roads … thanks mate for getting this across the line!

PR #6132, opened 2016-12-05
Fix sandbox not being disabled when exception is thrown

CarsonF:
See https://github.com/twigphp/Twig/pull/2282


GawainLynch:
Thanks mate, this is awesome!

PR #6135, opened 2016-12-07
Hotfix Twig current function

GawainLynch:
There is a slightly new-ish bug in the `{{ current() }}` Twig function that seems to have been introduced in d56905f9a4d79c31e390525ee315998fdf9bb45b

The short version is that [`Content::offsetExists()`](https://github.com/bolt/bolt/blob/release/3.2/src/Legacy/Content.php#L358) checks the `value` property, and the ContentType was never stored there, rather in its own named property. 

The unit test mocked the Content object therefore hiding the problem, hence the update to the test.

#1555 lives … it is real … it has found us! I warned you all of this day, now it has arrived :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/Twig/Handler/RecordHandler.php
+++ b/src/Twig/Handler/RecordHandler.php
@@ -71,12 +71,12 @@ class RecordHandler
             return true;
         }
 
-        if (!isset($content['contenttype'])) {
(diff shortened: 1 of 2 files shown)


GawainLynch:
FWIW, this turned up as the test itself is killing PHPUnit in `master` … the things you find.

PR #6145, opened 2016-12-08
Fix adjustSidebarHeight so it's longer than content for admin menus

bnelson-cainc:
Remember that fix I submitted for the secondary (black) menu on the left last week? I found a small issue with it.

It works great as long as the content is longer than the menu, but when the reverse is true, as it is on some of the admin menus, I saw menu items that weren't in the black anymore. I hardly ever use those menus anymore, which is why I forgot to check them.

This small adjustment to the original change fixes that, and hopefully it's perfect now.

Sorry for the extra churn!


bobdenotter:
> Sorry for the extra churn!

No worries! Thank you for the additional fix! 👍

PR #6148, opened 2016-12-09
Fixed: HTML Overflow in system log fixed.

bobdenotter:
See before screenshot: 

screen shot 2016-12-09 at 10 12 37


app/view/twig/components/panel-activity-system.twig
```diff
@@ -27,7 +27,7 @@
                 {% elseif log.route == 'logout' %}
                     <em>{{ macro.userlink(log.ownerid|default('')) }}</em> {{ __('panel.latest-activity.logged-out') }}
                 {% else %}
-                    <em>{{ log.message|raw }}</em>
+                    <em>{{ log.message|raw|shy }}</em>
```
GawainLynch: This is so shy, it blushes like I do :smile_cat:

PR #6133, opened 2016-12-06
Gud DI

CarsonF:
This ensures no DI services/parameters are retrieved until boot time.

PR #6157, opened 2016-12-12
Boot extension registered service providers

### LINKED ISSUE #6151: Paths to extensions added pages don't work
For 3.3, I can't access the pages added by an extension.

Details
-------

 - Relevant Bolt Version: [ 3.3 | master ]

Reproduction
------------

1. `git checkout release/3.2`.
2. Install "Labels" extension.
3. note you can access it from the backend. 

screen shot 2016-12-10 at 12 52 30

4. `git checkout release/3.3`.
5. Link in backend is the same, page however can't be found. 

screen shot 2016-12-10 at 12 53 52

GawainLynch:
Fixes #6151

tl;dr #6133 gone b0rked 'setnshun muntin', :koala: bru!

Early in Application we [register](https://github.com/bolt/bolt/blob/release/3.3/src/Application.php#L51) the extension service provider so that it will be the (closest to) first to boot.

When the extension service provider [boots](https://github.com/bolt/bolt/blob/release/3.3/src/Provider/ExtensionServiceProvider.php#L125-L130) we are now registering the extensions, which in the case of anything based on `\Bolt\Extension\SimpleExtension` will implement `ServiceProviderInterface` therefore requiring both `register()` and `boot()` to be called …

Now for the O.C.D. in the room …

[calling `$app->register()`](https://github.com/silexphp/Silex/blob/1.3/src/Silex/Application.php#L176) during `$app->boot()` will add providers to the `$this->providers` in Application, but the iterrator that is [used in `$app->boot()`](
https://github.com/silexphp/Silex/blob/1.3/src/Silex/Application.php#L196-L198) won't have it … so if we're adding that late, we need some way of cleanly invoking those boots.

…and yeah, adding a `boot()` method is killing my OCD too, @CarsonF, but happy to hear better ideas just got work that is D.O.A. right now as a result.

### DIFF
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -11,6 +11,7 @@ use Bolt\Filesystem\Handler\JsonFile;
 use Bolt\Logger\FlashLoggerInterface;
 use Bolt\Translation\LazyTranslator as Trans;
 use Silex\Application;
+use Silex\ServiceProviderInterface;
 use Symfony\Component\Debug\Exception\ContextErrorException;
 
 /**
@@ -33,6 +34,8 @@ class Manager
     private $flashLogger;
     /** @var Config */
     private $config;
+    /** @var bool */
+    private $booted = false;
     /** @var bool */
     private $loaded = false;
     /** @var bool */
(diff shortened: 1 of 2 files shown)


bobdenotter:
giphy-1

PR #6159, opened 2016-12-12
Deprecation and associated clean ups

GawainLynch:
First off the block is, erm, `block()` … So in Twig 2 (as of current) expects the only valid test to is `block() is defined`, as (apparently) defined [here](http://twig.sensiolabs.org/doc/functions/block.html)

Second just ended up being implementing `form_start` & `form_end` where `form_widget` was being used.

I started toying with the idea of the buttons too, but put that in the "bigger work" category.

### DIFF
--- a/app/view/twig/_base/_fieldset.twig
+++ b/app/view/twig/_base/_fieldset.twig
@@ -2,7 +2,7 @@
 
 {#=== INIT ===========================================================================================================#}
 
-{% set widget_name = block('fieldset_widget') %}
+{% set widget_name = block('fieldset_widget') is defined ? block('fieldset_widget') : '' %}
 
 {% if not widget_name %}
     {% set data_bolt_widget = '' %}
--- a/app/view/twig/_base/_page-nav.twig
+++ b/app/view/twig/_base/_page-nav.twig
@@ -13,7 +13,7 @@
 
 {% set page_hasnav = true %}
 
-{% set page_nav = block('page_nav')|default('NO NAV') %}
+{% set page_nav = block('page_nav') is defined ? block('page_nav') : 'NO NAV' %}
 
 {% block page_plain %}
     {{ data('omnisearch.placeholder', __('general.phrase.find')) }}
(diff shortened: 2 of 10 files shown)


app/view/twig/_base/_page-nav.twig
```diff
@@ -13,7 +13,11 @@
 
 {% set page_hasnav = true %}
 
-{% set page_nav = block('page_nav')|default('NO NAV') %}
+{% if block('page_nav') is defined %}
+    {% set page_nav = block('page_nav') %}
+{% else %}
+    {% set page_nav = 'NO NAV' %}
```
CarsonF: This could still be a ternary operation:

```twig
{% set page_nav = block('page_nav') is defined ? block('page_nav') : 'NO NAV' %}
```

CarsonF: Also doesn't `|default` skip the defined check?

GawainLynch: > This could still be a ternary operation:

Ternary works for me. 

> Also doesn't |default skip the defined check?

Wasn't happening any longer in `master` on Twig 2.x-dev … I did look into the code upstream at the time, but yeah … would have to again … 'cause :steam_locomotive: :boom:

app/view/twig/_base/_fieldset.twig
```diff
@@ -2,7 +2,11 @@
 
 {#=== INIT ===========================================================================================================#}
 
-{% set widget_name = block('fieldset_widget') %}
+{% if block('fieldset_widget') is defined %}
+    {% set widget_name = block('fieldset_widget') %}
+{% else %}
+    {% set widget_name = '' %}
+{% endif %}
```
CarsonF: Ah sorry. Ternary here too? 🙈

app/view/twig/_base/_page.twig
```diff
@@ -15,10 +15,10 @@
 {% set page_locale_short = page_locale_long|slice(0, 2) %}
 {% set page_timezone_offset = app.timezone_offset %}
 
-{% if block('page_subtitle') is empty %}
-    {% set page_title = block('page_title') %}
-{% else %}
+{% if block('page_subtitle') is defined %}
     {% set page_title = block('page_title') ~ ' » ' ~ block('page_subtitle') %}
+{% else %}
+    {% set page_title = block('page_title') %}
```
GawainLynch: Yeah, this one just got unreadable

PR #6162, opened 2016-12-13
[4.0] Update changelog

GawainLynch:
Basically sums up #6079 & #6083 so that we don't start losing stuff early :smile: 

@bobdenotter I've put the branch of the org, so we can massage layout (or add/remove) and then merge after that.


bobdenotter:
Good call.. Thanks! 👍

PR #6163, opened 2016-12-13
Adapted searchSingleContentType() to allow status = ... to be passed as a filter

graham73may:
Trying to use the searchContent() function in the backend to allow for content selection regardless of status (in an extension - in my particular instance this in the RelationList extension). E.g. if I set two things to be timed, the status=published where clause doesn't allow me to select them. 

'status' as a filter has been added to the searchSingleContentType() function with a fallback to `'published'` if a specific filter has not been defined. (Works the same as already handled in Core at [`src/Legacy/Storage.php`](https://github.com/bolt/bolt/blob/v3.2.3/src/Legacy/Storage.php#L1282-L1284))

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -694,6 +694,9 @@ class Storage
      */
     private function searchSingleContentType($query, $contenttype, $fields, array $filter = null, $implode = false)
     {
+        // Add status to fields
+        $fields['status'] = true;
+
         // This could be even more configurable
         // (see also Content->getFieldWeights)
         $searchableTypes = ['text', 'textarea', 'html', 'markdown'];
@@ -745,7 +748,14 @@ class Storage
 
         // Build actual where
         $where = [];
-        $where[] = sprintf("%s.status = 'published'", $table);
+
+        $request    = $this->app['request_stack']->getCurrentRequest();
+        $isFrontend = $request ? Zone::isFrontend($request) : true;
+
+        if ($isFrontend && empty($filter['status'])) {
+            $where[] = sprintf("%s.status = 'published'", $table);
+        }
+
         $where[] = '(( ' . implode(' OR ', $fieldsWhere) . ' ) ' . $tagsQuery . ' )';
         $where = array_merge($where, $filterWhere);
 
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -695,7 +695,7 @@ class Storage
     private function searchSingleContentType($query, $contenttype, $fields, array $filter = null, $implode = false)
     {
         // Add status to fields
-        $fields['status'] = true;
+        $fields['status'] = ['type' => 'status'];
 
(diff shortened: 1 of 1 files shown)


src/Legacy/Storage.php
```diff
@@ -694,6 +694,9 @@ function ($word) {
      */
     private function searchSingleContentType($query, $contenttype, $fields, array $filter = null, $implode = false)
     {
+        // Add status to fields
+        $fields['status'] = true;
+
```
GawainLynch: Something … finger …

Back up in `searchContent()` the [passed in fields](https://github.com/graham73may/bolt/blob/96f24521963422b43623495dc98c114c8c9ff076/src/Legacy/Storage.php#L873-L875) are ConentType field arrays.

The `foreach ($fields as $field => $fieldconfig) {` loop then wants `$fieldconfig['type']` and `$fieldconfig` is a boolean.

graham73may:
Hey @GawainLynch, I've use `type=status` as it doesn't make sense for status field to be searchable. 

The reason it was boolean was because initially I had it as type text, but it was caught by `$searchableTypes = ['text', 'textarea', 'html', 'markdown'];` and added to `$fieldsWhere` as a searchable field - which didn't make sense.

Thanks!

GawainLynch:
We [discussed this tonight at the weekly dev meeting](https://github.com/bolt/bolt/wiki/Dev-meeting-2016-12-13), and our consensus is that this should be both  feature based, and done via the new storage … so a non-stable branch.

There is just too many bodies buried in that class :laughing:

graham73may:
What do you mean by feature based? 😯

GawainLynch:
> What do you mean by feature based? 

That I suck at communication tonight :laughing: …actually I suck at hitting the "Comment" button

We meant "feature" as in that we don't believe it falls into the bug-fix/hot-fix category, so isn't suitable for 3.2. Also we feel these changers should be done as part of the on-going cut-over to the new storage  layer.

GawainLynch:
@rossriley putting this one in your court wrt moving forward in the new layer

bobdenotter:
@rossriley Does this need work in the new Storage layer? I think this already works in the new one, meaning that this one is good to go in, right?

rossriley:
Yes, in the new storage layer, the search just builds on top of the standard `getContent` query so you can pass in all normal filters by doing 

`$app['query']->getContent('pages/search', ['status'=>'published', 'filter'=>'lorem ipsum'])`

But yes, I don't see any problem with this going in now for those that still need to use the legacy storage.

GawainLynch:
> Yes, in the new storage layer, the search just builds on top of the standard getContent query so you can pass in all normal filters by doing

:+1: 

> But yes, I don't see any problem with this going in now for those that still need to use the legacy storage.

It was (unanimously?) agreed to not extend storage features in legacy for too many reasons … nACKing and we can re-open if needed

PR #6166, opened 2016-12-13
Path Resolver

CarsonF:
# Path Resolver
`PathResolver` is the replacement for the _paths_ piece of `ResourceManager`. There are a few differences I'll go into. 

## Standardized configuration (almost)

There are not multiple predefined paths configurations with PathResolver. The `Composer` configuration is the default for PathResolver.
This repo now defines its own `.bolt.yml` since it is non standard. Hopefully we can fix that in 4.0.

## Paths are resolved at access time
`PathResolver` _resolves_ paths when you access them, not when you set them.

For example, with `ResourceManager`:
```php
$resources->setPath('root', '/foo');
$resources->setPath('web', 'public');
$resources->getPath('web'); # /foo/public

// Now change the root path
$resources->setPath('root', '/bar');
// "web" path hasn't changed
$resources->getPath('web'); # /foo/public
```

Now with `PathResolver`:
```php
$resolver->define('root', '/foo');
$resolver->define('web', 'public');
$resolver->resolve('web'); # /foo/public

// Now change the root path
$resolver->define('root', '/bar');
// "web" path picks up this change
$resolver->resolve('web'); # /bar/public
```

## Variables

`PathResolver` takes this resolution of the root path at access time a step further and let's you use any path as a variable in another.

```php
$resolver->define('root', '/');
$resolver->define('web', 'public');
$resolver->define('files', '%web%/files');
$resolver->resolve('files'); # /public/files

// Now let's change web to something else
$resolver->define('web', 'mypublic');

$resolver->resolve('files'); # /mypublic/files
```

## Underlying Library

`PathResolver` uses [webmozart/path-util](https://github.com/webmozart/path-util) instead of our barely maintained fork of [pathogen](https://github.com/bolt/pathogen). path-util has just **19%** of the lines of code that pathogen has ([results](http://pastebin.com/SzhWu6Mh)). path-util is also used by `bolt/filesystem`, where pathogen isn't used by anything. Do I even need to mention that it's an unmaintained fork?

# How to use

You can get a predefined path like so:
```php
$app['path_resolver']->resolve('web');
```

Resolve a more complete path:
```php
$app['path_resolver']->resolve('%config%/config.yml');
```

Resolve a relative path onto the root path:
```php
$app['path_resolver']->resolve('./foo/bar');
```

Absolute paths can be resolved, but will be unchanged (assuming no variables):
```php
$app['path_resolver']->resolve('/var/log');
```

# Configuration

Probably half of the code in this PR is to ensure backwards compatibility with ResourceManager, which can be created several different ways. So in 4.0 this will be simplified up a ton.

## Recommended
The recommended way is with `.bolt.yml/.bolt.php` and this hasn't changed (other than allowing path variables):
```yaml
paths:
  web: mypublic
  files: "%web%/myfiles" # values starting with % need to be quoted
```
```php
<?php
return [
  'paths' => [
    'web'   => 'mypublic',
    'files' => '%web%/myfiles'
  ]
];
```

## Custom

This is for creating the application yourself and returning the instance in `.bolt.php`.

### Old Way

```php
<?php

$configuration = new Bolt\Configuration\Composer(__DIR__);
$configuration->setPath("web", "mypublic");
$configuration->setPath("files", "mypublic/files");
$configuration->setPath("themebase", "mypublic/theme");

$app = new Bolt\Application(['resources' => $configuration]);

$app->initialize();

return $app;
```

### New Way

```php
<?php

$app = new Bolt\Application();

$app['path_resolver.root'] = __DIR__;
$app['path_resolver.paths'] = [
  'web' => 'mypublic',
  // Note that "files" and "themebase" don't need to be redefined.
];

return $app;
```

# What's left

- [ ] RM tests are probably broken
- [ ] Replace RM calls with PathResolver calls throughout the codebase


GawainLynch:
> PathResolver uses webmozart/path-util instead of our barely maintained fork of pathogen.

So [this line](https://github.com/bolt/bolt/blob/master/composer.json#L14) will be disappearing from `master` soon after this is merged? :wink:

GawainLynch:
Overall I like, but still unsure about the bootstrapping … but I'll hold comment until more work has landed.

CarsonF:
I think this is ready to go now. Final thoughts anyone?

GawainLynch:
I :heart: … thanks mate!

PR #6170, opened 2016-12-14
Local Extension Replacement

rossriley:
OK, sorry this took longer than I wanted, had to do some rejigging after the recent extension boot changes.

This adds two new ways to use extensions. Firstly you can register extensions at an application level, that is ones that can be autoloaded via your root composer.json. To do this in your bootstrap, or anywhere else prior to the `$app->initialize()` call:

```
$app['extension']->register(new MyApp\Extensions\CleverExtension());
```

We also add a new root namespace for extensions that are found at `extensions/local`, these can't be instantiated until the extension boot process so these must be registered by class name. If you use a php bootstrap this works like:

```
#bolt.php
...
// This will load a class that is found at extensions/local/MyLocalExtension/ImageProcessExtension.php
$app['extension']->register('MyLocalExtension\ImageProcessExtension');
...
```

More normally for those using the `yml` bootstrap config we add support for an extensions array

```
#bolt.yml

paths:
    web: public
    themebase: public/theme
    files: public/files
    view: public/bolt-public/view
extensions:
    - MyextOne\LocalExtensionOne
    - MyextTwo\LocalExtensionTwo
    - MyextThree\LocalExtensionThree
```

The above example will load classes found at `extensions/local/MyextOne/LocalExtensionOne.php`, `extensions/local/MyextTwo/LocalExtensionTwo.php` and `extensions/local/MyextThree/LocalExtensionThree.php` respectively.

After this happens then the extensions will load normally and can have identical functionality to traditionally loaded extensions with two important distinctions.

1. They do not appear in the extensions UI
2. There's no magic for asset copying, local extensions are responsible for putting their assets in the right place, by default this will be at `public/extensions/<response of ->getId()>`

## Edited With Updated Syntax

Internally everything now uses the extension add method so the above bolt.php code is replaced with this:

```
#bolt.php
...
// This will load a class that is found at extensions/local/MyLocalExtension/ImageProcessExtension.php
$app['extensions']->add(new MyLocalExtension\ImageProcessExtension());
...
```

The extension will also be instantiable since the autoload for extensions is now included in the constructor.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -160,14 +160,27 @@ return call_user_func(function () {
         if ($service instanceof Silex\ServiceProviderInterface) {
             $app->register($service, $params);
         }
+
     }
 
+    $extensionArray = [];
     foreach ((array) $config['extensions'] as $extensionClass) {
-        if (is_string($extensionClass)) {
-            $app['extensions']->add(new $extensionClass());
-        } else {
-            $app['extensions']->add($extensionClass);
+        if (is_string($extensionClass) && is_a($extensionClass, ExtensionInterface::class, true)) {
+            $extensionClass = new $extensionClass();
         }
+        if ($extensionClass instanceof ExtensionInterface) {
+            $extensionArray[] = $extensionClass;
+        }
+    }
+
+    if (count($extensionArray)) {
+        $app['extensions'] = $app->share(
+            $app->extend('extensions', function ($extensions) use ($extensionArray) {
+                foreach ($extensionArray as $ext) {
+                    $extensions->add($ext);
+                }
+            })
+        );
     }
 
     return $app;
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -7,6 +7,7 @@ use Bolt\Configuration\ResourceManager;
 use Bolt\Configuration\Standard;
 use Bolt\Debug\ShutdownHandler;
 use Bolt\Exception\BootException;
+use Bolt\Extension\ExtensionInterface;
 use Silex;
 use Symfony\Component\Yaml\Yaml;
 
@@ -80,6 +81,7 @@ return call_user_func(function () {
         'resources'   => null,
         'paths'       => [],
         'services'    => [],
+        'extensions'  => [],
     ];
 
     if (file_exists($rootPath . '/.bolt.yml')) {
@@ -157,6 +159,14 @@ return call_user_func(function () {
         }
         if ($service instanceof Silex\ServiceProviderInterface) {
             $app->register($service, $params);
+        }
+    }
+
+    foreach ((array) $config['extensions'] as $extensionClass) {
+        if (is_string($extensionClass)) {
+            $app['extensions']->add(new $extensionClass());
+        } else {
+            $app['extensions']->add($extensionClass);
         }
     }
 
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -124,9 +124,13 @@ class Manager
         // Set paths in the extension
         if ($baseDir !== null) {
             $extension->setBaseDirectory($baseDir);
+        } else {
+            $extension->setBaseDirectory($this->extFs->getDir($extension->getId()));
         }
         if ($webDir !== null) {
             $extension->setWebDirectory($webDir);
+        } else {
+            $extension->setWebDirectory($this->webFs->getDir($extension->getId()));
         }
 
         // Determine if enabled
@@ -174,7 +178,6 @@ class Manager
             $this->addManagedExtension($descriptor);
         }
 
-        $this->loaded = true;
     }
 
     /**
@@ -229,6 +232,7 @@ class Manager
                 $provider->boot($app);
             }
         }
+        $this->loaded = true;
         $this->booted = true;
     }
 
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -121,7 +121,6 @@ class Manager
             throw new \RuntimeException('Can not add extensions after they are registered.');
         }
 
-
         // Set paths in the extension
         if ($baseDir === null) {
             $baseDir = $this->extFs->getDir($extension->getId());
--- a/src/Extension/Manager.php
+++ b/src/Extension/Manager.php
@@ -121,17 +121,17 @@ class Manager
             throw new \RuntimeException('Can not add extensions after they are registered.');
         }
 
+
         // Set paths in the extension
-        if ($baseDir !== null) {
-            $extension->setBaseDirectory($baseDir);
-        } else {
-            $extension->setBaseDirectory($this->extFs->getDir($extension->getId()));
+        if ($baseDir === null) {
+            $baseDir = $this->extFs->getDir($extension->getId());
         }
-        if ($webDir !== null) {
-            $extension->setWebDirectory($webDir);
-        } else {
-            $extension->setWebDirectory($this->webFs->getDir($extension->getId()));
+        $extension->setBaseDirectory($baseDir);
+
+        if ($webDir === null) {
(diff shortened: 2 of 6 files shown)


src/Provider/ExtensionServiceProvider.php
```diff
@@ -127,6 +131,8 @@ public function boot(Application $app)
         /** @var Manager $extensionService */
         $extensionService = $app['extensions'];
         $extensionService->addManagedExtensions();
+        $app['extension']->boot($app);
```
GawainLynch: This one can go

src/Provider/ExtensionServiceProvider.php
```diff
@@ -127,6 +131,8 @@ public function boot(Application $app)
...
+        $extensionService->triggerLoaded();
```
GawainLynch: Moved down

src/Provider/ExtensionServiceProvider.php
rossriley: That's the one I've just put in.

rossriley: It boots the new extension queue

GawainLynch: But it is called 3 lines down, and needs to be after the `register()`

rossriley:
Because there's only this tiny window where extensions are open for registering...

https://github.com/bolt/bolt/blob/release/3.2/src/Extension/Manager.php#L153-L174

And additionally we wanted an API where people can add extensions to an app, in 2.x we had `$app['extensions']->register(....)` in the first versions of 3.x we had `$app['extensions']->add()` but that doesn't work in later versions, so I suggested that we have a simple published API to add extensions that is separate to the implementation details that will stay stable across major and minor versions and at the same time we unify the mechanism for loading App extensions as well as local extensions.

It should hopefully make the documentation simpler becase all we have to document is a single call that takes a class name and is registered before the app initializes.

Whilst `$app->extend()` is ok for more advanced modifications it's not really ideal for end users who want to do something simple, and also as we said before, the more we encourage users to modify core services, the more inflexible we are to making changes to it.

rossriley:
@GawainLynch is that something I've done on the Travis tests, I've restarted the 5.5 job a few times and it still fails on the development/data folder error, yet the others don't.

CarsonF:
> Because there's only this tiny window where extensions are open for registering...

Extensions can be registered any time before boot. Invoking `extensions` service is not ideal, but that is ok for users if they want to.

CarsonF:
And are you saying `$app['extensions']->add()` doesn't work anymore? It should...

GawainLynch:
> @GawainLynch is that something I've done on the Travis tests, I've restarted the 5.5 job a few times and it still fails on the development/data folder error, yet the others don't.

Travis … Almost always on that MySQL test … don't start me :laughing: 

I've kicked it off again … there is no shortage of my days that are spent pressing that restart button for exactly this reason though :laughing:

CarsonF:
I'm realizing this window you're talking about is true up until #6133. Before that PR you couldn't add extensions after initialize() and the extensions service wasn't registered until initialize(). So the window is only open by extending App and overriding that method.

Now initialize() isn't needed, it's called in the App constructor, and extensions can be added any time before boot.

rossriley:
OK, so which is the best way forward. 

1. Modify the `Manager::add()` method to work with the above syntax. We'll need to modify the method to use the default base directories when not passed in. At the moment if you use `$app['extensions']->add(new MyExtension())` you get

```
LogicException in AbstractExtension.php line 53: Extension was not added with a base directory
``` 

That would be workable for normal application-level extensions but then we've still got the problem for local extensions, that the `extensions/local` directory is only autoloadable on boot, so we either need to include the autoload file on register which I'm not sure is a good idea, because this would potentially allow extension files to be included before main Bolt files.

Or if we want to keep all this functionality in the extensions manager class we could replicate the queue inside manager, but only for local extensions, eg ones passed in by string classname and then on boot, loop through the queue and instantiate them.

@CarsonF, @GawainLynch let me know the preferred approach and I'll get this wrapped up.

CarsonF:
Yeah I think that's fine to modify `add()` to have default logic for directories. 

--- 

> That would be workable for normal application-level extensions but then we've still got the problem for local extensions, that the extensions/local directory is only autoloadable on boot, so we either need to include the autoload file on register which I'm not sure is a good idea, because this would potentially allow extension files to be included before main Bolt files.

Two thoughts here:
1) From what I understand these are not meant for novices, thus we don't need to "help out here". I think the classes should be autoloaded from the project's composer.json not the extension one. 
2) Since Extensions isn't invoked until it is needed now, we could move the `addManagedExtensions` call into the service factory. I think this makes more sense than calling it at `boot()` in the service provider.

tests/phpunit/unit/Extension/ExtensionQueueTest.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Ross Riley <riley.ross@gmail.com>
+ */
+class ExtensionQueueTest extends AbstractExtensionsUnitTest
+{
+    public function testBasicExtensionRegisters()
+    {
+        $app = $this->makeApp();
+        $ext = new NormalExtension();
+        $app['extension']->register($ext);
```
GawainLynch: "extensions"

tests/phpunit/unit/Extension/ExtensionQueueTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $app->initialize();
+        $app->handle(Request::createFromGlobals());
+        $this->assertArrayHasKey('Bolt/Normal', $app['extensions']->all());
+    }
+
+    public function testLocalExtensionRegisters()
+    {
+        $this->localExtensionInstall();
+        $app = $this->makeApp();
+        $app['extension']->register('TestLocalExtension\Extension');
```
GawainLynch: "extensions"

app/bootstrap.php
```diff
@@ -160,5 +162,13 @@
...
     }
 
+    foreach ((array) $config['extensions'] as $extensionClass) {
+        if (is_string($extensionClass)) {
+            $app['extensions']->add(new $extensionClass());
+        } else {
+            $app['extensions']->add($extensionClass);
+        }
+    }
+
```
CarsonF: extensions services shouldn't be invoked here, extend the service. Maybe with an if check to see if there are any extensions to add. 

Also you should do `is_a` check for string and `instanceof` check for object, like we do for `services`, `resources` and `app`.

src/Extension/Manager.php
```diff
@@ -124,9 +124,13 @@ public function add(ExtensionInterface $extension, DirectoryInterface $baseDir =
         // Set paths in the extension
         if ($baseDir !== null) {
             $extension->setBaseDirectory($baseDir);
+        } else {
+            $extension->setBaseDirectory($this->extFs->getDir($extension->getId()));
```
CarsonF: I think it would make more sense to flip the condition here and skip the else block. 
```php
if ($baseDir === null) {
    $baseDir = $this->extFs->getDir($extension->getId());
}
$extension->setBaseDirectory($baseDir);
```

src/Extension/Manager.php
```diff
@@ -174,6 +178,13 @@ public function addManagedExtensions()
             $this->addManagedExtension($descriptor);
         }
 
+    }
+
+    /**
+     *  Marks as loaded, preventing additional registrations
+     */
+    public function triggerLoaded()
```
CarsonF: I don't see a need for this method.

rossriley:
All sorted in time for a Christmas merge 🎄 🎅

GawainLynch:
:santa: :santa: :santa:

PR #6171, opened 2016-12-14
Twig_Source path argument

GawainLynch:


CarsonF:
Ah. Found it. https://github.com/twigphp/twig/commit/44f0f9cecdd72df00039489e162557653e03dc5d#diff-e7bace45505091def27fd3bcf8592eafR251

GawainLynch:
> Passing $file->getFullPath() as the 3rd argument to TwigSource in the FilesystemLoader is what is causing the issue. Apparently it wants the real filesystem path there.

Which we don't have … but when I was looking previously it only wants something "unique" in this case. So any ideas?

CarsonF:
Maybe so when you were looking but then Nicolas made that commit. I would remove that parameter for now.

PR #6176, opened 2016-12-16
Extension Nut command registration

GawainLynch:
Recent changes to loading mean that Nut commands registered by extensions will never be registered … Same story at #6157

@CarsonF Yes, I know … Let's get it properly on v4 refactor?


CarsonF:
Ugh that change is having lots of effects. I don't know how to make it better though even when we get to v4.

PR #6184, opened 2016-12-18
Don't handle onclick in HTML, but rather in JS, when we're sure dependencies have loaded.

### LINKED ISSUE #6172: jQuery problems in exception render
Under certain circum stances, exceptions can be thrown **before** jQuery is sent, leaving this in the console:
```
Uncaught ReferenceError: $ is not defined
    at HTMLAnchorElement.onclick (my-page:49)
```

…and the exception trace unusable.

Is it possible to not use a library for this?

bobdenotter:
Fixes #6172


GawainLynch:
JavaScrap … :boom:

PR #6188, opened 2016-12-19
Add scoping for new storage getContent queries

rossriley:
This PR allows us to scope queries that are fetched via `getContent` queries. Primarily this will be used to allow queries to be tweaked depending on context. For instance when called via the `{% setcontent %}` Twig tag, we default to records where status is published, and we order by the field/direction defined in `contenttypes.yml`.

For now this leaves the existing `getContent` to be called without any scoping, but adds a method to the query object to fetch with a scope. So you can call:

```
$records = $app['query']->getContentByScope('frontend', '(entries,pages)');
```

This PR contains the frontend scope implementation, others may be added as needed.


GawainLynch:
Sorry mate, this got lost in EoY fun.

Issue #6190, opened 2016-12-20
[BUG/RFC] Slug generation with multiple fields doesn't go by array order


Maelstromeous:
Upon reading the documentation in https://docs.bolt.cm/3.2/fields/slug, it states you can use multiple fields to generate your slug.

One would expect that if you have the value: `uses: [reg, title]` that it would generate a slug using first the `reg` content then the `title`.

However, it appears that if you have `title` listed before `reg` in your `contenttypes.yml`, it uses that order (so title then reg) to generate the slug, rather than the defined order in the config.

Affects version: 3.2^
PHP Version: 5.6

PR #6191, opened 2016-12-20
Slug generation with multiple fields ordering

DennisSnijder:
See #6190
Fixes: #6190


bobdenotter:
Works like a charm! 👍

PR #6209, opened 2016-12-30
Fix HTTPS detection for canonical url

SahAssar:
Also don't include canonical in cached configs if it isn't set, since it can vary per request.

If one has host requirements in the routing.yml one doesn't want to set a canonical for the whole site (because it needs to be autodetected from the request since there are multiple valid hostnames for one site).

This fixes so that:

1. The canonical is only cached if it is set in the config.yml
2. HTTPS detection happens before the canonical is built, so that HTTPS detection for the canonical works.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -1175,7 +1175,7 @@ class Config
                 ],
             ],
             'liveeditor'                  => true,
-            'canonical'                   => !empty($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : '',
+            'canonical'                   => null,
             'cookies_use_remoteaddr'      => true,
             'cookies_use_browseragent'    => false,
             'cookies_use_httphost'        => true,
--- a/tests/phpunit/unit/Configuration/ResourceManagerTest.php
+++ b/tests/phpunit/unit/Configuration/ResourceManagerTest.php
@@ -176,10 +176,10 @@ class ResourceManagerTest extends \PHPUnit_Framework_TestCase
             )
         );
         $app = new Application(['resources' => $config]);
-        $this->assertEquals('cli', $config->getRequest('protocol'));
+        $this->assertEquals('http', $config->getRequest('protocol'));
         $this->assertEquals('bolt.dev', $config->getRequest('hostname'));
         $this->assertEquals('http://bolt.dev/bolt', $config->getUrl('canonical'));
-        $this->assertEquals('cli://bolt.dev', $config->getUrl('host'));
+        $this->assertEquals('http://bolt.dev', $config->getUrl('host'));
         $this->assertEquals('http://bolt.dev/', $config->getUrl('rooturl'));
     }
 
--- a/src/Configuration/ResourceManager.php
+++ b/src/Configuration/ResourceManager.php
@@ -333,43 +333,42 @@ class ResourceManager
         // This is where we set the canonical. Note: The protocol (scheme) defaults to 'http',
         // and the path is discarded, as it makes no sense in this context: Bolt always
         // determines the path for a page / record. This is not the canonical's job.
-        $canonical = $app['config']->get('general/canonical', '');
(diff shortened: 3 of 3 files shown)


GawainLynch:
Flagging that this will need a careful merge as all this changes in 3.3.

@CarsonF do you want to handle this merge, and the branch merge into 3.3 as you're closer to the ground on related lines?

CarsonF:
Yeah I would only change what's absolutely necessary here as it is all deprecated in 3.3

SahAssar:
I've updated some tests to expect cli:// for autogenerated canonicals since the tests are run via CLI.

tests/phpunit/unit/Configuration/ResourceManagerTest.php
```diff
@@ -178,9 +178,9 @@ public function testDefaultRequest()
         $app = new Application(['resources' => $config]);
         $this->assertEquals('cli', $config->getRequest('protocol'));
         $this->assertEquals('bolt.dev', $config->getRequest('hostname'));
-        $this->assertEquals('http://bolt.dev/bolt', $config->getUrl('canonical'));
+        $this->assertEquals('cli://bolt.dev/bolt', $config->getUrl('canonical'));
         $this->assertEquals('cli://bolt.dev', $config->getUrl('host'));
-        $this->assertEquals('http://bolt.dev/', $config->getUrl('rooturl'));
+        $this->assertEquals('cli://bolt.dev/', $config->getUrl('rooturl'));
```
GawainLynch: Won't these break tests in 3.3+?

SahAssar: No, not when I tested it on 3.3 but I'm gettings some differing results in https://github.com/bolt/bolt/pull/6209/files#diff-6eeac24760013e79aaf552a1ccb05ad9R236 (test.dev vs bolt.dev). I'll have a look.

CarsonF: Not from my changes. Only `canonical` or `canonicalurl` is changed to point to `Canoincal` class while inside of request cycle. The whole class is deprecated anyways.

SahAssar: @CarsonF: Yeah, my thing is basically just getting the tests to behave predictably and understanding what I misunderstood, I won't be putting that much time into it since it'll be removed soon anyway.

SahAssar:
This should be GTG now.

SahAssar:
https:// from canonical should now be fixed.

Doing cli:// urls came from a discussion with bob about how to handle the protocols when in cli, right around here: http://boltcms.slackarchive.io/general/-/1475223747.005477/1483360272.0032/1483360272003200/

Basically my thought was that if we treat cli as a protocol for `$request->getRequest('protocol')` shouldn't we do the same for other purposes too?

CarsonF:
Well I don't really agree with that. It's non standard and it would be inconsistent with urls everywhere else in the app. Also `$resources->getRequest('protocol')` isn't used anywhere.

SahAssar:
Alright, so basically we should do away with cli as a "protocol" altogether and default to http, right? I'm just looking to get some consistency in this, so cli or no cli does not really matter to me as long as we pick one.

CarsonF:
Yes. Checking if CLI is really easy, there's no need to put it into urls. I want consistency as well, and this is the odd ball.

src/Config.php
```diff
@@ -1175,7 +1175,6 @@ protected function getDefaults()
                 ],
             ],
             'liveeditor'                  => true,
-            'canonical'                   => !empty($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : '',
```
CarsonF: Mind if we leave this key and set value to `null`? That's what I have it as in 3.3. Just so we can keep track of keys in one spot.

GawainLynch:
:ship:

PR #6211, opened 2016-12-30
[Deprecations] Have {{ setcontent }} use FQCN of Bolt's Twig extension

GawainLynch:
Just a few deprecated calls …
```
Referencing the "Bolt" extension by its name (defined by getName()) is 
deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully 
Qualified Extension Class Name instead: 472x
    472x in Application::run from Codeception
```


GawainLynch:
This gets done a different way in #6166 … so if we're good with that, let's close this one.

PR #6214, opened 2017-01-01
Rouge Logic 1

GawainLynch:
This introduces Phase I of Symfony Forms use for the back-end in 3.4.

The end goal is to remove logic from controllers, templates, and other services, that is better handled by configurable upstream components.

* Remove form logic from `Bolt\Controller\Backend\Users` :tada: :balloon: :cake: 
* Custom validation constraints for:
  * Entity does not match a persisted entity type, based on one or more fields (lazy loaded)
  * Matching field values, e.g. user name !== password
  * Field value is contained inside another field value, e.g. password is not a part of a display name
* `UserRoleType` that handles disabling of roles that a user shouldn't modify
* Themed form widgets where needed

This is, in part, brought forward by keeping #5496 moving — which I have a  clean up branch for — as this would be a nice feature to cap-off the start of _this_ work.

Follow-up work:
* Login form & controllers
* Content record editor :gun: (lots of work)


bobdenotter:
Coo! 👍

GawainLynch:
I thought it worth adding this here for history …

> * Custom validation constraints for:
>    * Entity does not match a persisted entity type, based on one or more fields (lazy loaded)
>    * Matching field values, e.g. user name !== password
>    * Field value is contained inside another field value, e.g. password is not a part of a display name

### `\Bolt\Form\Validator\Constraints\ExistingEntity`

```
    $options = [
        'className'  => string,  // FQCN of an entity class known to Bolt's EntityManager
        'fieldNames' => array,  // Array of one of more field names to OR in the SELECT
    ];
```
```
    $constraint = new UsersAssert\ExistingEntity([
        'className'  => Entity\Users::class,
        'fieldNames' => ['field_name_1', 'field_name_2'],
    ]);
```

### `\Bolt\Form\Validator\Constraints\FieldEquals`

```
    $options = [
        'field'       => string,  // Other form field name to compare
        'insensitive' => boolean, // Do a case insensitive comparison
        'loose'       => boolean, // Do a loose equivalence comparison
    ];
```
```
    $constraint = new UsersAssert\FieldEquals([
        'field'       => 'email',
        'insensitive' => true,
    ]);
```

### `\Bolt\Form\Validator\Constraints\FieldContains`

```
    $options = [
        'field'       => string,  // Other form field name to compare
        'insensitive' => boolean, // Do a case insensitive comparison
    ];
```
```
    $constraint = new UsersAssert\FieldContains([
         'field'       => 'field_name',
         'insensitive' => true,
    ]);
```

bobdenotter:
Gawain says this is GTG, and who am I to argue.

GawainLynch:
Thank you very very much for all your help and guidance on this @CarsonF !

Issue #5307, opened 2016-05-12
[RFC] Repeater fields: Allow for multiple subtypes / blocks


psren:
The new repeater type is great.
You can repeat 1 to n elements of the same type.

It would be cool if you could select between different possible types there.
## Details
- Relevant Bolt Version: [ 3.1 | 4.0 | master ]
## Intention

Allow for something "pagebuilder-like"
## Idea

``` yaml
      repeater:
            type: repeater
            group: repeater
            limit: 3
            prefix: "<p>This allows you to create multiple sets of fields. Use the add button at the bottom to create a new empty set.</p>"
            fields:
                repeatabletexttype:
                   repeattitle:
                       type: text
                   repeatimage:
                       type: image
                       extensions: [ gif, jpg, png ]
                   repeatcontent:
                       type: html
               repeatableothertype:
                   repeattitle:
                       type: text
```

Implementation in Sulu CMF looks like that.

![bildschirmfoto 2016-05-12 um 13 25 28](https://cloud.githubusercontent.com/assets/1016798/15213182/d4570854-1845-11e6-9652-0fba3a3fc946.png)
(Like in Bolt 3.0)

![bildschirmfoto 2016-05-12 um 13 25 31](https://cloud.githubusercontent.com/assets/1016798/15213187/e13b724e-1845-11e6-8781-c3a6b650c8ea.png)
One can add multiple types in custom order. This allows for a "pagebuilder" experience with custom elements.

![bildschirmfoto 2016-05-12 um 13 26 00](https://cloud.githubusercontent.com/assets/1016798/15213198/fc050f36-1845-11e6-8dca-f4c17639dff7.png)
The type of a element can be changed, too.

Pinpickle:
This was in the original RFC for the repeating spec, and was known as "block" fields:

https://github.com/bolt/bolt/issues/3267

psren:
And has it been implemented? I couldn't find it anywhere in the Docs or relase notes

bobdenotter:
psren. Not yet, We're soon going to solidify the plans for 3.1 and onwards. For now, it's just the repeaters as-is, and the templatefields that are at your disposal.

psren:
Ah okay. Should i leave this open since the others RFC is closed?

SahAssar:
psren: yep, this should probably be kept open.

dantleech:
what would be nice here (and something that I wanted to do in Sulu) is to have just references to types which are defined elsewhere:

```yaml
repeater:
[...]

PR #6215, opened 2017-01-01
New Feature: Named Repeater Blocks

rossriley:
Happy New Year! and with it the arrival of named repeater blocks.

This has a few outstanding bugs and code improvement still to be completed along with test additions / fixes so will remain WIP for a while.

### In action

First of all, here's a mini screen recording of the feature in action. The concept builds on top of the existing repeater UI but allows adding multiple blocks by type.

Named Repeater Demo

Full length screen-record: http://recordit.co/0LYTbypUGr

### How to use

There's a new field-type `block` that takes an array of block / field configurations. Here's the configuration used in the demo above.

```yaml
        contentblocks:
            type: block
            label: Content Blocks
            group: Content
            fields:
                imagesection:
                    label: Image Section
                    fields:
                        title:
                            type: text
                        teaser:
                            type: html
                            height: 30px
                        image:
                            type: image
                twocolumn:
                    label: Two Column Block
                    fields:
                        title:
                            type: text
                        columnone:
                            type: html
                            label: Column One Content
                            height: 50px
                        columntwo:
                            type: html
                            label: Column Two Content
                            height: 50px
```

### Displaying in the template

At the template level these sets are almost identical to the existing repeater blocks, there's an extra attribute value available on the collection and on each field called `block` which tells you the name of the block, so we can do this in the template:

```twig
    {% for group in record.contentblocks %}
        {% if group.block == 'twocolumn' %}
        ...... render the twocolumn template...
        {% endif %}
        {% if group.block == 'imagesection' %}
        ...... render the imagesection template...
        {% endif %}
    {% endfor %}
```

Or perhaps even more succinctly by having a partial template for each defined block:

```twig
    {% for group in record.contentblocks %}
        {% include '_koala-block-' ~ group.block ~ '.twig' %} 
    {% endfor %}
```

### Feedback / Improvements

All gratefully received. It was mentioned in the original RFC(#5307) that we also support re-use of these block configurations, we can definitely address this in a follow-up PR and fine-tune the syntax too.


bobdenotter:
Very cool stuff! 👍

src/Storage/Mapping/MetadataDriver.php
```diff
@@ -264,19 +264,15 @@ public function setRepeaters($contentKey, $className, $inputData = null)
...
-
-                    if (isset($this->typemap[$value['type']])) {
-                        $value['fieldtype'] = $this->typemap[$value['type']];
-                    } else {
-                        $value['fieldtype'] = $this->typemap['text'];
+            if (in_array($data['type'], ['repeater','block'])) {
+                if ($data['type'] == 'repeater') {
+                    $this->normalizeFieldTypes($data['fields']);
+                }
+                if ($data['type'] == 'block') {
```
GawainLynch: Can this and the one above be strict compared?

GawainLynch:
Also, is it possible to squash some of the "attempts", and the asset builds out (merge conflicts are too easy)

GawainLynch:
Awesomesauce :+1:

eduardomart:
Looking at the latest alpha version, I was trying to do some testing but it seems Named Repeaters are broken, see here:

bryan-d:
> there's an extra attribute value available on the collection and on each field called block which tells you the name of the block, so we can do this in the template

Seems to no longer be the case in the latest alpha version. See here:
image

l0rn:
i also get null if i iterate like this:

```
    {% for group in record.contentblocks %}
        {{ dump(group.block) }}
    {% endfor %}

```

l0rn:
I did some digging, found out that in FieldCollection this function exists:

```
    /**
     * @return string
     */
    public function getBlock()
    {
        return $this->first()->getBlock();
    }
```

This function is neither in the FieldCollectionInterface nor implemented in the LazyFieldCollection ....

Since the renderer in 3.4 now uses LazyFieldCollections getBlock is not available. I would make a pull request but i think i have not enough overview of bolt to know the implications.

l0rn:
I guess a repeater inside a named repeater / block doesn't work.... It leads to funny admin ui crashes.... Can't figure out what the problem is so far

SahAssar:
@l0rn Repeaters inside repeaters are not implemented, and probably won't be, we should probably throw a warning if configured as such though.

bobdenotter:
I vote that instead of an exception, we just display this gif: 

giphy

l0rn:
I guessed so... since named repeaters / blocks are repeaters that passage about nesting applies to them..... Still I would think being able to nest those would have a lot of value in terms of editing flexibility... i already looked at the data model of bolt and see the core issue, since content is stored in a JSONField ... An Implementation therefore wouldn't be that easy i guess

SahAssar:
Repeater values are stored in the bolt_field_value table, not as json (unless they are part of templatefields, which is always stored as json). I don't think the real issue is storage though (and it'd be even easier if it was actually just dumped in a json column), it was just more or less agreed that nesting fields in that manner will lead to a lot of headaches and would get silly pretty quickly.

PR #6220, opened 2017-01-02
[3.4] [Updated] Users invitation

### LINKED ISSUE #4555: [Feature] Create users via invitation
It would be nice to be able to create new users via an invitation vs having to input all their info yourself (including a password). The invitation could be typing their email and having bolt email them a link, and/or just an a authorized link generated and displayed on backend that can be copy/pasted to them.

GawainLynch:
Closes #5496
Fixes #4555

This is a reworking of the PR done by @mangroud in #5496
This is **not ready** for merge yet ~~, as I'd like to squash my commits … when we decide collectively on approach … just want to (mostly) get this off my plate.~~

A UX addition in this one too:
screenshot from 2017-01-02 18-21-39


CarsonF:
What do you think about prefilling (possibly forcing) a username / email address / even display name with the invitation? Really the only thing I don't want to fill in for my users is their password. I'm not saying this should be forced from the admins POV.

GawainLynch:
> What do you think about prefilling (possibly forcing) a username / email address / even display name with the invitation? Really the only thing I don't want to fill in for my users is their password. I'm not saying this should be forced from the admins POV.

Can we? My understanding was that the intention was to make this an admin feature … that said, yes I think we should either enforce it as required, or make the fields validate.

That said, one of my massive concerns about this from the beginning was the security implications, as anyone with a valid link could (until said link is used) can create an account… so TBH, there should be some additional stop-gaps to attempt to curb abuse.

CarsonF:
Yes it is. I guess my use case is basically the same as create user and then send them a reset password link...

GawainLynch:
> Yes it is.

Derp, by "can we" I meant "is it implemented for non-admin users?"

> I guess my use case is basically the same as create user and then send them a reset password link...

This is pretty much where I am going. While it is a small vector, it is one that I can see potentially ending up in a CVE if we're not careful.

bobdenotter:
Does this need anything else than monkey-testing?

GawainLynch:
Yes, @CarsonF is working on more changes currently

GawainLynch:
@CarsonF has a copy of this branch … I am withdrawing my work on it, if it gts over the line, it won't be one by me

zomars:
Is this feature dead? I got lost between issues and PRs.

GawainLynch:
Not dead, just lacking time to finish. :disappointed:

PR #6223, opened 2017-01-03
dot dev is an actual TLD, so use dot test for testing instead

SahAssar:
Since .dev is an actual tld we should use a tld reserved for testing, like .test


GawainLynch:
:+1:

PR #6227, opened 2017-01-06
Handling of fragment in URL generator

GawainLynch:
* Fix fragments that have '/' & '?'
* Backport handling of fragment in URL generator @see https://github.com/symfony/symfony/pull/12979/commits/6d79a565e0e56b525134080687eba3961eab1a73

### DIFF
--- a/src/Routing/UrlGeneratorFragmentWrapper.php
+++ b/src/Routing/UrlGeneratorFragmentWrapper.php
@@ -33,16 +33,17 @@ class UrlGeneratorFragmentWrapper implements UrlGeneratorInterface, Configurable
      */
     public function generate($name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
     {
-        $fragment = null;
-        if (isset($parameters['#'])) {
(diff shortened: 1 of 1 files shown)


src/Routing/UrlGeneratorFragmentWrapper.php
```diff
@@ -33,16 +33,17 @@ public function __construct(UrlGeneratorInterface $wrapped)
      */
     public function generate($name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
     {
-        $fragment = null;
-        if (isset($parameters['#'])) {
+        $fragment = isset($parameters['_fragment']) ? $parameters['_fragment'] : null;
+        unset($parameters['_fragment']);
+        if ($fragment === null && isset($parameters['#'])) {
```
CarsonF: We still want to unset `#` key if it exists. Although I guess both of them are not going to be set at once.

GawainLynch: Moved the `unset()` down — as it's harmless regardless — to avoid the need for more condition(al)s

bobdenotter:
Looks good to me! :shipit:

PR #6230, opened 2017-01-07
Twig cleanup

CarsonF:


GawainLynch:
Epic!

src/Storage/Entity/ContentValuesTrait.php
```diff
@@ -646,9 +646,13 @@ protected function getTemplateFieldsContentType()
...
-            $template = $this->app['templatechooser']->record($this);
-            if (array_key_exists($template, $templateFieldsConfig)) {
-                return $templateFieldsConfig[$template];
+            $templateChooser = $this->app['templatechooser'];
+            /** @var \Bolt\TemplateChooser $template */
+            $templates = $templateChooser->record($this);
+            foreach ($templates as $template) {
+                if (array_key_exists($template, $templateFieldsConfig)) {
+                    return $templateFieldsConfig[$template];
+                }
```
CarsonF: Should this use twig to resolve the template like I did with `TemplateFieldsType`?

GawainLynch: Would be nice.

CarsonF: But....?

GawainLynch: No buts … 
make-it-so

CarsonF: #6237

PR #6231, opened 2017-01-08
[3.3][RFC] Adding security service provider

GawainLynch:
If agreed upon, this PR will add the Silex provider, and a provider of our own for Symfony Security.

This changes nothing about Bolt's current route security behaviour, but makes ` $app['security.firewalls']` and ` $app['security.access_rules']` shareable services that we know will be there from 3.3.0 onward (_read_: extensions)

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -210,6 +210,7 @@ class Application extends Silex\Application
             ->register(new Silex\Provider\ValidatorServiceProvider())
             ->register(new Provider\RoutingServiceProvider())
             ->register(new Silex\Provider\ServiceControllerServiceProvider()) // must be after Routing
+            ->register(new Provider\SecurityServiceProvider())
             ->register(new Provider\RandomGeneratorServiceProvider())
             ->register(new Provider\PermissionsServiceProvider())
             ->register(new Provider\StorageServiceProvider())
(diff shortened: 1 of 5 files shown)


CarsonF:
This enables the firewall and would probably conflict with changes to `permissions.yml`

GawainLynch:
> This enables the firewall and would probably conflict with changes to permissions.yml

`'security' => false,` … and monkey & acceptance tests pass

CarsonF:
Ah. Ok so what's the benefit of this? Seems like extensions can't do anything with this until we actually use it. Just drinking ☕️ sorry if it's obv.

GawainLynch:
> Ah. Ok so what's the benefit of this

Extensions could add their own firewalls to their routes, I may know one that is heading in that direction :wink:

CarsonF:
But you just said the firewall wasn't checked

GawainLynch:
c.f. [Silex docs](http://silex.sensiolabs.org/doc/2.0/providers/security.html#securing-a-path-with-http-authentication)

GawainLynch:
Caught the test problem :wink:

bobdenotter:
🚂  🚋

PR #6237, opened 2017-01-08
Resolve template for template fields

CarsonF:
Clean up from from #6230


GawainLynch:
Local pass and branch merges coming up

PR #6238, opened 2017-01-08
Fix infinite loop from 3.2 merge

CarsonF:


GawainLynch:
Passing locally … and need.

PR #6241, opened 2017-01-09
Fix 'Notice: Undefined index: persistant' when using redis for sessions

AlphaRecon19:
When using redis for sessions with errors enabled, Bolt throws the following exception:

error

Easy fix to get it working again. This also should enable persistent connections to redis again as it would just default to `connect`


GawainLynch:
Matches [vendor docs](https://github.com/phpredis/phpredis#php-session-handler) and that [classes use](https://github.com/bolt/bolt/blob/release/3.2/src/Provider/SessionServiceProvider.php#L418) :+1: 

Nice find, thanks mate

PR #6245, opened 2017-01-12
Typo in default value

GawainLynch:


bobdenotter:
What if whoever made that actually intended it to be 61 minutes?

GawainLynch:
I just asked, apparently not :stuck_out_tongue_winking_eye:

Issue #6248, opened 2017-01-12
[BUG] Vague exception when you've got an invalid contenttypes.yml


Maelstromeous:
By accident, I had an invalid config for `contenttypes.yml`.

However, the exception that was thrown was extremely vague, which basically gave me no information as to what I did wrong.

```
Twig_Error_Runtime in Template.php line 556:
Impossible to access an attribute ("all") on a null variable in "@bolt/exception/_request.twig" at line 46.
```

Used contenttype (notice the `type: image` is missing)

```
shields:
    name: Shields
    singular_name: Shield
    fields:
        title:
            type: text
        image:
            upload: shields
```

![exception](https://cloud.githubusercontent.com/assets/1776058/21895332/361d5c28-d8da-11e6-804b-56e4b723ab43.png)

Details
-------

 - Relevant Bolt Version: 3.2.5
 - Install type: Composer
 - PHP version: 5.6
 - Used web server: Apache

Reproduction
------------

1. Create an invalid contenttype
2. Attempt to load a page
3. Observe exception

bobdenotter:
On it!

mborn319:
I can confirm, this was a pain. Thanks for the fix @bobdenotter !

PR #6249, opened 2017-01-12
Adding some fallbacks

bobdenotter:
Alleviates the quirk with #6248, but doesn't actually solve it properly. 

The underlying issue is that a thrown exception here doesn't get caught.

https://github.com/bolt/bolt/blob/release/3.2/src/Config.php#L651

### DIFF
--- a/app/view/twig/exception/_trace.twig
+++ b/app/view/twig/exception/_trace.twig
@@ -1,7 +1,7 @@
 <h2>Stack trace</h2>
 
 <table class="table table-striped table-hover table-condensed">
-{% for t in exception.trace %}
+{% for t in exception.trace|default() %}
     <tr>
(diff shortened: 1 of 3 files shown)


GawainLynch:
> The underlying issue is that a thrown exception here doesn't get caught.

It gets [caught here](https://github.com/bolt/bolt/blob/release/3.2/src/Config.php#L454-L459). 

The underlying issue is that the [exception message](https://github.com/bolt/bolt/blob/release/3.2/src/Config.php#L458) gets saved (ugly hack) and then fetched [here](https://github.com/bolt/bolt/blob/release/3.3/src/Configuration/Validation/Configuration.php#L33-L38) passed to the exception controller. 

Matching PR in #6250

PR #6253, opened 2017-01-13
Don't set homepage template to `null`, if theme.yml is not present.

bobdenotter:


src/TemplateChooser.php
```diff
@@ -32,7 +32,9 @@ public function homepage($content)
         $templates = [];
 
         // First candidate: Theme-specific config.yml file.
-        $templates[] = $this->config->get('theme/homepage_template');
+        if ($this->config->get('theme/homepage_template')) {
```
CarsonF: Could save the second look up by saving the result here.

src/TemplateChooser.php
```diff
@@ -32,7 +32,9 @@ public function homepage($content)
...
+            $templates[] = $this->config->get('theme/homepage_template');
+        }
 
         // Second candidate: Global config.yml file.
         $templates[] = $this->config->get('general/homepage_template');
```
CarsonF: Will this one always exist?

bobdenotter: It should, unless somebody axplicitly deletes it from their config.yml

src/TemplateChooser.php
bobdenotter: I know, but I made it in line with the dozen or so other similar statements in that file. If we change this one, we should change all of them to keep it consistent.

PR #6258, opened 2017-01-14
Move Nut Import/Export to new Storage Engine and support multiple migrations

rossriley:
## What's New

The commands have been updated to run off the new storage engine and support has been added for both export and re-import of all fields:
  
  * Repeaters
  * Templatefields
  * Taxonomies
  * Relations

Additionally a new option has been added to both the Import and export command that adds data directory support, it's used like this: 

    ./vendor/bin/nut database:export -d app/data

    ./vendor/bin/nut database:import -d app/data

When exporting to a directory a sequential migration file will be generated using a timestamp, eg: `20170319104854.yml`.

Conversely when importing from a directory the migration files will be ran in sequential order.

Finally the import command introduces the overwrite flag which allows content in newer files to overwrite previously saved records, this is activated via:

     ./vendor/bin/nut database:import -d app/data --overwrite


JarJak:
Is there an ability to also make migrations with data and to add some logic into it (maybe not in yml format, but php?) as doctrine/migrations do?

PR #6259, opened 2017-01-14
Fixes of Type Hinting in Bolt\Storage\Query namespace

gander:


src/Storage/Query/ContentQueryParser.php
```diff
@@ -102,7 +103,7 @@ public function setQuery($query)
      *
      * @param array $params
      */
-    public function setParameters($params)
+    public function setParameters(array $params)
```
GawainLynch: @rossriley Do we want to be changing type hints on a stable branch?

src/Storage/Query/Handler/SelectQueryHandler.php
```diff
@@ -15,7 +16,7 @@ class SelectQueryHandler
     /**
      * @param ContentQueryParser $contentQuery
      *
-     * @return QueryResultset
+     * @return QueryResultset|Content|null
```
GawainLynch: That is not correct.

src/Storage/Query/ContentQueryParser.php
gander: `$params` is array type, require the correct data type

src/Storage/Query/Handler/SelectQueryHandler.php
gander: it's return `QueryResultset` or `Content`  or `null`. Find better, but don't set one type. Maybe mixed?

GawainLynch: Do you have test cases or is this from the IDE?

gander: 1) dump output for certain queries
2) its used for queries so should return set or content or null
3) look at line 51 & 53

GawainLynch:
@gander This was discussed at tonight's developer meeting, and we're happy to run with it. Only caveat being we'll merge it to 3.4 (probably in the morning), so as always :+1:

GawainLynch:
Merged into `release/3.4` via d7388ce795b3525edde16e3def4093ccd6cd1fdf

@gander as you submitted this from a release branch, you will need to hard reset that branch. If you're stuck with that, call out on IRC and I'll happily help :smile:

PR #6266, opened 2017-01-17
Set a default on user display name

GawainLynch:
Fixes occasional "single request crashes" in the back-end with debug enabled.

screenshot_from_2017-01-16_16-01-09

When the session hits the check timer, the user object can be empty and the nav render still gets called during that redirect. 

We'll address properly with the DTO/REST work.


GawainLynch:
Ping @madc :smile

PR #6275, opened 2017-01-19
Allow to use environment variable to change configuration settings

kouz75:
We use bolt on several environment (dev/test/production).
On each environment, we need to change some settings (DB,debug,API key etc).
using _local.yml allow to do the job but the files aren't save on git and we cannot share them. 

Our solution is to create a environnement variable to define the current plateform and use it to  read  setting files for the current environnement.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -321,14 +321,14 @@ class Config
     {
         // Read the config and merge it. (note: We use temp variables to prevent
         // "Only variables should be passed by reference")
-        $tempconfig = $this->parseConfigYaml('config.yml');
+        $tempConfig = $this->parseConfigYaml('config.yml');
         if (getenv('ENVIRONMENT')) {
-            $tempconfigenvironment = $this->parseConfigYaml('config_'.getenv('ENVIRONMENT').'.yml');
-            $tempconfig = Arr::mergeRecursiveDistinct($tempconfig, $tempconfigenvironment);
+            $tempConfigEnvironment = $this->parseConfigYaml('config_'.getenv('ENVIRONMENT').'.yml');
+            $tempConfig = Arr::mergeRecursiveDistinct($tempConfig, $tempConfigEnvironment);
         }
 
-        $tempconfiglocal = $this->parseConfigYaml('config_local.yml');
-        $general = Arr::mergeRecursiveDistinct($tempconfig, $tempconfiglocal);
+        $tempConfiGlocal = $this->parseConfigYaml('config_local.yml');
+        $general = Arr::mergeRecursiveDistinct($tempConfig, $tempConfiGlocal);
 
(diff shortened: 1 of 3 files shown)


src/Config.php
```diff
@@ -322,6 +322,12 @@ protected function parseGeneral()
         // Read the config and merge it. (note: We use temp variables to prevent
         // "Only variables should be passed by reference")
         $tempconfig = $this->parseConfigYaml('config.yml');
+        if(isset($_SERVER['ENVIRONMENT']))
```
GawainLynch: Super globals should never be used

src/Config.php
```diff
@@ -322,6 +322,12 @@ protected function parseGeneral()
...
+        {
+            $tempconfigenvironment = $this->parseConfigYaml('config_'.$_SERVER['ENVIRONMENT'].'.yml');
```
GawainLynch: PSR-2 on the variable names, e.g. `$tempConfigEnvironment`

src/Config.php
GawainLynch: The other problem I have with this approach is that it is putting config back into the request cycle to finish … @CarsonF any ideas on approach here?

GawainLynch:
As this is feature based, I am moving it to `release/3.4` as 3.2 is hotfix only, and 3.3 is in beta freeze.

rossriley:
This is a similar feature to #6208, not sure whether we need both approaches. 

If we do include this method too, then I'd also suggest that we use `getenv('environment')` rather than accessing the superglobal directly.

src/Config.php
kouz75: fix super globals

src/Config.php
kouz75: I have rewrite several variable name

CarsonF:
I think this is different enough from your other PR @rossriley. 

We at my work do something similar. We have sections in one config file and each environment can extend another so we don't have to duplicate common values. We use `PHP_ENV` though for our env key.

While I do think config_local.yml isn't good enough, it seems to work fine for everyone else. I'm wondering if this isn't something that should be customized for your own install. I'm pretty sure that logic will be more accessible after @gawainlynch's config refactor drops.

GawainLynch:
This was discussed at tonight's core [development meeting](https://github.com/bolt/bolt/wiki/Dev-meeting-2017-01-24) and we feel the fit as-is isn't correct. As noted above and in the meeting, similar functionality already exists, and a whole lot more will be possible after the config loading refactor lands in 3.4 next month, so this would only lock in BC requirements that are going to change very soon.

All that aside, thank you very much for the PR, all-in-all the idea is great, but the base of Bolt needs work to better do this in a supportable way :+1:

PR #6279, opened 2017-01-21
Fix extend page pre-release version display

GawainLynch:
That bug that takes you over a year to get to…

If someone wants to break up the UI a bit go for it, this at least (re)enables the functionality.

image


bobdenotter:
Does this re-introduce multiple "stable versions" in the top section? We made it show only one (the latest) on purpose.

GawainLynch:
> Does this re-introduce multiple "stable versions" in the top section? We made it show only one (the latest) on purpose.

Not at all … the screen shot is just *after* the "Show all" button was pressed otherwise you wouldn't see the pre-releases

bobdenotter:
Looks good! 👍

PR #6281, opened 2017-01-21
Updates to base-2016 for Bolt 3.2

bobdenotter:
- Get it updated with the 3.3 changes
 - Including the `partials/_sub_fields.twig` for `{{ fields() }}`.
 - Updating NPM / Bower shizzle
 - Conditionally include jQuery only.


theme/base-2016/partials/_footer.twig
```diff
@@ -17,6 +17,8 @@
 
 </footer>
 
-<script src="{{ asset('js/jquery.min.js', 'theme') }}"></script>
+{# Extensions and copypasta gleaned from other sources tend to cause multiple jQuery
+   includes. We only include it here conditionally, if not done so already. #}
+<script>window.jQuery || document.write("<script src='{{ asset('js/jquery.min.js', 'theme') }}'><\/script>");</script>
```
cdowdy: So you know. Google is going to start to block document write to inject scripts in chrome for slow connections, non asyn /  defer and a few more here https://developers.google.com/web/updates/2016/08/removing-document-write

Essentially jQuery will never load in chrome 55+ like this

bobdenotter: Interesting read, but i am not convinced it will block: 

> The script is not hosted on the same site. In other words, Chrome will not intervene for scripts with a matching eTLD+1 (e.g. a script hosted on js.example.org inserted on www.example.org).

cdowdy: Well it also says

>The document.write() is in a top level document. The intervention does not apply to document.written scripts within iframes

Which this is a top level doc

>the script in the document.write() is parser-blocking. Scripts with the 'async' or 'defer' attributes will still execute.

Which this is a parser blocking script


Also all that is superceded when a user is on a slow connection (and eventually slow wifi) it'll be removed and not execute


I just wanted to let you know thats all :)

bobdenotter: @cdowdy I am happy you're letting us know, and i'll happily consider alternatives.

I'm just not convinced that this is a problem yet. It says _all_ criteria need to be met:

> Specifically Chrome will not execute the <script> elements injected via document.write() when **all** of the following conditions are met:

If you have an alternative solution to this problem, i'm all ears.

bobdenotter:
@cdowdy By the way, I picked up this trick from Html5boilerplate, and they have an open issue about the same topic: 

https://github.com/h5bp/html5-boilerplate/issues/1896

theme/base-2016/partials/_footer.twig
SahAssar: Why not just inject a DOM node instead of writing HTML? that way we follow best practices and don't inspire others to write it "the wrong way".

bobdenotter:
I reverted the "conditional loading of jQuery". Should be GTG now, if tests pass.

cdowdy:
@bob yeah this is a pain IMO. I just wanted you all to be aware and since this is Bolt's theme that ships and anyone who would use this without modifications could wonder why this is breaking if it does break. In those instances people who don't modify the theme probably have no prior knowledge of speculative pre-parser, same origin, cross origin, that script and link tags block parsing, what async defer is etc haha.  

There is also some instances in crbug lists of people on cable LAN connections being blocked. 

The way to fix this is in my opinion is:

1. Leave it with document.write since there isn't a consensus currently on how to "fallback"
2. use a script loader for all js
3. make all assets (from extensions and themes) go through an asset component and combine them.

The third one you all have already decided is a no go since you want to keep bolt light.

The second one is more manageable. For instance load jquery with filament groups [loadJS](https://github.com/filamentgroup/loadJS)  

```twig  
<script>if (!window.jquery) {
  // include loadJS here...
  function loadJS( src ){ ... }
 loadJs({{ asset('js/jquery.min.js', 'theme') }} );
}
</script>
```

or you could use twig if/else conditionals which could be ugly 

Leaving it is in my opinion the last resort since we're hoisting bad practices into a default theme.

bobdenotter:
@cdowdy I've reverted the change for now. Too big of a risk to make potential breaking changes in an otherwise stable version. Let's revisit this for the new base-2017 theme, and get it right in that one. :-)

Using a simple script loader seems like a nice idea, though. 👍

SahAssar:
Btw, if we are going to go with a loader (which it looks like we aren't for now) something minimal but still compatible with current standards (commonJS) would be my go-to, like http://stuk.github.io/require1k/ or similar.

bobdenotter:
@SahAssar Looks interesting too! But, at first glance it seems like loadJS isn't much bigger. 

I am leaning towards _not_ using a loader in the default theme though. It's complicated enough for novices, as it is. :-)

SahAssar:
@bobdenotter Yeah, the point of require1k is more that it is mostly, sorta compatible with the commonjs require style of imports, rather than a loader-specific way of doing things (like a `loadJs` function).

Anyway, not using a loader seems like the better path :)

PR #6283, opened 2017-01-21
Move the getRepeaters call into the nohydrate block

rossriley:
Whoops, looks like I've been responsible for wasting a lot of internet electricity.

@graham73may watch your query count go down after this one.

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -1614,12 +1614,11 @@ class Storage
             $objects[$row['id']] = $this->getContentObject($contenttype, $row);
         }
 
-        $this->getRepeaters($objects);
-
         if ($getTaxoAndRel) {
             // Make sure all content has their taxonomies and relations
             $this->getTaxonomy($objects);
             $this->getRelation($objects);
+            $this->getRepeaters($objects);
         }
 
         return $objects;


graham73may:
I'll definitely give this a go and compare! 

Whilst digging the other day we noticed that some things like menu items were being hydrated. I think it was firing off more calls as well. Is it required when they're only returning basic info?

rossriley:
Yes, that was what I found when debugging this. In the backend at least the culprit was the permission check which for some reason seems to fire off a check for each record in the contenttype, not just the record that you are editing.

So this PR alone dropped 434 queries to 20 for me.

But thanks for designing that hideous contenttype 😄  because it's helped me find where the bodies are buried.

graham73may:
Ah great! I'll send you my new query numbers when I merge this in, should be interesting. :-p

GawainLynch:
> Whoops, looks like I've been responsible for wasting a lot of internet electricity.

You sure it was you? I could have sworn I just saw :koala: paw prints!

> the culprit was the permission check which for some reason seems to fire off a check for each record in the contenttype

Yes, I can't remember *exactly* when I attacked that last (2.2-dev or 3.0-dev), but I removed ~98-99% of those query calls (a few have since been reintroduced :frowning_face:). The persisting problem being how much business logic is still in our BE templates.

Part of this will go away with Symfony Security in v4.0, part of it will get further cleaned up with the work attached to forms & DTOs.

But that said, this is exactly why a couple of people have found me heavily disagreeing with enhancing "owner" properties at the moment.

graham73may:
So this is before (with the other PR to fix query times already merged):
image

This is after (so this PR + the other one):
image


Wow... and to think on Friday that was 160,000 ms!

PR #6286, opened 2017-01-23
Fix incorrect alias

rossriley:
The join aliased content_id to id but this is no longer the case in the subselect.

This updates the alias to point to content_id


rossriley:
Hold off on merge, just needs some more testing.

rossriley:
All seems good now.

PR #6287, opened 2017-01-23
Fix array_column for PHP 5.x

### LINKED ISSUE #6278: Remove array_column; doesn't work for PHP 5.5/5.6
In 68892bd285c9fc65558985d3586d472045029690 I started using `array_column`. However, this only works with objects in PHP 7.0+ 😭  So relations, taxonomies, and repeaters are broken for PHP 5.5 and 5.6.

CarsonF:
Fixes #6278

- Implemented `Arr::column`
- Deprecated `Arr::makeValuePairs`

### DIFF
--- a/tests/phpunit/unit/Helper/ArrTest.php
+++ b/tests/phpunit/unit/Helper/ArrTest.php
@@ -12,7 +12,7 @@ use Bolt\Tests\BoltUnitTest;
  */
 class ArrTest extends BoltUnitTest
 {
-    public function testMakeValuePairs()
+    public function testLegacyMakeValuePairs()
     {
         $test = [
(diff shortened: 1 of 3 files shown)


tests/phpunit/unit/Helper/ArrTest.php
```diff
@@ -133,4 +133,46 @@ public function testNonArraysAreNotIndexedOrAssociative()
...
+        $expected = [
+            'foo'   => 'bar',
+            'hello' => 'world',
+            '5'     => 'asdf',
+        ];
+        $this->assertEquals($expected, $result);
+    }
+}
+
+class TestColumn
```
CarsonF: Noe. You can't make me.

PR #6296, opened 2017-01-24
Fixed bug in Storage::saveContent() (slug)

### LINKED ISSUE #6295: [BUG] Storage::saveContent() slug is mandatory but that's wrong
Hey,

I'm using the boltforms plugin and save the form input to the content type. **The content type has no slug field**. If I save the record (by filling the form correctly), I don't see any error, but find an exception in the system logs
```
[Bolt Forms] An exception occurred saving submission to ContentType table `feedbacks` 
URI: /add-feedback
Source: /var/www/vhosts/bolt/vendor/bolt/bolt/src/Legacy/Storage.php :: 412 
```

I already know the fix I think but wanted you to see the issue first.

Details
-------

 - PHP version: 5.6
 - Used web server: Nginx

Reproduction
------------

Content type:
```
feedbacks:
    name: Feedbacks
    singular_name: Feedback
    fields:
        foo:
            type: integer
            group: main
        token:
            type: text
            group: security
    recordsperpage: 100
    searchable: false
    viewless: true
    default_status: held
```
Save the record with `Storage::saveContent()`.

sbani:
In `Storage:saveContent` the slug is used to create the uri, but
`$fieldValues['slug']` is not always present.
Added the check for missing slug to the class again because `getUri()`
is able to handle an empty slug.

I changed the simple if statement above because it's shorter and faster now. And I don't think it's worse readable. I can put it in an own commit, but there is no change at all. A revert would do no change there.

I can only find this commit e7516d643a1f879aa56c68f33a06e30291860513 which first introduced the **fix**, but I can't find the one which removes these required lines of code with blame.

Hope you can follow my explanations and thank you for reading and checking 👍 

Fixes #6295

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -401,10 +401,10 @@ class Storage
         }
 
         // Test to see if this is a new record, or an update
-        if (empty($fieldvalues['id'])) {
-            $create = true;
-        } else {
-            $create = false;
+        $create = empty($fieldvalues['id']);
+
+        if (!isset($fieldvalues['slug'])) {
+            $fieldvalues['slug'] = ''; // Prevent 'slug may not be NULL'
         }
 
         // We need to verify if the slug is unique. If not, we update it.


src/Legacy/Storage.php
```diff
@@ -401,10 +401,10 @@ public function saveContent(Content $content, $comment = null)
...
 
         // Test to see if this is a new record, or an update
-        if (empty($fieldvalues['id'])) {
-            $create = true;
-        } else {
-            $create = false;
+        $create = empty($fieldvalues['id']);
+
+        if (!isset($fieldvalues['slug'])) {
+            $fieldvalues['slug'] = ''; // Prevent 'slug may not be NULL'
```
GawainLynch: This will still fail under some circumstance.

sbani: Thank you @GawainLynch. What are these circumstances?
Do you have any tests in mind or do you want to give me a short introduction in slack?

I will help where I can

GawainLynch: Yep, and I love the work you do …

Well, it is more a "down the line" problem. The short version is that the tables expect slug to be "not null", and for all intensive purposes unique. We'd have set a unique flag in 3.0 but it would have potentially killed the BC path for some users.

The problem that I worry about (apart from touching legacy code) is that the slug value is one of two possible indexes we use to find record in some circumstances. That doesn't mean that this approach isn't an improvement of sorts, just that my worry is that it move the problem from one that can easily be identified, to an unknown fault down the line when another part of the code is using `empty()` or similar … don't start me on `empty()` being one of the greater design mistakes in PHP to date :laughing:

GawainLynch: …and apologies for the initial spare response, I was on my way out the door at the time

sbani: ah @GawainLynch. Definitely no problem. I know you have a lot to do and it's ok for me to wait.
Thank you for the kind words.

I know the slug should somehow be present and I'm not into the code at all anymore (maybe because of 🍺 ), but as far as I understand the slug is now required to save the record. At least in my case.

This means that there are two possibilities right now (if the user has no slug the content type):
1. create a slug (silently) in the background
2. update legacy code to work without slug.

When I interpret your answer correctly, only 1 can prevent BC breaks because 2 is too fragile.

All in all it sounds like a too big topic for this PR, isn't it?

Please let me know what I can contribute or how I can help.
Because a generated slug in the content type is the short fix for us at the moment and we are fine with that.

Thank you for your fast replies. 🥇

GawainLynch: > 1. create a slug (silently) in the background

Yes, I *think* this is the better way to go. You could just use the `getTitle()` helper to get the value of the first string field (or one called "title" if it exists), and Slugify that.

sbani: thanks. I will close this PR.
Do you think I should close the issue too?

GawainLynch: Oh, no you don't need to close either right now … this can can be adapted … the only thought is that not a lot should still use the legacy `saveContent()` … might be wroth hashing this over with Ross in Slack later

PR #6299, opened 2017-01-24
Modded core's select field to accommodate for multiple contenttypes v2

mrenigma:
Continuing on from #6270 and the discussion had with @SahAssar I've modified the original implementation of this slightly so instead of determining the return value based on the existence of `contenttype` it instead does it based on whether multiple contenttypes have been defined. This separates the left hand side of the values string from the right more distinctly and allows more refined choice for the developer.

```yml
featured_items:
    label: "Event/Event series"
    autocomplete: true
    type: select
    values: (event,eventseries)/title
```

Would create a select option with `<option value="event/1">Title</option>`.

While

```yml
featured_items:
    label: "Event/Event series"
    autocomplete: true
    type: select
    values: event/contenttype,title
```

Would create a select option with `<option value="1">Event / Title</option>`

I know that it's already been merged in as is - happy for this to be closed without being pulled if everyone is more happy with the first implementation. I'm also always looking for feedback :smile:

### DIFF
--- a/app/view/twig/editcontent/fields/_select.twig
+++ b/app/view/twig/editcontent/fields/_select.twig
@@ -30,7 +30,7 @@
     {% set wherefilter = field.filter|default({}) %}
     {% setcontent lookups = lookuptype where wherefilter order sortingorder nohydrate limit querylimit %}
     {% set valuefield = lookupfieldlist|default(lookupfield)|default('id') %}
-    {% set values = lookups|selectfield(valuefield, option.multiple, field.keys|default('id')) %}
+    {% set values = lookups|selectfield(valuefield, option.multiple, field.keys|default('id'), lookuptype) %}
 {% endif %}
 
 {# Get the current selection. Either a single value, or an array. #}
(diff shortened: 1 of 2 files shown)


SahAssar:
Sorry to be a thorn in your side, but wouldn't this break existing select fields that look like this:
```
featured_items:
    label: "Event/Event series"
    autocomplete: true
    type: select
    values: event/title
```
And expect id only values?

SahAssar:
You could use the `key` option, which currently defaults to `id`, and add an option to set it to `contenttype/id`. See more on the key option here: https://docs.bolt.cm/3.2/fields/select#populating-the-values-from-a-contenttype

mrenigma:
@SahAssar no it wouldn't because the values string is not requesting multiple content types.

`event/title` will always return an ID based value with this implementation.

I did look at the key option but I would have needed to rip more things out to make that work :grimacing: . I didn't want to touch more than I needed to!

SahAssar:
@mrenigma ah, great! 👍

PR #6303, opened 2017-01-26
Fix bugs with PathResolver/ResourceManager

CarsonF:
- BUG FIX: Checking `web` path exists when it used an alias threw exception. eg: `web: "%root%/public"`
- BUG FIX: `templatespath` not being initialized when needed (threw exception).
- Removed usage of `app` path everywhere. This will be re-added in v4 but for the user, not Bolt.
- Migrated some more code to use Filesystem abstraction or `PathResolver` instead of `ResourceManager`.
- Replaced `RM::getUrl` calls with `PathResolver` paths relative to `web` path.
- `view` path changes, see below.

I had been trying not to use `view` path in new code, because it is ambiguous (was it for the user or for Bolt?) It needed to be addressed due to bugs with `Configuration\Environment`.
**I changed the name to `bolt_assets`** (`view` is maintained for BC). This always points to the user's path for the synced bolt web assets. 
Open to changing the name to something similar. I thought about `bolt-assets` instead of an underscore. See usage in diff.
Our composer script handler now respects this path as well, instead of it being hardcoded.
This also cleans up the asset packages. The "bolt" filesystem cannot generate asset urls anymore as it may not be in the web root.


app/config/config.yml.dist
```diff
@@ -204,8 +204,8 @@ thumbnails:
     default_image: [ 1000, 750 ]
     quality: 80
     cropping: crop
-    notfound_image: bolt://app/view/img/default_notfound.png
-    error_image: bolt://app/view/img/default_error.png
+    notfound_image: bolt_assets://img/default_notfound.png
+    error_image: bolt_assets://img/default_error.png
```
bobdenotter: Is this not causing a BC break?

CarsonF: It was changed to `bolt://` earlier in 3.3, that won't work anymore (but not BC break). The previous value, for 3.2-, is managed [here](https://github.com/bolt/bolt/blob/release/3.3/src/Provider/ThumbnailsServiceProvider.php#L70-L86).

PR #6309, opened 2017-01-27
Minor feature: Contenttype singletons

SahAssar:
This is the "singleton contenttype" feature that @bobdenotter, me and a few others have talked about in slack.

It checks if there is only one record and the current user has no permissions to create new records and will then consider that contenttype a "singleton" that we link directly to the only record, skipping the submenu.

### DIFF
--- a/app/view/twig/_nav/_secondary-content.twig
+++ b/app/view/twig/_nav/_secondary-content.twig
@@ -1,7 +1,17 @@
 {% set definedSubs = [] %}
 
 {% for slug, contenttype in app.config.get('contenttypes')  %}
-    {% if isallowed('contenttype:' ~ slug) %}
+    {% setcontent records = slug limit 4 nohydrate orderby '-datechanged' %}
+    {% if not isallowed('contenttype:' ~ slug ~ ':create') and isallowed('contenttype:' ~ slug) and records|length == 1 and contenttype.show_in_menu == "true" %}
+        <li>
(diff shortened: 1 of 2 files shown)


SahAssar:
@bobdenotter Changed to 3.4 now. The check for singletons is "if the user isn't allowed to create records and there is only one record", so there is no additional configuration needed besides how we already handle things in permissions.yml.

SahAssar:
The test failure looks unrelated to me (especially since they only occur on mysql)...

GawainLynch:
> The test failure looks unrelated to me (especially since they only occur on mysql)...

Actually, you've triggered a #1555 bug.

bobdenotter:
> The check for singletons is "if the user isn't allowed to create records and there is only one record"

I think i'd prefer an optional `singleton: true` setting in contenttypes.yml, for ease of use. A ton of people develop (perhaps unknowingly) with the "root" account, so they won't trigger those restrictions, thinking it's broken.

leup:
Not sure to understand what a "singleton" is here.

If I understand this correctly, what's the PR is doing here:
- User cannot create more records (not allowed to) and there is only one record created => skip the view list and link to the record

But it's still possible that another user could create more records from that content type.

While having a `singleton: true` in the content type would be:
- Content type is a singleton and should only have one record, no one being able to create more

SahAssar:
Alrightio, this was more a aligning the UI with the existing permissions/content. I'll open a new PR with the `singleton: true` instead 🙂

PR #6310, opened 2017-01-27
Deprecate Render

CarsonF:
# Deprecate `Render`

`Render::render` is the last non-deprecated method in the class, and it is now just a shortcut to render template to `TemplateResponse`. As a shortcut, it makes more sense to do the logic in the Base Controller with the other shortcut methods.

It also gets simpler once the globals go away (to be deprecated later in 3.x).

---

I also changed `TemplateResponse` to take template name instead of template object.
Name is really the only thing useful from the object since it's already rendered.
Not to mention `Twig_Template` is marked as internal.

### DIFF
--- a/src/Controller/Base.php
+++ b/src/Controller/Base.php
@@ -2,6 +2,7 @@
 namespace Bolt\Controller;
 
 use Bolt\AccessControl\Token\Token;
+use Bolt\Response\TemplateResponse;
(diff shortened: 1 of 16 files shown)


GawainLynch:
> I also changed TemplateResponse to take template name instead of template object.

Yeah, I was in two minds about that too, when it was implemented. But TBH the string is all I personally would need from there, so :+1:

PR #6324, opened 2017-01-30
Add support for extension definition in bolt.yml file

rossriley:
Allows locally autoloadable extensions to be added via the bolt.yml file.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -82,6 +82,7 @@ return call_user_func(function () {
         'resources'   => null,
         'paths'       => [],
         'services'    => [],
+        'extensions'  => [],
     ];
 
     if (file_exists($rootPath . '/.bolt.yml')) {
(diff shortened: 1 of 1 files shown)


bobdenotter:
👍

app/bootstrap.php
```diff
@@ -184,5 +186,27 @@
 
     }
 
+    $extensionArray = [];
+    foreach ((array)$config['extensions'] as $extensionClass) {
+        if (is_string($extensionClass)) {
+            $extensionClass = new $extensionClass();
+        }
+        if ($extensionClass instanceof ExtensionInterface) {
+            $extensionArray[] = $extensionClass;
```
GawainLynch: Can we invert and throw an exception as discussed in Slack?

CarsonF:
Would it be better to do that logic inside of the extend closure? Wouldn't have to have a temp array. Maybe there would be some other benefit from doing it lazily?

bobdenotter:
Sweet!

PR #6339, opened 2017-02-01
Clarify message in system log, and don't log the message for user root.

### LINKED ISSUE #6334: "Configuration error: root is not granted to any roles"
When a non-root user is logged into the backend, a lot of these notices show up in the system log: 

> Configuration error: root is not granted to any roles

Which sort-of makes sense, because there _are_ no roles assigned to root. However, Bolt should not add these notices, because root is allowed to do stuff, without having to have a role. 

Details
-------

 - Relevant Bolt Version: [ 3.2 | 3.3 | master ]

bobdenotter:
Fixes #6334.

### DIFF
--- a/src/AccessControl/Permissions.php
+++ b/src/AccessControl/Permissions.php
@@ -69,7 +69,7 @@ class Permissions
     {
         // Log the message if enabled
         if ($this->app['config']->get('general/debug_permission_audit_mode', false)) {
-            $this->app['logger.system']->addInfo($msg, ['event' => 'authentication']);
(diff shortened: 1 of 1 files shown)


CarsonF:
Would  you mind changing the method to `info` instead of `addInfo`?

bobdenotter:
@CarsonF Done.

CarsonF:
I don't think that logic is right. You still want to `return false` if `$roles` is not an array.

bobdenotter:
I think you're right. Good catch!

PR #6341, opened 2017-02-01
DI Error & Exception Handlers

CarsonF:
This moves the error and exception handler calls into DI. There should be basically zero changes in functionality. This just offers more flexibility with how these handlers are configured. There's lots of comments in code if you care to know more.


bobdenotter:
> There's lots of comments in code if you care to know more.

Thanks! Very helpful! 👍

GawainLynch:
This re-introduces White Screen of Death when problems happen in bootstrap … which is very common.

CarsonF:
@gawainlynch what are those from?

GawainLynch:
Any number of things that could go wrong during bootstrap for reasons that often equate to "webhost".

My test has just been to throw an exception in bootstrap, like the one we now throw if you try and add a class that doesn't meet contract, that will just WSoD now, rather than informing the developer they messed up.

CarsonF:
Ah ok. Those exceptions can be converted manually. But besides those I don't think anything else could go wrong.

GawainLynch:
> Those exceptions can be converted manually. 

Make it so :wink: 

> But besides those I don't think anything else could go wrong.

Sure, we can reassess based on what gets reported as issues, or (hopefully) not.

CarsonF:
Will do once I get to work. My thought the only thing that can go wrong is the autoloader, which is already accounted for. Anything else is a bug which we should already know about.

CarsonF:
Ok just checked that file. The autoloader exception emits HTML so that doesn't need to be changed. The only other exception in that file is the `LogicException` inside of the `extensions` extend closure, which is ran after error handler is registered.

So no changes needed.

PR #6342, opened 2017-02-02
Fix to #6293

### LINKED ISSUE #6293: [BUG] record.taxonomy data is different under preview.
Details
-------

 - PHP version: 7.0.9
 - Used web server: Apache

Reproduction
------------
 1. **Bug summary**: 
`record.taxonomy` arrays return differently in preview mode, in turn breaking taxonomies outputs in theme files. 

 2. **Specifics**:
`setFromPost()` at
`C:/MAMP/htdocs/nuffieldtrust.org.uk/vendor/bolt/bolt/src/Controller/Frontend.php::210`

Combined with the taxonomy loop at
`C:/MAMP/htdocs/nuffieldtrust.org.uk/vendor/bolt/bolt/src/Storage/Entity/ContentValuesTrait.php::482`

 3. **Steps to reproduce**:
Create a taxonomy with terms that have multiple words, tag some content. 
Dump out record.taxonomy to see the values don't match up. 

This in turn is causing issues as record.taxonomy is being used to show taxonomies and breaks under preview. 

 4. **Expected result**: 
To receive the same array values as when viewing a published item .

image

 5. **Actual result**: 
The "slug" values of each taxonomy are returned / displayable.

image

mrenigma:
So I came across @graham73may 's issue while dealing with a bug myself so I thought I'd take a stab at fixing it.

My solution is to try and retrieve the taxonomy options from the config, if it exists apply the `setTaxonomy` method to each taxonomy with values obtained from the options otherwise provide the only value we have `$value` (this is for tags).

All are having the default `sortorder` set to 0 unless it's an array in which case the `sortorder` is set to the index key (is there ever an instance where the key is an associative?).

Fixes #6293

### DIFF
--- a/src/Storage/Entity/ContentValuesTrait.php
+++ b/src/Storage/Entity/ContentValuesTrait.php
@@ -487,8 +487,21 @@ trait ContentValuesTrait
                     }
                 }
 
-                $this->taxonomy[$taxonomytype] = $value;
+                $taxonomyOptions = $this->app['config']->get('taxonomy/' . $taxonomytype . '/options');
+
+                if ($taxonomyOptions && is_array($value)) {
+                    foreach ($value as $k => $v) {
+                        if (isset($taxonomyOptions[$v])) {
+                            $this->setTaxonomy($taxonomytype, $v, $taxonomyOptions[$v], $k);
+                        }
+                    }
+                } else if ($taxonomyOptions && isset($taxonomyOptions[$value])) {
+                    $this->setTaxonomy($taxonomytype, $value, $taxonomyOptions[$value], 0);
+                } else {
+                    $this->setTaxonomy($taxonomytype, $value, $value, 0);
+                }
             }
+
             unset($values['taxonomy']);
             unset($values['taxonomy-order']);
         }


src/Storage/Entity/ContentValuesTrait.php
```diff
@@ -487,7 +487,19 @@ public function setFromPost($values, $contenttype)
                     }
                 }
 
-                $this->taxonomy[$taxonomytype] = $value;
+                $taxonomy_options = $this->app['config']->get('taxonomy/' . $taxonomytype . '/options');
```
GawainLynch: PSR-2 on variable name

GawainLynch:
This function is/has been removed in 3.4, so not overall sure what the benefit of doing this in stable is.

mrenigma:
@GawainLynch I've updated the variable name - although the PSR's not define how properties/variables should be named (studly, camel or underscored) I agree it should be consistent with the bolt style.

This is a fix for the 3.2 branch because it currently does not work - if you could point me in the right direction to the replacement methods for 3.4 I am more than happy to PR a fix targeting the new methods too :smile:

GawainLynch:
> @GawainLynch I've updated the variable name - although the PSR's don't define how properties/variables should be named (studly, camel or underscored) I agree it should be consistent with the bolt style.

Well, PSR-1, and PSR-2 covers it with the voting, as does Symfony & Bolt code styles that all define it.

GawainLynch:
> if you could point me in the right direction to the replacement methods for 3.4 I am more than happy to PR

That you won't be able to do, as it is waiting on 3 different branches to fully land, so I guess we can send it in here, it just won't work in 3.4 again.

mrenigma:
@GawainLynch Fair enough! I will make sure to read the Bolt code style, I wonder if you can create separate formatting code profiles in PHPStorm on a per project basis? :thinking:

GawainLynch:
Best bet is to start with [Coding Standards](http://symfony.com/doc/current/contributing/code/standards.html) and **ignore**:

* No spaces around concatenation
* Use Yoda conditions when checking a variable against an expression to avoid an accidental assignment inside the condition statement (this applies to ==, !=, ===, and !==);

mrenigma:
Will do.

Yoda conditions - ergh! Feels like a C hangover which I wish would just go away

GawainLynch:
Thank you as always :smile:

Issue #5974, opened 2016-10-30
[BUG] "contentsCss" pref for CKEditor not honored


bintoro:
Specifying `wysiwyg/ck/contentsCss` in `config.yml` has no effect unless I comment out the relevant lines introduced in 6034fedd477926481bd43aeb92a5698b0cbacdd9. After that it works like a charm.

I haven't delved very deeply into Bolt's source, but it seems that `getDefaults()` sets the exact same `contentsCss` paths as `setCKPath()`, so perhaps the latter need not touch them at all.

Observed on v3.0 and v3.2.0-rc1.

tompato:
This still appears to be an issue. Regardless of your setting in 'general/wysiwyg/ck/contentsCss' the defaults are always used as @bintoro mentions above.

I think it's just a legacy thing as there are defaults for CK Editor to fallback to using defined in the very same file ('vendor/bolt/bolt/src/Config.php') so if the user didn't define anything it shouldn't break.

CarsonF:
This is fixed in 3.3

PR #6343, opened 2017-02-02
Fixes: #5974

tompato:
Prevents defaults from overriding user supplied values for ContentsCss field and appends user supplied ones to the defaults - Fixes: #5974

**What I did:**
Removed code from setCKPath function that was overriding user supplied
config values with defaults again and altered parseGeneral function so
that it appends the user supplied value(s) for the contentsCss field in
a way that will maintain the defaults when the mergeRecursiveDistinct
function runs.

**Explanation**
It's a bit strange as the recursive merge that occurs will override the default values if I didn't index everything from two onwards (as there are two default css files in the default config settings). I think the array_unshift and unset is the quickest way to do this if the user supplies an array of strings in their config.yml file for the contentsCss field, but feel free to correct me if there's a better way.

Tested it locally with no user supplied contentsCss field in config.yml, a single string and an array of strings (in case the user wishes to supply more than one css file).

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -330,7 +330,7 @@ class Config
 
         // Make sure old settings for 'contentsCss' are still picked up correctly
         if (isset($general['wysiwyg']['ck']['contentsCss'])) {
-            if(is_array($general['wysiwyg']['ck']['contentsCss'])) {
+            if (is_array($general['wysiwyg']['ck']['contentsCss'])) {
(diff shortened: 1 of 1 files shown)


CarsonF:
This is fixed in 3.3. The fix is a little bit more in depth so I would rather not back port it.

Thank you for your work though 👍

tompato:
No worries, good to see it's fixed, I shall have to update to 3.3 when it's released! :)

PR #6344, opened 2017-02-02
Move Canonical to Routing namespace

CarsonF:
Because I'm OCD


GawainLynch:
Say no to early BC lock-in! :wink:

PR #6353, opened 2017-02-04
Remove old, unused `hash_strength`

### LINKED ISSUE #6346: [Config] Remove hash_strength
`hash_strength` doesn't do anything anymore.

bobdenotter:
Fixes #6346

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -377,13 +377,6 @@ cookies_lifetime: 1209600
 # to '.example.org'.
 cookies_domain:
 
-# The hash_strength determines the amount of iterations for encrypting
-# passwords.
-#
-# A higher number means a harder to decrypt password, but takes longer to
-# compute. '8' is the minimum, '10' is the default, '12' is better.
-hash_strength: 10
-
 # Bolt sets the `X-Frame-Options` and `Frame-Options` to `SAMEORIGIN` by
 # default, to prevent the web browser from rendering an iframe if origin
 # mismatch (i.e. iframe source refers to a different domain).
--- a/src/Config.php
+++ b/src/Config.php
@@ -1172,7 +1172,6 @@ class Config
                 'only_aliases'      => false,
             ],
             'accept_file_types'           => explode(',', 'twig,html,js,css,scss,gif,jpg,jpeg,png,ico,zip,tgz,txt,md,doc,docx,pdf,epub,xls,xlsx,csv,ppt,pptx,mp3,ogg,wav,m4a,mp4,m4v,ogv,wmv,avi,webm,svg'),
-            'hash_strength'               => 10,
             'branding'                    => [
                 'name'        => 'Bolt',
                 'path'        => '/bolt',
--- a/src/Provider/AccessControlServiceProvider.php
+++ b/src/Provider/AccessControlServiceProvider.php
@@ -23,12 +23,6 @@ class AccessControlServiceProvider implements ServiceProviderInterface
             }
         );
 
-        $app['access_control.hash.strength'] = $app->share(
-            function () use ($app) {
-                return max($app['config']->get('general/hash_strength'), 8);
-            }
-        );
-
         $app['access_control'] = $app->share(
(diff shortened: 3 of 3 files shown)


GawainLynch:
Just one this. While it is completely correct according to the request & discussion in #6346, further discussion in #development on Slack was that in 3.4 we'd only use PasswordLib for legacy PHPass validation (at that point, an updated has is saved in the user's record), and use `password_hash()` and `password_verify()` … as when we implemented PasswordLib as a bridge, we were minimum PHP 5.4 and the other polyfills hadn't appeared, or been on our radar.

Basically looking to do something like this in 3.4:
```
$cycles = $app['access_control.hash.strength'];
$hash = password_hash($password, PASSWORD_DEFAULT, ['cost' => $cycles]);

$userEntity->setPassword($hash);
$repo->save($userEntity);
```

… as @rossriley put it 
> 8 might be ok for someone on shared hosting but I'd much rather use 12 or 13 on beefier hardware

bobdenotter:
As far as I could tell, `access_control.hash.strength` is not used in the current codebase. 

And doesn't this take care of re-hashing the old hashed passwords? 

```
        // Rehash password if not using Blowfish algorithm
        if (!Blowfish::detect($userAuth->getPassword())) {
            $userEntity->setPassword($this->passwordFactory->createHash($password, '$2y$'));
            try {
                $this->getRepositoryUsers()->update($userEntity);
            } catch (NotNullConstraintViolationException $e) {
                // Database needs updating
            }
        }
```

So, new ones (plus old ones that are recognized, but then converted) are hashed using Crypt, which uses enough rounds. What am i missing?

GawainLynch:
> So, new ones (plus old ones that are recognized, but then converted) are hashed using Crypt, 

Yes, but the line
```php
         if (!Blowfish::detect($userAuth->getPassword())) {
```

… is the only one we want to use PasswordLib for in v3.4+, and as we discussed the other day, removing it completely from v4.

> which uses enough rounds. What am i missing?

That "enough" is subjective, and people like @rossriley (and historically myself too), wind the cycles up on hardware that can cope with it, conversely if you're on a machine that really struggles, lower it … "you" obviously get both broken bits at that point :wink:

bobdenotter:
> … is the only one we want to use PasswordLib for in v3.4+, and as we discussed the other day, removing it completely from v4.

Correct, but it doesn't use the configured number of rounds in any way. 

>  @rossriley (and historically myself too), wind the cycles up on hardware that can cope with it

`passwordFactory->createHash` doesn't seem to allow for a strength / rounds / cost parameter. `password_hash` does, but we don't use that yet. Are you saying that instead of removing, we should keep it around for future (re)use?

GawainLynch:
> Correct, but it doesn't use the configured number of rounds in any way.

Exactly, which is why the conversation turned around in Slack. When we implemented PasswordLib, our minimum PHP requirement was 5.4 and we didn't have polyfills. Now we require PHP 5.5.9+ that hass all the needed bits.


Hence my original comment:

> Basically looking to do something like this in 3.4:
```
$cycles = $app['access_control.hash.strength'];
$hash = password_hash($password, PASSWORD_DEFAULT, ['cost' => $cycles]);

$userEntity->setPassword($hash);
$repo->save($userEntity);
```

bobdenotter:
I'm sorry, but i don't understand where you're going with this? should we keep it around? Should I change the PR?

GawainLynch:
OK, communication failure day … :disappointed: 

I am saying:
  * The commit *currently* in this PR is removing what we decided to keep, so needs to be dropped
  * Anything in core that is doing password hashing should be moved to use `password_hash($password, PASSWORD_DEFAULT, ['cost' => $cycles])` (where `$cycles = $app['access_control.hash.strength']`)
  * That can be done as part of an update and branch move for this PR (hence my failed attempt to explain an approach)
  * I'm right here for questions
  * We're at the 95% point of having all this done I think :smile:

bobdenotter:
Ok, understood. I'll close this PR, and #6346 as well.

Issue #6291, opened 2017-01-24
[BUG] Sortable select fields don't remain sorted.


graham73may:
Details
-------

 - Relevant Bolt Version: 3.2.6
 - Install type: Composer install
 - PHP version: 7.0.9
 - Used web server: Apache
 - For UX/UI issues: Chrome latest

Reproduction
------------
​
 1. **Bug summary**: 
 
Sortable select fields change to being alphabetically sorted on save. You can re-order the items after the page has been saved. 

 2. **Specifics**:

 3. **Steps to reproduce**:

Add a sortable select field to `contenttypes.yml`

```
        authors:
            group: "Meta"
            label: "Authors"
            multiple: true
            sortable: true
            autocomplete: true
            type: select
            values: people/title
```

Load up the contenttype your fields are on, add some items in a specific order. Press save. 

 4. **Expected result**: 
Expect to see the items saved in the order they appeared in the field. 

 5. **Actual result**: 
The items in the field become alphabetically sorted. 

It is however possible to sort the items once you have saved the page once and re-drag the items.

rossriley:
Looks like a fix for another bug has impacted on this, but it may be version specific since I'm not sure what point release this went in.

In #6096 the call that sorts the values based on the selected options (the unique twig helper was originally brought over from the extension) got removed so I'm pretty sure that will stop the sortable working altogether.....
https://github.com/bolt/bolt/pull/6096/files#diff-b0666fa264092eada1a5d5a18c75a4faL45

I can't get my head around this block:

```
 foreach ($merged as $val) {
     $compiled[$val[0]] = $val;
 }
```

I'm sure it must have worked at some point since it's been part of the extension from the beginning but you are right, as things stand now it seems to be pulling the first character of a string out, where I'm sure it must have previously been pulling out the index position instead.

So rough guess is that things in the underlying selectvalues have moved around somehow. Crux is it will need reworking and what you pasted in Slack is probably a good place to start.

mrenigma:
@rossriley I think that was one of our bugs which resulted in that pr :smile: 

I thought I would put what I found on here as well as slack so it doesn't get lost.

> I've come a cross an interesting issue and I don't know if it's intended. It looks like a standard sortable multiple select field with an `id,title` value tries to use the `ArrayHandler` unique twig function and assigns `$val[0]` to be the key even when `$val` is a string.
>
> Resulting in the following (twig variables dumped in order: `newvals`, `values`, `options`, `selection`):

[...]

PR #6355, opened 2017-02-04
Fixed: Maintain order for multiple-select fields.

### LINKED ISSUE #6332: Removed `unique` in options needs to be restored.
See #6096

From the dev meeting: 

rossriley
bopp, one for you to look at: https://github.com/bolt/bolt/pull/6096/files#diff-b0666fa264092eada1a5d5a18c75a4faL45 in that PR the call to the unique helper got removed…. was that just accidental or some more behind it?

Bopp
@rossriley I think that was accidental, but would need to verify.. Are you bumping into issues?

rossriley
what that did was pushed the selected values to the top of an array of options, so that sorted/selected values stayed in the order they should be

Bopp
Ah, ok! I'll open a quick issue, and restore that line.

rossriley:
eg, if you had selected and sorted [f,e] and the available options were [a,b,c,d,e,f] then the options were rendered in order [f,e,a,b,c,d]

bobdenotter:
Fixes #6332

### DIFF
--- a/app/view/twig/editcontent/fields/_select.twig
+++ b/app/view/twig/editcontent/fields/_select.twig
@@ -34,6 +34,19 @@
     {% set selection = [ selection ] %}
 {% endif %}
 
+{# if we have (sorted) options present already, make sure their order is
+   maintained. We do this by first selecting the current values, and then
+   merging that with the entire array of values. See #6332. #}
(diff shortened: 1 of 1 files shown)


bobdenotter:
Also fixes #6291

PR #6357, opened 2017-02-04
[WIP|RFC] Silex 2

dantleech:
This PR attempts to upgrade Bolt to Silex 2.0.

The changes:

- Services are shared by default, `$app->share()` has been removed.
- The provider `boot()` method has been segrated into an interface.
- Silex introduced a `LocaleServiceProvider`
- `\Pimple` renamed to `Container`
- Silex interfaces now in `Silex\Api`

Current status:

- Routes prefixes do not seem to be registered (i.e. `/bolt/login` matches `/{alias}/{file}` from the thumbnail controller.

--

This is a BC break unless a BC layer is introduced before the next major release (i.e. stubb \Pimple etc).

--

Hopefully I will have time to work on this further if it makes sense :)


GawainLynch:
Already in progress, but we're holding off on the SP changes until the last step. Thanks though :+1:

src/Provider/TwigServiceProvider.php
```diff
@@ -10,26 +10,44 @@
...
 use Symfony\Bridge\Twig\Extension\HttpFoundationExtension;
+use Pimple\Container;
 
 class TwigServiceProvider implements ServiceProviderInterface
 {
-    public function register(Application $app)
+    public function register(Container $app)
     {
+        // Twig options
+        $app['twig.options'] = function () use ($app) {
```
dantleech: Pimple does not let you override frozen services, and after the TwigServiceProvider is registered, these options seem to be frozen. Moved before we instantiate the TwigServiceProvider below, however that suggests that it might be registerred before.

dantleech:
Oh well. Should have checked before starting :)

bobdenotter:
You're of course more than welcome to help out on the migration to Silex 2 / Symfony 3.x for Bolt 4. If you do wish to help, best to coordinate with Gawain and Carson, since they have the best overview of the plan and overall approach. :-)

PR #6361, opened 2017-02-07
Decouple exceptions from exception handling

CarsonF:
Probably a couple things left to finish, but here's the idea.


src/Exception/Database/DatabaseException.php
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param string         $driver
+     * @param string         $message
+     * @param Exception|null $previous
+     * @param int            $code
+     */
+    public function __construct($driver, $message = "", Exception $previous = null, $code = 0)
```
GawainLynch: My OCD would like to know if that is copy/pasta? :wink

CarsonF: Haha nope

GawainLynch:
CC builds are having issues with our ruleset … ignoring.

Thanks for picking up the last of this chunk of work @CarsonF

PR #6362, opened 2017-02-07
Add sub and super to default allowed tags

evertalbers:
For those of us who have clients that write C0<sub>2</sub> all the time.  Donald Trump is firmly opposed to this pull request.


rossriley:
Make typography great again.

PR #6371, opened 2017-02-09
Refactor lazy logic out of FieldCollection

CarsonF:
This refactors `LazyFieldCollection` out of `FieldCollection`. `FieldCollection` is now just a mapping of `FieldValue` objects with some additional methods. `LazyFieldCollection` handles fetching references from db, hydrating results to `FieldValue` objects, and giving those to a `FieldCollection`. Both implement the new `FieldCollectionInterface`.

The motivation for this was that `FieldCollection::add` and `FieldCollection::set` did different things (as does `ArrayCollection`). `ArrayCollection` can be a `list` or `map`, but `FieldCollection` is always a `map`. Now the `add` and `set` methods of `FieldCollection` do the same thing. This also eliminates the need to override `offsetGet/Exists` since the objects are always stored in the collection correctly. Hopefully, this will reduce confusion.

### DIFF
--- a/src/Storage/Field/Collection/LazyFieldCollection.php
+++ b/src/Storage/Field/Collection/LazyFieldCollection.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace Bolt\Storage\Field\Collection;
+
+use Bolt\Storage\Entity\FieldValue;
+use Bolt\Storage\EntityManager;
+use Doctrine\Common\Collections\AbstractLazyCollection;
+
+/**
+ * This class is used by lazily loaded field values. It stores a reference to an array of rows and
+ * fetches from the database on demand.
+ *
+ * @author Ross Riley <riley.ross@gmail.com>
+ */
+class LazyFieldCollection extends AbstractLazyCollection implements FieldCollectionInterface
+{
+    /** @var int[] */
+    protected $references = [];
+    /** @var EntityManager|null */
+    protected $em;
+    /** @var int */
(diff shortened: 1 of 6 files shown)


src/Storage/Field/Collection/LazyFieldCollection.php
```diff
@@ -0,0 +1,108 @@
...
+            foreach ((array) $instances as $val) {
+                $fieldtype = $val->getFieldtype();
+                $field = $this->em->getFieldManager()->getFieldFor($fieldtype);
+                $type = $field->getStorageType();
+                $typeCol = 'value_' . $type->getName();
+
+                // Because there's a potential for custom fields that use json storage to 'double hydrate' this causes
+                // json_decode to throw a warning. Here we prevent that by replacing the error handler.
+                set_error_handler(
+                    function ($errNo, $errStr, $errFile) {},
```
bobdenotter: @CarsonF Wanna fix this, or tell Nitpick CI to shove it?

CarsonF: Nitpick can shove it :)

GawainLynch: But doesn't this become an exception in PHP 7+?

CarsonF: I'm not sure. I was just trying to leave that logic alone. Git says it's new but it's not. I didn't write it.

PR #6373, opened 2017-02-09
Add warning when relation name clashes with field name

### LINKED ISSUE #6372: [BUG] Naming a field within a contenttype the same as a relationship results in no relationships saving
If you have a field within a contentype that has the same name as another contenttype which you're linking via a relationship, you can't add relations.

Contentype setup:

```
services:
    name: Services
    singular_name: Service
    fields:
        addons:
            label: Special addons
            type: repeater
            fields:
                name:
                    type: text
                price:
                    type: text
            group: Addons
    relations:
        addons:
            multiple: true
            label: Addons for service
            order: -id
addons:
    name: Addons
    slug: addon
    singular_name: Addon
    fields:
        name:
            type: text
            required: true
            group: Details
        price:
            type: text
            required: true
            group: Details
```

The above setup breaks.

However, renaming the `services` field named `addons` to something else such as `extras` results in the relationships working again.

Details
-------

 - Install type: Composer
 - PHP version: 7.0
 - Used web server: Nginx
 - For UX/UI issues: Chrome v56

Reproduction
------------

1. Create the above contenttypes and add a record to the `addons` contentype.
2. Attempt to add a relationship to a `services` contenttype record.
3. Refresh your record and and observe the relationship vanishing and no database entry

rossriley:
Fixes #6372


Maelstromeous:
👍

PR #6375, opened 2017-02-10
Fix mcrypt deprecation warning

CarsonF:
I know this is fixed in 3.3, but Travis won't even finish the runs for 7.1. So this fixes it for the mean time. It won't even be in 3.3.


SahAssar:
Merging!

PR #6376, opened 2017-02-11
Prepare release Bolt 3.2.7

bobdenotter:


composer.json
```diff
@@ -10,9 +10,9 @@
     ],
     "license": "MIT",
     "require": {
-        "bolt/filesystem": "^2.0",
+        "bolt/filesystem": "^2.1",
         "bolt/pathogen": "~0.6",
-        "bolt/thumbs": "^3.1.1",
+        "bolt/thumbs": "^3.2",
         "brandonwamboldt/utilphp": "^1.0",
```
GawainLynch: While this does a lot for your OCD, it really pisses on people that use this a library, and might have problems with specific versions … I've built off my own fork more than one for this.

Minimum supported version is simply that. "minimum supported version", not ""bleeding edge" :cry:

bobdenotter: I can revert this if you prefer. no problem. :-)

GawainLynch: 462b8d3180815cbcad4471ba2a1fed7544d6593b

PR #6381, opened 2017-02-16
Update pager template to use blocks

CarsonF:
This abstracts the logic around creating a pager into simple blocks.

See the [pager/default.twig](https://github.com/CarsonF/bolt/blob/feature/pager-blocks/app/theme_defaults/pager/default.twig) for an example.

### DIFF
--- a/src/Storage/Field/Collection/FieldCollection.php
+++ b/src/Storage/Field/Collection/FieldCollection.php
@@ -146,16 +146,6 @@ class FieldCollection extends AbstractLazyCollection
     }
 
     /**
-     * Overrides the default array access method and passes on to the get method
-     * @param mixed $offset
-     * @return mixed
-     */
-    public function offsetGet($offset)
-    {
-        return $this->get($offset);
-    }
-
-    /**
      * Handles the conversion of references to entities.
      */
     protected function doInitialize()
--- a/app/theme_defaults/_sub_pager.twig
+++ b/app/theme_defaults/_sub_pager.twig
@@ -1,37 +1,5 @@
-{% set link = pager.makelink() %}
-
-{% if pager.totalpages > 1 %}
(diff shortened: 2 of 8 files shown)


bobdenotter:
Nice! I'd just like to add some html/classes to some items, for semantics and so that they can be styled.

src/Storage/Field/Collection/FieldCollection.php
```diff
@@ -146,16 +146,6 @@ public function offsetGet($offset)
     }
 
     /**
-     * Overrides the default array access method and passes on to the get method
-     * @param mixed $offset
-     * @return mixed
-     */
-    public function offsetGet($offset)
```
rossriley: Should this be part of the PR?

CarsonF: It's from a previous merge conflict. It could be committed separately

CarsonF:
@bob can you expand on that?

src/Storage/Field/Collection/FieldCollection.php
GawainLynch: @rossriley what @CarsonF said … it was my bad in the first place, for bigger branch merges I usually run tests before committing the merge, but you know that time you drop process 'cause "It'll be fiiiiiiiiiiiiiiiiiiiine!"

Admittedly it isn't best practice and I can move to a new PR in a few hours when I get up for real. My motivation is that @CarsonF is away for the week, and my time zones sucks too coordinate merges, and I was super excited about this PR. 

Not meant to excuse, just an explanation :smile:

rossriley:
No worries. Just checking it wasn't a mistake. I've actually just run into
the fatal so let's get it merged.

GawainLynch:
Sure … Opened a #6383 so that we can follow up on the request by @bobdenotter

PR #6384, opened 2017-02-16
Add "extension" to list of contexts

xiaohutai:
There's an `extension` option according to https://docs.bolt.cm/3.2/internals/container-service-references#app-logger-system which was missing:

systemlog


GawainLynch:
Thanks mate, really appreciated :+1:

PR #6385, opened 2017-02-16
[3.2] Force mbstring functions call on root namespace

GawainLynch:
In cases where someone hasn't enabled/installed PHP's mbstring extension, the following exception will be  thrown:

```
PHP Fatal error: Uncaught Error: 
Call to undefined function Bolt\mb_internal_encoding() 
in /var/www/vendor/bolt/bolt/app/bootstrap.php:31
```

The `Bolt\mb_internal_encoding()` part of the error is confusing as it indicates to the average user that the problem is with Bolt, and not the PHP installation.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -28,8 +28,8 @@ use Symfony\Component\Yaml\Yaml;
  */
 return call_user_func(function () {
     // Use UTF-8 for all multi-byte functions
-    mb_internal_encoding('UTF-8');
-    mb_http_output('UTF-8');
+    \mb_internal_encoding('UTF-8');
+    \mb_http_output('UTF-8');
 
     // Resolve Bolt-root
     $boltRootPath = realpath(__DIR__ . '/..');


rossriley:
I don't think it's an issue with this particular case, but it's worth noting that we often take advantage of function namespacing to help make some PHP functions testable.

This is especially the case with things like file operations where we can swap out `mkdir`, `file_put_contents` etc for our own test functions.

Like I said, I don't think we do this with mbstring functions anywhere but worth remembering so we don't do this as a general practice.

GawainLynch:
@rossriley Yes, mate and thanks for bringing it up … was literally banging my head against the wall on some of those mocks yesterday, and quite frankly, I have your back on  this wrt to the broader codebase.

My internal self-justification on this one was:
* It is prior to autoloaders
* It is the "hottest" part of the hot path
* For all intensive purposes we don't treat `app/bootsrap.php` as a SUT

PR #6386, opened 2017-02-17
[4.0] Use commit hash to prevent SAT resolver loop pre-Silex 2

GawainLynch:
Currently master will install vendor requirements fine on a git install where `silex/web-profiler` is required by the root project. But when Bolt is required as a dependency, the Composer SAT gets into a permanent loop with CPU pinned around 99%.

This is a simple hack that will be obvious to remove when we're ready for the Silex 2 cutover, but allow Composer installs for ~~idiots~~ brave people like myself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "PasswordLib/PasswordLib": "^1.0@beta",
         "php": "^5.6 || ^7.0",
         "silex/silex": "^1.3",
-        "silex/web-profiler": "1.0.x@dev",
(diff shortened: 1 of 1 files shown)


GawainLynch:
> ¯\_(ツ)_/¯

Took me hours to trace this one. I'd call corner case bug in Composer, but then I'd have Rob slap me about 5 minutes later in #composer I'm sure, either way, it is a transitory adjustment until we can finally cutover to Silex 2, which by the way in it's current dependency state is going to have to wait for Silex 2.1 (running 2.1-dev now on my personal site), or project based installs …
youre-going-to-have-a-bad-time

PR #6389, opened 2017-02-19
[3.3] Core deprecation notice logging

GawainLynch:
the deprecations are strong in this one

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -45,6 +45,8 @@ class Cache extends FilesystemCache
      */
     public function clearCache()
     {
+        @trigger_error(sprintf('%s is deprecated and will be removed in version 4.0. Use flushAll() instead.', __METHOD__));
+
         $this->flushAll();
 
         return [
--- a/src/Asset/Target.php
+++ b/src/Asset/Target.php
@@ -79,6 +79,7 @@ class Target
      */
     public function listAll()
     {
+        @trigger_error(sprintf('%s is deprecated and will be removed in version 4.0.', __METHOD__), E_USER_DEPRECATED);
         $reflection = new \ReflectionClass($this);
 
         return $reflection->getConstants();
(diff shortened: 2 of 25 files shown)


GawainLynch:
Deprecation counts based on unit test execution:
 * Before: 203
 * After: 69,213

bobdenotter:
Won't this affect performance adversely?

GawainLynch:
> Won't this affect performance adversely?

Micro, and obviously only with debugging enabled.

This is how Symfony, Silex, and Twig (etc) all do, and literally where @CarsonF and I have been getting all of our recent low hanging ~~pineapples~~ deprecation removal up targets from 

… and interestingly enough the discussion between Carson and I **two years ago** about being able to do this on core, was the start of the very long running attempts to get our exception handling sorted out.

In very basic terms, this is why languages like PHP — in this example — have `trigger_error('', E_USER_DEPRECATED);` so that developers can easily log and find parts of their code need to migrate to new calls/components/etc. 

The `@` silencer in this case become the _extra_ magic.

What we now get with this, is a **very** clear picture of what it to be done, and how close we're getting, towards replacing **both** upstream and core use of legacy code.

GawainLynch:
It just occurred to me to leave this here for those that need it:

If you do not want the deprecation notices, nor the small overhead that it introduces **when debugging is turned on only**, you can adjust your `debug_error_level` setting in `app/config/config.yml` like so:

```yaml
debug_error_level: 16383  # equivalent to E_ALL &~ E_USER_DEPRECATED
```

src/Cache.php
```diff
@@ -45,6 +45,8 @@ public function __construct($directory, $extension = self::EXTENSION, $umask = 0
      */
     public function clearCache()
     {
+        @trigger_error(sprintf('%s is deprecated and will be removed in version 4.0. Use flushAll() instead.', __METHOD__));
```
CarsonF: Missing `E_USER_DEPRECATED` here.

PR #6399, opened 2017-02-21
[4.0] Remove PasswordLib

GawainLynch:
Open to discussion on pretty much everything.

This simply removes PasswordLib (which is abandoned), and introduces a configurable service that advanced users can customise.

We're still 10 months out from release, but let the dominoes start to fall :wink: 

**UPDATE:** For people referring back to this, the concept of this PR & the `PasswordHashManager` service is a transitory **hack** during 4.0-dev until we can cut over to both Silex 2 & Symfony Security proper.

### DIFF
--- a/src/Exception/PasswordLegacyHashException.php
+++ b/src/Exception/PasswordLegacyHashException.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Bolt\Exception;
+
+/**
+ * Exception thrown when attempting to use PHPass hashes.
+ */
+class PasswordLegacyHashException extends PasswordHashException
+{
+    /**
+     * Constructor.
+     */
+    public function __construct()
+    {
+        $this->message = 'This user password has is stored using the legacy PHPass algorithm, and is no longer secure to use. ' .
+            'Please use the password reset functionality to set a new password.';
+    }
+}
--- a/src/Provider/AccessControlServiceProvider.php
+++ b/src/Provider/AccessControlServiceProvider.php
@@ -3,7 +3,7 @@
 namespace Bolt\Provider;
 
 use Bolt\AccessControl;
-use PasswordLib\Password\Factory as PasswordFactory;
+use Bolt\AccessControl\PasswordHashManager;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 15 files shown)


src/AccessControl/PasswordHash.php
```diff
@@ -0,0 +1,94 @@
...
+
+use Bolt\Exception\PasswordHashException;
+use Bolt\Exception\PasswordLegacyHashException;
+
+/**
+ * Password hash management service.
+ *
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
+class PasswordHash
```
CarsonF: Can we name this something else? It sounds like an entity from first glance, but it's actually a service. What about `PasswordHasher`?

src/Provider/AccessControlServiceProvider.php
```diff
@@ -67,7 +67,7 @@ function ($app) {
 
         $app['password_factory'] = $app->share(
```
CarsonF: `_factory` is confusing as this is a shared service not a _factory_.

tests/phpunit/unit/EventListener/StorageEventListenerTest.php
```diff
@@ -40,14 +39,14 @@ public function setUp()
         $this->schemaManager = $this->prophesize(SchemaManagerInterface::class);
         $this->urlGenerator = $this->prophesize(UrlGeneratorInterface::class);
         $this->flashLogger = $this->prophesize(FlashLoggerInterface::class);
-        $this->passwordFactory = $this->prophesize(PasswordFactory::class);
+        $this->passwordHash = $this->prophesize(PasswordHash::class);
```
CarsonF: Would the interface be better to mock here?

src/AccessControl/PasswordHash.php
GawainLynch: :+1:

src/Provider/AccessControlServiceProvider.php
GawainLynch: Ugh, screwed up a commit

src/Exception/PasswordLegacyHashException.php
```diff
@@ -0,0 +1,23 @@
...
+ * Password legacy hash verification exception.
+ */
+class PasswordLegacyHashException extends PasswordHashException
+{
+    /**
+     * Exception thrown when attempting to use PHPass hashes.
+     *
+     * @return PasswordLegacyHashException
+     */
+    public static function legacyPHPass()
```
CarsonF: Why not put this in the constructor? I don't see this class being used for anything else.

GawainLynch: Until the next legacy (looks to be 7.3 for the next round)

GawainLynch:
OK, GTG

PR #6400, opened 2017-02-21
fail safe for missing file

dforstercon:
Duplicated failsafe check from not alias version to this function.
May also needed in other branches.

### DIFF
--- a/src/Twig/Handler/ImageHandler.php
+++ b/src/Twig/Handler/ImageHandler.php
@@ -41,7 +41,7 @@ class ImageHandler
     {
         //Check if it's an alias as the only parameter after $filename
         if ($width && !$height && !$crop && $this->isAlias($width)) {
-            return $this->getAliasedUri($filename, $width);
+            return $this->getAliasedUri($fileName, $width);
         }
 
         if ($width || $height) {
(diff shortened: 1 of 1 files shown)


GawainLynch:
> May also needed in other branches.

We cascade our branch merges, so no stress there.

src/Twig/Handler/ImageHandler.php
```diff
@@ -271,6 +271,10 @@ private function getThumbnailUri(Thumbnail $thumb)
      */
     private function getAliasedUri($filename, $alias)
     {
+        if ($filename == null){
```
GawainLynch: Can you make the comparison strict please

dforstercon:
I changed the comparison to strict. 
But as I said, I just copied the failsafe from the getThumbnailUri function, which is non-strict either. :)

GawainLynch:
Yeah, we have a few of those … I'll look deeper in the morning, but the defaults on the two feeder functions are nulls, and we're trying to rid the code base of loose junk.

Thanks very much for the effort :+1:

GawainLynch:
I had a look back through the git history, and the `getThumbnailUri()` comparison was deliberately loosened for BC.

This should be GTG however.

PR #6401, opened 2017-02-21
Fix AND lookups on relationship/join searches

rossriley:
So this was a lot of fun, it's actually a feature that hasn't worked on Legacy storage since forever (as far as I can see) but it's come up twice in the last few weeks with users.

Whilst AND queries work fine on normal fields, eg: `setcontent records = 'pages' where { 'order': '>1 && < 10' }` it doesn't work when you are searching for both parts of a join to be set. 

For example on tags: `setcontent records = 'pages' where { 'tags': 'fun && gifs' }` should return any page that is tagged with both 'fun' and 'gifs'.

This PR does not address the legacy storage but it does work for the new engine. (Whether it is worth backporting can be a separate discussion).

So all of these examples work correctly now:

`$records = $app['query']->getContent('pages', ['tags'=>'fun && gifs']);`

On both taxonomies and on both relations too:
`$records = $app['query']->getContent('pages', ['tags'=>'fun && gifs', 'entries'=>'2 && 8']);`

And also mixed:
`$records = $app['query']->getContent('pages', ['tags'=>'fun && gifs', 'entries'=>'2 || 8']);` 

Note also that there's a small refactor of methods specific to join tables, now moving to a dedicated JoinTypeBase abstract class.


GawainLynch:
> This PR does not address the legacy storage but it does work for the new engine. (Whether it is worth backporting can be a separate discussion).

FWIW my vote is a firm "NO!" on backporting to legacy.

GawainLynch:
:shipit:

SahAssar:
Awesome! Being able to use queries like `$records = $app['query']->getContent('pages', ['tags'=>'fun && gifs', 'entries'=>'2 || 8']);` look great!

bobdenotter:
🎉

PR #6410, opened 2017-02-24
[3.3][Deprecation] Switch internal URL generator use of '#' to '_fragment'

GawainLynch:


GawainLynch:
Just in case someone asks, upstream: https://github.com/symfony/symfony/pull/20001

PR #6411, opened 2017-02-24
[3.3] Slugify slugs on edit save

### LINKED ISSUE #6408: [BUG] User created slugs are not slugified or verified
- Relevant Bolt Version: [ 3.2.6 ]
 - Install type: [ GitHub checkout ]
 - PHP version: [ 7.0 ]
 - Used web server: [ Apache 2 ]

1. **Summary**:When authors do a mistake in the slug/permalink field in the backend, i.e. leaving a trailing slash accidentally, or typing something like '/I/do/not/get/the/routing/system' as a slug, the record could be saved but then the backend has gone, because of the invalid slug.
 2. **Specifics**:
    * Throws an twig error for the backend, i.e. for http://www.example.org/bolt, after a record got an invalid slug saved
 3. **Steps to reproduce**:
    * Edit a slug or permalink with an invalid slug
 4. **Expected result**: 
    * Trying to save an invalid slug should throw an error message to the authors before the record is saved
 5. **Actual result**: 
    * After saving a record with an invalid slug a twig exception is thrown and the backend has gone, 

Here is an example of an error with an invalid trailing slash saved:

Uncaught Exception: Twig_Error_Runtime

Twig_Error_Runtime in Template.php line 447:
An exception has been thrown during the rendering of a template ("Parameter "slug" for route "contentlink" must match "[^/]++" ("est-enim-effectrix-multarum-et-magnarum-voluptatum/" given) to generate a corresponding URL.") in "@bolt/editcontent/_buttons.twig" at line 60.

GawainLynch:
Fixes #6408

### DIFF
--- a/src/Provider/StorageServiceProvider.php
+++ b/src/Provider/StorageServiceProvider.php
@@ -267,7 +267,8 @@ class StorageServiceProvider implements ServiceProviderInterface
                     $app['logger.change'],
                     $app['logger.system'],
                     $app['logger.flash'],
-                    $app['url_generator.lazy']
+                    $app['url_generator.lazy'],
+                    $app['slugify']
                 );
 
                 return $cr;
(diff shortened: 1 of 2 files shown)


bobdenotter:
Sweet! 

:shipit:

PR #6414, opened 2017-02-24
Created "Deprecated" Helper class

CarsonF:


GawainLynch:
Thanks for picking this up mate, really really really appreciated!

PR #6422, opened 2017-02-26
Replacing deprecated function calls

bobdenotter:


GawainLynch:
There is a lot to be done, now you an see how visible it all is hey :wink:

PR #6428, opened 2017-02-27
Change Symfony Security to use lazy url generator.

CarsonF:
This prevents routes from being flushed early.


GawainLynch:
Ping @CarsonF :laughing:

PR #6432, opened 2017-02-28
Deprecation updates

CarsonF:
- Removed deprecation notices for code that we still call. These don't help users at all since there is nothing they can do about it.
- Fixed some usages of deprecated code.
- Updated Twig Widget functions to check strict vars from Twig_Environment instead of from `twig.options` DI parameter. `twig.options` may not have the `strict_vars` key since it is optional and defaults to opposite of `debug`.
- Updates to `Deprecated` helper class
  - Added tests
  - Magic methods now give appropriate warnings.
     e.g. `Foo::__call('bar')` -> `Foo::bar() is deprecated`
  - Using `::method()` in constructor now acts like entire class is deprecated
- Moved `DeepDiff::diff` -> `Arr::deepDiff` (marked as internal).
- Marked tests that test deprecated code as `legacy` group. This puts them a separate list and doesn't count against our deprecated usage count.
- Added deprecation notices for:
  - `DeepDiff::diff`
  - `Library::getExtension`
  - `Library::safeFilename`
  - `Library::path`
  - `Library::redirect`
  - `Library::simpleredirect`
  - `Library::getExtension`
  - `ResourceManger::get/setUrl`
  - `ResourceManger::get/setRequest`
  - `safe_twig` service and `SafeEnvironment` class
  - `render` & `safe_render` services and `Render` class
  - Resolving extension objects for controller with class name


src/Provider/TranslationServiceProvider.php
```diff
@@ -137,7 +137,7 @@ public static function addResources(Application $app, $locale)
              * @var \SplFileInfo $fileInfo
              */
             foreach ($iterator as $fileInfo) {
-                $ext = Lib::getExtension((string) $fileInfo);
+                $ext = pathinfo($fileInfo, PATHINFO_EXTENSION);
```
GawainLynch: Wondered how many years until someone finally changed this line to the native function … mind youm the native one was only introduced in PHP 4.0.3 (yes, v4)

GawainLynch:
> Marked tests that test deprecated code as legacy group. This puts them a separate list and doesn't count against our deprecated usage count.

:heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart: :heart:

GawainLynch:
Passing locally and zero idea when Amazon S3 is going to be functional enough to do tests … :shipit:

PR #6434, opened 2017-03-01
[3.2] Mark Arr::isEmptyArray() @internal

GawainLynch:
Please please please can we not add this to the API to support in a tagged version.

I get we need something for repeaters, and I don't have a better solution myself but this has support liability attached we don't want to have to live with even for 3.x

### DIFF
--- a/src/Helpers/Arr.php
+++ b/src/Helpers/Arr.php
@@ -2,9 +2,6 @@
 
 namespace Bolt\Helpers;
 
-use RecursiveArrayIterator;
-use RecursiveIteratorIterator;
-
 class Arr
 {
     /**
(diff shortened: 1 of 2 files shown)


CarsonF:
Can it just be replaced with a private method in the class it's used instead?

GawainLynch:
> Can it just be replaced with a private method in the class it's used instead?

Care Factor Zero … the last few weeks (as you well know) have been hard work porting around stuff that we've got BC on and should never have had, creating extra work for us we shouldn't need to have had to worry about

GawainLynch:
c.f. #6421

rossriley:
On reflection, I'd prefer we just mark this internal for now... I'm not too keen on this being done in the Save controller service, I'd rather the individual fields were able to control their data normalisation it maybe that it is better somewhere else to make it accessible to fields in which case I'll refactor for 3.3 branch.

CarsonF:
Me either, but since this is the only use case I can think of for it, it should be there. Since it's private it can be refactored to individual fields when they are ready for it. 

The codebase is already hard to follow with all the traits, magic, and lack of phpdocs. I would rather keep it close to the use of it.

rossriley:
Cool, let's merge this then just to get things working in 3.2 and we can think more anon.

GawainLynch:
Thanks @rossriley … Really really appreciate you meeting half way on this :+1:

PR #6435, opened 2017-03-01
Fix renaming files in sub-subfolders

SahAssar:
Renaming file in sub-subfolders didn't work since we only prepended the namespace, not the parentpath. This fixes that.


CarsonF:
@GawainLynch FYI, for when you are merging, this already fixed/refactored/replaced in 3.3.

PR #6436, opened 2017-03-01
Adding codemirror features

SahAssar:
This adds the following codemirror features to the backend fileeditor:

 * Proper syntax highlighting for HTML/twig mixed. Up until now we only highlighted HTML, and twig was all treated the same as text. Done via a overlay mode that combines the modes.

 * Codemirror codefolding, based on the indent strategy. This is very nice especially for long contenttypes where you can collapse at certain levels (like for example hide all the fields or the whole contenttype definition). Hotkeyed to Ctrl-Q to fold the current line.

 * Forces tabs to be converted to 4 spaces. This was previously done on save, but we've seen some people paste code into slack/whereever where they used tab chars, which made the indentation look wrong there. Better to be consistent.

 * Save with Ctrl-S.

 * Replace with Ctrl-H

To try it out you need to do both a updateLib and updateBolt.


SahAssar:
@bobdenotter I guess it could go either way, we've been using it for a quite a while (since 2.2) so I don't think the bug risk is very big, but I can move it 3.3 if that's preferable :)

bobdenotter:
It's not so much the risk of blowing things up, but more that we're trying to get some structure in releases, and try to only add new features to the next minor updates. 

So, I think 3.3 would be better. :-)

CarsonF:
I would rather see this in 3.4. 3.3 is in beta.

SahAssar:
There, changed to 3.4 and cleaned up :)

src/Controller/Backend/FileManager.php
```diff
@@ -359,26 +359,28 @@ private function getRelatedFiles(FileInterface $file)
...
             case 'less':
-                return ['css'];
+                return array_merge($all, ['css']);
             case 'js':
-                return ['javascript'];
+                return array_merge($all, ['javascript']);
             default:
-                return [];
+                return $all;
         }
```
CarsonF: I would rather assign to a variable and merge `$all` afterwords instead of duplicating that logic for each case.

SahAssar: Fixed (or did you want to keep $all as a variable too?)

PR #6439, opened 2017-03-02
Deprecated ResourceManager Completely

CarsonF:
This removes the remaining usages of `ResourceManager` and adds a ton of deprecation notices.

I know this is late to be in 3.3. This would be best grouped in this release, because `PathResolver` has been added and `ResourceManager` has had most of its usage removed and deprecated.

-----

The only other notable change in this PR is 1b19486d978ba8519a907c3c7a00f0de8c4a70ed. The `theme` filesystem is added to `filesystem` service at boot, instead of when the object is created. This allows the `filesystem` service to be accessed whenever, without fear of an infinite loop. The Twig loader was also changed to add the theme path at boot since it now doesn't exist before that point.


GawainLynch:
Before merging this, I would like to extend a very public "thank you" to @rossriley who originally authored `ReourceManager`.

`ReourceManager` was a transitional concept that he invested a lot of time and hard work in to get Bolt into a more flexible position starting with Bolt 2.0-dev, moving BOlt away from using hard coded file system paths … and a lot of people would not grasp exactly how far it took this project.

Bolt has grown as a platform considerably since then, and for quite a while now RM has needed to e replaced for a bunch of reasons that do not matter past "we want to make Bolt better, more powerful, and more flexible."

The functionality introduced in this PR marks the end of an era in a way, so :beers: for all

PR #6441, opened 2017-03-02
What's a ResourceManager?

CarsonF:
Remember that time I said there was no more usage of ResourceManager (#6439)? I lied.

**Now** there are no more usages (including tests).


bobdenotter:
This will probably need to be updated properly in the docs as well, right?

CarsonF:
Yeah I'm thinking a migration guide might be needed. And any usages of RM in docs should be removed as well.

bobdenotter:
> And any usages of RM in docs should be removed as well.

Yes, please. ;-)

Issue #6382, opened 2017-02-16
[RFC] Improving Controller Reusability


rossriley:
I think this is mainly in reference to #6058 

The original reason for coming up with the BoltResponse object was so that we could have a single response object that was passed around and modified until it came time to render when we would have a single object with a template name, some context variables and some globals which could then render.

I'm just in the process of doing some migrations to 3.3 and it doesn't look like we've got any capability now to do in-controller modification. Here's a few simplified examples of what was possible before.....

```php
# in MyApp/Controller/Frontend
public function record(Request $request, $contenttypeslug, $slug = '')
{
        $response = parent::record($request, $contenttypeslug, $slug);
        $response->addGlobals(['siteconfig'=>$this->getCustomSiteConfig()]);
    
        return $response;
}


# A more indepth example
public function record(Request $request, $contenttypeslug, $slug = '')
{
	$response = parent::record($request, $contenttypeslug, $slug);
	if($record = $render->getGlobalContext()['record']) {
            $showRelated = $record->get('showother');
            if ($showRelated) {
                $otherRecords = $this->app['storage']->getContent($showRelated.'/latest/4');
    		$response->addGlobals(['other'=>$otherRecords]);
            }   
	}
    
	return $response;
}
```

What this allowed was for you to override the default Frontend controller methods without having to completely rewrite them.

By pre-rendering the template we're now only left with the option of completely re-implementing the entire 60ish lines of method code which isn't really ideal when we don't really need to change things like the record lookup, the templatechooser etc.

Plus it means that if we change anything in the base controller methods without having a decent public API to extend more gracefully, then we lose the ability to update how things are done without breaking a load of application code in the wild.

So, whilst there's good reasons with Twig deprecations etc to need to do this, is there anything we can put in its place to keep the principle of BoltResponse in the controller code?

GawainLynch:
Would updating the [render()](https://github.com/bolt/bolt/blob/release/3.3/src/Controller/Base.php#L75-L90) with an event suffice?

e.g.
```php
    protected function render($template, array $context = [], array $globals = [])
    {
        $twig = $this->app['twig'];

        $template = $twig->resolveTemplate($template);

        foreach ($globals as $name => $value) {
            $twig->addGlobal($name, $value);
        }
        $globals = $twig->getGlobals();

        // Event class with getter/setters for $template, $context & $globals
        $event = new ControlerPreRenderEvent(template, $context, $globals); 
[...]

PR #6443, opened 2017-03-03
TemplateView

CarsonF:
In regards to #6382

Working on how this can be done without breaking BC, but it's proving challenging.

The only response method called in Frontend was the `X-XSS-Protection` header in record route. So I moved it out.

Now you can do something like:
```php
class MyFrontend extends Frontend
{
    protected function render($template, array $context = [], array $globals = [])
    {
        return new TemplateView($template, array_replace($context, $globals));
    }
}
```

and then replace the `controller.frontend` service with that class.

Open to ideas about how to integrate this further. I don't think changing `Controller\Base::render` directly is wise though, since others can base their controllers on it. But maybe if it's something just for Frontend / Backend classes that would be ok. I'm not sure if a separate shortcut even makes much since given how terse the line actually is.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -454,3 +454,10 @@ hash_strength: 10
 #        disable: true     # Disable news panel. Defaults to false. "Alerts" will still be shown.
 #    stack:
 #        disable: true     # Disable stack usage. Defaults to false.
+
+# Options that will be forced in next major version
+compatibility:
+    # Whether to return TemplateView instead of TemplateResponse from Controller\Base::render()
+    # Response methods cannot be used on TemplateView objects.
+    # Setting this value to false is deprecated.
+    template_view: true
--- a/src/Controller/Backend/FileManager.php
+++ b/src/Controller/Backend/FileManager.php
@@ -42,11 +42,7 @@ class FileManager extends BackendBase
             ->assert('namespace', '[^/]+')
             ->value('namespace', 'files')
             ->bind('fileedit')
-            ->after(function (Request $request, Response $response) {
-                if ($request->isMethod('POST')) {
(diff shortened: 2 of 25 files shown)


CarsonF:
`TemplateView` allows template & context to be _modified_ before it is rendered.
`TemplateResponse` allows _read-only_ access to template & context but also allows modification to the response headers or body (string).

Let me give some examples of how this can be used.

Say you want to modify a record:
```php
$app->view(function (TemplateView $view, Request $request) {
    if ($view->context->has('record')) {
        $record = $view->context->get('record');

        // modify record

        $view->context->set('record', $record);
    }
}, 10 /* above default */);
```

Say you want to add a header for a certain record:
```php
$app->after(function (Request $request, Response $response) {
    if ($response instanceof TemplateResponse) {
        $context = $response->getContext();
        $record = isset($context['record']) ? $context['record'] : null;

        if (/* if record is foo */) {
            $response->headers->set('Foo', 'Bar');
        }
    }
});
```

Either of these could be used to retrieve the `record` and do something with it.

SahAssar:
👍 This looks great!

GawainLynch:
FWIW the conceptual approach, as it stands, has my :+1:

GawainLynch:
Welcome to Monday, @CarsonF … Your friendly, weekly, "already-asking-too-much-of-a-person" nudge on this :hugs:

bobdenotter:
No movement on this for 5 weeks. Please don't forget, @CarsonF. :-)

CarsonF:
There is, it's just local. Haven't forgotten.

CarsonF:
This is ready for review now.

bobdenotter:
As discussed during the dev-meeting: :shipit:

PR #6445, opened 2017-03-03
[3.2] Ensure config-cache.json gets flushed

GawainLynch:
Today's tally:

:koala: — 2 
Gawain — 0


GawainLynch:
Just a note to add that this being branched-merged into `release/3.3` is a release blocker for beta 

… it's corner case when/where it causes a problem in 3.3, but someone going to just randomly hit it and get very confused when everything breaks in ways that I don't want to detail, as the safety of the worlds kittens depends on this staying out of public knowledge!

PR #6453, opened 2017-03-04
The continued war on slugs

### LINKED ISSUE #6438: [3.3] Repeaters mismatch between slug and key..
There's still some mismatch between the keys and slugs for contenttypes, when it comes to repeaters:

Details
-------

 - Relevant Bolt Version: [ 3.2 | 3.3 | master ]

Reproduction
------------

Create a simple contenttype, like: 

```
foo:
    name: Bars
    slug: bars
    singular_name: Bar
    singular_slug: bar
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        images:
            type: repeater
            fields:
                repeattitle:
                    type: text
                repeatimage:
                    type: image
```

Add a record. 

Edit the record.twig template: 

```
{% block main %}

    <h1>{{ record.title }}</h1>

<hr>

{% for section in record.images %}
    {{ dump(section) }}
    image: {{ section.get('repeatimage')|join(',')  }} <br>
    title: {{ section.get('repeattitle') }}

{% else %}

    <p>No repeaters.</p>

{% endfor %}

<hr>
```

In the browser it now looks like this: 

screen shot 2017-03-01 at 17 57 44

Now, edit the contenttype, to look like this: 

```
bars:  # note: foo -> bars
    name: Bars
    slug: bars
    singular_name: Bar
    singular_slug: bar
    fields:
        title:
....
```

Go back to editing the record. Add a repeater block or two, and back to the browser: 

screen shot 2017-03-01 at 18 00 46

@rossriley Is this fixable? Or should we add a warning for this, for now?

rossriley:
Fixes #6438

The legacy patch that fills in repeaters to old content objects was slacking on the job. We should give priority to the contenttype 'key' property only falling back to the slug if the key isn't set (which also means that the slug is the same as the key)

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2598,7 +2598,11 @@ class Storage
 
         // Get the contenttype from first $content
         $first = reset($content);
-        $contenttypeslug = $first->contenttype['slug'];
+        // Try with the key first, if this isn't available then fall back to using the slug
+        $contenttypeslug = $first->contenttype['key'];
+        if ($contenttypeslug === null) {
+            $contenttypeslug = $first->contenttype['slug'];
+        }
         $contenttype = $this->getContentType($contenttypeslug);
         $repo = $this->app['storage']->getRepository('Bolt\Storage\Entity\FieldValue');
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
slug

bobdenotter:
Travis says "no" on this one. :-)

rossriley:
@GawainLynch playing a strong meme game today.

PR #6457, opened 2017-03-04
Add missing `delete` permission.

### LINKED ISSUE #6068: [BUG] When a non-root user views the edit page of a piece of content, the Delete button doesn't appear
Non-root users who view the edit page of a piece of content and have the delete permission for that contenttype, don't see a Delete button, even though they can delete the content by other means (such as via the dropdown arrow and checkboxes on the content list view).

Additionally, an error appears in the logs whenever such a visitor views the edit page.

`Info	authentication	editcontent	/bolt/editcontent/pages/1	Configuration error: delete is not granted to any roles.
bolt\vendor\bolt\bolt\src\AccessControl\Permissions.php line: 315
`

Details
-------

 - Relevant Bolt Version: [ 3.0.11, 3.1.5, 3.2.2 ]
 - PHP version: [ 7.0 ]
 - Used web server: [ Apache ]

Reproduction
------------
​
 1. **Steps to reproduce**:
    * Grant delete permissions to a role
    * Assign a user to that role
    * Login as that user
    * When viewing a list of content, and clicking the dropdown arrow, the Delete options appears as expected
    * Click to edit a piece of content
    * The log entry mentioned above now appears in the System Log
    * The user does not see the Delete button
 2. **Expected result**: 
    * The Delete button appears (screenshot for root user)
image

 3. **Actual result**: 
    * The Delete button does not appear (screenshot for non-root user who has delete permissions) 
image

bobdenotter:
Fixes #6068

### DIFF
--- a/app/config/permissions.yml.dist
+++ b/app/config/permissions.yml.dist
@@ -67,6 +67,7 @@ global:
     dbcheck: [ admin, developer ]
     dbupdate: [ admin, developer ]
     dbupdate_result: [ admin, developer ]
+    delete: [admin, developer, chief-editor]    
     extensions: [ developer ]
     extensions:config: [ developer ]
     fileedit: [ admin, developer ]


GawainLynch:
@bobdenotter I don't think we should approach this fix in this way.

The global permissions "loosely" track matching routes, and the issue is to do with ContentType permissions, a completely separate section & concern.

I had meant to look into this, but time, etc … but if you want to double check it is probably just needing a route name to binding name added to one of the `->before()` in one of he bases … or may actually be a more insidious problem hiding under the surface

bobdenotter:
This popped up on another website too, today: 

screen shot 2017-03-08 at 15 31 00

GawainLynch:
Which means we've probably got a misconfigration in core that we'd just hide with this change. My guess is a copy/pasta :koala:

I'll look today after I get back from medical appointments.

GawainLynch:
It is indeed a bug in core that would have gotten rather hard to track down with this PR.

This check in [_aside.twig](https://github.com/bolt/bolt/blob/release/3.2/app/view/twig/editcontent/_aside.twig#L15)
```
{% if isallowed('delete', context.contenttype) and context.content.id is not empty %}
```

Is passing a `null` for `context.contenttype`, causing the permissions class to check for it as a `global` property, which it most certainly isn't, rather than a ContentType property, which it very much is. :wink:

GawainLynch:
Amending the change to this fixes the problem
```diff
diff --git a/src/Twig/Handler/UserHandler.php b/src/Twig/Handler/UserHandler.php
index 61edf3d..80a49f6 100644
--- a/src/Twig/Handler/UserHandler.php
+++ b/src/Twig/Handler/UserHandler.php
@@ -72,6 +72,8 @@ class UserHandler
             // It's a content record
             $contenttype = $content;
             $contentid = $content['id'];
+        } elseif ($content instanceof \Bolt\Storage\Mapping\ContentType) {
+            $contenttype = (string) $content;
         } elseif (is_array($content)) {
             // It's a contenttype
             $contenttype = $content;
```

GawainLynch:
Oh, obviously above patch was on 3.2 so will need to be done by hand in 3.3, but TBH I think should actually go to 3.2 as it is a legitimate bug fix

bobdenotter:
I'm thinking we should close this one, and just do a new PR for 3.2. (because the bug is present there too). Waht do you think?

bobdenotter:
Oh, wait. No reading comprehension, before i've had coffee. Yeah, new PR incoming.

GawainLynch:
Would you like some git magic instead? :smile: 

I only didn't … too late

bobdenotter:
You'd need to git magic both the change and the target. This way i'll get to monkeytest it too, before it lands. :-)

GawainLynch:
Nah, just checkout the branch, reset it against `release/3.2`, `git apply` the diff above, commit and force push.

bobdenotter:
It's too early for common sense. Stop it. ;-)

GawainLynch:
:hugs:

PR #6463, opened 2017-03-07
Raw should only be used on user data when intended

SahAssar:
Raw should only be used on user data when intended (or at all).

I don't think I need to say more.


bobdenotter:
We'll have to see how to fix the 'records' view, which outputs escaped HTML, because of this change: 

screen shot 2017-03-09 at 16 46 55

SahAssar:
@bobdenotter: I'm guessing that isn't a text field, which means that it should use the twig markup object to output it, right? relying on `|raw` here seems odd (and something we should fix).

PR #6470, opened 2017-03-10
Update DatabaseExceptionTrait.php

HeinDR:
I ran into an error when I'd configured a wrong database name for a mysql database. This threw an exception as expected but the exception threw an error because it could not access the `$driveNames` property. I could fix this by setting the visibility to protected.

### DIFF
--- a/src/Exception/Database/DatabaseExceptionTrait.php
+++ b/src/Exception/Database/DatabaseExceptionTrait.php
@@ -7,7 +7,7 @@ trait DatabaseExceptionTrait
     /** @var string */
     protected $driver;
 
-    private static $driverNames = [
+    protected static $driverNames = [
         'pdo_mysql'   => 'MySQL',
         'mysql'       => 'MySQL',
         'mysql2'      => 'MySQL',


src/Exception/Database/DatabaseExceptionTrait.php
```diff
@@ -7,7 +7,7 @@
     /** @var string */
     protected $driver;
 
-    private static $driverNames = [
+    protected static $driverNames = [
```
GawainLynch: This should not be called directly from outside the trait, there is the `getPlatform()` for that.

Can you provide a trace for the exception you were seeing as I am unable to reproduce.

HeinDR: The trace is empty. The `$driverNames` property is called from the `getPlatform` method. This is the error I got:
`FatalErrorException in DatabaseExceptionTrait.php line 47: 
Error: Cannot access property Bolt\Exception\Database\DatabaseConnectionException::$driverNames`
To reproduce you can change the `databasename` in the `config.yml` to a non existing databasename and you should get the error.

GawainLynch: > To reproduce you can change the databasename in the config.yml to a non existing databasename and you should get the error.

That is exactly what I was doing, with `debug: nuclear` … joking about the "nuclear", that is -1 (`E_ALL`). :laughing:

GawainLynch: FWIW I get a chain of the correct 4 exceptions from PDO though to DBAL for a bad DB name.

HeinDR: Hm that is strange. When I change the name I still get this error. Could it be a PHP version thing? I'm using 5.6.20.

HeinDR: When I do a stacktrace I see that `getPlatform` is called from:
`Bolt\Exception\Database\DatabaseException->getPlatform() called at [/var/www/hderuiter/incadat/vendor/bolt/bolt/src/EventListener/ExceptionListener.php:170]`

GawainLynch: Ugh … OK … we were both off on tangents :laughing: 

Short version: inheritance. So yes, this PR is actually correct. I am however going send it to `release/3.3` … as :bug:

HeinDR: Oke cool.

GawainLynch:
Sorry for the initial mix up on my part, and thank you very much for the PR :+1:

Issue #6472, opened 2017-03-10
Discrepancy in `{{ record.foo }}` vs `{{ record.get('foo') }}`


bobdenotter:
There's a discrepancy in how these are handled.. The latter isn't output as `\Twig_Markup`.

Example: 

```
    <h1>{{ record.title }}</h1>

    <hr>
    <p>record.teaser:</p>

    {{ record.teaser }}

    <hr>

    <p>record.get('teaser'):</p>

    {{ record.get('teaser') }}

    <hr>
```

Result: 

![screen shot 2017-03-10 at 17 24 01](https://cloud.githubusercontent.com/assets/1833361/23803295/63a1449e-05b6-11e7-9336-94e7e0e66e99.png)

Details
-------

 - Relevant Bolt Version: [ 3.2 | 3.3 ]

werner-rammer:
I experienced a (probably) similar problem: I upgraded from 3.0 to 3.2.9 and found that the HTML tags were visible in the output (the same as the ```reord.get('teaser')``` example above).
I used this in ```entry.twig``` (which worked fine in 3.0):
```
{{ fields(exclude = ['image']) }}
```
After some experimenting I modified this to:
```
    {{ record.teaser }}
    {{ record.body }}
```
i.e., include the fields explicitly. This resolved the problem (see ```record.teaser``` above).

CarsonF:
Pretty sure this is only for repeaters, as they don't render twig/markdown or mark content as safe yet.

bobdenotter:
@CarsonF Look at the code example above. It's not just repeaters.

bobdenotter:
The discrepancy is still there, but the risk of introducing new bugs when changing this was too high, so instead we did what we did in #6474.

PR #6473, opened 2017-03-11
[WIP] Api compatibility with legacy content objects

rossriley:
As agreed this PR adds one method to each of the Repeater collection and Repeater field objects.

It matches the API of dealing with existing content items.

### DIFF
--- a/src/Storage/Field/Collection/LazyFieldCollection.php
+++ b/src/Storage/Field/Collection/LazyFieldCollection.php
@@ -68,6 +68,19 @@ class LazyFieldCollection extends AbstractLazyCollection implements FieldCollect
     }
 
     /**
+     * Returns the type of a given $fieldName
+     *
+     * @param $fieldName
+     * @return string|null
+     */
+    public function getFieldType($fieldName)
+    {
+        $this->initialize();
+
+        return $this->collection->getFieldType($fieldName);
+    }
+
+    /**
      * Handles the conversion of references to entities.
      */
     protected function doInitialize()
--- a/src/Storage/Field/Collection/LazyFieldCollection.php
+++ b/src/Storage/Field/Collection/LazyFieldCollection.php
@@ -81,6 +81,20 @@ class LazyFieldCollection extends AbstractLazyCollection implements FieldCollect
     }
 
     /**
+     *  Alias to the standard get method that matches compatibility with the Legacy content entity.
+     *  This can be removed once the deprecation of legacy content is complete.
+     *
+     * @param $fieldName
+     * @return mixed
+     */
+    public function getDecodedValue($fieldName)
+    {
+        $this->initialize();
+
+        return $this->collection->getDecodedValue($fieldName);
+    }
+
+    /**
      * Handles the conversion of references to entities.
      */
     protected function doInitialize()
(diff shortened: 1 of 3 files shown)


bobdenotter:
`getdecodedValue()` works the other way around from #6472 now. :-)

screen shot 2017-03-11 at 12 25 37

bobdenotter:
I can't get `fieldtype` to work yet.. 

```
        {% for repeater in record.repeater %}
            {% for key, repeaterfield in repeater %}
                {{ repeater.fieldtype(key) }}
            {% endfor %}
        {% endfor %}
```

gives: 

screen shot 2017-03-11 at 12 26 30

rossriley:
Sorry, forgot we have two classes now. Added to the LazyCollection too.

bobdenotter:
Now i'm getting this: 

screen shot 2017-03-11 at 12 41 37

Would it help if i threw pasted the contenttype and template to reproduce?

bobdenotter:
I'm using the default `showcase` contenttype, and the template looks like this: 

```
{% extends 'partials/_master.twig' %}

{% block main %}

    <h1>{{ record.title }}</h1>

    {% for repeater in record.repeater %}

        {% for key, repeaterfield in repeater %}

            {% set fieldtype = repeater.fieldtype(key) %}
            {% set value = repeater.getDecodedValue(key) %}

            <p>field is: {{ fieldtype }}</p>
            {{ dump(value) }}

        {% endfor %}

    {% endfor %}
```

rossriley:
OK, try again.

bobdenotter:
Halfway there. :-)

`{% set fieldtype = repeater.fieldtype(key) %}` -> works
`{% set value = repeater.decodedValue(key) %}` -> gives error: 

screen shot 2017-03-11 at 13 01 44

bobdenotter:
Almost perfect. :-)

`{{ repeater.decodedValue(key) }}` still does the same thing as `{{ repeater.get(key) }}`, as in it returns escaped HTML: 

screen shot 2017-03-11 at 12 25 37

GawainLynch:
Also if you're pushing more, gentle nudge on the PHPDoc types :wink:

bobdenotter:
I wouldn't mind adding the PHPDoc myself in a consecutive PR. 

Doe ik ook nog eens wat. ;-)

GawainLynch:
Your call … just thinking about it, these should be properly marked up as `@deprecated`, no?

bobdenotter:
Yes, they should be marked `@deprecated` too. :-)

CarsonF:
Also missed the interface.

GawainLynch:
Also wondering if they should be marked `@internal` as otherwise we're something locking into the external API that a) never worked; b) is scheduled to be ripped out inside the 3.x lifecycle

src/Storage/Field/Collection/LazyFieldCollection.php
```diff
@@ -68,6 +68,33 @@ public function setGrouping($grouping)
...
+     *  This can be removed once the deprecation of legacy content is complete.
+     *
+     * @param $fieldName
+     * @return mixed
+     */
+    public function getDecodedValue($fieldName)
+    {
+        $this->initialize();
+
+        return $this->collection->getDecodedValue($fieldName);
```
GawainLynch: Shouldn't this just:

```php
return $this->collection->get($fieldName);
```

GawainLynch:
@bobdenotter pulled this into #6474, so closing

PR #6474, opened 2017-03-11
🚧 🔨 Fix/clean up  sub fields

### LINKED ISSUE #6472: Discrepancy in `{{ record.foo }}` vs `{{ record.get('foo') }}`

bobdenotter:
This PR makes sure the vilified-but-necessary `_sub_fields`: 

 - don't rely on `|raw`, `|twig`, `template_from_string()` or any such hackishness
 - have the same treatment for regular fields, template fields and repeaters

~~Might needs some final touch-ups after #6472 has landed.~~

**Edit:** 

 - Works correctly for HTML, text, textarea and Markdown fields now.
 - Works correctly for repeaters now. (note: No support for `allowtwig: true` in repeaters. It was never supported, but this PR doesn't add it either)
 - Incorporates @rossriley 's changes in #6473. 
 - Closes #6472, as this PR provides a viable workaround for that issue, without changing stuff that might have unforeseen consequences. 

After this is merged, I will update the docs for 3.3 accordingly, because iterating over repeaters can be done more similar to how it's done for "normal" fields or templatefields.

### DIFF
--- a/app/theme_defaults/_sub_fields.twig
+++ b/app/theme_defaults/_sub_fields.twig
@@ -32,7 +32,7 @@
 
 {# Output a text-like field with the wrapper for live editing. #}
 {% macro textfield(object, key) %}
-    <div data-bolt-field="{{key}}">{{ object.decodedValue(key) }}</div>
+    <div data-bolt-field="{{key}}">{{ object.renderedValue(key) }}</div>
 {% endmacro %}
 
 {% macro imagelistfield(images) %}
--- a/app/theme_defaults/_sub_fields.twig
+++ b/app/theme_defaults/_sub_fields.twig
@@ -30,15 +30,9 @@
 
 {# SECTION 2: MACROS #}
 
-{# If 'allowtwig' is true for this field, we'll need to parse it as
-   Twig here. Note, that we're doing this inside a macro, so the
-   snippet has a limited scope. No global variables, etc. #}
-{% macro parsetextfield(key, value, contenttype) %}
-    {% if contenttype.fields[key].allowtwig is defined and contenttype.fields[key].allowtwig == true %}
-        {% set value = include(template_from_string(value), sandboxed = true) %}
-    {% endif %}
-
-    <div {% if key is not empty %}data-bolt-field="{{key}}"{% endif %}>{{ value }}</div>
+{# Output a text-like field with the wrapper for live editing. #}
+{% macro textfield(object, key) %}
+    <div data-bolt-field="{{key}}">{{ object.decodedValue(key) }}</div>
 {% endmacro %}
 
 {% macro imagelistfield(images) %}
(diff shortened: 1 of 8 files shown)


app/theme_defaults/_sub_fields.twig
```diff
@@ -30,15 +30,9 @@
...
-   snippet has a limited scope. No global variables, etc. #}
-{% macro parsetextfield(key, value, contenttype) %}
-    {% if contenttype.fields[key].allowtwig is defined and contenttype.fields[key].allowtwig == true %}
-        {% set value = include(template_from_string(value), sandboxed = true) %}
-    {% endif %}
-
-    <div {% if key is not empty %}data-bolt-field="{{key}}"{% endif %}>{{ value }}</div>
+{# Output a text-like field with the wrapper for live editing. #}
+{% macro textfield(object, key) %}
+    <div data-bolt-field="{{key}}">{{ object.getDecodedValue(key) }}</div>
```
GawainLynch: Twig style just needs `{{ object.decodedvalue(key) }}` 

… which TBH still doesn't sound like what it is … decoded from what to what, etc?

theme/base-2016/partials/_sub_fields.twig
```diff
@@ -30,15 +30,9 @@
...
-   snippet has a limited scope. No global variables, etc. #}
-{% macro parsetextfield(key, value, contenttype) %}
-    {% if contenttype.fields[key].allowtwig is defined and contenttype.fields[key].allowtwig == true %}
-        {% set value = value|twig %}
-    {% endif %}
-
-    <div {% if key is not empty %}data-bolt-field="{{key}}"{% endif %}>{{ value }}</div>
+{# Output a text-like field with the wrapper for live editing. #}
+{% macro textfield(object, key) %}
+    <div data-bolt-field="{{key}}">{{ object.getDecodedValue(key) }}</div>
```
GawainLynch: Twig style just needs `{{ object.decodedvalue(key) }}`

app/theme_defaults/_sub_fields.twig
GawainLynch: OK, so expecting that I'd probably "Done a Gawain" and not communicated well enough, I went reading back though Twig internals to find the [part of the code that does to UPPER/lower case & `is`/`get` handling](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Template.php#L616-L657)

…then the :sweat_drops: :fish: … this works exactly the same:

```twig
<div data-bolt-field="{{key}}">{{ attribute(object, key) }}</div>
```

… and in turn no need for #6473 IMHO

bobdenotter:
@GawainLynch Simple contenttype to reproduce: 

```
bars:
    name: Bars
    slug: bars
    singular_name: Bar
    singular_slug: bar
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        body:
            type: html
            height: 300px
        repeato:
            type: repeater
            fields:
                repeattitle:
                    type: text
                repeatimage:
                    type: image
    record_template: record.twig
```

Fill it with some content, and it gives me: 

screen shot 2017-03-13 at 08 59 29

bobdenotter:
> Twig style just needs `{{ object.decodedvalue(key) }}` which TBH still doesn't sound like what it is

I agree that something like `{{ object.parsedValue(key) }}` would make more sense, but that would _also_ mean more code that needs work when 4.0 comes around. I don't know what would be best for now.

app/theme_defaults/_sub_fields.twig
bobdenotter: Nope, `{{ attribute(object, key) }}` won't work here. It gives the same result as #6472

bobdenotter:
Screenshot of Repeater when editing content: 

screen shot 2017-03-21 at 22 04 14

Results in frontend: 

screen shot 2017-03-21 at 22 07 53

bobdenotter:
> Can you add those two new methods to FieldCollectionInterface?

Will do. 👍 

> Also that doesn't handle rendering Twig.

As already noted at the top of the PR. ;-)

bobdenotter:
> Implement getRenderedValue to replace getDecodedValue

Better, because it does what it says now. 👍

rossriley:
Just one more note on this, whilst calling it `getRenderedValue` is more clear, it still needs to be marked strictly for Bolt use only, the presentation needs to be delegated down to the fields, but this is fine as a stop-gap until we have something better.

PR #6483, opened 2017-03-14
[3.2] First check user

### LINKED ISSUE #6481: [3.2+] Changing Database Settings Not Resetting Session
Spotted this one and while the security implications are very minimal there's an outside chance that someone could retain escalated privileges on a db they shouldn't have access to.

The main problem is that you can also move to a new database, retain your login and in the case of the new DB being empty then you don't get redirected to the userfirst page but can carry on creating content etc.

To reproduce: just change the database setting in config/config_local.yml when already logged in. Then carry on navigating the backend and you will still be logged in but start working on the new database.

GawainLynch:
Fixes #6481

### DIFF
--- a/src/AccessControl/AccessChecker.php
+++ b/src/AccessControl/AccessChecker.php
@@ -150,7 +150,7 @@ class AccessChecker
         try {
             // Only show this flash if there are users in the system.
             // Not when we're about to get redirected to the "first users" screen.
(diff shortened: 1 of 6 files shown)


GawainLynch:
The row count idea was @SahAssar's btw

GawainLynch:
> Users repository already has a `hasUsers` method doing the same thing.

So it does … I couldn't see for looking apparently

PR #6484, opened 2017-03-15
Session Handlers Phase II

CarsonF:
# Configuring Handlers

In my previous attempt, I tried to normalize all connection parameters and options for `Memcache`, `Memcached`, `Redis`, and `Predis`. In reality this became confusing and didn't work well.

Now all that logic has been moved into factory classes. In addition there were several bugs in how each client was configured that have been resolved. All the factories are under unit tests as well 👏 

These do still need to be documented I know. I was hesitant to document the configuration before, because I didn't like how it turned out and wanted to change it. But there's only so much one can handle looking at undocumented PHP C source code 😉 
Now I'm happy(er) with it, and will take a stab at the docs.

Also it's important to note that `predis` is its own type now separate from the `redis` type which is just for the php extension.

# IniBag

Why is everything builtin to PHP garbage?

`IniBag` abstracts away the `ini_get`, `ini_set`, `ini_get_all` methods and provides helpful error messages when something goes wrong.

It's currently used for [importing session ini options](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Provider/SessionServiceProvider.php#L175-L190) and configuring Memcached options ([here](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Session/Handler/Factory/MemcachedFactory.php#L42) and [here](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Session/Handler/Factory/MemcachedFactory.php#L162-L188)).

# Cache Handlers

Added session handlers for [Doctrine Cache](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Session/Handler/DoctrineCacheHandler.php), [PSR-6 Cache](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Session/Handler/PsrCacheHandler.php), and [PSR-16 Simple Cache](https://github.com/CarsonF/bolt/blob/improvements/sessions/src/Session/Handler/PsrSimpleCacheHandler.php).

# Misc

- Moved session handler creation to its own container key which allows all factory logic to be skipped.
```php
$app['session.handler'] = $app->share(function () {
    return MyCustomSessionHandler();
});
```
- Throw exception when trying to register a bag after the session has started.
- Removed unused session options

### DIFF
--- a/src/Session/Handler/Factory/AbstractFactory.php
+++ b/src/Session/Handler/Factory/AbstractFactory.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Bolt\Session\Handler\Factory;
+
+use Bolt\Helpers\Deprecated;
+use Bolt\Session\OptionsBag;
+use GuzzleHttp\Psr7\Uri;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Abstract Factory with common functions to help create connections to services.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ */
+abstract class AbstractFactory
+{
+    /**
+     * Parse the session options to connection parameters.
+     *
+     * @param OptionsBag $sessionOptions
+     *
+     * @return OptionsBag[]
+     */
+    public function parse(OptionsBag $sessionOptions)
+    {
+        if ($sessionOptions['save_path']) {
+            return $this->parseConnectionsFromSavePath($sessionOptions['save_path']);
+        }
+
+        return $this->parseConnectionsFromOptions($sessionOptions);
+    }
+
+    /**
+     * Parses session save_path into a list of connection parameters.
+     *
+     * @param string $savePath
+     *
+     * @return OptionsBag[]
+     */
+    protected function parseConnectionsFromSavePath($savePath)
+    {
+        $connections = (array) explode(',', $savePath);
+        $connections = array_map('trim', $connections);
+
+        return array_map([$this, 'parseConnectionItemFromSavePath'], $connections);
+    }
+
+    /**
+     * @param string $path
+     *
+     * @return OptionsBag
+     */
+    abstract protected function parseConnectionItemFromSavePath($path);
+
+    /**
+     * Parses session options with array configuration into a list of connection parameters.
+     *
+     * @param OptionsBag $options
+     *
+     * @return OptionsBag[]
+     */
+    protected function parseConnectionsFromOptions(OptionsBag $options)
+    {
+        if ($options->has('connections')) {
+            $connections = $options->get('connections');
+        } elseif ($options->has('connection')) {
+            $connections = [$options->get('connection')];
+        } elseif ($options->has('host') || $options->has('port')) {
+            Deprecated::warn('Specifying "host" and other options directly in session config', 3.3, 'Move them under the "connection" key.');
+
+            $connections = [$options->all()];
+        } else {
+            $connections = [[]];
+        }
+
+        return array_map([$this, 'parseConnectionItemFromOptions'], $connections);
+    }
+
+    /**
+     * @param array|string $item
+     *
+     * @return OptionsBag
+     */
+    abstract protected function parseConnectionItemFromOptions($item);
+
+    /**
+     * @param string $uri
+     *
+     * @return UriInterface
+     */
+    protected function parseUri($uri)
+    {
+        if ($uri && stripos($uri, 'unix://') === 0) {
+            // parse_url() doesn't like "unix:///foo", but parses "unix:/foo" fine.
+            $uri = str_ireplace('unix://', 'unix:', $uri);
+
+        } elseif ($uri && $uri[0] !== '/' && !preg_match('#^\w+://?.+#', $uri)) {
+            // If no scheme given and path exists and isn't absolute, assume tcp scheme is wanted.
+            // This allows both "127.0.0.1" and "/foo.sock" to work without specifying scheme.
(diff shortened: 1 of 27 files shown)


src/Session/Handler/Factory/MemcacheFactory.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param string|null $class Memcache class name. Mostly for tests.
+     */
+    public function __construct($class = Memcache::class)
+    {
+        if (!extension_loaded('memcache')) {
+            throw new RuntimeException('Unable to use "memcache" session handler as redis extension is not installed and enabled. Install the extension or try the "memcached" session handler instead.');
```
GawainLynch: > as redis extension is not installed

s/redis/memcache

src/Session/Handler/Factory/MemcachedFactory.php
```diff
@@ -0,0 +1,336 @@
...
+    /**
+     * Constructor.
+     *
+     * @param ParameterBag|null $ini   ini values. Mostly for tests.
+     * @param string|null       $class Memcached class name. Mostly for tests.
+     */
+    public function __construct(ParameterBag $ini = null, $class = Memcached::class)
+    {
+        if (!extension_loaded('memcached')) {
+            throw new RuntimeException('Unable to use "memcached" session handler as redis extension is not installed and enabled. Install the extension or try the "memcache" session handler instead.');
```
GawainLynch: I'm detecting a copy/pasta trend here …

CarsonF: brain 💀

tests/phpunit/unit/Session/Handler/Factory/MemcachedFactoryTest.php
```diff
@@ -0,0 +1,636 @@
...
+
+use Bolt\Session\Handler\Factory\MemcachedFactory;
+use Bolt\Session\OptionsBag;
+use Bolt\Tests\Session\Handler\Factory\Mock\MockMemcached;
+use Memcached;
+use PHPUnit_Framework_TestCase as TestCase;
+use Symfony\Component\HttpFoundation\ParameterBag;
+
+/**
+ * @requires extension memcached 2.0.0
```
GawainLynch: For some reason it is the "`2.0.0`" that is triggering the same errors for me, as it is Travis.

Removing "`2.0.0`" causes the tests to correctly skip locally.

CarsonF: Ok I adding another `@requires` without the version seems to fix that.

src/Session/Handler/Factory/AbstractFactory.php
```diff
@@ -0,0 +1,121 @@
...
+     * @param string $uri
+     *
+     * @return UriInterface
+     */
+    protected function parseUri($uri)
+    {
+        // parse_url() doesn't like "unix:///foo", but parses "unix:/foo" fine.
+        if ($uri && stripos($uri, 'unix://') === 0) {
+            $uri = str_ireplace('unix://', 'unix:', $uri);
+        }
```
CarsonF: Happy to change this just don't know where the comment would go.

GawainLynch: It's the split on the `} else {` … I ended up just "letting go", but one option would be to put the `else` specific comments in the logic block …

CarsonF:
I think I got all the tests fixed now. But Travis is struggling.

GawainLynch:
So my last question(s) before green lighting this:

1. Should we add this to Travis' config so at least Redis & Memcached get tested
```diff
diff --git a/.travis.yml b/.travis.yml
index 226f7c1..14f769a 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ before_install:
   # Start SMTP listener
   - sudo python -m smtpd -n -c DebuggingServer localhost:25 2>&1 > /dev/null &
   # Increase PHP memory limit. Required for Composer & PHPUnit.
-  - echo "memory_limit=2G" ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+  - echo "memory_limit = 2G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+  - echo "extension = memcached.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+  - echo "extension = redis.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
 
 before_script:
   # Set up Composer
```

2. Should we set up the **PECL build** for Memcache so that is tested on Travis as well

3. Should we follow-up PR `master` to bump Memcached (read: 'd') to a minimum of 3.0.0 and remove Memache

CarsonF:
:+1: For Travis

I can do the PR to master to remove the deprecated code and bump versions. 

Do we think we need to add a warning that memcache won't be supported in 4.0?

GawainLynch:
> I can do the PR to master to remove the deprecated code and bump versions.

All yours, flagging for process.

> Do we think we need to add a warning that memcache won't be supported in 4.0?

I think so. The "deprecation" as such is due to many years of dormancy upstream, and potentially a **lot** of work to get upstream to be usable in PHP 7.x … and upstream is a pile of bit-rot now anyway.

GawainLynch:
> Now I'm happy(er) with it, and will take a stab at the docs.

Holding you to this :wink:

PR #6486, opened 2017-03-15
One char typo

rarila:
Beware of #1555 :-)


GawainLynch:
Moved to 3.2 as it's in stable+ … branch merge will get it down the line.

PR #6490, opened 2017-03-17
[3.3] Add extra PHP module requirements to composer.json

GawainLynch:
A lot of people are setting up on new VMs and occasionally omitting PHP extensions from their install.

As these are the group(s) that would typically use Composer, this will short-cut some of their problem solving as they will now see something like:

```
$ composer update --no-dev
Loading composer repositories with package information
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested PHP extension ext-openssl * is missing from your system. Install
      or enable PHP's openssl extension.
```
Should someone want to install on a cut-down machine to use as a deploy base, they can simply `composer update --no-dev --ignore-platform-reqs`


GawainLynch:
> Doesn't thumbs require exif and gd? We shouldn't need it here then.

[Only GD](https://github.com/bolt/thumbs/blob/master/composer.json#L19)

That aside, I am in two minds … Yes, DRY OCD … but also having them in one place is a simple canonical reference for all parties … Or is it too much for your OCD? :stuck_out_tongue_winking_eye:

PR #6494, opened 2017-03-18
[3.3][Tests] Deprecations … another sweep

GawainLynch:
* Move `setExpectedException*()` to annotations (deprecated in PHPUnit 5.5.2)
* Move FQCN PHPUnit class name use into imports with an `as` that matches the PHPUnit 6.x namespace
* Class name resolution OCD

### DIFF
--- a/tests/phpunit/BoltListener.php
+++ b/tests/phpunit/BoltListener.php
@@ -3,6 +3,10 @@
 namespace Bolt\Tests;
 
 use Bolt\Application;
+use PHPUnit\Framework\BaseTestListener;
+use PHPUnit_Framework_Test as Test;
+use PHPUnit_Framework_TestCase as TestCase;
+use PHPUnit_Framework_TestSuite as TestSuite;
 use Symfony\Component\Console\Input\StringInput;
 use Symfony\Component\Filesystem\Filesystem;
 
@@ -11,7 +15,7 @@ use Symfony\Component\Filesystem\Filesystem;
  *
  * @author Gawain Lynch <gawain.lynch@gmail.com>
  */
-class BoltListener implements \PHPUnit_Framework_TestListener
+class BoltListener extends BaseTestListener
 {
     /** @var array */
     protected $configs = [
@@ -38,7 +42,7 @@ class BoltListener implements \PHPUnit_Framework_TestListener
     /**
      * Called on init of PHPUnit exectution.
      *
-     * @see PHPUnit_Util_Configuration
+     * @see \PHPUnit_Util_Configuration
      *
      * @param array   $configs Location of configuration files
      * @param bool    $theme   Location of the theme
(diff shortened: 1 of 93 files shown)


tests/phpunit/unit/Controller/FrontendTest.php
```diff
@@ -240,15 +245,17 @@ public function testNoRecord()
             ->will($this->returnValue(false));
         $this->setService('storage', $storage);
 
-        $this->setExpectedException('Symfony\Component\HttpKernel\Exception\HttpException', 'not found');
-
         $response = $this->controller()->record($this->getRequest(), 'pages');
 
         $this->assertTrue($response instanceof TemplateResponse);
         $this->assertSame('record.twig', $response->getTemplateName());
         $this->assertNotEmpty($response->getGlobals());
```
GawainLynch: from the `$app->aborts()`

tests/phpunit/unit/Storage/EntityManagerTest.php
```diff
@@ -26,42 +31,44 @@ public function testCreateQueryBuilder()
...
         $em = $app['storage'];
-        $repo = $em->getRepository('showcases');
+        $em->getRepository('showcases');
 
         // The first check should work, this one should fail because the factory has not been set.
-        $this->setExpectedException('RuntimeException');
         $em = new EntityManager($app['db'], $app['dispatcher'], $app['storage.metadata']);
-        $repo = $em->getRepository('showcases');
+        $em->getRepository('showcases');
     }
```
GawainLynch: Done. GitHub detection fail.

tests/phpunit/unit/Storage/Query/QueryParameterParserTest.php
```diff
@@ -78,9 +81,8 @@ public function testCompositeOrAnd()
         $this->assertEquals('(id > :id_1) AND (id <= :id_2)', $filter->getExpression());
         $this->assertEquals(['id_1' => '29', 'id_2' => '37'], $filter->getParameters());
 
-        $this->setExpectedException('Bolt\Exception\QueryParseException');
         $p = new QueryParameterParser($expr);
-        $filter = $p->getFilter('ownerid', '>1||<4 && <56');
+        $p->getFilter('ownerid', '>1||<4 && <56');
     }
```
GawainLynch: Done. GitHub detection fail.

tests/phpunit/unit/Storage/StorageTest.php
```diff
@@ -64,7 +67,6 @@ public function testGetContentObject()
             ->setMockClassName('Failing')
             ->getMock()
         ;
-        $this->setExpectedException('Exception', 'Failing does not extend \Bolt\Legacy\Content.');
         $storage->getContentObject(['class' => 'Failing', 'fields' => $fields]);
     }
```
GawainLynch: Done. GitHub detection fail.

tests/phpunit/unit/Storage/StorageTest.php
```diff
@@ -114,14 +119,17 @@ public function testSaveContent()
         $this->assertEquals(1, $postsave);
     }
```
GawainLynch: Done and expanded slightly. Still a GitHub detection fail.

tests/phpunit/BoltListener.php
```diff
@@ -11,7 +17,7 @@
  *
  * @author Gawain Lynch <gawain.lynch@gmail.com>
  */
-class BoltListener implements \PHPUnit_Framework_TestListener
+class BoltListener implements TestListener
```
CarsonF: Since you are here, what do you think about extending `BaseTestListener` instead? That way we don't have to have all these empty method bodies.

CarsonF:
Also can these commits be squashed down into their change, not their class? i.e. `replace `setExpectedException` and `update phpunit namespace`. `[Tests] Update Bolt\Tests\Users` doesn't tell me anything when I'm looking through the git log.

tests/phpunit/BoltListener.php
GawainLynch: Done

GawainLynch:
> So I found that PHPUnit has compatibility classes for the commonly used ones:

Updated 

> Also, maybe this should be targeting 3.2 instead? .... Not sure where this falls between feature and bug fix.

3.2 should close soon with 3.3.0. But also I think I targeted 3.3 as that was where some extras went in, so on PR, keep moving, etc …

> Also can these commits be squashed down into their change, not their class?

On to it

GawainLynch:
Everything GTG now.

bobdenotter:
Whoop. 👍

PR #6495, opened 2017-03-18
Fix frontend taxonomy lookups where key is different to slug

### LINKED ISSUE #6228: [BUG] Mismatch between legacy and new storage?
Details
-------

 - Relevant Bolt Version: 3.2.5
 - PHP version: 7.0
 - Used web server: Apache version 2.4

Reproduction
------------

From the Slack channel:

zomars [10:12]  
My `contentypes.yml`file:

```yaml
portfolio:
    name: Portafolio
    singular_name: Portafolio
    slug: portafolio
    singular_slug: portafolio
    tablename: portafolio
```

zomars [10:12]  
Gets me this:
pasted image at 2017_01_06 10_12 am

My tablename is portafolio but the taxonomy contenttype is portfolio. This shows me the correct taxonomy in the Back-End, but `null` in the Front-End.

I’ve manually changed the DB to this:

pasted image at 2017_01_06 10_14 am

zomars [10:14 AM]  
And now it gives me the correct taxonomies in the Front-End BUT `null` in the Back-End

sahassar [10:14 AM]  
Aaaah... Might be worth opening an issue about that, since it's a mismatch betweeen legacy and new storage (I'd guess)

rossriley:
Fixes #6228 

Adds a check to see if the key is different to the slug when deciding how to lookup taxonomy joins.

Note this will need to be reworked slightly for 3.3 as the implementation has changed slightly (dropping util:: functions ) so it may need a separate PR.

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2381,7 +2381,7 @@ class Storage
         }
 
         // Get the contenttype from first $content
-        $first = $content[util::array_first_key($content)];
+        $first = reset($content);
         $config = $first->contenttype;
         $contenttype = (isset($config['key'])) ? $config['key'] : $config['slug'];
 
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -2381,7 +2381,9 @@ class Storage
         }
 
         // Get the contenttype from first $content
-        $contenttype = $content[util::array_first_key($content)]->contenttype['slug'];
+        $first = $content[util::array_first_key($content)];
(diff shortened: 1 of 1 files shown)


GawainLynch:
> Note this will need to be reworked slightly for 3.3 as the implementation has changed slightly (dropping util:: functions ) so it may need a separate PR.

Why not just back-port [this line](https://github.com/bolt/bolt/blob/release/3.3/src/Legacy/Storage.php#L2384) and problem solved

PR #6496, opened 2017-03-18
Add option to only dump the SQL from the database:update command

dantleech:
This PR adds a `dump-sql` option to the `database:update` command.

Using this option will dump a list of SQL statements *instead* of performing the update on the database.

This is useful when generating migrations.

```
$ app/nut database:update --dump-sql                                                                                                                                                                   
ALTER TABLE bolt_field_value ADD value_boolean TINYINT(1) DEFAULT '0' NOT NULL
DROP INDEX IDX_4C524BC3EA112943 ON bolt_relations
DROP INDEX IDX_4C524BC378CED90B ON bolt_relations
DROP INDEX IDX_4C524BC35ACD2645 ON bolt_relations
DROP INDEX IDX_4C524BC330354A65 ON bolt_relations
CREATE INDEX IDX_4C524BC3EA11294378CED90B ON bolt_relations (from_contenttype, from_id)
CREATE INDEX IDX_4C524BC35ACD264530354A65 ON bolt_relations (to_contenttype, to_id)
CREATE INDEX IDX_ABAA1200989D9B62 ON bolt_taxonomy (slug)
...
```

If this is OK I will add the test :)

### DIFF
--- a/src/Nut/DatabaseRepair.php
+++ b/src/Nut/DatabaseRepair.php
@@ -4,6 +4,7 @@ namespace Bolt\Nut;
 
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputOption;
 
 /**
  * Nut command to repair/update database schema
@@ -17,6 +18,7 @@ class DatabaseRepair extends BaseCommand
     {
         $this
             ->setName('database:update')
+            ->addOption('dump-sql', null, InputOption::VALUE_NONE, 'Dump the SQL, do not execute the query')
             ->setDescription('Repair and/or update the database.')
         ;
     }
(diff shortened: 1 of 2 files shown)


GawainLynch:
OK, defo feature … so we'll sneak into 3.3 … this is the "Gawain Sucks at UX" question, is this better in `database:update` or `database:check` … other than that question :+1:

dantleech:
> is this better in database:update or database:check 

Good question - I didn't realise that `database:check` existed.

I would say it more naturally belongs here, as it dumps the SQL which would otherwise be executed on the database by the command.

Coming from a Doctrine background the analagous command is `./app/console doctrine:schema --dump-sql`.

dantleech:
But will add a test :)

GawainLynch:
> Coming from a Doctrine background the analagous command is

Good enough for me … right or wrong, am very much for keeping tact with upstream

> But will add a test :)

:+1: :+1: :+1: :+1: :+1: :+1: :+1:

GawainLynch:
Thanks mate, and **massive thanks** for adding a test!

FYI I took the liberty of attacking the test fails as there was some other O.C.D. type clean ups I wanted to add :wink:

dantleech:
awesome. I blind pushed yesterday as I had to leave, so wasn't sure if the tests worked. Thanks for fixing :)

PR #6500, opened 2017-03-21
Allow Canonical Url to be overridden

CarsonF:
`Canonical::setOverride` can now be used _during_ the request cycle to change the canonical url. Absolute and relative paths are resolved with the current request.

It was also refactored to not depend on `RequestStack` and `RequestContext`.

Test coverage is 100%.


bobdenotter:
Should this be documented, so that people actually know this can be done and how?

CarsonF:
Possibly....IMO it's an advanced use case. SEO extension is the only thing I can think of that needs this code.

GawainLynch:
> Test coverage is 100%.

Is there any better way to a merger's heart :smile:

PR #6502, opened 2017-03-21
Renamed local extensions to bundled extensions

CarsonF:
This is a follow up to #6419.

As discussed in Slack, we don't want to refer to extensions managed by users as local extensions since the concepts have some differences and we want to reduce confusion.

The name that seemed to be the most agreed upon was "Bundled".
The name opposite of that that seems to make the most sense to me is still "Managed".

Managed Extensions:
- Are installed via Composer (via UI/Nut)
- Have version constraints
- Handle syncing assets to the user's web directory

Bundled Extensions:
- Are registered by users (via code or `.bolt.yml`)
- Users manage autoloading classes
- Do not have version constraints - Users need to make sure the extension code is compatible
- Users need to manage assets themselves
- Cannot be added/removed via UI or Nut
- Can be packages from project's composer.json

**NOTE:** The only difference between managed and bundled (besides asset syncing) is the install method. Extensions can and _should_ be written agnostic of this.

/cc @bolt/backend-developers


GawainLynch:
> I'm still not convinced "Bundled" is the absolute best name for this

To its actual target group, I'd say it is a lot better … closer to what SF uses, and more familiar.

PR #6503, opened 2017-03-21
[4.0] Changelog & deprecation removals

GawainLynch:


CarsonF:
Are we organizing by change sets? It just a bit confusing to have "php requirement 5.6+" and "php requirement 7.0+"

GawainLynch:
> Are we organizing by change sets?

No, as was discussed, I am just dumping as much as I can in the 4.x section as we go to; a) try and lighten the load on Bob near release; b) make the job of compiling the change list easier at some point.

tl;dr Leave it alone or volunteer to do it :wink:

PR #6504, opened 2017-03-22
Make thrown SecurityPolicy in Twig-in-content show Exception, instead of outputting HTML on page. 

bobdenotter:
Instead of outputting a SecurityPolicy error for Twig-in-content inline in the HTML, we now throw an exception.

### DIFF
--- a/src/Legacy/Content.php
+++ b/src/Legacy/Content.php
@@ -226,16 +226,9 @@ class Content implements \ArrayAccess
 
         $snippet = html_entity_decode($snippet, ENT_QUOTES, 'UTF-8');
 
-        try {
-            $template = $this->app['twig']->createTemplate((string) $snippet);
+        $template = $this->app['twig']->createTemplate((string) $snippet);
 
-            return twig_include($this->app['twig'], $this->getTemplateContext(), $template, [], true, false, true);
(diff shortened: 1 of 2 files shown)


bobdenotter:
@CarsonF Yeah, that's much better. It allows you to find where it comes from, as well as the possibility to disable it in production. 👍 

screen shot 2017-03-22 at 15 53 43

CarsonF:
I was thinking we could have a custom Exception class, maybe `Bolt\Exception\Content\RenderSecurityException`, that has 'fieldName` property and requires a Twig SecurityError exception as the previous class.
Then have special logic for this in our ExceptionListener.

bobdenotter:
Yeah, I'm not sure if i'm the right person to build that, then.

CarsonF:
I can commit an outline for you to run with if you want?

bobdenotter:
Sure, that'd help a lot! :-)

CarsonF:
Ok that's at least a direction for you to work with. I'm sure it needs rough edges worked in both PHP and Twig.

bobdenotter:
I think the direction here is good, but we shouldn't be adding this much code this late in the 3.3 phase. 

I propose we merge in the "simple" fix, and bump this to 3.4.

PR #6505, opened 2017-03-22
Clarify error message for missing assets in extensions.

bobdenotter:
We assume that the action to take is evident, but in practice it needs pointing out quite often.

### DIFF
--- a/src/Extension/AssetTrait.php
+++ b/src/Extension/AssetTrait.php
@@ -196,7 +196,8 @@ trait AssetTrait
         }
 
         $message = sprintf(
-            "Couldn't add file asset '%s': File does not exist in either %s or %s directories. Make sure the file exists in either of these locations, by placing the file there manually (for Bundled Extensions) or by uninstalling / installing te extension again (for Managed Extensions).",
+            "Couldn't add file asset '%s': File does not exist in either %s or %s directories. Make sure the file exists in either of these locations, by ".
+            "placing the file there manually (for Bundled Extensions) or by uninstalling / installing the extension again (for Managed Extensions).",
             $path,
             $this->getWebDirectory()->getFullPath(),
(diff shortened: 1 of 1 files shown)


bobdenotter:
Currently, the only way I know is by "Removing and re-adding the extension", because you can't trigger the process using one of the `app/nut` commands. Right, @GawainLynch ?

GawainLynch:
Yeah, we're not offering any "sync" so that part is incorrect to add.

CarsonF:
Technically, for managed extensions, there's no need to. As their assets should be static, not changing, and they are synced when installed or updated.

bobdenotter:
Updated.

CarsonF:
Shouldn't this exception only show up for extension devs that make a mistake specifying their assets?

bobdenotter:
In an ideal world, yes. I've had it happen a few times though, that i "lost" my assets. Often my own fault, but the pointer makes sense.

CarsonF:
I just worry that this is now misguiding extension devs

bobdenotter:
I think the current version is OK. Let's merge it, and onwards. :-)

GawainLynch:
Well, I am not seeing the confusing part myself … but the ordering is wrong. The bulk of our users should be using normal extensions, so should be mentioned first IMHO.

bobdenotter:
> The bulk of our users should be using normal extensions, so should be mentioned first IMHO.

Agreed, but it will happen a _lot_ less often with managed extensions, while it happens quite often when someone is poking around creating their own. That's why I put the usecase i expect happens most at the front.

GawainLynch:
> Agreed, but it will happen a lot less often with managed extensions

Where are we setting up the tally board?

It's a system log message … better one's in later PRs welcome I guess.

PR #6518, opened 2017-03-26
Force saving of repeaters, even when empty

### LINKED ISSUE #6086: [BUG] When deleting image repeatsets, deleting all sets results in 500 OR saves but doesn't apply
Currently if you delete all sets from a repeatset for a record, the record cannot be saved and returns a 500 error.

However, if you delete all but 1 of the sets, and clear the image link, it allows a save and performs successfully.

The Delete Set button is greyed out, but the elments are still removed upon clicking. Maybe this is the cause of the bug?

Theres is also a scenario where it allows a save, but doesn't apply it. So upon refresh the data is there like it's never been edited or applied.

Details
-------

 - Relevant Bolt Version: 3.2.1
 - Install type: Composer

rossriley:
Fixes #6086

This adds an empty hidden value to repeater fields which ensures that a value is always posted even if there are no other sets posted.

Fixes a bug where deleting all the sets did not work because the field was not present.

### DIFF
--- a/app/view/twig/editcontent/fields/_repeater.twig
+++ b/app/view/twig/editcontent/fields/_repeater.twig
@@ -23,6 +23,9 @@
         </div>
     {% endif %}
 
+    {# This ensures that an empty value is always submitted even if there are no subsequent repeater sets #}
+    <input type="hidden" name="{{ name }}[]">
+
     <div class="repeater-slot">
         <script type="text/template">
             {% set index = '#' %}


bobdenotter:
Should this not go into 3.2 as well?

GawainLynch:
Also merged down into `release/3.2` via d28fb42403ee2e0b60d238c1f6e2790ff92c6bfd

PR #6519, opened 2017-03-26
[4.0] Remove more interal deprecations

GawainLynch:


GawainLynch:
> The only thing I'm wondering about is whether more of CallbackResolver should be removed, like the classmap logic.

Yeah, looked at that and thought, "Have to go make diner, and this one's not as easy as it looks … sending in as-is." :laughing:

CarsonF:
@GawainLynch Can you also add a deprecation note for classmap usage to 3.3?

PR #6520, opened 2017-03-27
[3.2] Version compare PHPDoc OCD-off

GawainLynch:
The discussion just didn't sit with me, so test case:

```php
<?php

namespace Koala;

use Symfony\Component\Console\Helper\Table;
use Symfony\Component\Console\Output\ConsoleOutput;

include_once __DIR__ . '/vendor/autoload.php';

final class Version
{
    //const VERSION = '3.2.0';
    const VERSION = '3.2.0 beta 9';

    public static function compare($version, $operator)
    {
        $currentVersion = str_replace(' ', '', strtolower(static::VERSION));
        $version = str_replace(' ', '', strtolower($version));

        return version_compare($version, $currentVersion, $operator);
    }
}
$operators = [
    //'<',
    //'>',
    //'<=',
    '>=',
];
$versions = [
    '3.1',
    '3.1.0',
    '3.2',
    '3.2.0-beta8',
    '3.2.0-beta9',
    '3.2.0',
    '3.2.1',
    '3.3',
    '3.3.0',
];
$output = new ConsoleOutput();
$output->writeln('');
$output->writeln(sprintf('Bolt version: "%s"', Version::VERSION));
foreach ($operators as $operator) {
    $table = new Table($output);
    $table->setHeaders(['Compared Version', '', 'Result']);

    foreach ($versions as $version) {
        $table->addRow([
            $version,
            $operator,
            Version::compare($version, $operator) ? 'true' : 'false'
        ]);
    }

    $table->render();
}

```
… plus updated PHPDoc :rofl:

### DIFF
--- a/src/Version.php
+++ b/src/Version.php
@@ -32,11 +32,29 @@ final class Version
     }
 
     /**
-     * Compares a version to Bolt's version.
+     * Compares a semantic version (x.y.z) against Bolt's version, given a
+     * specified comparison operator.
      *
-     * Note: Be sure to include the `.z` number in the version given, as omitting it can give inconsistent results.
-     * For example, if the current version is `3.3.0`, `compare('3.3', '>=')` returns false. In reality you want
-     * the check to return true for this case, and `compare('3.3.0', '>=')` _does_ return true.
+     * Note 1:
+     * Be sure to include the `.z` number in the version given, as
+     * omitting it can give inconsistent results.
+     *
+     * e.g. If the version of Bolt was '3.2.0' (or greater), then:
+     *     `Version::compare('3.2', '>=');`
+     * is NOT equal to, or greater than, Bolt's version.
+     *
+     * Note 2:
+     * Pre-release versions, such as 3.2.0-beta1, are considered lower
+     * than their final release counterparts (like 2.3.0). As you may notice,
+     * the difference being that Bolt '3.2.0-beta1' is considered LOWER than
+     * the `compare($version)` value of '3.2.0'.
+     *
+     * e.g. If the version of Bolt was '3.2.0 beta 1', then:
+     *     `Version::compare('3.2.0', '>=');`
+     * is equal to, or greater than, Bolt's version.
+     *
+     * @see http://semver.org/ For an explanation on semantic versioning.
+     * @see http://php.net/manual/en/function.version-compare.php#refsect1-function.version-compare-notes Notes on version_compare
      *
      * @param string $version  The version to compare.
      * @param string $operator The comparison operator: <, <=, >, >=, ==, !=


src/Version.php
```diff
@@ -32,11 +32,56 @@ public static function isStable()
...
+     * | 3.2.0            | >= | true   |
+     * | 3.2.1            | >= | true   |
+     * | 3.3              | >= | true   |
+     * | 3.3.0            | >= | true   |
+     * +------------------+----+--------+
+     *
+     * Note that pre-release versions, such as 3.2.0-beta, are considered lower
+     * than their final release counterparts (like 2.3.0).  As you may notice,
+     * the difference being that '3.2.0-beta9' is considered LOWER than '3.2.0',
+     * and '3.2'.
```
bobdenotter: And 3.2 what? 🤔 

I'm guessing that comma on the previous line should be removed.

src/Version.php
```diff
@@ -32,11 +32,56 @@ public static function isStable()
...
+     * | 3.2.0-beta8      | >= | false  |
+     * | 3.2.0-beta9      | >= | true   |
+     * | 3.2.0            | >= | true   |
+     * | 3.2.1            | >= | true   |
+     * | 3.3              | >= | true   |
+     * | 3.3.0            | >= | true   |
+     * +------------------+----+--------+
+     *
+     * Note that pre-release versions, such as 3.2.0-beta, are considered lower
+     * than their final release counterparts (like 2.3.0).  As you may notice,
```
bobdenotter: `like 2.3.0` or `like 3.2.0`?

Two spaces after the period. You old-fashioned, you.

bobdenotter:
I don't mind having them in. In practice I've never thought "Gee, I wish this was documented less". ;-)

src/Version.php
```diff
@@ -32,11 +32,56 @@ public static function isStable()
...
-     * For example, if the current version is `3.3.0`, `compare('3.3', '>=')` returns false. In reality you want
-     * the check to return true for this case, and `compare('3.3.0', '>=')` _does_ return true.
+     * @see http://semver.org/ For an exmplaination on semantic versioning.
+     *
+     * Note: Be sure to include the `.z` number in the version given, as
+     * omitting it can give inconsistent results. See below at the results in
+     * Tables 1 & Table 2 for more detail.
+     *
+     * Note: Depending on the stability of the VERSION constant, the output of
+     * this may give, initially, unexpected results.
```
CarsonF: I don't agree with this. How does the stability of the version give unexpected results?

GawainLynch:
OK, so updated and trimmed down … let me know what's thought and I'll adjust and squash to get a workable result.

src/Version.php
```diff
@@ -32,11 +32,56 @@ public static function isStable()
...
 
     /**
-     * Compares a version to Bolt's version.
+     * Compares a semantic version against Bolt's version, given a specified
+     * comparison operator.
      *
-     * Note: Be sure to include the `.z` number in the version given, as omitting it can give inconsistent results.
-     * For example, if the current version is `3.3.0`, `compare('3.3', '>=')` returns false. In reality you want
-     * the check to return true for this case, and `compare('3.3.0', '>=')` _does_ return true.
+     * @see http://semver.org/ For an exmplaination on semantic versioning.
```
CarsonF: IntelliJ/PHPStorm doesn't parse this inline, and thus everything below is left out.

You could do an inline tag:
```
* {@see http://semver.org/ See SemVer} for an explanation on semantic versioning.
```

Or no tag:
```
* See http://semver.org/ for an explanation on semantic versioning.
```

Or put it at the bottom of the phpdoc. 

Or leave it out. 

Personally, I think not even mentioning semver here is better, and just link to [version_compare](http://php.net/manual/en/function.version-compare.php)

----

Also "exmplaination" is misspelled.

GawainLynch: > IntelliJ/PHPStorm doesn't parse this inline

Then it needs to [update](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/see.html)

CarsonF: Do you see any description below those tags? Nope.

GawainLynch: > Personally, I think not even mentioning semver here is better, and just link to version_compare

My thinking is the world is heading semver … any chance we get to clarify that, the better.

CarsonF: > A DocBlock roughly exists of 3 sections:
>
> Summary; a one-liner which globally states the function of the documented element.
> Description; an extended description of the function of the documented element; may > contain markup and inline tags.
> Tags; a series of descriptors for properties of this element; such as @param and @return.
https://www.phpdoc.org/docs/latest/references/phpdoc/basic-syntax.html#sections

src/Version.php
GawainLynch: Copy/pasta … fixing

src/Version.php
```diff
@@ -32,11 +32,46 @@ public static function isStable()
...
+     * | 3.2          | >= | false  |
+     * | 3.2.0-beta8  | >= | false  |
+     * | 3.2.0-beta9  | >= | true   |
+     * | 3.2.0        | >= | true   |
+     * +--------------+----+--------+
+     *
+     * Note: Pre-release versions, such as 3.2.0-beta, are considered lower
+     * than their final release counterparts (like 2.3.0). As you may notice,
+     * the difference being that '3.2.0-beta9' is considered LOWER than both
+     * Bolt '3.2.0', and '3.2'.
```
CarsonF: I would just keep this paragraph and remove the two paragraphs and the table above it. They are all saying the same thing.

src/Version.php
GawainLynch: Well, more what it is trying to say is, if there is a non-stable part to the version … blah

src/Version.php
GawainLynch: WFM … How about now?

src/Version.php
```diff
@@ -32,11 +32,28 @@ public static function isStable()
...
+     * | 3.2      | >= | false  |
+     * | 3.2.0    | >= | true   |
+     * +----------+----+--------+
+     *
+     * Note: Pre-release versions, such as 3.2.0-beta, are considered lower
+     * than their final release counterparts (like 2.3.0). As you may notice,
+     * the difference being that Bolt '3.2.0-beta9' is considered LOWER than
+     * both `$version` '3.2.0', and '3.2'.
+     *
+     * @see http://semver.org/ For an exmplaination on semantic versioning.
```
CarsonF: exmplaination => explanation

GawainLynch: I've fuxed it nuw.

GawainLynch:
> I think examples of a couple use cases would help the most.

Updated to add some, a feedbag welcome … or is that feedback, brain go :boom: :boom: 

:stuck_out_tongue_winking_eye:

bobdenotter:
Sorry, but this is still confusing me. 
I think we're overthinking it, because there's one expected usecase that I can think of: Say a developer is working on an extension or a bit of custom code. All they want do is this: 

```
if (Version::compare('3.2.0', '>=')) {
  // This gets executed if the end user is on Bolt 3.2.0 or up. 
}

if (Version::compare('3.2.0', '<')) {
  // This gets executed if the end user is on a Bolt version _lower_ than 3.2.0
}
```

(Note: I have no idea if the above is correct, without actually trying it out.)

GawainLynch:
> because there's one expected usecase that I can think of …

* A specific version needs to react a certain way
* A range of specific versions needs to react a certain way
* …

bobdenotter:
Shall we just add those examples, and get this merged in?

bobdenotter:
Looks good to me! :shipit:

PR #6525, opened 2017-03-29
Fix bug in repeater buttons

eduardomart:
Problem:
Currently, when reaching the `limit` set in a repeater definition, the *[Add new Repeater set]* button gets grayed out but clicking it still adds a new repeater set, and the *[Duplicate set]* button gets wrongly grayed out.

Fix:
The `.disabled` class was being applied to the *[Duplicate set]* buttons, instead of adding the `“disabled”` property to the *[Add new Repeater set]* button.

### DIFF
--- a/app/src/js/widgets/fields/fieldRepeater.js
+++ b/app/src/js/widgets/fields/fieldRepeater.js
@@ -246,10 +246,10 @@
 
             if (this._count >= this.options.maximum) {
                 this._ui.add.addClass('disabled');
-                this.element.find('.duplicate-button').addClass('disabled');
+                this._ui.add.prop('disabled', 'disabled');
             } else {
                 this._ui.add.removeClass('disabled');
-                this.element.find('.duplicate-button').removeClass('disabled');
+                this._ui.add.prop('disabled', false);
             }
 
             if (this._count <= this.options.minimum) {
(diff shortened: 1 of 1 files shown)


app/src/js/widgets/fields/fieldRepeater.js
```diff
@@ -246,10 +246,10 @@
 
             if (this._count >= this.options.maximum) {
                 this._ui.add.addClass('disabled');
-                this.element.find('.duplicate-button').addClass('disabled');
+                this._ui.add.prop('disabled', 'disabled');
```
rossriley: Can you also add this line underneath:
    
    this.element.find('.duplicate-button').prop('disabled', 'disabled');

app/src/js/widgets/fields/fieldRepeater.js
```diff
@@ -246,10 +246,10 @@
...
             } else {
                 this._ui.add.removeClass('disabled');
-                this.element.find('.duplicate-button').removeClass('disabled');
+                this._ui.add.prop('disabled', false);
```
rossriley: and this line underneath here:

    this.element.find('.duplicate-button').prop('disabled', false);

eduardomart:
@ross I should have known that, thanks for catching it!

ross:
> @ross I should have known that, thanks for catching it!

I think you want @rossriley

eduardomart:
@ross you're right, sorry, I am used to his Slack handle :-)

eduardomart:
@rossriley 

I updated the files. Just pinging you since I am not sure you got a notification from git, I didn't.

bobdenotter:
@rossriley This should be GTG now, right?

PR #6528, opened 2017-03-29
Fix LowlevelChecks::disableApacheChecks() not passing value to RM

### LINKED ISSUE #6455: [3.3] Regression: No .htaccess throws exception when Apache checks are disabled.
This is a pretty new one, only just showing up after a recent pull of release/3.3

Loading up the frontend now throws the following:
screen shot 2017-03-04 at 14 12 08

This is still happening when the Apache checks are disabled as detailed in the exception.

CarsonF:
Fixes #6455

### DIFF
--- a/src/Configuration/LowlevelChecks.php
+++ b/src/Configuration/LowlevelChecks.php
@@ -59,10 +59,9 @@ class LowlevelChecks
 
     public function __set($property, $value)
     {
-        if ($this->config && $property === 'disableApacheChecks' && $value === true) {
-            Deprecated::method(3.3, 'Extend "config.validator" service and remove the apache check there.');
-
-            $this->config->disableApacheChecks();
+        if ($property === 'disableApacheChecks' && $value === true) {
+            $this->disableApacheChecks();
+            return;
         }
 
         $this->$property = $value;
@@ -205,7 +204,12 @@ class LowlevelChecks
 
     public function disableApacheChecks()
     {
+        Deprecated::method(3.3, 'Extend "config.validator" service and remove the apache check there.');
+
         $this->disableApacheChecks = true;
+        if ($this->config) {
+            $this->config->disableApacheChecks();
+        }
     }
 
     /**


GawainLynch:
LGTM in theory, but have no Apache to test on.

CarsonF:
Should we change the deprecation message to
`Set $app['config.validator.apache_enabled'] = false`?

I guess it depends on what we want to do with that check in 4.0.
Personally I think it should be moved to `bolt/starter-kit`.

GawainLynch:
> Should we change the deprecation message to Set `$app['config.validator.apache_enabled'] = false`?

Easier

> Personally I think it should be moved to bolt/starter-kit

Which is now https://github.com/bolt/configuration-notices

CarsonF:
I kinda wonder if the check is even relevant. Is it supposed to check for misconfiguration? If so and it doesn't exist can you even get to that point in code? (Assuming it's not configured via etc conf files which wouldn't need the check either). @bobdenotter?

bobdenotter:
I have no clue how to trigger this. Is there a hidden setting in `config.yml` for this?

bobdenotter:
Ok, the code says: 

```
     * If you see this error and want to disable it, call $config->getVerifier()->disableApacheChecks();
     * inside your bootstrap.php file, just before the call to $config->verify().
```

My `bootstrap.php` doesn't have `$config->verify()`. Not sure where to put it.

CarsonF:
> My bootstrap.php doesn't have `$config->verify()`. Not sure where to put it.

Did you see the error? I'm asking if the check is worth anything, how does disabling it help that? Also, `verify()` does nothing now.

GawainLynch:
Setting up Apache … give me a few :disappointed:

GawainLynch:
> I'm asking if the check is worth anything

Right now, working back-end after removing `.htaccess`
screenshot from 2017-03-30 08-01-56

With PR, `.htaccess` **only** removed (i.e. changes not properly moved to vhost file), and `$app['config.validator.apache_enabled'] = false;` added to `index.php` before the `$app->run();
screenshot from 2017-03-30 08-05-53
… the image 404 is obviously expected as I haven't move the rewrites.

The back-end is still inaccessible.

bobdenotter:
But that's the point of this "feature", right? Bolt gives a warning about a missing `.htaccess` to novice users who sometimes misplace it. This "feature" basically allows you to say "Don't tell me about the missing .htaccess, i will handle it myself, because I know better." 

So, In that regard, it does exactlly what it's told: You broke your install, but Bolt won't tell you about it, because you specifically requested it not to. 

That install, without the `.htaccess` will now probably work correctly on nginx.

GawainLynch:
> But that's the point of this "feature", right? Bolt gives a warning about a missing .htaccess to novice users who sometimes misplace it. 

Yes, we get that … but it is only so effective … as my point, it will never work if you're trying to access a rewritten URL, which is "Part A" of why `bolt/configuration-notices` might be a better place for it.

Part B is that to put this mildly … `.htaccess` should **never ever ever** be used! Those that are still stuck with Apache (poor souls) move their rewrites, etc, to the vhost and disable `.htacess` use for both performance and security … Nginx users just laugh at you for using Apache, laugh to the point of tears when you use `.htaccess`.

So @CarsonF's question was around how much use is there even having the check, as it **literally** blocks everyone who is _not a novice_ from configuring their server correctly.

GawainLynch:
…not to mention, this check over the last year has absorbed so much dev time, it isn't funny, for something that only partly works, for a percentage of the user population.

bobdenotter:
The alternative, if we remove the check, is that people on nginx will just have to suffer the unneeded `.htaccess` in the webroot.

GawainLynch:
OK, I need more :coffee: … or one of us does :laughing: 

The idea of the check is to see if the file is present. Removing the check (whatever the reason … run with me) means nobody "needs" the `.htaccess` file present: e.g. intermediate+ Apache users, and Nginx users

Background to Carson's question. He wanted to nuke that test ages ago, but I resisted as I expected negative push back from those representing the novice group. 

However, regardless the notice itself is inconsistent, should the file be missing:
  * If you try to hit a rewritten URL, you will only get a **404 page**. Enabled or disabled
  * If notice is enabled, and you hit a normal route, you will get an **exception page**
  * If notice is disabled, and you hit a normal route, **normal page** appears, **minus any rewritten assets**

So thinking about i, `bolt/configuration-notices` won't work either, as that is admin section and requires rewrites to work.

bobdenotter:
Ah, like that. I meant "the extra rule to bypass the check". The check itself needs to stay, like you say. 

> If you try to hit a rewritten URL, you will only get a 404 page. Enabled or disabled

That's the notice, alerting the user that the file is missing. Seems correct.

> If notice is enabled, and you hit a normal route, you will get an exception page

That's because htaccess is missing, and it can't rewrite. Seems correct

> If notice is disabled, and you hit a normal route, normal page appears, minus any rewritten assets

This is the "I know better" part kicking in: No notice/exception. It's up to you to fix your rewrites in nginx.conf or apache vhost. Also correct behaviour.

> So thinking about it, bolt/configuration-notices won't work either, as that is admin section and requires rewrites to work.

Exactly, it won't work, because the novice user won't get to that point.

GawainLynch:
> However, regardless the notice itself is inconsistent, should the file be missing:

Yes, we know these three are correct. The point I am raising is that even as a novice, the results are inconsistent … and with BC is just `<insert rude word here>` painful to work with … which is where I go back again to what (I think) Carson is trying to get to the bottom of.

1. Can we break BC and be done with it? — I vote yes, disabling it is **not** a novice feature, we'll just co… co… co… communicate 
2. Is the test *really* useful? — I vote "no", even for novices
3. If the answer to 2. is agreed to be "no", can we remove it? — I vote "yes"

GawainLynch:
@CarsonF also, the BC disable does seem to work OK using this as the project `app/bootstrap.php`, called from `public/index.php` on Apache:
```php
return call_user_func(
    function () {
        require_once __DIR__ . '/../vendor/autoload.php';

        $configuration = new Bolt\Configuration\Composer(dirname(__DIR__));
        $configuration->getVerifier()->disableApacheChecks();
        $configuration->verify();

        $app = new Bolt\Application(array('resources' => $configuration));
        $app->initialize();

        return $app;
    }
);
```

```
$ ls -al
total 64
drwxrwx---+ 7 apache apache 4096 Mar 30 09:23 .
drwxrwx---+ 6 apache apache 4096 Mar 30 07:31 ..
drwxrwx---+ 3 apache apache 4096 Mar 30 07:28 bolt-public
drwxrwx---+ 2 apache apache 4096 Mar 25 09:48 extensions
drwxrwx---+ 2 apache apache 4096 Mar 30 07:28 files
-rw-rw----+ 1 apache apache  507 Mar 30 09:19 index.php
drwxrwx---+ 3 apache apache 4096 Mar 30 07:28 theme
drwxrwx---+ 2 apache apache 4096 Mar 25 09:48 thumbs
```

GawainLynch:
Works, merging, debating forward steps later

bobdenotter:
👍 

I don't think it needs debating. It works, and we have bigger fish to fry. This is (at best) an edge-case that impacts user who purposefully remove the `.htaccess` file. They can deal with it. 
My main concern was the novice users, and for them it'll still work as before, so i'm all good.

GawainLynch:
> I don't think it needs debating. It works, and we have bigger fish to fry.

Well, that's where you might want to get on page. 

The validators as they are now are a stop-gap to get the validation working with BC, and to get them out of the pre-registration cycle, as that was invoking boot before registration.

In v4 we want to nuke the whole darn lot, as all we have left is:
  * `Validation\Apache`
  * `Validation\Cache`
  * `Validation\Configuration`
  * `Validation\Database`

But:
  * `Validation\Configuration` goes in 3.4.
  * `Validation\Cache` goes with the new caching in 4.0 (coming to a PR near you in July)
  * `Validation\Database` is better handled via the exception handler

… see a straggler? 

Also, as I said, there are lots and lots of reasons that exception would never appear in real use for a novice. So we end up with a whole validation logic event system in v4, that was there to maintain BC, to support a validation check that only works some of the time, for some of the target groups.

Also, one of the reasons people disable/remove `.htaccess` in Apache (instead of sensibly moving to Nginx) is for the slight performance hit of reading that file. This validation _method_ kinda goes against that … micro-op, I know, but the entire back-end team are in that "group" that do this :wink: 

i.e. Open to suggestions on better ways of handling this "problem" for novices … Nerf bats comes to mind :stuck_out_tongue_winking_eye: … but unfortunately this is not "bigger fish to fry".

PR #6533, opened 2017-04-01
[3.3][MySQL] Set database character set & collation as configured

### LINKED ISSUE #6430: [BUG] MySQL connection charset/collation hardcoded in DoctrineListener
As I need to save 4-byte utf8 values like emojis in content, I need to change the connection charset to utf8mb4. Though there is a possibility to change charset & collation in the database section of config.yml  (see #4485 and #4498), this does only apply for schema changes. The connection charset and collation is hardcoded  in [DoctrineListener](https://github.com/bolt/bolt/blob/6307ce400eb9e74b7f6140bc7a8e684ce38b50c4/src/EventListener/DoctrineListener.php#L82).

I don't know the correct/clean way to fix this directly in Bolt, but I'm currently using following workaround extension to overwrite the default DoctrineListener and allowing to set the connection charset & collation trough config.yml: https://gist.github.com/maetthu/ae4a13db20c5a0480c7d975361f2b2c6

GawainLynch:
Fixes #6430

We'll need to follow up with docs on `utf8mb4` "fun"

### DIFF
--- a/src/EventListener/DoctrineListener.php
+++ b/src/EventListener/DoctrineListener.php
@@ -2,6 +2,7 @@
 
 namespace Bolt\EventListener;
 
+use Bolt\Config;
 use Bolt\Events\FailedConnectionEvent;
 use Bolt\Exception\Database\DatabaseConnectionException;
 use Doctrine\Common\EventSubscriber;
@@ -20,13 +21,17 @@ class DoctrineListener implements EventSubscriber
 {
     use LoggerAwareTrait;
 
+    /** @var Config */
+    private $config;
+
     /**
      * Constructor.
      *
      * @param LoggerInterface $logger
      */
-    public function __construct(LoggerInterface $logger)
+    public function __construct(Config $config, LoggerInterface $logger)
     {
+        $this->config = $config;
         $this->setLogger($logger);
     }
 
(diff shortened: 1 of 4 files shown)


src/EventListener/DoctrineListener.php
```diff
@@ -65,13 +70,10 @@ public function postConnect(ConnectionEventArgs $args)
...
-            // Set utf8 on names and connection, as all tables have this charset. We don't
-            // also do 'SET CHARACTER SET utf8', because it will actually reset the
-            // character_set_connection and collation_connection to @@character_set_database
-            // and @@collation_database respectively.
-            // see: http://stackoverflow.com/questions/1566602/is-set-character-set-utf8-necessary
-            $db->executeQuery('SET NAMES utf8');
-            $db->executeQuery('SET CHARACTER_SET_CONNECTION = utf8');
+            // Set database character set & collation as configured
+            $charset   = $this->config->get('general/database/charset', 'utf8');
+            $collation = $this->config->get('general/database/collate', 'utf8_general_ci');
```
bobdenotter: The defaults are handled in Config.php. We can remove them here.

GawainLynch: Copy/pasta … gone

bobdenotter:
:shipit:

PR #6540, opened 2017-04-04
[WIP] Update _files.twig

matesko:
If there is incomplete upload of image and file size is 0 bolt throws error when accessing width and height

### DIFF
--- a/app/view/twig/files/_files.twig
+++ b/app/view/twig/files/_files.twig
@@ -64,7 +64,7 @@
                     <span class="buic-moment" title="{{ file.size }} bytes ({{ file.sizeformatted(true, false) }})">
                         <b>{{ file.sizeformatted(true, true) }}</b></span>,
                     {{ buic_moment(file.timestamp) }}
-                    {% if file.image %}
+                    {% if file.image and file.size > 0 %}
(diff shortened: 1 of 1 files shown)


CarsonF:
Is the image content actually empty?

matesko:
Yes the image is actually empty, file size is 0

CarsonF:
What I'm wondering is why the upload left an empty file. Can you reproduce that case?

matesko:
I was unable to reproduce it. Some user uploaded files, and two were empty...

bobdenotter:
It can happen that the image is 0 bytes. When you go over quota on shared hosting for example. Bolt shouldn't break if that happens, so I think this is a useful addition. 👍

CarsonF:
I do too. But I think this is the wrong spot. If the image is 0 bytes the upload process should fail.

bobdenotter:
People can also upload images by FTP or whatever. Bolt should not break if this happens, in any case.

CarsonF:
That's true. Still if we are saying 0 byte images are ok, then `Image\Info` should handle it.

bobdenotter:
True.

CarsonF:
Fixed with https://github.com/bolt/filesystem/pull/47

Thanks @matesko for bringing it to our attention 👍

PR #6541, opened 2017-04-04
replaced use of deprecated request object

ToBe998:
The template used the now deprecated global `request('foo')` object, which causes a twig exception when rendering. This exception can not be overridden if happening while rendering a widgets template, causing a crash in the backend.

Replaced (as specified by @CarsonF) with `global.request.get('foo')`.

*Note*: There are other occurrences which use `app.request.query.get('foo')`. Should probably use same new method?

### DIFF
--- a/app/theme_defaults/_sub_searchbox.twig
+++ b/app/theme_defaults/_sub_searchbox.twig
@@ -2,7 +2,7 @@
 
     <form method="get" action="/search" id="searchform" enctype="text/plain">
 
-        <input type="text" name="search" value="{{ app.request.get('search') }}" placeholder="{{ __('general.phrase.search-for-ellipsis') }}" >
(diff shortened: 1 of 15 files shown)


bobdenotter:
Hi @ToBe998, 

Thanks! 

> Note: There are other occurrences which use app.request.query.get('foo'). Should probably use same new method?

Yes, I would think so. :-)

CarsonF:
Note that `request.get()` is a shortcut for `request.query.get()`. The shortcut method checks GET parameters, then POST parameters, then request attributes. It's probably better to include the `.query` part as that's more specific to what you are actually looking for.

But yes, any usage of `app.request` should be replaced with `global.request`.

GawainLynch:
The :koala: did it …

PR #6543, opened 2017-04-04
[3.3.] Nut debug tweaks

GawainLynch:
Mostly tweaks to allow different sorting, and a cople of table "pretty"


src/Nut/DebugEvents.php
```diff
@@ -21,6 +23,7 @@ protected function configure()
         $this
             ->setName('debug:events')
             ->setDescription('System events, and target callable debug dumper.')
+            ->addOption('sort-listener', null, InputOption::VALUE_NONE, 'Sort events in order of listener name.')
```
CarsonF: Would it be better to have a `sort` option that can be given values like `name`, `default`, etc.?

src/Nut/DebugRoutes.php
```diff
@@ -36,9 +37,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $table = new Table($output);
         $table->setHeaders([
-            'Bind name',
-            'Path matching parameter',
-            'Method(s)',
+            [new TableCell('System routes', ['colspan' => 3])],
+            ['Bind Name', 'Path Match Pattern', 'Method(s)']
```
CarsonF: `Bind Name` - Maybe just `Name`? You call `bind()` to set the _route name_, and "bind" isn't used anywhere else.

`Path Match Pattern` - Maybe just `Path`? Symfony's referred to the _path_ as [_pattern_](https://github.com/symfony/routing/blob/2.8/Route.php#L150), but it is deprecated.

src/Nut/DebugServiceRegistration.php
```diff
@@ -30,27 +34,43 @@ protected function configure()
...
+    private function displayRegistration(InputInterface $input, OutputInterface $output)
     {
         $table = new Table($output);
+        $rightAligned = new TableStyle();
+        $rightAligned->setPadType(STR_PAD_LEFT);
+
         $table->setHeaders([
-            '#',
-            'Provider',
+            [new TableCell('System service provider registration boot order', ['colspan' => 2])],
```
CarsonF: How about just `Service Providers`? You have a lot of extra words in there that don't mean anything :)

src/Nut/DebugRoutes.php
GawainLynch: Going for "Route Name" and "Path" then

src/Nut/DebugServiceRegistration.php
GawainLynch: WFM

CarsonF: Oh wait, I guess it should be singular, right?

src/Nut/DebugEvents.php
GawainLynch: Have a look at the three … it only makes sense to give so many sort options, and this way is easy too, no?

src/Nut/DebugServiceRegistration.php
GawainLynch: Why, it is the title block for the service providers table.

CarsonF: Yep, you're right nvm.

src/Nut/DebugEvents.php
CarsonF: I guess so. In that case `sort-name` makes more sense. Same for service providers command.

src/Nut/DebugEvents.php
```diff
@@ -29,33 +32,44 @@ protected function configure()
...
+    private function displayListeners(InputInterface $input, OutputInterface $output)
     {
         $table = new Table($output);
-        $table->setHeaders(array(
-            'Event Name',
-            'Priority',
-            'Listener',
-        ));
+        $table->setHeaders([
+            [new TableCell('System events, and target callable', ['colspan' => 3])],
```
CarsonF: `System events` vs other kinds of events?

src/Nut/DebugServiceRegistration.php
```diff
@@ -22,6 +25,7 @@ protected function configure()
         $this
             ->setName('debug:services')
             ->setDescription('System service provider registration boot order debug dumper.')
```
CarsonF: Can we change the name to `debug:providers` or `debug:service-providers`? Individual services are not listed here. Also update the class name to match. 

Description: Dumps service providers and their order.

src/Nut/DebugServiceRegistration.php
```diff
@@ -30,27 +34,43 @@ protected function configure()
...
     {
         $table = new Table($output);
+        $rightAligned = new TableStyle();
+        $rightAligned->setPadType(STR_PAD_LEFT);
+
         $table->setHeaders([
-            '#',
-            'Provider',
+            [new TableCell('Service Providers', ['colspan' => 2])],
+            ['Provider Class Name', 'Boot Order']
```
CarsonF: Can `Boot Order` be changed to just `Order`? Order affects services registered, booted, event subscriptions, and controllers mounted.

src/Nut/DebugEvents.php
```diff
@@ -29,33 +32,44 @@ protected function configure()
...
     }
 
     /**
      * Output a table of listeners, and their callbacks.
      *
+     * @param InputInterface  $input
      * @param OutputInterface $output
      */
-    private function displayListeners(OutputInterface $output)
+    private function displayListeners(InputInterface $input, OutputInterface $output)
```
CarsonF: What's with the separate method?

GawainLynch: Gone now … copy/pasta

GawainLynch:
When we're happy, I'll rebase it all up.

src/Nut/DebugEvents.php
```diff
@@ -20,7 +22,8 @@ protected function configure()
     {
         $this
             ->setName('debug:events')
-            ->setDescription('System events, and target callable debug dumper.')
+            ->setDescription('Events, and target callable, debug dumper.')
```
CarsonF: Dumps event listeners

src/Nut/DebugEvents.php
```diff
@@ -20,7 +22,8 @@ protected function configure()
...
+            ->addOption('sort-callable', null, InputOption::VALUE_NONE, 'Sort events in order of callable name.')
```
CarsonF: `sort-name`

src/Nut/DebugEvents.php
```diff
@@ -29,33 +32,33 @@ protected function configure()
...
-    private function displayListeners(OutputInterface $output)
-    {
         $table = new Table($output);
-        $table->setHeaders(array(
-            'Event Name',
-            'Priority',
-            'Listener',
-        ));
+        $table->setHeaders([
+            [new TableCell('Events, and target callable', ['colspan' => 3])],
```
CarsonF: Event Listeners

src/Nut/DebugRoutes.php
```diff
@@ -36,9 +37,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $table = new Table($output);
         $table->setHeaders([
-            'Bind name',
-            'Path matching parameter',
-            'Method(s)',
+            [new TableCell('System routes', ['colspan' => 3])],
```
CarsonF: Routes

src/Nut/DebugServiceProviders.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('debug:service-providers')
+            ->setAliases(['debug:providers'])
+            ->setDescription('System service provider registration boot order debug dumper.')
```
CarsonF: Dumps service providers and their order.

src/Nut/DebugServiceProviders.php
```diff
@@ -0,0 +1,70 @@
...
+            ->addOption('sort-class', null, InputOption::VALUE_NONE, 'Sort by provider class names.')
```
CarsonF: sort-name

src/Nut/DebugEvents.php
GawainLynch: The column is the FQCN::method() of the event name being listener to

GawainLynch:
> Would dump: be better than debug: for these?

I tossed that over with @bobdenotter, he pointed out that SF use `debug:` so we thought we'd go with that

src/Nut/DebugRoutes.php
```diff
@@ -23,9 +24,9 @@ protected function configure()
         $this
             ->setName('debug:routes')
             ->setDescription('System route debug dumper.')
-            ->addOption('sort-bind', null, InputOption::VALUE_NONE, 'Sort in order of bind name (default).')
+            ->addOption('sort-route', null, InputOption::VALUE_NONE, 'Sort in order of route name (default).')
             ->addOption('sort-pattern', null, InputOption::VALUE_NONE, 'Sort in order of URI patterns.')
-            ->addOption('sort-methods', null, InputOption::VALUE_NONE, 'Sort in order of HTTP method grouping allowed.')
+            ->addOption('sort-method', null, InputOption::VALUE_NONE, 'Sort in order of HTTP method grouping allowed.')
```
CarsonF: This is where I was saying it seems to make more sense to have a value here instead of an option for each sorting. `--sort name/path/method`.

src/Nut/DebugEvents.php
```diff
@@ -29,36 +33,39 @@ protected function configure()
...
-        $table->setHeaders(array(
-            'Event Name',
-            'Priority',
-            'Listener',
-        ));
+        $rightAligned = new TableStyle();
+        $rightAligned->setPadType(STR_PAD_LEFT);
+        $table->setHeaders([
+            [new TableCell('Event names, target callable listeners, and their priority', ['colspan' => 3])],
+            ['Event Name', 'Callable', 'Priority'],
```
CarsonF: The first header is the exact same info as the second, no?

I think `Listener` would be better than `Callable`.

GawainLynch: Got to get you to actually use some code one day …
```
$ ./app/nut debug:events
+-----------------------+---------------------------------------------------------------------------------------+--------------------+
| Event names, target callable listeners, and their priority                                                                         |
+-----------------------+---------------------------------------------------------------------------------------+--------------------+
| Event Name            | Callable                                                                              |           Priority |
+-----------------------+---------------------------------------------------------------------------------------+--------------------+
| kernel.request        | Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure()         |               2048 |
| kernel.request        | Symfony\Component\HttpKernel\EventListener\ProfilerListener::onKernelRequest()        |               1024 |
| kernel.request        | Bolt\EventListener\ConfigListener::onRequestEarly()                                   |                512 |
```

src/Nut/DebugRoutes.php
GawainLynch: Yeah, but was thinking of consistency with the others

src/Nut/DebugEvents.php
CarsonF: I know what it outputs. I meant it conveys the same information

GawainLynch: How is "event name", and "callable" (or "listener", which I don't agree with in this case) is conveying the exact same thing?

GawainLynch: I've dropped the (first) title row and changed "Callable" to "Listener" … c.b.f :wink:

GawainLynch:
Even added test coverage! #PR

PR #6546, opened 2017-04-05
Clear 'exception' and 'translation' too.

bobdenotter:
When clearing the cache using `app/nut`, some cruft stayed behind: 

```
[🐹::~/Sites/bolt-origin]$ du -h -d1 app/cache
4.0K	app/cache/.sessions
8.0K	app/cache/composer
  0B	app/cache/development
8.3M	app/cache/exception
  0B	app/cache/profiler
144K	app/cache/trans
8.5M	app/cache
```

This PR flushes those folders as well.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -76,9 +76,12 @@ class Cache extends FilesystemCache
             }
 
             // Clear our own cache folder.
+            $this->flushDirectory($this->filesystem->getFilesystem('cache')->getDir('/composer'));
             $this->flushDirectory($this->filesystem->getFilesystem('cache')->getDir('/development'));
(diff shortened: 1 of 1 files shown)


bobdenotter:
Ok, i'll revert 'composer'. 

How about `exception`? That's where the most cruft accumulates, in my quick tests.

GawainLynch:
> How about `exception`? That's where the most cruft accumulates, in my quick tests.

Somewhat a fair point, but the idea was never completed as I couldn't get someone's feedback … leave it in, we'll deal with it later :wink:

bobdenotter:
Done.

PR #6565, opened 2017-04-10
Fix content changelog field handling

rossriley:
I'm not sure exactly where the bug starts manifesting itself but sometimes the changelog will record a diff for relation and taxonomy fields, even if there are no taxonomy or relations set on a contenttype.

The root cause of this was that the metadata for the field was not set on the LogChange entity although it was expecting the property to be available.

This fixes that issue by injecting the metadata into the entity object after it has been fetched in the Repository.

If the field is not then set, the transformation will be skipped.

I've tested this on 3.3 and 3.4 and this seems to work.

### DIFF
--- a/src/Storage/Entity/LogChange.php
+++ b/src/Storage/Entity/LogChange.php
@@ -2,6 +2,8 @@
 
 namespace Bolt\Storage\Entity;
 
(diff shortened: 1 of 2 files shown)


rossriley:
any chance of merging this into 3.4 when you get a chance too.

PR #6569, opened 2017-04-12
Bag and ImmutableBag

CarsonF:
This is something I've been wanting to do for about a year and a half now. We all know working with arrays in PHP is not the best experience. `isset` checks are annoying and `array_*` methods make code hard to read. 

Introducing the solution™:
# `Bag` and `ImmutableBag`

These are OO implementations of arrays. The goal of these classes:
- Provide functionality on par with built-in array methods
- Provide useful functionality lacking from built-in array methods
- Provide a fluent-like experience
- Make implementing code more readable

Examples:
```php
$arr = [
    'debug' => true,
    'color' => 'blue',
    'db' => [
        'driver' => 'sqlite',
    ],
];

$bag = Bag::from($arr)
    ->defaults([
        'debug' => false,
        'color' => 'green',
    ])
    ->replace([
        'color' => 'red',
    ])
;
$bag->get('debug'); // true
$bag->getPath('db/driver'); // "sqlite"
$bag->keys()->join(', '); // "debug, color, db"
$bag->isAssociative(); // true
```
```php
$colors = Bag::from(['red', 'blue', 'yellow'])
    ->merge(['green', 'orange'])
;
$colors->isIndexed(); // true
$colors->indexOf('yellow'); // 2
$colors[2]; // "yellow"

$colors->prepend('purple');
$colors[] = 'pink';

$colors->first(); // "purple"
$colors->last(); // "pink"

$colors->shuffle()->first(); // one of the colors

$colors->chunk(2); // Bags represented as arrays:
// [ ['purple', 'red'], ['blue', 'yellow'], ['green, 'orange'], ['pink'] ]

$colors->removeFirst(); // "purple"
$colors->removeFirst(); // "red"
```
These examples only cover half of the functionality.

All methods asking for a collection will accept other `Bags`, `arrays`, `stdClass`, and `Traversable` objects. This makes it very easy work with any collection-like object. `Nulls` and `mixed` values are also supported which are converted to empty arrays and an array with one item respectively.

The methods are clearly documented in code but here's a summary 
<details>
<summary>Methods</summary>

```php
static from(mixed $collection): Bag
static fromRecursive(mixed $collection): Bag
static combine(array|Traversable $keys, array|Traversable $values): Bag
toArray(): array
toArrayRecursive(): array
mutable(): Bag
immutable(): ImmutableBag
has(int|string $key): bool
hasPath(string $path): bool
hasItem(mixed $item): bool
get(int|string $key, mixed $default): mixed
getPath(string $path, mixed $default): mixed
count(): int
isEmpty(): bool
indexOf(mixed $item): int|string|false
first(): mixed|null
last(): mixed|null
join(string $separator): string
sum(): number
product(): number
isAssociative(): bool
isIndexed(): bool
keys(): Bag
values(): Bag
map((int|string $key, mixed $value) => mixed): Bag
mapKeys((int|string $key, mixed $value) => mixed): Bag
filter((int|string $key, mixed $value) => bool): Bag
clean(): Bag
replace(array|Traversable $collection): Bag
replaceRecursive(array|Traversable $collection): Bag
defaults(array|Traversable $collection): Bag
defaultsRecursive(array|Traversable $collection): Bag
merge(array|Traversable $list): Bag
slice(int $offset, int $length = null, bool $preserveKeys = false): Bag
partition((int|string $key, mixed $value) => bool): [Bag, Bag]
column(string $columnKey, string $indexKey = null): Bag
flip(): Bag
reduce((mixed $carry, mixed $item) => mixed, mixed $initial): mixed
unique(): Bag
chunk(int $size, bool $preserveKeys = false): Bag[]
reverse(bool $preserveKeys = false): Bag
shuffle(): Bag
add(mixed $item): void
prepend(mixed $item): void
set(int|string $key, mixed $value): void
setPath(string $path, mixed $value): void
clear(): void
remove(int|string $key, mixed $default = null): mixed
removeItem(mixed $item): mixed
removeFirst(): mixed|null
removeLast(): mixed|null
```
</details>

### Also I know this looks like a lot of code, but there's only 115 lines of logical source code. 
<details>

```
phploc 3.0.1 by Sebastian Bergmann.

Size
  Lines of Code (LOC)                              983
  Comment Lines of Code (CLOC)                     517 (52.59%)
  Non-Comment Lines of Code (NCLOC)                466 (47.41%)
  Logical Lines of Code (LLOC)                     115 (11.70%)
```
</details>

## What's `ImmutableBag` for?

All of the methods in `ImmutableBag` are queries of the current state or return a _new_ modified Bag. Thus an `ImmutableBag` cannot be mutated...no duh. `Bag` extends `ImmutableBag` and provides methods to add, modify, and remove items. 

In PHP, arrays are always passed/returned by value, but objects are passed/returned by reference. 
Let's look at an example:
```php
class Foo
{
    /** @var array */
    private $items;

    public function getItems()
    {
        return $this->items;
    }
}

$foo = new Foo();
$items = $foo->getItems();
$items[] = 'hello';
```
Since `Foo::$items` is an array, the `$items` returned from `getItems()` is a different array. Thus appending "hello" to the array does not modify `Foo::$items`.

Now pretend `Foo::$items` is a `Bag`. Since it is an object, the `$items` returned from `getItems()` is the same object as `Foo::$items`. This means you can modify `Foo`'s internal state externally. `Foo` may not want this, so it can return an `ImmutableBag` instead. 

It could be argued that a clone of the Bag could be returned from the method instead. This is true, but it would lead to more processing for each query of the bag, and could lead to WTF's if the user is trying to modify the clone and it's not applying to the object.

## Hasn't this been done already?
Obviously others think PHP arrays suck as well and have attempted to resolve this. Symfony's `ParamaterBag` is a good basic _map_ (associative arrays) container but is lacking when it comes to mutating the items around and working with lists. Doctrine's `ArrayCollection` is also another, more robust, option. It works well for maps and lists, but still has limited functionality due to needing to interface with a database collection. It also has some annoyances, like `getKeys()` returns an `array` instead of another `ArrayCollection` instance.

# `Arr` updates

`Arr` now has:
```php
has(array|ArrayAccess $data, string $path): bool
get(array|ArrayAccess $data, string $path, mixed $default = null): mixed
set(array|ArrayAccess $data, string $path, mixed $value): void
```
These methods work with "paths" to provide access to nested objects. This is the same as how `Config::get/set` works and it has even been updated to just use these `Arr` methods.
`Bag::hasPath/getPath/setPath` are powered by these functions as well.

Also `replaceRecursive/isIndexed/isAssociative` have been updated to support `Traversable` objects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,7 @@
         "symfony/intl": "^2.8",
         "symfony/monolog-bridge": "^2.8",
         "symfony/options-resolver": "^2.8",
+        "symfony/polyfill-php71": "^1.3",
         "symfony/process": "^2.8",
         "symfony/property-access": "^2.8",
         "symfony/routing": "^2.8",
--- a/src/Collection/Bag.php
+++ b/src/Collection/Bag.php
@@ -1,0 +1,202 @@
+<?php
+
+namespace Bolt\Collection;
+
+use Bolt\Helpers\Arr;
+
+/**
+ * This is an OO implementation of almost all of PHP's array functionality.
+ *
+ * Generally only methods dealing with a single item mutate the current bag,
+ * all others return a new bag.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ */
+class Bag extends ImmutableBag
+{
+    /**
+     * Returns an immutable bag with the items from this bag.
+     *
+     * @return ImmutableBag
+     */
+    public function immutable()
+    {
+        return new ImmutableBag($this->items);
+    }
+
+    // region Mutating Methods
+
+    /**
+     * Adds an item to the end of this bag.
+     *
+     * @param mixed $item The item to append.
+     */
+    public function add($item)
+    {
+        $this->items[] = $item;
+    }
+
+    /**
+     * Adds an item to the beginning of this bag.
+     *
+     * @param mixed $item The item to prepend.
+     */
+    public function prepend($item)
+    {
+        array_unshift($this->items, $item);
+    }
+
+    /**
+     * Sets a item by key.
+     *
+     * @param string $key   The key
+     * @param mixed  $value The value
+     */
+    public function set($key, $value)
+    {
+        $this->items[$key] = $value;
+    }
+
+    /**
+     * Sets a value at the path given.
+     * Keys will be created as needed to set the value.
+     *
+     * This function does not support keys that contain "/" or "[]" characters
+     * because these are special tokens used when traversing the data structure.
+     * A value may be appended to an existing array by using "[]" as the final
+     * key of a path.
+     *
+     *     // Set an item at a nested structure.
+     *     setPath('foo/bar', 'color');
+     *
+     *     // Append to a list in a nested structure.
+     *     setPath('foo/baz/[]', 'a');
+     *     setPath('foo/baz/[]', 'b');
+     *     getPath('foo/baz'); // returns ['a', 'b']
+     *
+     * Note: To set values not directly under ArrayAccess objects their
+     * offsetGet() method needs to be defined as return by reference.
+     *
+     *     public function &offsetGet($offset) {}
+     *
+     * @param string $path  The path to traverse and set the value at
+     * @param mixed  $value The value to set
+     */
+    public function setPath($path, $value)
+    {
+        Arr::set($this->items, $path, $value);
+    }
+
+    /**
+     * Remove all items from bag.
+     */
+    public function clear()
+    {
+        $this->items = [];
+    }
+
+    /**
+     * Removes and returns the item at the specified key from the bag.
+     *
+     * @param string|integer $key     The kex of the item to remove.
+     * @param mixed|null     $default The default value to return if the key is not found.
+     *
+     * @return mixed The removed item or default, if the bag did not contain the item.
+     */
+    public function remove($key, $default = null)
+    {
+        if (!$this->has($key)) {
+            return $default;
+        }
+
+        $removed = $this->items[$key];
+        unset($this->items[$key]);
+
+        return $removed;
+    }
+
+    /**
+     * Removes the given item from the bag if it is found.
+     *
+     * @param mixed $item
+     */
+    public function removeItem($item)
+    {
+        $key = array_search($item, $this->items, true);
+
+        if ($key !== false) {
+            unset($this->items[$key]);
+        }
+    }
+
+    /**
+     * Removes and returns the first item in the list.
+     *
+     * @return mixed|null
+     */
+    public function removeFirst()
+    {
+        return array_shift($this->items);
+    }
+
+    /**
+     * Removes and returns the last item in the list.
+     *
+     * @return mixed|null
+     */
+    public function removeLast()
+    {
+        return array_pop($this->items);
+    }
+
+    // endregion
+
+    // region Internal Methods
+
+    /**
+     * Don't call directly. Used for ArrayAccess.
+     *
+     * @internal
+     *
+     * @inheritdoc
+     */
+    public function &offsetGet($offset)
+    {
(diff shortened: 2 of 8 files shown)


CarsonF:
Thanks @bobdenotter ❤️ 

I was thinking 3.3 so we could use these for `TemplateView` and `TemplateResponse`.

bobdenotter:
On the one hand i'd say "sure, i see no harm, let's put it in". On the other had, we've been postponing 3.3 for ages now, and we need to get it out. 

I think i'd prefer to get it in with 3.4, which should be a much shorter cycle with a few other nice things. Good for PR. :-)

CarsonF:
I don't disagree with you. But knowing that this feature is coming, I don't want to implement `TemplateView` and `TemplateResponse` a different way. We could revert to using `BoltResponse` until 3.4. Moving away from `BoltResponse` is already a BC break, I don't want to do it again in 3.4 with `TemplateView` and `TemplateResponse`.

bobdenotter:
Let's hear what @GawainLynch has to say :-)

GawainLynch:
Well … this could take some typing :stuck_out_tongue_winking_eye: 

First thing, high-five @CarsonF … we've both spent a lot of time discussing, and reviewing each other's failed attempts, but this finally gets us what I think we've been aiming for … this stuff is simple, but powerful, and where needed safe (i.e. `ImmutableBag`).

## Code Location

The "bags" part is simple, powerful, and have broad use … but is only **almost** self-contained.

So what about:

  * Breaking `Bag`, `ImmutableBag`, needed parts of `\Bolt\Helpers\Arr` into it's own contained repository project (`bolt/bags`?)
    * Broadens the audience of use, therefore contribution, testing, Olympic glory
    * Finer grained control over smaller, edge-case, fix/release updates (think Composer user and up, then don't have to update `bolt/bolt`, to fix "the bug", only `composer update bolt/bags`)
    * K.S.M. — SOLID
  * Logic moved/removed from `\Bolt\Helpers\Arr` would be in a similar class in `bolt/bags`, e.g. `\Bolt\Bags\Arr`
  * Have `\Bolt\Helpers\Arr extend \Bolt\Bags\Arr` 

## Branch Target

For what I am doing, 3.4 is WFM … this connects to the config work I am about to get back to (that is for 3.4 anyway) … however …

Juggling `TemplateView` and `TemplateResponse` for `release/3.3` hackishly, then trying to continually merge those changes up to `release/3.4` just makes for a job of policing code flow harder, and also having to support two approaches for the life cycle of 3.3.

TBH, we've called `TemplateView`/`TemplateResponse`blockers for 3.3, and this really helps with that, so from that PoV :+1:

As far as "Do we know how well the code is tested?" … see below, but it exceeds our general average for new work by a **long way**, so that gives me confidence.

## Testing

I'll let the graphics do the talking…

PHPUnit test on Collection only
Collection
Bag
ImmutableBag

GawainLynch:
> Ok, let's get this in!

Sure, but let's see what Carson thinks about the "breakout", then go from there.

CarsonF:
Ok Gawain and I have talked and we are going to move this to a new _bolt/collection_ library.

GawainLynch:
Just an update on progress …

* The bag/collection code is now in [bolt/collection](https://github.com/bolt/collection)
  * README in [PR](https://github.com/bolt/collection/pull/1)
  * Need to tag a 1.0.0
* I have a rebased branch of this one (down to 15 additions and 306 deletions) for Carson to go over, and integrate with this branch, performing git magic as needed
* Merge … time … almost … here …

bobdenotter:
Cool! :-)

CarsonF:
Ok closing this. The code is up at [bolt/collection](http://github.com/bolt/collection). Will open new PR to integrate library.

PR #6572, opened 2017-04-14
Collection Library

CarsonF:
This is the follow up to #6569 

- `bolt/collection` library has been required
- Replace usages of `Arr` class with version from library
- Moved `Arr::deepDiff`
- Deprecated `Arr` class

I want to wait a few days to see how things follow in core before tagging a stable release on bolt/collection. We'll need to tag it before next beta though.


src/Logger/Handler/RecordChangeHandler.php
```diff
@@ -211,4 +210,40 @@ private function initialize()
...
         $this->initialized = true;
     }
+
+    /**
+     * @param array $a
+     * @param array $b
+     *
+     * @return array [key, left, right][]
+     */
+    private function diff(array $a, array $b)
```
GawainLynch: Erm, you might have forgotten about B.C on [`Bolt\DeepDiff`](https://github.com/bolt/bolt/blob/release/3.3/src/DeepDiff.php#L25)

CarsonF: Errrumm what's a BC? 🤦‍♂️

CarsonF: Fixed using your suggestion of reflection

bobdenotter:
Cool stuff! 🎊 🎉

PR #6574, opened 2017-04-16
[3.3] Avoid Nut BC derp

GawainLynch:
So, when Carson and I split the config commands into a base class, I meant to override the `initialize()` in that abstract, instead I put it into `BaseCommand` … :man_facepalming: 

Anyway, as we're using `OutputStyle` (`SymfonyStyle`) in quite a few places, and I'd guess mroe to come :wink: 

… and it would be good to have consistently formatted output that is logical for the programmer to write, so I am setting a `null` `$output` to a `ConsoleOutput` (normal) wrapped in a `SymfonyStyle`

### DIFF
--- a/src/Nut/AbstractConfig.php
+++ b/src/Nut/AbstractConfig.php
@@ -11,6 +11,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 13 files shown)


bobdenotter:
> and I'd guess more to come 😉

As in "Hold off on merging, because there will be more commits"?

GawainLynch:
No, as in I am (target core branch undecided) getting out outputting consistent, including providing, OMG, return values to the console!

bobdenotter:
In that case, merging in! :shipit:

PR #6577, opened 2017-04-18
Tweaked Error/Exception Handler Configuration

CarsonF:
See commit messages for details.

:koala: edit:
  * Moved invoking/subscribing handler listener to second phase, so it gets the correct parameters from Config
  * Replaced `display_errors` and `reporting_level` with `throw_at` and `log_at` DI parameters
    * ErrorHandler's "displayErrors" logic is deprecated and we are instructed to use "throw at" logic instead.
      This is also better because instead of a boolean we can specify which errors we want thrown (aka displayed to user).
    * Additionally, PHP has `display_errors` ini setting which is similar but not the same as ErrorHandler's "display errors" logic.
      This is now automatically disabled (if error handler is enabled) since it is not helpful to users.
You can see the error with the logger you've configured and optionally see it inline as an exception message.
    * PHP's `reporting_level` is automatically configured to report all errors (if error handler is enabled).
This is because Symfony's ErrorHandler has its own logic for configuring how errors are handled: "throw at" and "log_at".
  * Reorder listener code below handlers to visual represent it is executed later
  * Remove debug. prefix from error_handler and exception_handler keys since they are separate from debugging.
  * Move logger out into its own service so it can be configured separately.
    * Set the logger ourselves so we can tell the handler to not log levels we don't want logged.
    * This removes the bootstrap BufferLogger which could cause a memory leak for long running processes.

### DIFF
--- a/src/Provider/DebugServiceProvider.php
+++ b/src/Provider/DebugServiceProvider.php
@@ -23,7 +23,7 @@ use Symfony\Component\HttpKernel\EventListener\DebugHandlersListener;
  * 1. There is no error / exception handlers during app creation and registration stage. This is a very small window,
  *    since closures are just being registered (no logic). Extensions are not loaded either.
  * 2. App Boot
- *   2.a. Debug 1st phase: Error & exception handlers are registered (if enabled) based on `debug.early`'s value.
+ *   2.a. Debug 1st phase: Error & exception handlers are registered based on `debug.early`'s value.
  *        There should be no logic required to get this value.
  *   2.b. Extensions 1st phase: Extensions are registered.
  *   2.c. Debug 2nd phase: The "real" `debug` value is retrieved from config, which means all the logic to setup config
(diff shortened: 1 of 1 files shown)


src/Provider/DebugServiceProvider.php
```diff
@@ -92,91 +93,47 @@ public function register(Application $app)
...
             }
-        );
+        };
+
+        // Logged errors in an integer bit field of E_* constants.
+        $app['error_handler.log_at'] = E_ALL; // TODO Should this be the same as throw_at? Previously it was via config env_error_level.
+
+        $app['error_handler.logger'] = function ($app) {
+            return $app['logger'];
+        };
```
GawainLynch: I :heart: the logger change so much!

src/Provider/DebugServiceProvider.php
```diff
@@ -92,91 +93,47 @@ public function register(Application $app)
...
+            if ($app['debug']) {
+                return $app['config']->get('general/debug_error_level', E_ALL);
+            } else {
+                return $app['config']->get('general/production_error_level', 0);
             }
-        );
+        };
+
+        // Logged errors in an integer bit field of E_* constants.
+        $app['error_handler.log_at'] = E_ALL; // TODO Should this be the same as throw_at? Previously it was via config env_error_level.
```
GawainLynch: `s/env_error_level/debug_error_level/`

Nitpick aside, and onto O.C.D., adding a TODO :cry: 

I think "thrown" and "logged" levels are somewhat related, but should be separable.

However, for example if I have my `debug_error_level` set to `E_ALL &~ E_DEPRECATED &~ E_USER_DEPRECATED`, aren't my (by default `$app['logger']`) log target going to potentially get flooded with deprecations, making their own errors harder to find?

tl;dr ` $app['error_handler.log_at']` is not invoked until `boot()` so probably should match config's `debug_error_level` for now … and **maybe** we later make it a separate config later?

GawainLynch:
Other than the question above, if you grant me some asynchronous Slack time so I can add relevant stuff to the "Debugging Bolt" part of the docs, then this gets a :+1:

src/Provider/DebugServiceProvider.php
CarsonF: `env_error_level` because it could be `debug_error_level` or `production_error_level`.

I think they are separate too, but I'm fine changing the default to use the same config value.

CarsonF:
> Other than the question above, if you grant me some asynchronous Slack time so I can add relevant stuff to the "Debugging Bolt" part of the docs, then this gets a 👍

Of course!

PR #6524, opened 2017-03-28
[WIP] Repeater enhancements

eduardomart:
- adds expand/collapse for repeater items
- makes repeaters drag-sortable

PR #6580, opened 2017-04-18
[3.3] Nut IO BC & derps

GawainLynch:
So, @eduardomart reported this:
io-error

Which is fixed in the first commit, but that brings about problem No. 2.

`ProcessHelper::run()` used in `server:run` expects an `OutputInterface`, but still calls [`Output::isDebug()`](https://github.com/symfony/console/blob/2.8/Helper/ProcessHelper.php#L59) …

So second commit, we just extend our own style, and forward calls where appropriate … that's where the RFC comes in @CarsonF


eduardomart:
Not sure if it's relevant but I was testing the 3.4 branch, not 3.3

GawainLynch:
> Not sure if it's relevant but I was working with the 3.4 branch, not 3.3

@eduardomart I think you are referring to #6524

eduardomart:
@GawainLynch yes, which is a 3.4 feature that's why I checked out the 3.4 branch (again... 🐨)

GawainLynch:
@eduardomart Yes, this PR will go into 3.3 and then we'll merge down to 3.4+ and they'll work again, just waiting on @CarsonF to review this, but his work schedule at the moment is a bit full :smile:

GawainLynch:
> I'm not sure I really agree with breaking the interface to support OutputStyle. 

This still implements `Symfony\Component\Console\Output\OutputInterface`, so by default, you have what the application gives you.

> But besides that since isDebug is the only method that breaks the interface I think it would be better to just define that, rather than have the magic call. 

Yeah, but there is the BC on other functions that other's use too, when they're assuming they have the standard `Output` class.

> Saves from having to override constructor and properties too.

`$input` and `$output` are private, so we need the constructor to store a local reference to the objects to forward the call … or reflection, your call :stuck_out_tongue_winking_eye:

GawainLynch:
> I'm not sure I really agree with breaking the interface to support OutputStyle.

Also, in 3.4 I want to get Nut command output all consistent, and stylised … it's everywhere at the moment.

CarsonF:
> This still implements Symfony\Component\Console\Output\OutputInterface, so by default, you have what the application gives you.

Right but the other way around. You're now assuming $output is OutputStyle and calling methods on OutputStyle. It's just making me twitch. I'll let you know if I think of something better.

> Yeah, but there is the BC on other functions that other's use too, when they're assuming they have the standard Output class.

I didn't see usage of `isQuiet`, `isVerbose`, `isVeryVerbose` anywhere. Only usage of `isDebug`. 
I still would rather implement those directly instead of the magic call.

> $input and $output are private, so we need the constructor to store a local reference to the objects to forward the call … or reflection, your call 😜

What I was saying is if you are not making the magic call at all you don't need to store $input and $output properties.

GawainLynch:
> I'll let you know if I think of something better.

OK, but we need something quick now, and something so we can guarantee consistent styling of our output, so we can look … you know, pro :wink: 

> I didn't see usage of isQuiet, isVerbose, isVeryVerbose anywhere. Only usage of isDebug.

Extensions may still chose to call them, I've seen it out there somewhere.

> I still would rather implement those directly instead of the magic call.

If that be your wish, consider it done.

CarsonF:
Hope you don't mind I made some changes. Also I hijacked the branch for some more things, but I can separate it if wanted. Open to talking about any and all changes.

GawainLynch:
Status quo, I've rebase this, and squashed some to keep tracking easier.

> Also I hijacked the branch for some more things

Yes, and I get to delete a branch I was (sort of) working on as a result :+1: 

----

Dear Merge Reviewer,
 * Command completion change should have happened for 3.0, isn't BC, and is micro-performance (and "expert audience") — :australia: vote  :+1: 
 * `ContainerHelper` was coming from me in a 3.4 branch after this went in. I was going to target 3.4, but I personally see zero harm in it landing here — :australia: vote :+1: 

Regards,

:australia: & :us:

Issue #6587, opened 2017-04-20
[RFC] Fix `enforce_ssl` - two options.


bobdenotter:
As pointed out by @eduardomart on Slack: Right now, the description for `enforce_ssl` in config.yml says: 

```
# Enforcing the use of SSL. If set, all but the front-end pages will enforce an
# SSL connection (and redirect to HTTPS if you attempt to visit the backend over
# plain HTTP).
enforce_ssl: true
```

This is incorrect. _All_ pages get redirected to their https equivalent, both front- and backend. How should we fix this? 

 - 1) Fix the description in config.yml, so that it matches the behaviour.
 - 2) Change the behaviour, so that it matches the description: Don't redirect to https on pages in the frontend.

GawainLynch:
> Fix the description in config.yml, so that it matches the behaviour.

:+1:

bobdenotter:
Closed by #6589

PR #6589, opened 2017-04-20
Update description for enforce_ssl setting

eduardomart:
As decided on #6587

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -410,7 +410,7 @@ hash_strength: 10
 #         optimize-autoloader: false     # Optimize autoloader during autoloader dump.
 #         classmap-authoritative: false  # Autoload classes from the classmap only. Implicitly enables `optimize-autoloader`.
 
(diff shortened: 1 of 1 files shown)


app/config/config.yml.dist
```diff
@@ -410,7 +410,7 @@ hash_strength: 10
 #         optimize-autoloader: false     # Optimize autoloader during autoloader dump.
 #         classmap-authoritative: false  # Autoload classes from the classmap only. Implicitly enables `optimize-autoloader`.
 
-# Enforcing the use of SSL. If set, all but the front-end pages will enforce an
+# Enforcing the use of SSL. If set, front-end and back-end pages will enforce an
 # SSL connection (and redirect to HTTPS if you attempt to visit the backend over
```
bobdenotter: "backend" here isn't really accurate anymore.. Perhaps "any page", instead?

eduardomart: Sounds good to me

PR #6591, opened 2017-04-22
Implement getValues in new Storage Entity

### LINKED ISSUE #6579: [BUG] New content object has old getValues() method which doesnt work anymore
The new Content object `Bolt\Storage\Entity\Content` has a method `getValues()` which returns `NULL`.

The problem is that this class now stores it's values inside a property `$_fields`, which is managed inside the `MagicAttributesTrait`. The content class therefore has a couple of refactored methods that use this trait's magic getters/setters to access the new values.
The new content class also has for example a new `setValues()` method.

**But** the new content class also still uses the `ContentValuesTrait`, that assumes that values are still stored inside a `$values` property, which is no longer the case. Therefore, all it's methods, including the `getValues()` method no longer work.

**Suggestion**
@ross suggested that the new way to access values is by using `$content->toArray()`. This is perfecly fine. But we then need one of these changes:

1) Remove the ContentValuesTrait and force developers to use the new method.
2) Remove the ContentValuesTrait and at least reimplement the getValues() method inside the Content class (same as with the setValues() method)
3) Refactor the ContentValuesTrait to use the new storage location or make it use the magic getter/setters.

(I can also take this one, if you point out which solution to use)

Details
-------
 - Relevant Bolt Version: 3.1, 3.2, 3.3, 3.4
 - Install type: Composer Install
 - PHP version: 5.5
 - Used web server: Apache 2.4

Reproduction
------------
- Fetch a Content object by one of the new methods: `$this->container['storage']->getRepository($slug)->findOneBy(["id" => $id])`
- Dump it's values with `dump($content->getValues());` 
- Ensure the values are actually there by `dump($content->_fields);`

rossriley:
Fixes #6579

Implements the getValues method to work on the new storage entity in the same way the old Content object did.

Includes test.

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -322,12 +322,9 @@ class Content extends Entity
             return [];
         }
 
-        $fields = array_keys($contentType['fields']);
         $allValues = $this->toArray();
 
-        return array_filter($allValues, function ($key) use ($fields) {
-            return in_array($key, $fields);
-        }, ARRAY_FILTER_USE_KEY);
+        return array_intersect_key($allValues, ($contentType['fields']));
     }
 
     /**
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -312,6 +312,25 @@ class Content extends Entity
     }
 
     /**
+     * Helper to return an array of user-defined values from the Entity.
+     * This excludes meta fields set by Bolt.
+     */
+    public function getValues()
+    {
+        $contentType = $this->getContenttype();
+        if (!isset($contentType['fields'])) {
+            return [];
+        }
+
+        $fields = array_keys($contentType['fields']);
+        $allValues = $this->toArray();
+
+        return array_filter($allValues, function ($key) use ($fields) {
+            return in_array($key, $fields);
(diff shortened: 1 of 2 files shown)


src/Storage/Entity/Content.php
```diff
@@ -312,6 +312,25 @@ public function setValues(array $values)
...
+        if (!isset($contentType['fields'])) {
+            return [];
+        }
+
+        $fields = array_keys($contentType['fields']);
+        $allValues = $this->toArray();
+
+        return array_filter($allValues, function ($key) use ($fields) {
+            return in_array($key, $fields);
+        }, ARRAY_FILTER_USE_KEY);
```
GawainLynch: `ARRAY_FILTER_USE_KEY` [Added in PHP 5.6.0.](http://php.net/manual/en/array.constants.php)

GawainLynch: Might want to add a `@deprecated` around the support hack, and we'll pull it out of `master`

CarsonF: Just use a `Bag` ;)

CarsonF: Actually using `array_intersect_keys` would be easier.

PR #6595, opened 2017-04-26
[Travis] Remove remaining legacy sudo use

GawainLynch:
The guys at Travis are [EOL-ing `sudo` enabled builds](https://blog.travis-ci.com/2017-04-17-precise-EOL), so this just removes the remnants that remained to keep us on their legacy environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ matrix:
   allow_failures:
     - php: hhvm
 
+sudo: false
+
 services:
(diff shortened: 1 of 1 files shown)


CarsonF:
Shouldn't this be merged with 3.2 as well?

GawainLynch:
I actually created this branch on a `release/3.2` based checkout, but it got interesting quickly.

PR #6603, opened 2017-04-30
[RFC] Makes deleting saved thumbnails when clearing cache optional

jkazimir:
This adds an option to config.yml which allows disabling the clearing of the saved thumbs folder when cache is cleared.

Details
-------

While in many cases it's okay and perhaps desired to clear saved thumbnails along with other caches, there are some cases where keeping them around may be better.  Product/photo galleries for example may get quite large, and when looking to not restrict the uploader, full images from cameras can get quite hefty in size where the required processing needed to scale them would increase proportionally.  A low powered VPS instance, which may otherwise be suitable for hosting such a gallery, may not be as up to the task of rescaling listing pages full of images on demand if they're cleared.  This change keeps the current default of clearing the saved images, while allowing a developer to disable this behavior if desired.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -199,6 +199,8 @@ changelog:
 #                     the requested dimensions.
 # browser_cache_time: Sets the amount of seconds that the browser will cache
 #                     images for. Set it to activate browser caching.
+# clear_with_cache:   Used alongside save_files. Setting to false prevents
+#                     deletion of saved thumbs when clearing bolt cache.
 #
 # Note: If you change these values, you might need to clear the cache before
 #       they show up.
@@ -213,6 +215,7 @@ thumbnails:
     allow_upscale: false
     exif_orientation: true
     only_aliases: false
+    clear_with_cache: true
 #    browser_cache_time: 2592000
 
 # Define the HTML tags and attributes that are allowed in 'cleaned' HTML. This
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -27,6 +27,9 @@ class Cache extends FilesystemCache
     /** @var AggregateFilesystemInterface */
     private $filesystem;
 
+    /** @var bool */
+    private $clearThumbs;
+
     /**
      * Cache constructor.
      *
@@ -34,11 +37,13 @@ class Cache extends FilesystemCache
      * @param string                       $extension
      * @param int                          $umask
      * @param AggregateFilesystemInterface $filesystem
+     * @param bool                         $clearThumbs
      */
-    public function __construct($directory, $extension = self::EXTENSION, $umask = 0002, AggregateFilesystemInterface $filesystem = null)
+    public function __construct($directory, $extension = self::EXTENSION, $umask = 0002, AggregateFilesystemInterface $filesystem = null, $clearThumbs = true)
     {
         parent::__construct($directory, $extension, $umask);
         $this->filesystem = $filesystem;
+        $this->clearThumbs = $clearThumbs;
     }
 
     /**
@@ -85,8 +90,10 @@ class Cache extends FilesystemCache
             $this->flushDirectory($this->filesystem->getFilesystem('cache')->getDir('/profiler'));
             $this->flushDirectory($this->filesystem->getFilesystem('cache')->getDir('/trans'));
 
-            // Clear the thumbs folder.
-            $this->flushDirectory($this->filesystem->getFilesystem('web')->getDir('/thumbs'));
+            if ($this->clearThumbs) {
+                // Clear the thumbs folder.
+                $this->flushDirectory($this->filesystem->getFilesystem('web')->getDir('/thumbs'));
+            }
         }
 
         return $result;
--- a/src/Config.php
+++ b/src/Config.php
@@ -1127,6 +1127,7 @@ class Config
                 'notfound_image'    => 'bolt_assets://img/default_notfound.png',
                 'error_image'       => 'bolt_assets://img/default_error.png',
                 'only_aliases'      => false,
+                'clear_with_cache'  => true,
             ],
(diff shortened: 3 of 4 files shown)


GawainLynch:
Hehe … @CarsonF and I have been wanting to find free time to hear me out on a broader approach here too, so let's use this as a starting point!

A current environment cache at 3.3+ looks something like:
cache

Currently dot files/directories are protected (e.g `.sessions/`), as is `composer/` as 99.99999% of the time you do **not** want to flush those on a regular cache flush.

Similarly, you might want to flush the development cache, but not the production one … the HTTP reverse proxy cache, but not Twig's potentially expensive template rebuild, or you know … thumbnails.

To the point …

We need to discuss/decide on a forward looking approach, and I am happy to use this as step 1, but the first thing is "config settings" vs "GUI/CLI options" to the flushing procedure … and also how do we do/keep this forward compatible.

bobdenotter:
I think making all aspects of this configurable is overkill, but this particular addition makes sense. 

I would like to add an option (in the UI) though, to do a "regular flush" or a "full flush". With a 'confirm' dialog, that will also flush `thumbs/` (regardless of this setting) as well as `.session` and the others. So, only when you need to clear all. 

For nut it could be `app/nut cache:clear --full`. 

A suggestion for the UI: 



<small>\</joking\></small>

jkazimir:
Haha, love the UI suggestion :+1:

To me, at least for now, it makes sense to have configuration at least for the default "clear".

Of course, the UI/CLI control make complete sense, and once services exist for more granular cache clearing, then only have the absolute necessary caches cleared during things that'll currently clear everything, such as a `composer install`.  Perhaps a single form with a checklist, "select all", and confirmation would be a sufficient UI to expose that?

GawainLynch:
> Further thought is needed here, it just hasn't been the highest priority.

Personally happy to continue it here BTW, especially if you're up for it @jkazimir … I feel the right parties are involved to move this forward :smile:

GawainLynch:
> I think making all aspects of this configurable is overkill, but this particular addition makes sense.

Try deleting `app/cache/.sessions/` on a very busy production site where people are submitting 'n' number of forms per second … or a bunch of editors are logged in. That doesn't mean there aren't times you want a way to do this, just shouldn't fire on a "normal" cache flush … same with `composer/`, same with `thumbs/`.

jkazimir:
Had a feeling a simple toggle wouldn't cut it, hence the [RFC] :-)  Happy to take it on in my free time.

How in depth should a refactor here go?  It sounds like at a minimum as far as 3.x compatibility on Bolt\Cache, clearCache / doFlush methods should stick around, but with a limited scope (only symfony & config cache, maybe?). There would then also be more refined local methods publicly exposed to handle any more granular clearing.

CarsonF:
My thought was using Symfony Cache with tags. But I also think we should have separate cache instances for major components. That way they can be swapped out easier.

Haven't thought through BC at all, and IMO that should happen after we agree on the best next solution.

CarsonF:
The other thing we Gawain and I have talked about before is matching Symfony with a `var/` folder. A folder that is written to programmatically, but not necessarily cache files.
```
.
|-- app
|-- public
|-- src
|-- tests
`-- var
    |-- cache
    |   |-- composer
    |   |-- profiler
    |   |-- translations
    |   `-- twig
    |-- logs
    |-- sessions
    `-- spool
```

GawainLynch:
> The other thing we Gawain and I have talked about before is matching Symfony with a var/ folder. A folder that is written to programmatically, but not necessarily cache files.

I'd debate one tiny change there:

```
.
|-- app
|-- public
|-- src
|-- tests
`-- var
    |-- cache
    |   |-- profiler
    |   |-- translations
    |   `-- twig
    |-- composer
    |-- logs
    |-- sessions
    `-- spool
```
Composer has its own thing going on, and that `composer` is the same as you'd find in `~/.composer` or 
 `~/.config/composer`. So repository mirrors, config defaults/overrides, auth keys … Though it is fair to call/say it's volatile, it shouldn't be considered/treated in the same way many other `var/` components would be … 

e.g. After every full flush would you want to have to go and set up your extension Composer environments, & private repo auth keys, so you can switch back to your normal CLI deployment?

GawainLynch:
> Haven't thought through BC at all, and IMO that should happen after we agree on the best next solution.

Hold my :beer:, I got this :stuck_out_tongue_winking_eye: `</humour>` 

Flagging that this might be another one for a compatibility flag? (pun intended, it's 7 am)

> My thought was using Symfony Cache with tags. 

Agreed :+1: 

Oh, and to head off mistaken impressions … Symfony Cache **starts** at v3.1.0, inline with the SF release it was with. From release, in the 3.2-HEAD [`composer.json` requirements](https://github.com/symfony/cache/blob/3.2/composer.json#L22-L24) (i.e. stable branch), and even master is also PHP 5.5.9+ only.

Additionally, it gives us our PSR-6 implementation! Welcome to one more standardised and replaceable section of our product :smile: 

> But I also think we should have separate cache instances for major components. That way they can be swapped out easier.

Agreed :+1: 

Also, I would like them to be able to respond/listen to various "needs" (thinking out loud here)
  * "Nuclear Flush", e.g.:
    * Use case: Preparing a site build/upgrade for deployment
      * Result: Empty `var/`
    * Use case: In-place updates of Bolt versions
      * Result: Empty `var/cache/`
  * User role-based needs, e.g.:
    * Use case: Template designer installs updated template
      * Result: Empty `var/cache/twig/`
  * Package installation event, e.g.
    * Use case: Bolt v4 extension enables a bunch of supported changes to Bolt's back-end templates (FYI I used "Bolt v4" as part of the example, not to indicate where I think it should initially land)
      * Result: Empty `var/cache/twig/`

PR #6608, opened 2017-05-01
Use getter to get id. Fixes exception when using datetime in repeater fields

### LINKED ISSUE #6607: "datetime" fields in Repeaters throw exception.
To reproduce, use: 

```
bars:
    name: Bars
    slug: bars
    singular_name: Bar
    singular_slug: bar
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        body:
            type: html
            height: 300px
        activity_programma:
            type: repeater
            fields:
                repeattitle:
                    type: datetime
    record_template: record.twig
```

See screenshot below. Fix incoming. 

screen shot 2017-05-01 at 16 24 07

nestordedios:
Fixes #6607

### DIFF
--- a/app/view/twig/editcontent/fields/_date.twig
+++ b/app/view/twig/editcontent/fields/_date.twig
@@ -6,7 +6,7 @@
     label:     field.label,
     info:      field.info|default(''),
     options:   field.options.datepicker|default(''),
-    default:   (context.content.get('id') == null and field.default) ? field.default|date('Y-m-d') : null,
+    default:   (context.content.id|default and field.default) ? field.default|date('Y-m-d') : null,
     required:  field.required|default(false),
     errortext: field.error|default(''),
 } %}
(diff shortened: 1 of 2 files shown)


bobdenotter:
This PR is changing it to be the same as it is on a `date` field, where @rossriley made the exact same change: https://github.com/bolt/bolt/blame/release/3.2/app/view/twig/editcontent/fields/_date.twig#L9

Should they both be changed to `(context.content.id is defined and field.default)` ?

CarsonF:
> Should they both be changed to (context.content.id is defined and field.default) ?

Yes. Unless the field could exist but be null/falsey and you want to check that too. That could be checked with the `|default` filter.
```twig
(context.content.id|default and field.default)
```
The `default` filter and `is defined` test both trigger the same logic for accessing attributes on an object.

GawainLynch:
:+1:

PR #6609, opened 2017-05-01
Allow overriding `description: ` in `contenttypes.yml`.

bobdenotter:
As pointed out by @eduardomart today, you can set the `description:` for custom contenttypes, but if you have created your own `pages:` for example, it will _always_ use the baked-in description, without allowing you to override it. This PR changes the order: Use the description in contenttypes.yml, and fall back to the baked-in one if it wasn't set.

### DIFF
--- a/app/view/twig/overview/_panel-actions_overview.twig
+++ b/app/view/twig/overview/_panel-actions_overview.twig
@@ -54,8 +54,10 @@
 
     </form>
 
-
-    {% set description = __(['contenttypes', context.contenttype.slug, 'description'], {DEFAULT: context.contenttype.description|default()}) %}
+    {# Use the `description:` from contenttypes.yml if set, otherwise fall back
+       to the (localized) default descriptions that are available for the
+       built-in contenttupes like 'pages' and 'blocks'. #}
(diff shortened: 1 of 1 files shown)


CarsonF:
This is a hot topic for me and I'm about to start refactoring the translation layer for 3.4.

I don't think we should be providing translations for _user content_ at all. Are these files just for examples? If so, they shouldn't be loaded into the app.

bobdenotter:
@CarsonF I agree completely. Which is why i find it annoying that the "example" (as you correctly call it) couldn't be overridden. 

This PR will set it to the correct behaviour (imho) and won't interfere with your plans, if i understand correctly.

bobdenotter:
Ok, cool! 

I expect we'll discuss how to tackle it for 3.4, because when we discussed it previously, I know you and I mostly agreed, but not everyone on the core team did. Since he's been pretty much absent for a long while, I think we should push forward with your plans. 

But for now: yes, this is a hack, but it fixes a bug for right now. ;-)

GawainLynch:
Fine … I'll do it :stuck_out_tongue_winking_eye:

PR #6616, opened 2017-05-04
[3.2] Fix incompatible extensions showing

GawainLynch:
* Fix "option" key handling in `extension/composer.json`
* This sends the "Composer" version of Bolt to the marketplace so it can remove packages without matching version constraints

FYI: `simpleforms` is the one I was testing on, as there is nothing available (even `dev-master` is for < 3.0 as it was retired in place of BoltForms)

**NOTE:** This will need an asset rebuild post-merge

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -31,7 +31,7 @@ theme: base-2016
 # locales, ensure the first is a standard locale.
 locale: en_GB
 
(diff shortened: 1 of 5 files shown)


bobdenotter:
> NOTE: This will need an asset rebuild post-merge

I'll handle that, and do a new 3.2.x soon.

PR #6619, opened 2017-05-05
[CodeClimate] Enable grep engine

GawainLynch:
CodeClimate have introduced a [grep engine](https://docs.codeclimate.com/v1.0/docs/grep) that might just solve one of my concerns about keeping track of re-introduction of deprecated, or deprecation prevention, code.

Firstly, I have added a check on new tests for PHPUnit deprecations … we can go from there.

As CodeClimate test on `release/3.3` and PRs only, I am going to use this one as a testing ground until happy, i.e. …

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,8 +1,59 @@
 engines:
     grep:
         enabled: true
+        exclude_paths:
+            - "!tests"
         config:
             patterns:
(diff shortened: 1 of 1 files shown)


GawainLynch:
OK, should be GTG now. The CC fails **should** get assigned to this PR, so shouldn't repeat again.

Issue #6530, opened 2017-03-30
[RFC] Allowing Users to Simply Opt-in to Features


rossriley:
Obviously the main driver of this is to allow advanced users to move onto the new storage system so I've used a rough outline of how this could work in this RFC but the same principle will apply to any experimental feature that we want to allow users to opt in to.

## Step 1 - Enabling the Experimental Feature Service

The configuration for this is done in the `.bolt.yml` file. The syntax can look something like this.

```yml
application: MyApp\Application
paths:
    web: public
    themebase: public/theme
    files: public/files
    view: public/bolt-public/view
services:
  -  Bolt\Provider\FeatureProvider:
        feature.storage_twig_new: true
```

Listing the service here enables users to opt into the load of this provider. Whilst we can ship this in `src/Bolt/Provider/FeatureProvider.php` it will not be loaded by default in a standard application it needs the user to put the class into their Bolt config to get it loaded.

The services definition in `.bolt.yml` optionally takes an array of values too, so we can use these to toggle individual features by setting to true. So `feature.storage_twig_new: true` allows us to opt into this feature but there could also be a list of several defined at this point.

## Step 2 - Enabling the Experimental Feature

Because we now have a way for the user to opt into features we can optionally extend the default services in our new provider. For instance the FeatureProvider class may look a little like this...

```php
<?php

namespace Bolt\Provider;

use Silex\ServiceProviderInterface;
use Silex\Application;
use Twig_Environment;

/**
 *
 */
class FeatureProvider implements ServiceProviderInterface
{

    public function register(Application $app)
    {
        if ($app['feature.storage_twig_new']) {
            $app['twig'] = $app->share(
                $app->extend(
                    'twig',
                    function (Twig_Environment $twig, $app) {
                        $twig->addTokenParser(new Experimental\SetcontentTokenParser());
                    }
                )
            );
        }


    }

    public function boot(Application $app)
    {
    }
}
```

We wrap the extension definition in an if block to only modify it the feature has been toggled on.

## Possible Risks

When implementing feature toggles there is always an extra layer of complexity that is being introduced, feasibly one user will see different behaviour depending on which features are on. But if we try and centralise the available features and define the extensions in a single provider then at least we have a very easy-to-scan definition of what can be opted into.

When we want to adopt these features into core,  we can just remove the entire if block from the FeatureProvider and the user's config will just do nothing any more.

[...]

PR #6622, opened 2017-05-05
Allow opt in to new setcontent queries

rossriley:
As previously discussed this uses the new compatibility section in config.yml to switch from the current to the new getContent query engine.

Note that this is definitely not ready for use by anyone other than core developers at the moment, but it will allow us to spot and fix any incompatibilities as we go through the 3.4 development period.

It is envisaged that all this code will be removed in version 4 assuming that the inconsistencies can be ironed out.


src/Controller/Base.php
```diff
@@ -378,6 +378,12 @@ protected function getRepository($repository)
      */
     protected function getContent($textquery, $parameters = [], &$pager = [], $whereparameters = [])
     {
+        $queryVersion = $this->getOption('compatibility/setcontent_version', 1);
```
GawainLynch: Unless I am missing something we've done, the first parameter is missing `general/`, as is [the two here](https://github.com/bolt/bolt/blob/release/3.3/src/Controller/Base.php#L84-L91)

Testing without defaults, or debug assignments, shows this to be the case :smile:

rossriley: sorry, yes missed a commit from my local version

src/Twig/Extension/BoltExtension.php
```diff
@@ -24,19 +25,23 @@ class BoltExtension extends Extension implements GlobalsInterface
...
     /**
      * Constructor.
      *
      * @param EntityManagerInterface $em
      * @param Config                 $config
      * @param PathsProxy             $paths
+     * @param Query                  $query
      */
-    public function __construct(EntityManagerInterface $em, Config $config, PathsProxy $paths)
+    public function __construct(EntityManagerInterface $em, Config $config, PathsProxy $paths, Query $query)
```
GawainLynch: So this might cause a problem. The entity manager passed in here is the lazy-loading version, but `Query` activates the real EM, and this class is constructed during registration.

I had to go do a bit of checking to make sure I wasn't off of a path then :laughing:

app/config/config.yml.dist
```diff
@@ -460,3 +460,5 @@ compatibility:
     # Response methods cannot be used on TemplateView objects.
     # Setting this value to false is deprecated.
     template_view: true
+    # Allows specifying a newer version of the setcontent parser
+    setcontent_version: 1
```
bobdenotter: Should we use a string for this, instead of a (seemingly) unsemantic number?

GawainLynch: What if we end up with a version 4 in the short term? Numerical comparison is easier, correct?

Mind you, wouldn't this be more suitable as a `compatibility` key?

CarsonF: IMO the key is going away in 4.0 anyways as the new engine will be the only thing available. That's why I chose booleans for the other two keys. I doubt we are going to try to manage more than one storage engine BC breaking change before 4.0.

GawainLynch: Sure, but don't you still think it's better suited to a `compatibility` key?

CarsonF: @GawainLynch It is a compatibility key...

GawainLynch: Yeah, I suck at reading :disappointed:

bobdenotter:
Once this lands, we effectively have solved #6530 too, right?

GawainLynch:
> Once this lands, we effectively have solved #6530 too, right?

Didn't #6443 took care of that?

bobdenotter:
> Didn't #6443 took care of that? 

Yes. :-)

GawainLynch:
@rossriley 3.4 is due to start beta phase on the 25th. How is progress on this coming?

GawainLynch:
:shipit:

PR #6624, opened 2017-05-05
[3.3] Fix TemplateView handling

GawainLynch:
* Don't crash when TwmplateView is really enabled
  * Fix config lookup of "compatibility" config values
  * TemplateView BC for routes accessing a rendered response, assuming a Response object
* Render email with Twig, as Swiftmailer messages need strings
* Controller helper to add a middleware that sets the response caching parameters
* Render maintenance mode template directly, as before() is before other listeners
* Dispatch view events for stack add() renders, as they are returned inside JSON
* [Tests] Update for template view, and test legacy on front-end

### DIFF
--- a/src/Controller/Async/General.php
+++ b/src/Controller/Async/General.php
@@ -5,7 +5,6 @@ namespace Bolt\Controller\Async;
 use Bolt;
 use Bolt\Extension\ExtensionInterface;
 use Bolt\Filesystem;
-use Bolt\Response\TemplateView;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Repository;
 use GuzzleHttp\Exception\RequestException;
@@ -33,14 +32,26 @@ class General extends AsyncBase
             ->bind('changelogrecord');
 
         $c->get('/dashboardnews', 'dashboardNews')
-            ->bind('dashboardnews');
+            ->bind('dashboardnews')
+            ->after(function (Request $request, Response $response) {
+                $response->setSharedMaxAge(3600);
+            })
+        ;
 
         $c->get('/lastmodified/{contenttypeslug}/{contentid}', 'lastModified')
             ->value('contentid', '')
-            ->bind('lastmodified');
+            ->bind('lastmodified')
+            ->after(function (Request $request, Response $response) {
+                $response->setSharedMaxAge(60);
+            })
+        ;
 
         $c->get('/latestactivity', 'latestActivity')
-            ->bind('latestactivity');
+            ->bind('latestactivity')
+            ->after(function (Request $request, Response $response) {
+                $response->setSharedMaxAge(3600);
+            })
+        ;
 
         $c->get('/makeuri', 'makeUri')
             ->bind('makeuri');
@@ -58,6 +69,9 @@ class General extends AsyncBase
                 }
 
                 return $extension;
+            })
+            ->after(function (Request $request, Response $response) {
+                $response->setSharedMaxAge(180);
             })
         ;
 
@@ -130,7 +144,6 @@ class General extends AsyncBase
         ];
 
         $response = $this->render('@bolt/components/panel-news.twig', ['context' => $context]);
-        $this->setResponseMaxAge(3600);
 
         return $response;
     }
@@ -177,7 +190,6 @@ class General extends AsyncBase
                 ],
             ]
         );
-        $this->setResponseMaxAge(3600);
 
         return $response;
     }
@@ -291,7 +303,6 @@ class General extends AsyncBase
         $html = $this->app['markdown']->text($readme);
 
         $response = new Response($html);
-        $this->setResponseMaxAge(180);
 
         return $response;
     }
@@ -472,10 +483,7 @@ class General extends AsyncBase
             'filtered'    => $isFiltered,
         ];
 
-        $response = $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
-        $this->setResponseMaxAge(60);
-
-        return $response;
+        return $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
     }
 
     /**
@@ -498,9 +506,6 @@ class General extends AsyncBase
             'contenttype' => $contenttype,
         ];
 
-        $response = $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
-        $this->setResponseMaxAge(60);
-
-        return $response;
+        return $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
     }
 }
--- a/src/Controller/Async/General.php
+++ b/src/Controller/Async/General.php
@@ -5,6 +5,7 @@ namespace Bolt\Controller\Async;
 use Bolt;
 use Bolt\Extension\ExtensionInterface;
 use Bolt\Filesystem;
+use Bolt\Response\TemplateView;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Repository;
 use GuzzleHttp\Exception\RequestException;
@@ -129,7 +130,7 @@ class General extends AsyncBase
         ];
 
         $response = $this->render('@bolt/components/panel-news.twig', ['context' => $context]);
-        $response->setSharedMaxAge(3600)->setPublic();
+        $this->setResponseMaxAge(3600);
 
         return $response;
     }
@@ -176,7 +177,7 @@ class General extends AsyncBase
                 ],
             ]
         );
-        $response->setPublic()->setSharedMaxAge(3600);
+        $this->setResponseMaxAge(3600);
 
         return $response;
     }
@@ -290,7 +291,7 @@ class General extends AsyncBase
         $html = $this->app['markdown']->text($readme);
 
         $response = new Response($html);
-        $response->setSharedMaxAge(180)->setPublic();
+        $this->setResponseMaxAge(180);
 
         return $response;
     }
@@ -472,7 +473,7 @@ class General extends AsyncBase
         ];
 
         $response = $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
-        $response->setPublic()->setSharedMaxAge(60);
+        $this->setResponseMaxAge(60);
 
         return $response;
     }
@@ -498,7 +499,7 @@ class General extends AsyncBase
         ];
 
         $response = $this->render('@bolt/components/panel-lastmodified.twig', ['context' => $context]);
-        $response->setPublic()->setSharedMaxAge(60);
+        $this->setResponseMaxAge(60);
 
         return $response;
     }
--- a/src/Controller/Async/Stack.php
+++ b/src/Controller/Async/Stack.php
@@ -7,6 +7,7 @@ use Bolt\Response\TemplateResponse;
 use Silex\ControllerCollection;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 /**
  * Async controller for Stack async routes.
@@ -19,11 +20,12 @@ class Stack extends AsyncBase
     protected function addRoutes(ControllerCollection $c)
     {
         $c->post('/stack/add', 'add')
-            ->assert('filename', '.*')
-            ->bind('stack/add');
+            ->bind('stack/add')
+        ;
 
         $c->get('/stack/show', 'show')
-            ->bind('stack/show');
+            ->bind('stack/show')
+        ;
     }
 
     /**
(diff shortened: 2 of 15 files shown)


GawainLynch:
Looks like we got us some test fix-un-up do do, @CarsonF :wink:

GawainLynch:
@CarsonF can you have a look over this please.

src/Controller/Async/General.php
```diff
@@ -129,7 +130,12 @@ public function dashboardNews(Request $request)
         ];
 
         $response = $this->render('@bolt/components/panel-news.twig', ['context' => $context]);
-        $response->setSharedMaxAge(3600)->setPublic();
+        if ($response instanceof TemplateView) {
+            $this->setResponseCaching(3600);
+        } else {
+            /** @deprecated Deprecated since 3.3, to be removed in 4.0. */
+            $response->setSharedMaxAge(3600)->setPublic();
+        }
```
CarsonF: Why have both?  The first works for both cases.

src/Controller/Async/Stack.php
```diff
@@ -38,21 +41,28 @@ public function add(Request $request)
...
         /** @var FileInterface|null $removed */
         $file = $stack->add($filename, $removed);
 
         $panel = $this->render('@bolt/components/stack/panel-item.twig', ['file' => $file]);
+        $panelEvent = new GetResponseForControllerResultEvent($this->app['kernel'], $request, HttpKernelInterface::SUB_REQUEST, $panel);
+        $dispatcher->dispatch(KernelEvents::VIEW, $panelEvent);
+
         $list = $this->render('@bolt/components/stack/list-item.twig', ['file' => $file]);
+        $listEvent = new GetResponseForControllerResultEvent($this->app['kernel'], $request, HttpKernelInterface::SUB_REQUEST, $list);
+        $dispatcher->dispatch(KernelEvents::VIEW, $listEvent);
```
CarsonF: This feels hacky to me. I think we should either call twig directly here or do a sub request.

src/Controller/Base.php
```diff
@@ -465,4 +465,16 @@ protected function resources()
...
+     * Add a middleware that sets the response caching parameters.
+     *
+     * @param int $maxAge Number of seconds to set `Cache-Control s-maxage`
+     */
+    protected function setResponseCaching($maxAge)
+    {
+        $this->app->after(function (Request $request, Response $response) use ($maxAge) {
+            $response->setSharedMaxAge($maxAge)->setPublic();
+        });
+    }
```
CarsonF: This isn't good because it applies the middleware to all requests. It would be better to set on route definition so it only gets applied to that route. 

I'm also worried that boiling it down to `$maxAge` doesn't get enough flexibility. Maybe changing the method name to setResponseMaxAge instead that way another cache helper could be created if needed.

src/Controller/Frontend.php
```diff
@@ -62,8 +62,8 @@ public function before(Request $request)
         if ($this->getOption('general/maintenance_mode')) {
             if (!$this->isAllowed('maintenance-mode')) {
                 $template = $this->templateChooser()->maintenance();
-                $response = $this->render($template);
-                $response->setStatusCode(Response::HTTP_SERVICE_UNAVAILABLE);
+                $html = $this->app['twig']->render(reset($template));
```
CarsonF: If the theme specific template doesn't exist there's no fallback to the global one now. Call twig->resolveTemplate->render().

Also you could still use a TemplateResponse here.

tests/phpunit/unit/Controller/FrontendTest.php
```diff
@@ -405,6 +480,20 @@ public function testSearchListing()
...
         $response = $this->controller()->search($this->getRequest());
 
+        $this->assertTrue($response instanceof TemplateView);
+        $this->assertSame('search.twig', $response->getTemplate());
+    }
+
+    /**
+     * @group legacy
+     */
+    public function testSearchLegacy()
```
CarsonF: If you do `testLegacySearch` you can skip the group annotation.

src/Controller/Frontend.php
GawainLynch: > Also you could still use a TemplateResponse here.

No you can't, it is too early in the cycle and throws excepptions

tests/phpunit/unit/Controller/FrontendTest.php
GawainLynch: Are you freaking serious! That is awesome!

src/Controller/Async/General.php
GawainLynch: Consolidated. 

I know this outcome from me, it is one where I think the problem most of the way though, then get partially side tracked in the process by a linked problem, come back and think I must have had it right so failing to take the thought pattern just that *one last step* :laughing:

src/Controller/Async/Stack.php
GawainLynch: Haha … this is one of those that I was desperate for you to wake up to go over … let me expand …

We need to send a JSON response object … that, um, contains fields with HTML … C.D.O. level 11 but still. So the initial version of the commit just used Twig directly.

However, when reviewing my change(s), it dawned on me … `TemplateView`'s goal in life is to allow people to respond to render events, and while to some the Stack is a small feature, to others the can't live without it … or me who'd like to be able to remove it from **my** account's rendering … i.e. it is one of those renders we want to have `KernelEvents::VIEW` dispatched for.

I'm going to dig this morning, and see how much work to add a sub request, but TBH for 3.3 due out in days, I'm more than a little cautious about adding routes to do this at this stage … 3.4 a different story.

src/Controller/Frontend.php
GawainLynch: Tried switching back to `$response = $this->render($template);`, and this is what we get:

```
RuntimeException in MiddlewareListener.php line 60:

A before middleware for route "homepage" returned an invalid response value. Must 
return null or an instance of Response.

in MiddlewareListener.php line 60
at MiddlewareListener->onKernelRequest(object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher))
at call_user_func(array(object(MiddlewareListener), 'onKernelRequest'), object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher)) in WrappedListener.php line 61
at WrappedListener->__invoke(object(GetResponseEvent), 'kernel.request', object(EventDispatcher))
at call_user_func(object(WrappedListener), object(GetResponseEvent), 'kernel.request', object(EventDispatcher)) in EventDispatcher.php line 184
at EventDispatcher->doDispatch(array(object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener), object(WrappedListener)), 'kernel.request', object(GetResponseEvent)) in EventDispatcher.php line 46
at EventDispatcher->dispatch('kernel.request', object(GetResponseEvent)) in TraceableEventDispatcher.php line 140
at TraceableEventDispatcher->dispatch('kernel.request', object(GetResponseEvent)) in HttpKernel.php line 125
at HttpKernel->handleRaw(object(Request), '1') in HttpKernel.php line 64
at HttpKernel->handle(object(Request), '1', true) in Application.php line 586
at Application->handle(object(Request)) in Application.php line 563
at Application->run(object(Request)) in Application.php line 96
at Application->run() in index.php line 12
```

So switching to `$html = $twig->resolveTemplate($template)->render([]);`

tests/phpunit/unit/Controller/FrontendTest.php
GawainLynch: I'm voting for the annotation in the end, as:
  * Consistency
  * `phpunit --filter testSearch` will run both, so if I am debugging a change I get the legacy test running without thinking about it

But "tested" the name change and it does as you say, and I think that's freaking cool nonetheless!

src/Controller/Async/Stack.php
GawainLynch: Sub requests it is! It's tiny and works. :+1:

GawainLynch:
@CarsonF if you're good where it is now, let me know and I'll drop the helper commit and squash the cache middleware change into the original.

src/Controller/Frontend.php
CarsonF: `TemplateResponse` not `TemplateView`. The only extra functionality it would give is the template name, so it's small.

src/Controller/Async/Stack.php
```diff
@@ -20,10 +21,20 @@ protected function addRoutes(ControllerCollection $c)
     {
         $c->post('/stack/add', 'add')
             ->assert('filename', '.*')
```
CarsonF: This is probably a derp on my part from last time I modified this route. I don't think the assert is needed here anymore.

src/Controller/Async/Stack.php
```diff
@@ -85,4 +127,22 @@ public function show(Request $request)
...
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    private function makeItemSubRequest(Request $request, $routeName, $fileName)
+    {
+        $panelUri = $this->generateUrl($routeName, ['fileName' => urlencode($fileName)]);
+        $subRequest = Request::create($panelUri, Request::METHOD_GET, [], $request->cookies->all(), [], $request->server->all());
+        $subRequest->setSession($request->getSession());
+
+        return $this->app->handle($subRequest, HttpKernelInterface::SUB_REQUEST, false);
+    }
```
CarsonF: Eww I don't like that at all. I would just go back to rendering with twig directly.

src/Controller/Async/General.php
```diff
@@ -32,14 +32,26 @@ protected function addRoutes(ControllerCollection $c)
             ->bind('changelogrecord');
 
         $c->get('/dashboardnews', 'dashboardNews')
-            ->bind('dashboardnews');
+            ->bind('dashboardnews')
+            ->after(function (Request $request, Response $response) {
+                $response->setSharedMaxAge(3600)->setPublic();
```
CarsonF: shared max age sets public for you.

src/Controller/Backend/Authentication.php
```diff
@@ -59,10 +60,19 @@ public function getLogin(Request $request, $resetCookies = false)
...
+                if ($resetCookies) {
+                    $response->headers->clearCookie($this->app['token.authentication.name']);
+                }
+            });
+        } else {
+            /** @deprecated Deprecated since 3.3, to be removed in 4.0. */
+            $response->setVary('Cookies', false)->setMaxAge(0)->setPrivate();
+            if ($resetCookies) {
+                $response->headers->clearCookie($this->app['token.authentication.name']);
+            }
```
CarsonF: This should be changed to a route middleware too. You can just check if the `$response->isRedirect()` before modifying headers.

tests/phpunit/unit/Controller/FrontendTest.php
CarsonF: I've already used the legacy prefix instead of the annotation. Your argument for filter is so you can run `testSearchLegacy` and `testSearchWithFilters` with one filter? Ok :expressionless:

src/Controller/Frontend.php
GawainLynch: Derp … OK, I can't read … updated.

src/Controller/Async/Stack.php
GawainLynch: This is how [Silex documents it](https://silex.sensiolabs.org/doc/2.0/cookbook/sub_requests.html), and as per the original thread on this bit, the events are still desired … So?

CarsonF: I still don't think it's necessary for sub-renders like this. If someone really needs to get into the middle of this they can override the controller. This is an "internal" route anyways.

src/Controller/Async/General.php
GawainLynch: :woman_facepalming: … forgot, good call. Updated them all

src/Controller/Backend/Authentication.php
GawainLynch: Please tell me I didn't miss that whole block :confused: 

Updated, and working beautifully

src/Controller/Async/Stack.php
GawainLynch: Removed

tests/phpunit/unit/Controller/FrontendTest.php
GawainLynch: Stalemate on personal preference I guess, It gets ripped out in `master` anyway

GawainLynch: Conceded … changed.

GawainLynch:
You're welcome! #teamwork!

PR #6626, opened 2017-05-06
[3.3] Only set environment for Nut Twig commands on CLI

GawainLynch:
This will prevent the classes from building on web requests :woman_facepalming:


bobdenotter:
🤷‍♀️

PR #6630, opened 2017-05-07
Make sure `wysiwyg` fields have the correct label, strip trailing `:` from alerts.

### LINKED ISSUE #6629: [BUG] Validation field names for text areas
It seems that the system name for the field is used instead of the field label for text area fields:

The field labels are as follows:

Details
-------

 - Relevant Bolt Version: v3.2.9
 - PHP version: 7.0
 - For UX/UI issues: Firefox

bobdenotter:
Fixes #6629

### DIFF
--- a/app/src/js/obj-validation.js
+++ b/app/src/js/obj-validation.js
@@ -86,13 +86,16 @@ Bolt.validation = (function () {
      */
     function showAlertbox(id, field, msg) {
         var alertbox;
+        var fieldname = $('label[for="' + field.id + '"]').contents().first().text().trim() || field.name;
+
+        fieldname = fieldname.replace(/(.*):+$/g, "$1" );
 
         alertbox = Bolt.data(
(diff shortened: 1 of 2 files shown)


app/src/js/obj-validation.js
```diff
@@ -86,13 +86,16 @@ Bolt.validation = (function () {
      */
     function showAlertbox(id, field, msg) {
         var alertbox;
+        var fieldname = $('label[for="' + field.id + '"]').contents().first().text().trim() || field.name;
+
+        fieldname = fieldname.replace(/(.*):+$/g, "$1" );
```
GawainLynch: I might be missing something here … but why are we even getting these colons in the first place? Can we attack root cause?

bobdenotter: It takes the labels, as set in `label: "Title:"`. I tend to include a colon in there, other people might not.

PR #6634, opened 2017-05-07
Set default values for Video field, prevent exceptions

### LINKED ISSUE #6601: [BUG] Adding video with height but no width produces error: "Notice: Undefined index: width"
After saving a video with a height set, but the width field empty, editing or viewing the record produces an error. In production this would usually won't be a problem since it's only a notice. Though, when `debug: true` and `debug_error_level: -1` the site breaks as such:

screencapture at fri apr 28 14 51 14 cest 2017

The problem is that in `ContentValuesTrait:388` the aspect ratio is calculated. Only when there is a height set. It doesn't check if a width is available.

Details
-------

 - Relevant Bolt Version: 3.2.11
 - PHP version: 7.1.4
 - Used web server: Apache

Reproduction
------------

1. Set `debug:true` and `debug_error_level: -1`
2. Add field of type `video` to a contenttype.
3. Create a record and use the following url for the video field:

   `https://www.mixcloud.com/NTSRadio/the-do-you-breakfast-show-w-charlie-bones-19th-april-2017/`

   The height field will be filled, but the width field will stay empty. This is probably because [mixcloud provides a width](https://www.mixcloud.com/oembed/?format=json&light=1&url=https%3A%2F%2Fwww.mixcloud.com%2FNTSRadio%2Fthe-do-you-breakfast-show-w-charlie-bones-19th-april-2017%2F) of `"100%"` which doesn't validate as an integer.
4. Save the record
5. You will be unable to view the record on your site
6. You will be unable to edit the record in the bolt control panel
​

Expected result:
An embed with it's height set, but no width set, or an error while trying to save the record.

Actual result:
An Exception.

bobdenotter:
Fixes #6601

### DIFF
--- a/src/Storage/Entity/ContentValuesTrait.php
+++ b/src/Storage/Entity/ContentValuesTrait.php
@@ -374,7 +374,15 @@ trait ContentValuesTrait
             }
 
             if ($this->fieldtype($key) === 'video' && is_array($this->values[$key]) && !empty($this->values[$key]['url'])) {
-                $video = $this->values[$key];
+                $defaultValues = [
+                    'html' => '',
+                    'responsive' => '',
+                    'width' => '1',
+                    'height' => '1',
+                    'ratio' => '1',
+                ];
+
+                $video = array_replace($defaultValues, $this->values[$key]);
 
                 // update the HTML, according to given width and height
                 if (!empty($video['width']) && !empty($video['height'])) {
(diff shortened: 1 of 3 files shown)


src/Storage/Entity/ContentValuesTrait.php
```diff
@@ -374,7 +374,15 @@ public function setValues(array $values)
...
-                $video = $this->values[$key];
+                $default_values = [
+                    'html' => '',
+                    'responsive' => '',
+                    'width' => '1',
+                    'height' => '1',
+                    'ratio' => '1',
+                ];
+
+                $video = array_replace($default_values, $this->values[$key]);
```
GawainLynch: snake_case … please tell me you didn't :stuck_out_tongue_winking_eye:

bobdenotter: I did not do any such thing! Where did you get that preposterous idea!

PR #6636, opened 2017-05-08
[3.2] Switch Embed.ly to oEmbed API

### LINKED ISSUE #6635: Replace baked-in embed.ly with something else
Recently, embed.ly changed their business model, so we can't use it in Bolt anymore. At least, not in the way we used to. We need to replace it asap. 

More information: [embedly faq](http://docs.embed.ly/docs/new-pricing-faq) / [blogpost with marketing bingo](https://blog.embed.ly/the-best-is-yet-to-come-f5aac2bd6b13)

Current options on the table are: 

 - 1. Fork over $99 per month to keep the service running as is. We're going to do this for May, to give us some time to implement another option.

 - 2. Switch over to (third party / hosted) noembed.com, urlembed.com or iframely.com. We can do this rather quickly, and include it in a 3.2.x release rather painlessly. 

 - 3. Provide configuration options in Bolt, so people can use either the "default" provider that is to be determined, or their own third-party or self-hosted option

 - 4. Integrate `embed/embed` with Bolt, so we can use an "internal" provider. 

 - 5. Self-host a noembed installation on embed.bolt.cm. This would seem "professional", and we're less dependant on third parties, but it will require extra maintenance from us on another module. Additionally, there might be privacy concerns, because people might question why embedded URL's are sent to a bolt.cm subdomain. 

Currently i'm leaning towards option 1 (for the short term), and get a 3.2.x out of the door with option 2 and 3. 
If we then also provide either option 4 or 5 later on, it can still work as a configurable option in Bolt: Use what we provide out-of-the-box, or select your own self-hosted or third-party solution.

GawainLynch:
Does the job, gets us through, relies on the SF Forms token to be valid in editcontent …
… RFC

Fixes #6635

### DIFF
--- a/app/src/js/widgets/fields/fieldVideo.js
+++ b/app/src/js/widgets/fields/fieldVideo.js
@@ -107,11 +107,14 @@
          */
         _update: function () {
             var self = this,
-                url = 'https://api.embed.ly/1/oembed',
+                url = bolt.data('endpoint.embed'),
+                form = $('#id').closest('form'),
+                token = form.find('input[name="bolt_csrf_token"]').val(),
                 request = {
-                    format: 'json',
-                    key:    '51fa004148ad4d05b115940be9dd3c7e',
-                    url:    self._ui.url.val()
+                    format:          'json',
+                    url:             self._ui.url.val(),
+                    bolt_csrf_token: token,
+                    provider:        'oembed'
                 };
 
             // If val is emptied, clear the video fields.
@@ -121,7 +124,7 @@
                 this._ui.refresh.prop('disabled', true);
                 self._ui.spinner.addClass('fa-spin');
 
-                $.getJSON(url, request)
+                $.post(url, request)
                     .done(function (data) {
                         self._set(data);
                     })
--- a/app/view/twig/editcontent/fielddata/_video.twig
+++ b/app/view/twig/editcontent/fielddata/_video.twig
@@ -1,0 +1,2 @@
+
+{{ data('endpoint.embed', url('embedRequestEndpoint')) }}
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "composer/composer": "^1.2",
         "doctrine/cache": "^1.5",
         "doctrine/dbal": "^2.5",
+        "embed/embed": "^3.0",
         "erusev/parsedown-extra": "~0.7",
         "ext-mbstring": "*",
         "ext-pdo": "*",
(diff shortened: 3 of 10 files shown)


app/src/js/widgets/fields/fieldVideo.js
```diff
@@ -107,11 +107,13 @@
          */
         _update: function () {
             var self = this,
-                url = 'https://api.embed.ly/1/oembed',
+                url = bolt.conf('paths.async') + 'oembed/request',
```
CarsonF: I've been trying to get away from string concatenation with paths. Can you assign url generator result to a data attribute in twig and fetch that here?

src/Request/Oembed/GuzzleDispatcher.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Bolt\Request\Oembed;
```
CarsonF: I know this is making a request to (O)embed, but I don't think it belongs here. It has nothing to do with our Symfony HTTP Requests. Instead it's a service we are calling. I think `Bolt\(O)embed` would be better.

src/Controller/Async/Oembed.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param Url $url
+     *
+     * @return \Embed\Adapters\Adapter
+     */
+    protected function getUrlAdapter(Url $url)
+    {
+        $oEmbedFactory = $this->app['oembed.factory'];
+
+        return $oEmbedFactory($url);
+    }
```
CarsonF: I kinda think having our own service class that wraps the ombed api that the controller could call here would be better. Then the controller doesn't have to know anything about the `Embed\*` code. But if there's a rush don't worry about it.

GawainLynch: It needs to be a factory, and I have over spend hugely my allowed computer time today.

So if someone wants to pick it up tonight, go for it … but a) this is 3.2; b) urgent

src/Request/Oembed/GuzzleDispatcher.php
GawainLynch: Tomorrow, or anyone with push access

app/src/js/widgets/fields/fieldVideo.js
GawainLynch: All of it is getting re-done … I just went with what is there … maybe tomorrow, but I think that will have to wait for 3.3 TBH

GawainLynch: Done :wink:

src/Controller/Async/Oembed.php
GawainLynch: Done … I got a better sense of what you meant after sleeping on it, and yeah :smile:

src/Request/Oembed/GuzzleDispatcher.php
GawainLynch: Done

bobdenotter:
Merging in! :shipit:

PR #6654, opened 2017-05-11
Update messages.es_ES.yml

zomars:
Some grammar fixes, "Limpiar caché" means "To clean the cache" which is not right for a success message. I propose something like "Cache cleanup completed."


bobdenotter:
Thanks, @zomars! 👍

PR #6655, opened 2017-05-11
Implement missing method `LogSystemRepository::countSystemLog()`

bobdenotter:
Analogous to https://github.com/bolt/bolt/blob/release/3.3/src/Storage/Repository/LogChangeRepository.php#L48-L71

This method is needed for a check i'm adding to `bolt/configuration-notices`.


GawainLynch:
There is a [`count()` method implemented](https://github.com/bolt/bolt/blob/release/3.3/src/Storage/Repository.php#L97-L106) for all repositories … the one in log change should be deprecated.

PR #6657, opened 2017-05-12
[3.2] Fix orphan taxonomy entities on delete

### LINKED ISSUE #6651: [BUG] Tags are not deleted from the database when piece of content is deleted
It seems as though a tag is not deleted in the database when it's assigned content gets deleted through the cms. I've talked to @bobdenotter about this and he says a tag _should_ be deleted when the only piece of content it's assigned to is deleted, but the reference to the piece of content stays in the db, although the content isn't there anymore

Details
-------
 - Bolt Version: 3.2.11 
 - Install type: The 'Easy way' / Recommended way
 - PHP version: 5.6
 - Used web server: Apache

Reproduction
------------

1. Create a taxonomy type tags, or just use the default
2. Assign to a content type instance a tag named 'supertag' and save the content type instance
3. View the content instance, view the tags listing, they're there!
4. Go back to Bolt and delete the content instance, so the tag is not coupled to any piece of content any more
5. View the database and see the tag is still there, assigned to the piece of content's id
6. if there is an overview of tags, the tag also is still visible, and subsequently if you visit it's listing page you get an error because of no slug 'supertag' in taxonomy 'tags'

GawainLynch:
Fixes #6651 by adding a pre-delete event to delete an entity's taxonomy entities.

/me waves at @sbonardt

### DIFF
--- a/src/EventListener/StorageEventListener.php
+++ b/src/EventListener/StorageEventListener.php
@@ -8,8 +8,10 @@ use Bolt\Events\StorageEvents;
 use Bolt\Exception\AccessControlException;
 use Bolt\Logger\FlashLoggerInterface;
 use Bolt\Request\ProfilerAwareTrait;
+use Bolt\Storage\Collection;
 use Bolt\Storage\Database\Schema;
 use Bolt\Storage\Entity;
+use Bolt\Storage\EntityManagerInterface;
 use Bolt\Storage\EventProcessor;
 use Bolt\Translation\Translator as Trans;
 use PasswordLib\Password\Factory as PasswordFactory;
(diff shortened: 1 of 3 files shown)


bobdenotter:
We should probably do the same for "relations", right? 

Did a quick test, and relations aren't removed either, when you delete a record.

GawainLynch:
> We should probably do the same for "relations", right?

Yeah, easy addition.

anketwokings:
Awesome! Thanks for the quick fix!

GawainLynch:
You're very welcome! :smile_cat:

PR #6663, opened 2017-05-13
[3.4] Base entity get()/set()

GawainLynch:
This gives all entities the ability to use `$entity->get($fieldName')` or `$entity->set($fieldName, $fieldValue')`.

Previously only ContentType entities could use this, and the `set()` is preserved in the CT entity as it additionally filters the `$value` array, and while that's fine for our ContentTypes, it is not a decision we should make for other entities.

Also with the current state, should you make the mistake of calling `get()` or `set()` on a non-CT entity currently, the magic method strips the prefix, ending up with an empty key to tray and find a value with.

A note on the forwarding of the calls, a number of entity getters & setters handle mutation.

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -47,31 +47,13 @@ class Content extends Entity
     protected $sortorder;
 
     /**
-     * Getter for templates using {{ content.get(title) }} functions.
-     *
-     * @param string $key
-     *
-     * @return mixed
-     */
-    public function get($key)
-    {
-        if ($key === 'title') {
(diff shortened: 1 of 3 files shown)


rossriley:
I think it would be more semantically correct to use 

`$method = 'get' . ucfirst($this->camelize($key));` 

at the moment the magic `__call` should handle the variation but camelizing here will make sure that a property that is named `my_entity_var` is called with the `->getMyEntityVar()` method should the end user have provided the method directly.

Other than that all good.

GawainLynch:
Done!

PR #6679, opened 2017-05-16
Fix PathResolver infinite recursion

CarsonF:
- A `PathResolutionException is now thrown when infinite recursion is detected.
- An `InvalidArgumentException` is now thrown when trying to define a path that references itself.
- Added `names()` which returns a list of defined names.
- Removed `resolveAll()` it wasn't used and if needed one can loop using `names()` method.


GawainLynch:
:shipit:

PR #6681, opened 2017-05-16
ScriptHandler updates

CarsonF:
- [x] Added new `ScriptHandler::updateProject` handler where all changes we want to run on updates can go. Added code to automatically add this handler to user's `composer.json` file (or tell them how if we can't modify the file).
- [x] Added code (in `updateProject`) to update user's `.bolt.yml` file for 3.3 `PathResolver` changes.
- [x] Updated create project handler to use `PathResolver` and interactively allow path customization (thanks to #6680).
- [x] Removed check for `.bolt.yml` file since it is optional now.
- [x] Removed syncing `.gitignore`. Version in `composer-install` has been fixed.
- [x] Update `installAssets` to use `PathResolver`.
- [x] Update `installThemeAndFiles` to use `PathResolver`.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -1,0 +1,179 @@
+<?php
+
+namespace Bolt;
+
+use Bolt\Configuration\Composer;
+use Bolt\Configuration\MyComposer;
+use Bolt\Configuration\PathResolverFactory;
+use Bolt\Configuration\ResourceManager;
+use Bolt\Configuration\Standard;
+use Bolt\Extension\ExtensionInterface;
+use LogicException;
+use Silex;
+use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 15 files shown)


CarsonF:
I think this is ready to go now, but I probably still needs some monkey testing.

src/Bootstrap.php
```diff
@@ -0,0 +1,179 @@
...
+use Webmozart\PathUtil\Path;
+
+/**
+ * Handles creating the Application with .bolt.[yml|php] configuration.
+ *
+ * This does not handle autoloading.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ */
+class Bootstrap
```
GawainLynch: Total and utter joy :+1: :heart:

PR #6685, opened 2017-05-18
[3.3] Remove remaining Twig global use

GawainLynch:
As per discussion in Tuesday's meeting.

@CarsonF On the `BackendBase` controller one, I know, I know … but scope … believe me they are used all over backend templates, and refactoring the entire backend template logic this close to release is just not on 

… my O.C.D. too :disappointed:

### DIFF
--- a/app/view/twig/_base/_listing.twig
+++ b/app/view/twig/_base/_listing.twig
@@ -301,13 +301,13 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
                     {% for taxonomyslug, values in content.taxonomy %}
                         {% if values|length > 1 %}
                             <li>
-                                <a class="nolink">{{ config.get('taxonomy')[taxonomyslug].name }}:
+                                <a class="nolink">{{ app.config.get('taxonomy')[taxonomyslug].name }}:
                                     <i class="fa fa-tag"></i> {{ values|join(", ")|excerpt(24) }}
                                 </a>
                             </li>
                         {% else %}
                             <li>
-                                <a class="nolink">{{ config.get('taxonomy')[taxonomyslug].singular_name }}:
+                                <a class="nolink">{{ app.config.get('taxonomy')[taxonomyslug].singular_name }}:
                                     <i class="fa fa-tag"></i> {{ values|first|excerpt(24) }}
                                 </a>
                             </li>
--- a/app/view/twig/_nav/_footer.twig
+++ b/app/view/twig/_nav/_footer.twig
@@ -3,7 +3,7 @@
     <footer id="bolt-footer" class="hidden-xs {% if global.request.cookies.get('sidebar') %}bolt-footer-hidden{% endif %}">
         {{ __('general.phrase.provided-by-colon') }}
         {{ app.config.get('general/branding/provided_link')|raw }} &ndash;
-        <i class="fa fa-cog"></i> <b>Bolt {{ bolt_version }}</b>:
+        <i class="fa fa-cog"></i> <b>Bolt {{ context.bolt_version }}</b>:
         <i class="fa fa-heart"></i > <a href="{{ path('about') }}">{{ __('general.about') }}</a> &ndash;
         <i class="fa fa-external-link-square"></i> <a href="http://bolt.cm" target="_blank">Bolt.cm</a>
     </footer>
@@ -11,7 +11,7 @@
 {% else %}
 
     <footer id="bolt-footer" class="hidden-xs bolt-footer-mini {% if global.request.cookies.get('sidebar') %}bolt-footer-hidden{% endif %}">
-        <i class="fa fa-cog"></i> <a href="http://bolt.cm" target="_blank">Bolt {{ bolt_version|replace({'alpha': 'α', ' beta ': 'β'}) }}</a> &ndash;
+        <i class="fa fa-cog"></i> <a href="http://bolt.cm" target="_blank">Bolt {{ context.bolt_version|replace({'alpha': 'α', ' beta ': 'β'}) }}</a> &ndash;
         <i class="fa fa-heart"></i > <a href="{{ path('about') }}">{{ __('general.about') }}</a>
     </footer>
 
--- a/src/Controller/Backend/BackendBase.php
+++ b/src/Controller/Backend/BackendBase.php
@@ -42,6 +42,14 @@ abstract class BackendBase extends Base
             $context = ['context' => $context];
         }
 
+        /** @deprecated Deprecated since 3.0, to be removed in 4.0. */
+        $oldGlobals = $this->app['twig']->getGlobals();
+        foreach (['user', 'users', 'bolt_version'] as $key) {
+            if (!isset($context['context'][$key])) {
+                $context['context'][$key] = $oldGlobals[$key];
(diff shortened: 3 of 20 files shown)


app/view/twig/_base/_listing.twig
```diff
@@ -301,13 +301,13 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
                     {% for taxonomyslug, values in content.taxonomy %}
                         {% if values|length > 1 %}
                             <li>
-                                <a class="nolink">{{ config.get('taxonomy')[taxonomyslug].name }}:
+                                <a class="nolink">{{ app.config.get('taxonomy')[taxonomyslug].name }}:
```
CarsonF: Are we deprecating the `config` global?

app/view/twig/_nav/_footer.twig
```diff
@@ -3,15 +3,15 @@
     <footer id="bolt-footer" class="hidden-xs {% if global.request.cookies.get('sidebar') %}bolt-footer-hidden{% endif %}">
         {{ __('general.phrase.provided-by-colon') }}
         {{ app.config.get('general/branding/provided_link')|raw }} &ndash;
-        <i class="fa fa-cog"></i> <b>Bolt {{ bolt_version }}</b>:
+        <i class="fa fa-cog"></i> <b>Bolt {{ context.bolt_version }}</b>:
```
CarsonF: The constant can be used here instead.

src/Controller/Backend/BackendBase.php
```diff
@@ -42,6 +42,14 @@ protected function render($template, array $context = [], array $globals = [])
             $context = ['context' => $context];
         }
 
+        /** @deprecated Deprecated since 3.0, to be removed in 4.0. */
+        $oldGlobals = $this->app['twig']->getGlobals();
+        foreach (['user', 'users', 'bolt_version'] as $key) {
+            if (!isset($context['context'][$key])) {
+                $context['context'][$key] = $oldGlobals[$key];
+            }
+        }
```
CarsonF: Why? These are already globals that can be lazily eval'ed.

theme/base-2016/extrafields.twig
```diff
@@ -4,7 +4,7 @@
 
     <h1>{{ record.title }}</h1>
 
-    {{ fields(template = 'partials/_sub_fields.twig') }}
+    {{ fields(record = record, template = 'partials/_sub_fields.twig') }}
```
CarsonF: This still doesn't handle the case where record is left out.

bobdenotter: In what scenario is `record` not present for a single record page?

CarsonF: I mean when the user doesn't pass record. i.e. The PHP code still tries to pull it from globals.

GawainLynch: > I mean when the user doesn't pass record. i.e. The PHP code still tries to pull it from globals.

Yes, but that is going to be gone in master … the point I wanted to make by setting that parameter is that it is going to be required in v4, start using it now. :wink:

app/view/twig/_base/_listing.twig
GawainLynch: You had it marked as so, and the template base was split between them … what's it going to be?

theme/base-2016/extrafields.twig
CarsonF: Ok so we should add a deprecation message to that function then. And throw exception if twig_globals is disabled and a record isn't passed in.

src/Controller/Backend/BackendBase.php
GawainLynch: If you set globals compatibility off, they fail … I'm all ears.

app/view/twig/_nav/_footer.twig
GawainLynch: ¿que?

theme/base-2016/extrafields.twig
GawainLynch: Starting a new branch now :+1:

app/view/twig/_nav/_footer.twig
CarsonF:
```twig
{{ constant('Bolt\\Version::VERSION') }}
```

src/Controller/Backend/BackendBase.php
CarsonF: https://github.com/bolt/bolt/blob/release/3.3/src/EventListener/ConfigListener.php#L158-L159

app/view/twig/_base/_listing.twig
CarsonF: Where did I have it marked as deprecated?

src/Controller/Backend/BackendBase.php
GawainLynch: So `user` & `users` will remain as globals?

app/view/twig/_nav/_footer.twig
GawainLynch: Done

theme/base-2016/extrafields.twig
GawainLynch: Done

src/Controller/Backend/BackendBase.php
CarsonF: Idk. Where are they used? Maybe not `users`, but I could see `user` being useful. Maybe a function would be better though that way access can be checked when requested?

CarsonF: If we don't want them as globals, and we want to pass them for all backend routes (which I may have a problem with), then they should be pulled the proper way, not from Twig globals.

app/view/twig/_base/_listing.twig
GawainLynch: That was my reading/remembering of https://github.com/bolt/bolt/blob/release/3.3/src/Twig/Extension/BoltExtension.php#L88-L90

src/Controller/Backend/BackendBase.php
GawainLynch: Yeah, fine … what I am trying to get here is stuff in place so I can cut master over with as little friction between now and when 3.x EoLs … 

As I mention in the PR text, there is scope on this … I am **not** going to refactor the entire backend templates this close to release … what is the middle ground here? :smile:

app/view/twig/_base/_listing.twig
GawainLynch: If we're keeping config global, I am happy enough with that, but please make a call, following your decision tree isn't always easy over the wire :rofl:

src/Controller/Backend/BackendBase.php
GawainLynch: > Where are they used

Oh, and as per original PR text, _all over the place_ … hence the problem needing a solution

CarsonF: > As I mention in the PR text, there is scope on this … I am not going to refactor the entire backend templates this close to release … what is the middle ground here? 😄

The leave them as globals, they work fine that way regardless of twig_globals flag.

GawainLynch: Done :+1:

PR #6686, opened 2017-05-18
[3.3] Update global use

GawainLynch:
Follow-up on discussions in Slack & #6685

### DIFF
--- a/app/view/twig/_sub/_messages.twig
+++ b/app/view/twig/_sub/_messages.twig
@@ -2,7 +2,7 @@
     {% import _self as self %}
 
     {# we only want to show these flashes to logged on users. #}
-    {% if user != null %}
+    {% if user|default %}
         {{ self.messages('danger') }}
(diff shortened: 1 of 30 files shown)


app/view/twig/_sub/_messages.twig
```diff
@@ -2,7 +2,7 @@
     {% import _self as self %}
 
     {# we only want to show these flashes to logged on users. #}
-    {% if user != null %}
```
CarsonF: Does this mean that globals are allowed in macros?

GawainLynch: You didn't see my comment in Slack yesterday I take it … yeah, this is [the same point in 3.2](https://github.com/bolt/bolt/blob/release/3.2/app/view/twig/_sub/_messages.twig#L5)

I thought that was the case from my earlier work on that branch, but was happy to capitulate until I had something to go on :wink:

PR #6702, opened 2017-05-23
[3.3] Pre-release code style pass

GawainLynch:


src/Extension/AssetTrait.php
```diff
@@ -196,8 +196,8 @@ private function normalizeAsset(FileAssetInterface $asset)
         }
 
         $message = sprintf(
-            "Couldn't add file asset '%s': File does not exist in either %s or %s directories. Make sure the file exists in either of these locations, by ".
-            "placing the file there manually (for Bundled Extensions) or by uninstalling / installing the extension again (for Managed Extensions).",
+            "Couldn't add file asset '%s': File does not exist in either %s or %s directories. Make sure the file exists in either of these locations, by " .
+            'placing the file there manually (for Bundled Extensions) or by uninstalling / installing the extension again (for Managed Extensions).',
```
bobdenotter: I see where this is coming from, but it looks off, with the line directly above it with `"`.

GawainLynch: Coding standards that we've used for a while, not to mention good practice.

PR #6704, opened 2017-05-24
Add pager to base-2016 search template

anketwokings:
The include of the pager seemed to be missing, the comment was the only thing there


GawainLynch:
Definitely a #1555

PR #6726, opened 2017-05-30
[4.0] Silex 2 … and WebLink

GawainLynch:
**DISCLAIMER**: For the short term, this is a WiP. There **will be changes required** in review!

I have been maintaining the first two commits since February, attempting to track how much work it was for the core branch merger (i.e. me!) to do in the end. It turns out that I was able to keep this running for all that time with almost no time investment. So the reason we have been holding off was largely that.

## Why now?

For a while, with the then upcoming Symfony's 3.3 things started to get a bit interesting, as Silex 1.3.x —
 which is [officially EoL](https://github.com/silexphp/Silex/tree/1.3) — have locked the upper constraint on its Symfony components at [`"~2.3|3.0.*"`](https://github.com/silexphp/Silex/blob/1.3/composer.json). It is blocking more than one upcoming feature. 

Symfony 3.3.0 was released last night, tests have all been passing for a while, so "**Game on!**"

## … and

It is also probably a good point-in-time to start to give ourselves some solid code to target.

## WebLink

This also brings us [WebLink](http://symfony.com/blog/new-in-symfony-3-3-weblink-component) & [asset preloading with HTTP/2 Push](https://symfony.com/blog/new-in-symfony-3-3-asset-preloading-with-http-2-push) … Browser support is [pretty good](http://caniuse.com/#search=prefetch) … anyone getting excited yet?

I have added `{{ prefetch() }}` & `{{ dns_prefetch() }}` to core & `base-2016` assets to play with for a while. **Note:** I've found Firefox surprising gives better data if you want to observe the requests.

## Known Issues 

* Profiler it going mental, saving massive log & dumper metadata files. This can cause it to fail on some pages, depending on how busy your CPU/IO is
* Profiler is slow … see above
* Slow with debugging turned on … see above
* Web toolbar's authentication information tells you there is no user … that is upcoming in future branches for Symfony Security

## Next

This might get some excited, and rightly so. 

But can I point out a single **very important** thing: This only brings one more of our components up to the next major version, it is not a sign of "completion" for 4.0!

There is still **tonnes** of work before 4.0.0, and to be honest, I am not sure it will see until early 2018. We'll see, eyes are on November, but thousands of hours of people's personal free-time has been invested in getting us this far. Keep that in mind :wink:

@CarsonF Go for it if you want to push commits, just leave a note for me in Slack so I don't force push over your work. Also, the `Application` thing … PoC! Just want things in a place where we can discuss changes/concepts/approaches and CRUD from there.

Also, everything is up for discussion, I am not imaging there isn't going to be some amount of changes required before merge … and :koala:


bobdenotter:
Silex 2, whooop whooop

GawainLynch:
@bobdenotter As discussed yesterday, I've dropped the PoC service provider and pushed … lets send this in and work on it in situ. Two months is enough time to review :smile:

bobdenotter:
Here we go! 

tenor

PR #6727, opened 2017-05-30
Preparing 3.3.0 beta 9.

bobdenotter:


src/Version.php
```diff
@@ -19,7 +19,7 @@
      *   Stable      — 3.0.0
      *   Development — 3.1.0 alpha 1
      */
-    const VERSION = '3.3.0 beta 8';
+    const VERSION = '3.3.0 beta 9';
```
GawainLynch: But it needs to be **OVER 9000**

GawainLynch:
Trust me when I say that merging right now, and still generating a Travis failure is proof that #1555 is upon us!

PR #6728, opened 2017-05-30
Fix BC for old thumbnail config paths

### LINKED ISSUE #6648: Smoother transition for "notfound_image" for bolt 3.2 -> 3.3
If you update a site from bolt 3.2 to bolt 3.3, thumbnails will no longer work. 

screen shot 2017-05-10 at 17 11 50

Right-click them, and you'll see the error: 

screen shot 2017-05-10 at 17 09 28

The cause is this: 

```
thumbnails:
    ...
    notfound_image: view/img/default_notfound.png
    error_image: view/img/default_error.png
```

which needs to be updated to: 

```
    notfound_image: bolt_assets://img/default_notfound.png
    error_image: bolt_assets://img/default_error.png
```

Seeing how we have tons of people coming in, who we need to tell "Right click the thumbnail, read the error and install the module it's telling you about", I can see that this is going to cause a lot of headaches when we release 3.3

I suggest we change the code to make this transition a bit smoother.. Perhaps something like this, in `function findDefaultImage()`: 

```
        // Take old-style (pre 3.3) paths into account:
        // view/img/foo.png => bolt_assets://img/foo.png
        if (strpos($path, '://') === false) {
            $path = str_replace('view/', 'bolt_assets://', $path);
        }
```

btw: If an image exists, why does the thumbnailer even need to read the `notfound_image` ?

CarsonF:
Fixes #6648

### DIFF
--- a/src/Provider/ThumbnailsServiceProvider.php
+++ b/src/Provider/ThumbnailsServiceProvider.php
@@ -8,6 +8,7 @@ use Bolt\Filesystem\Exception\DefaultImageNotFoundException;
 use Bolt\Filesystem\Exception\FileNotFoundException;
 use Bolt\Filesystem\Handler\ImageInterface;
 use Bolt\Filesystem\Matcher;
+use Bolt\Helpers\Deprecated;
 use Bolt\Thumbs;
 use Bolt\Thumbs\ImageResource;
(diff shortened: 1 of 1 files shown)


bobdenotter:
👍

PR #6731, opened 2017-05-31
[2.2] Be more specific with DBAL dependency versions

rossriley:
The 1.* version requirement pulls in the latest versions of both annotations and collections and these require PHP 5.6 or later.
Specifying 1.2 and 1.3 respectively ensures the versions match the platform requirements of Bolt.


GawainLynch:
PHP 5.3 failure unrelated

PR #6739, opened 2017-06-03
[4.0] Phase I: Symfony 3 & Silex 2 project filesystem layout

GawainLynch:
tl;dr Keep volatile/generated data out of `app/` and in `var/` … as it done on real operating systems! (yes @SahAssar, that dig/joke was especially for your amusement :wink:)

## Background

In Symfony 3, things got a little more sane as far as the location of different filesystem targets, full details on the [Symfony Best Practices guide](http://symfony.com/doc/current/best_practices/creating-the-project.html#structuring-the-application), but in brief a **project's** layout should look something like this (slightly adjusted for Bolt)

```
%site%/
├─ app/
│  ├─ config/
│  └─ Resources/
├─ bin
│  └─ nut
├─ src/
│  └─ {bundled(s)}
├─ var/
│  ├─ cache/
│  ├─ logs/
│  └─ sessions/
├─ tests/
├─ vendor/
└─ web/
```
Where:

| | Description
|-|----------------
| `app/` | The application configuration, templates and translations.
| `bin/` | Executable files (e.g. bin/console).
| `src/` | The project's PHP code.
| `tests/` | Automatic tests (e.g. Unit tests).
| `var/` | Generated files (cache, logs, etc.).
| `vendor/` | The third-party dependencies.
| `web/` | The web root directory.

## PR Changes
- Add 'var' mount point to filesystem/path resolver
- Move `app/cache` to a `var/cache`
- Move session store to an **un-hidden directory** in var
- Move COMPOSER_HOME to `var/composer/`
- Move debug log file(s) to `var/log/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 # Bolt specific stuff. Don't put config files, databases or 'vendor' in Git.
-app/cache/*
 app/config/*.yml
 app/config/extensions/
 app/database/*
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ app/database/*
 files/*
 extensions/*
 thumbs/
+var/*
 vendor/
 theme/*
 !theme/base-*
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -275,9 +275,10 @@ debug_trace_argument_limit: 4       # Determine how many steps in the backtrace 
 #production_error_level: 8181 # = E_ALL &~ E_NOTICE &~ E_WARNING &~ E_DEPRECATED &~ E_USER_DEPRECATED
 
 # System debug logging
+#
 # This will enable intensive logging of Silex functions and will be very hard on
-# performance and log file size.    The log file will be created in your cache
-# directory.
+# performance and log file size. The log file will be created in your site's
+# var/log/ directory.
 #
 # Enable this for short time periods only when diagnosing system issues.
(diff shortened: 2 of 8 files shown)


bobdenotter:
Looks good to me. Afaict it's an improvement and closer to what Symfony says is "best practices", but the impact to users doesn't seem too large. 👍

GawainLynch:
> but the impact to users doesn't seem too large. 

Not at all :smile: 

The obvious catch is just making sure the new `var/` is writable, as with `app/cache/` … but most of the migration we have a base to handle now, thanks to groundwork recently by @CarsonF

GawainLynch:
Oh, and moving `app/nut` to `bin/nut` is out of scope for this PR, unless it is something we can all agree on, then I'll add it :smile:

CarsonF:
I know this is late...but I don't agree with `app/Resources`. IMO everything in the `app` dir is a "resource". `app/translations` and `app/templates` makes much more sense to me.

GawainLynch:
Not late as such, but these are the types of comments I need/want from people … 

One to see who's looking, and two as I am not happy with currently PR layout in some areas too … e.g. the `Application` thing **must** DIAF as-is. But it is all intended to get people looking, thinking and discussing.

So bring it on, I'll ping you in Slack during the week and maybe we can quickly has a few points like `Resources` out … I think we're both pretty much on the same page, just need to get our context straight with each other :smile:

nlemoine:
Symfony Flex introduces new folder structure convention, read more here: https://www.symfony.fi/entry/porting-a-symfony-3-application-to-flex

PR #6750, opened 2017-06-13
[3.x] Serializable & cacheable entities

GawainLynch:
## Problem A

Due to legacy (in most cases) entity objects in a lot of cases can't `serialize()`, e.g. for things like caches. 
as mostly for legacy reasons. they contain closures like `Application` (e.g. by `Config` being needed via proxy for taxonomy, or `EntityManager for relations`).

This is causing breakage in a number of areas, and causing me to chase :ghost: on two upcoming branches :smile: 

I am seeing broken serialised objects in session caches when the 1% random log-out that isn't caused by config settings, is **actually** serialisation breaking.

Also, as an example of where this is blocking things… 75-90% of the current dashboard loading time (assuming warm caches) is data that could be tag-cached, even in memory … but instead a ridiculous amount of logic in re-re-re-repeated on every page load. Of the example, 3/4 of that is just the menus … see where I am going with this :wink: 

## Problem B

`EntitySerializeTrait` provides the interface contract methods for entity serialisation

`JsonSerializable` :+1:, but not for `Serializable` .

Obviously adding `unserialize()` is a proverbial no-brainer.

However, currently `serialize()` returns an _array_, the interface requires the implementer to do the pre-processing, and return a serialised _string_. Changing that becomes a BC break.

## Proposed Solution

OK, let's talk … This PR makes the serialisation work, but does potentially break some corner-case use. 

There is probably something else I have missed, and am not "sold" on the chosen function name/use of the renamed functions in the collections classes

### DIFF
--- a/src/Storage/Collection/LazyCollection.php
+++ b/src/Storage/Collection/LazyCollection.php
@@ -2,6 +2,7 @@
 
 namespace Bolt\Storage\Collection;
 
+use Bolt\Storage\Entity\Content;
 use Bolt\Storage\EntityProxy;
 use Doctrine\Common\Collections\ArrayCollection;
 
@@ -26,12 +27,13 @@ class LazyCollection extends ArrayCollection
     }
 
     /**
-     *  Force loads the proxy objects and returns the real objects
+     * Force loads the proxy objects and returns the required data for Twig.
      */
-    public function serialize()
+    public function getProxy()
     {
         $output = [];
         foreach ($this as $element) {
+            /** @var Content $proxy */
             $proxy = $element->getProxy();
             $output[] = $proxy->getContenttype() . '/' . $proxy->getSlug();
         }
--- a/src/Storage/Collection/Relations.php
+++ b/src/Storage/Collection/Relations.php
@@ -3,17 +3,19 @@
 namespace Bolt\Storage\Collection;
 
 use Bolt\Exception\StorageException;
+use Bolt\Helpers\Deprecated;
 use Bolt\Storage\Entity;
 use Bolt\Storage\EntityManager;
 use Bolt\Storage\EntityProxy;
 use Doctrine\Common\Collections\ArrayCollection;
+use Serializable;
 
 /**
  * This class stores an array collection of Relations Entities.
  *
  * @author Ross Riley <riley.ross@gmail.com>
  */
-class Relations extends ArrayCollection
+class Relations extends ArrayCollection implements Serializable
 {
     protected $em;
 
@@ -234,13 +236,39 @@ class Relations extends ArrayCollection
         return $collection;
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function serialize()
     {
-        $output = [];
-        foreach ($this as $k => $existing) {
-            $output[$existing->getToContenttype()][] = spl_object_hash($existing);
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::method(3.4, 'toArray');
+
+            $output = [];
+            foreach ($this as $k => $existing) {
+                $output[$existing->getToContenttype()][] = spl_object_hash($existing);
+            }
+
+            return $output;
         }
 
-        return $output;
+        $data = [];
+        foreach ($this as $element) {
+            $data[] = $element;
+        }
+
+        return serialize($data);
+    }
+
+    /**
+     * @internal
+     */
+    public function unserialize($serialized)
+    {
+        $data = unserialize($serialized);
+        foreach ($data['elements'] as $k => $v) {
+            $this->set($k, $v);
+        }
     }
 }
--- a/src/Storage/Collection/Taxonomy.php
+++ b/src/Storage/Collection/Taxonomy.php
@@ -2,19 +2,23 @@
 
 namespace Bolt\Storage\Collection;
 
+use Bolt\Configuration\ConfigurationValueProxy;
+use Bolt\Helpers\Deprecated;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Mapping\MetadataDriver;
 use Closure;
 use Cocur\Slugify\Slugify;
 use Doctrine\Common\Collections\ArrayCollection;
+use Serializable;
 
 /**
  * This class stores an array collection of Taxonomy Entities.
  *
  * @author Ross Riley <riley.ross@gmail.com>
  */
-class Taxonomy extends ArrayCollection
+class Taxonomy extends ArrayCollection implements Serializable
 {
+    /** @var array|null|ConfigurationValueProxy */
     public $config;
 
     /**
@@ -122,7 +126,7 @@ class Taxonomy extends ArrayCollection
     public function getGrouped()
     {
         $types = [];
-        $elements = $this->toArray();
+        $elements = parent::toArray();
         /** @var Entity\Taxonomy $element */
         foreach ($elements as $element) {
             $type = $element->get('taxonomytype');
@@ -241,6 +245,9 @@ class Taxonomy extends ArrayCollection
         }
     }
 
+    /**
+     * @internal
+     */
     public function serialize()
     {
         $output = [];
@@ -248,7 +255,30 @@ class Taxonomy extends ArrayCollection
             $output[] = ['slug' => $existing->getSlug(), 'name' => $existing->getName()];
         }
 
-        return $output;
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::method(3.4, 'toArray');
+
+            return $output;
+        }
+
+        $data = [
+            'elements' => $this->toArray(),
+            'config'   => $this->config,
+        ];
+
+        return serialize($data);
     }
 
+    /**
+     * @internal
+     */
+    public function unserialize($serialized)
+    {
+        $data = unserialize($serialized);
+        $this->config = $data['config'];
+        foreach ($data['elements'] as $k => $v) {
+            $this->set($k, $v);
+        }
+    }
 }
--- a/src/Storage/Entity/Entity.php
+++ b/src/Storage/Entity/Entity.php
@@ -4,12 +4,13 @@ namespace Bolt\Storage\Entity;
 
 use ArrayAccess;
 use JsonSerializable;
+use Serializable;
 
 /**
  * An abstract class that other entities can inherit. Provides automatic getters and setters along
  * with serialization.
  */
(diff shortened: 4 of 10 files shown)


bobdenotter:
Shouldn't we altogether avoid `serialize()` because of security concerns? As in: if somebody manages to pass in an object to `unserialize`, PHP will invoke its `__wakeup()` with arbitrary code.

GawainLynch:
> Shouldn't we altogether avoid serialize

OK, so that rules out using Doctrine, Symfony & Laravel cache components

GawainLynch:
@rossriley if you think the BC break is unacceptable, @CarsonF has suggested switching to detecting what's calling `serialize()`, and responding appropriately … I just have been too short on time to update you, or this PR :smile:

CarsonF:
Yeah let's create a new method `toArray` or something and move all internal calls to that. Then we can detect who's calling serialize by checking if `file` is defined in the first frame of a backtrace. If there is then trigger warning and call toArray. Else serialize and return string.

GawainLynch:
Updated as requested.

src/Storage/Collection/LazyCollection.php
```diff
@@ -26,16 +29,50 @@ public function add($element)
...
     }
+
+    /**
+     * @internal
+     */
+    public function serialize()
+    {
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::warn('Calling serialize() on collections to return an array', 3.4, 'Use toArray() instead.');
```
CarsonF: You could use method instead. 
```php
Deprecated::method(3.4, 'toArray');
```

src/Storage/Entity/Entity.php
```diff
@@ -73,8 +77,36 @@ public function setId($id)
...
         return (string) $this->getId();
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function toArray()
+    {
+        $data = [];
+        foreach ($this as $k => $v) {
```
GawainLynch: It's not my code, just moved.

src/Storage/Collection/LazyCollection.php
GawainLynch: I was aiming for more context, but fine. Done.

src/Storage/Collection/Taxonomy.php
```diff
@@ -241,7 +245,10 @@ public function getSortorder($field, $slug)
         }
     }
 
-    public function serialize()
+    /**
+     * {@inheritdoc}
+     */
+    public function toArray()
     {
```
GawainLynch: Ugh, I want to argue you down on this one … and I'm not 100% convinced of the LSP break. However, that's a moot point in my mind as your follow-up question/point makes the most sense … 

So done!

src/Storage/Entity/Entity.php
CarsonF: @rossriley?

src/Storage/Field/Collection/RepeatingFieldCollection.php
```diff
@@ -283,13 +285,33 @@ public function getEmptySet()
...
-        return $output;
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::method(3.4, 'toArray');
+
+            return $output;
+        }
+
+        return serialize($output);
+    }
```
CarsonF: What about `$em`, `$mapping`, and `$name` properties?

src/Storage/Collection/Taxonomy.php
```diff
@@ -204,7 +208,7 @@ public function getField($fieldName)
      */
     public function filter(Closure $p)
     {
-        $elements = $this->toArray();
+        $elements = parent::toArray();
```
CarsonF: change isn't needed.

src/Storage/Collection/Relations.php
```diff
@@ -234,13 +236,39 @@ public function offsetGet($offset)
...
+        }
+
+        $data = [];
+        foreach ($this as $element) {
+            $data[] = $element;
         }
 
-        return $output;
+        return serialize($data);
+    }
```
CarsonF: Where is the change to not call this method anymore? And all serialize() methods for that matter.

src/Storage/Collection/Relations.php
```diff
@@ -122,7 +124,7 @@ public function update(Relations $collection)
     public function getGrouped()
     {
         $types = [];
-        $elements = $this->toArray();
+        $elements = parent::toArray();
```
CarsonF: Change is not needed.

src/Storage/Collection/Relations.php
```diff
@@ -234,13 +236,39 @@ public function offsetGet($offset)
...
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::method(3.4, 'toArray');
+
+            $output = [];
+            foreach ($this as $k => $existing) {
+                $output[$existing->getToContenttype()][] = spl_object_hash($existing);
+            }
+
+            return $output;
```
CarsonF: This logic is different than what `toArray` would return. We should not suggest it or move this logic to another method someone can call.

src/Storage/Collection/Taxonomy.php
```diff
@@ -241,14 +245,40 @@ public function getSortorder($field, $slug)
...
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1);
+        if (isset($trace[0]['file'])) {
+            Deprecated::method(3.4, 'toArray');
+
+            return $output;
+        }
+
+        $data = [
+            'elements' => $this->toArray(),
+            'config'   => method_exists($this->config, 'getData') ? $this->config->getData() : $this->config,
```
CarsonF: `ConfigurationValueProxy` should be made `Serializable` instead.

src/Storage/Collection/LazyCollection.php
```diff
@@ -26,16 +29,50 @@ public function add($element)
...
     {
         $output = [];
         foreach ($this as $element) {
+            /** @var Content $proxy */
             $proxy = $element->getProxy();
             $output[] = $proxy->getContenttype() . '/' . $proxy->getSlug();
         }
 
         return $output;
     }
```
CarsonF: This should return EntityProxies...or I guess the objects they are proxing..maybe..as that's what every other method on the class works with. 

With the way this is currently coded `get()` could return an object, an EntityProxy, or a string. Huge red flag 👎

src/Storage/Field/Collection/RepeatingFieldCollection.php
GawainLynch: `$em` & `$mapping` are not serializable

src/Storage/Collection/Taxonomy.php
GawainLynch: Yeah, forgot to pull those … I'll grab 'em in the morning

src/Storage/Collection/Taxonomy.php
GawainLynch: WFM

src/Storage/Collection/Relations.php
GawainLynch: As above, derp and changing incoming

src/Storage/Field/Collection/RepeatingFieldCollection.php
CarsonF: Then neither is this class.

src/Storage/Collection/LazyCollection.php
GawainLynch: Yeah, good point actually … Let me dig.

src/Storage/Collection/Taxonomy.php
GawainLynch: Oh, so much nicer already for that one, thanks :smile:

src/Storage/Collection/Relations.php
GawainLynch: That, currently, I need feedback from @rossriley on

src/Storage/Entity/Entity.php
GawainLynch: (Ignore the visual references to 3.3, it is just my 3.3 worktree checked out with this branch)

entity backtrace

GawainLynch: … and believe it or not, the `getUsers()` thing was BC, no idea if what was requiring it to maintain the BC is fixed, and the class, globals, and a tonne else, is going in `master` already or soon.

src/Storage/Collection/LazyCollection.php
GawainLynch: OK, removed the interface and implementation methods, and changed the name of that to `getProxy()` to match `\Bolt\Storage\EntityProxy` … I'll keep digging as to what's calling what internally.

src/Storage/Collection/Relations.php
GawainLynch:
```shell
$ ag "'serialize'" src/
src/Storage/Field/Type/FieldTypeBase.php
94:        $valueMethod = 'serialize' . ucfirst($this->camelize($attribute));

src/Storage/Field/Type/TemplateFieldsType.php
81:        $valueMethod = 'serialize' . ucfirst($key);

src/Storage/Migration/Export.php
113:                if (is_callable([$val, 'serialize'])) {

src/Storage/Entity/MagicAttributeTrait.php
76:        if (strncasecmp($method, 'serialize', 9) == 0) {

src/Storage/Entity/Entity.php
101:            $method = 'serialize' . $k;
106:            $method = 'serialize' . $k;

src/Storage/Entity/TemplateFields.php
44:            if (is_callable([$val, 'serialize'])) {
```

```shell
$ ag ">serialize\(" src/
src/Composer/Package/Dependency.php
62:        return $this->serialize();
72:        return $this->serialize();

src/Composer/Package/DependencyRelationship.php
30:        return $this->serialize();
40:        return $this->serialize();

src/Storage/Field/Type/TemplateFieldsType.php
88:            $value = $this->serialize($value, $metadata);

src/Storage/Field/Collection/RepeatingFieldCollection.php
295:            $output[$collection] = $vals->serialize();

src/Storage/Migration/Export.php
114:                    $val = $val->serialize();

src/Storage/Entity/TemplateFields.php
45:                $val = $val->serialize();

src/Session/SessionStorage.php
379:        $data = $this->serializer->serialize($this->data);
```

GawainLynch: So basically [`\Bolt\Storage\Entity\MagicAttributeTrait::__call()`](https://github.com/bolt/bolt/blob/release/3.4/src/Storage/Entity/MagicAttributeTrait.php#L76-L80) is catching method called, or prefixed with, "serialize" …

PR #6754, opened 2017-06-15
$content type issue in UploadContainer->save

siada:
Cast $content in save to a string to avoid quirks in external modules where $content is an integer

Details
-------

There's an issue in the `flysystem-rackspace` package that causes the lock file upload to fail because the $content of the lock file is generated from `time()` and flysystem-rackspace is expecting that to be a string.

### DIFF
--- a/src/Filesystem/UploadContainer.php
+++ b/src/Filesystem/UploadContainer.php
@@ -41,7 +41,7 @@ class UploadContainer implements ContainerInterface
      */
     public function save($file, $content)
     {
-        $this->filesystem->put($file, $content);
+        $this->filesystem->put($file, (string)$content);
     }
 
     /**


GwendolenLynch (changes requested):
As mentioned (albeit on a mobile device, hence the spare response), this isn't correct … and as you point out, is an inconsistency (?) in the upstream adaptor

bobdenotter:
We discussed this in #general the other day, and while it is an inconsistency upstream, it might be easier to fix here. 

Should @siada first attempt to get it fixed in https://github.com/thephpleague/flysystem-rackspace ?

siada:
My issue actually stems from here:
https://github.com/siriusphp/upload/blob/master/src/Handler.php#L261

the $content at this point it coming through as an int returned by `time()` which moves on down the callstack and eventually ends up here:
https://github.com/guzzle/guzzle/blob/v3.4.0/src/Guzzle/Http/EntityBody.php#L39

I fully understand that that's a very old version of guzzle, however the flysystem-rackspace package depends on php-opencloud which is where the very old Guzzle dependency comes from. 

So there's lots of places it _could_ be fixed, for example php-opencloud could update to the latest version of guzzle which doesn't use EntityBody but the php-opencloud project has been stagnant for a while and has only recently been picked up again however they've still only updated to Guzzle 3.8 not 6.0+ 👎

GawainLynch:
Regardless, making core then cast all data types is just going to break things, and we don't support Guzzle 3.x, and 5.x is only just hanging in there and has been dropped for us in 4.0-dev, Guzzle 6+ :+1: :smile: 

For your use case, you should be able to achieve an override on the mount (which we don't have an adaptor for in `bolt/filesystem` so again, not something we can support)

GawainLynch:
@siada Many thanks. I am not sure if you were silently following along in tonight's meeting, but we've discussed this and as it impacts core in ways that are not desirable to fix an upstream problem that really shouldn't exists in 2017, we're going to close this off. I really hate that part of this job :disappointed: 

Your involvement in the community on a daily basis is wonderful however, and I am sure you'll understand nonetheless. Please also let me know if there is possibly ways to help with interfacing with upstream. I'd be happy to lend a small hand in nudging them in a sensible direction :+1:

siada:
At this stage for me it's become a non-issue anyway as we switched to using the AWS provider instead as it's much more up to date and far better supported compared to the Rackspace extensions.

I fully understand not wanting to merge though, as @bobdenotter said, it would've been easier to fix here, but in the grand scheme of things it wasn't the right option at the end of the day adding a workaround because another package doesn't work right.

GawainLynch:
Funnily enough, I was just helping someone in #composer who was having a problem with Guzzle 5 vs 6 on a pair of packages … I couldn't be help point out this problem.

But yeah, AWS has a lot going for it, like a RFS in Bolt 3.3+ :wink:

JarJak:
This can be fixed by adding autocinfirm setting as an option in #7511

PR #6755, opened 2017-06-15
$content type issue in UploadContainer->save

siada:
Cast $content in save to a string to avoid quirks in external modules where $content is an integer

Details
-------

There's an issue in the `flysystem-rackspace` package that causes the lock file upload to fail because the $content of the lock file is generated from `time()` and flysystem-rackspace is expecting that to be a string.


bobdenotter:
It looks like this change causes tests to fail: https://travis-ci.org/bolt/bolt/jobs/243175759

siada:
opened against incorrect baseline, discussion moved to #6754

PR #6757, opened 2017-06-15
fix complex search crash bug

JannieT:
This PR fixes a bug where a search like "a nice day" will crash the site if there is any content containing "day". 

Details
-------

The crash is triggered because single character search words like "I" or "a" are filtered out before the search in a way that leaves the search words index dirty. When matches are found and the index is used to weight the results, the app crashes.

PS: I don't like brown M&Ms

### DIFF
--- a/src/Storage/Entity/ContentSearchTrait.php
+++ b/src/Storage/Entity/ContentSearchTrait.php
@@ -155,9 +155,9 @@ trait ContentSearchTrait
 
         $wordMatches = 0;
         $cntWords = count($words);
-        for ($i = 0; $i < $cntWords; $i++) {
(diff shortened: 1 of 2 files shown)


GawainLynch:
First up, thank you for you work … and including tests!

I've moved the logic down a notch, to `\Bolt\Storage\Entity\ContentSearchTrait::weighSearchResult()`, as the way it was handling arrays … well … legacy :rofl: 

The updated change is at least twice as fast in my testing of millions of array keys … micro-op, but still better than assuming we know the array keys, as we obviously can't … it invokes the Samuel L. Jackson Law of Assumption.

So again, thank you :+1:

PR #6760, opened 2017-06-17
[3.3] Prevent filelist exceptions when file not found

GawainLynch:
Currently in 3.3-HEAD, if you add a file to a filelist field, and that file is removed afterwards, the page will crash and be un-editable except for directly accessing the database, or replacing (which?) the file :wink: 

```
Twig_Error_Runtime in Template.php line 230:
An exception has been thrown during the rendering of a template ("File not found at path: 2017-06/refactor-wars.png") in "@bolt/editcontent/fields/_filelist.twig" at line 67.
```

This PR:
  - Adds a method to the path bridging `Matcher` class to check for the file's existence, prior to accessing it, which if not found **correctly** throws an exception.
  - Display's "known" information from the database when not found, and changes the icon to a `?`
    - Obviously that is UI, so "go for it" requesting changes to that, I just picked the first thing that came to mind to get the PR in and discussion moving

### DIFF
--- a/app/view/twig/editcontent/fields/_filelist.twig
+++ b/app/view/twig/editcontent/fields/_filelist.twig
@@ -35,7 +35,7 @@
 
 {% set template_item %}
     <div class="item ui-state-default">
-        <div class="icon"><i class="fa fa-file-o"></i><span>%EXT_E%</span></div>
+        <div class="icon"><i class="fa %ICON%"></i><span>%EXT_E%</span></div>
         <input type="text" class="title" value="%TITLE_A%">
         <input type="hidden" class="filename" value="%FILEPATH_A%">
         <a href="#" class="remove"><i class="fa fa-times"></i></a>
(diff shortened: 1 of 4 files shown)


GawainLynch:
> Matching the file to a filesystem isn't the most performant thing. I would hate to duplicate that loop...esp since it is called from a file list loop.

Yep, and originally went for your suggestion, as I thought you'd object … but hey … let's party like it's 2017 :wink: 

Second problem just found, is that `imagelist` has the same problem, but it calls `Matcher::getImage()` … so adding two methods? This says to me, our approach needs modification here (short term).

GawainLynch:
@CarsonF, I have added a "what about this?" approach in the latest commit, if you're good with that I'll clean up the branch commits

GawainLynch:
OK, added one more as RFC seeing as the branch will get rebased anyway … less `app` in Twig.

CarsonF:
Sure that works for me. I want it all to go away anyways. We just need to update storage layer to handle FS objects instead of these strings without mount points.

GawainLynch:
Rebased and GTG

PR #6761, opened 2017-06-18
Extensions assets path

CarsonF:
It makes much more sense to have a `extensions_assets` path for ...extensions assets..than having it map to the web directory directly.

This doesn't change much in practice other than this:
```twig
{{ asset('extensions/carson/extension/file.js', 'extensions') }}
```
to
```twig
{{ asset('carson/extension/file.js', 'extensions') }}
```

/cc @SahAssar

### DIFF
--- a/src/Extension/AssetTrait.php
+++ b/src/Extension/AssetTrait.php
@@ -181,15 +181,17 @@ trait AssetTrait
         }
 
         $app = $this->getContainer();
+        /** @var \Bolt\Filesystem\Manager $filesystem */
+        $filesystem = $app['filesystem'];
 
-        $file = $app['filesystem']->getFile("extensions_assets://{$this->getId()}/$path");
+        $file = $filesystem->getFile("extensions_assets://{$this->getId()}/$path");
         if ($file->exists()) {
             $asset->setPackageName('extensions')->setPath($file->getPath());
 
             return;
         }
 
-        $themeFile = $app['filesystem']->getFile("theme://$path");
+        $themeFile = $filesystem->getFile("theme://$path");
         if ($themeFile->exists()) {
             $asset->setPackageName('theme');
 
--- a/src/Extension/AssetTrait.php
+++ b/src/Extension/AssetTrait.php
@@ -180,16 +180,16 @@ trait AssetTrait
             return;
         }
 
-        $file = $this->getWebDirectory()->getFile($asset->getPath());
+        $app = $this->getContainer();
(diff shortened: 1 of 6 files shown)


src/Extension/AssetTrait.php
```diff
@@ -180,16 +180,16 @@ private function normalizeAsset(FileAssetInterface $asset)
             return;
         }
 
-        $file = $this->getWebDirectory()->getFile($asset->getPath());
+        $app = $this->getContainer();
+
+        $file = $app['filesystem']->getFile("extensions_assets://{$this->getId()}/$path");
```
GawainLynch: This breaks extensions installed via Composer (CLI or API), as Composer will install into `{path}/{composer_vendor}/{composer_name}/`, so the problem is, that `getId()` will return `{path}/{BoltVendorFromNS}/{BoltNameFromClass}/` (i.e. CamelCase), which on file systems that are case-sensitive, it won't be found and is skipped.

CarsonF: Yeah I found that as I was finishing up yesterday. Extension assets are copied to this directory by their composer id, right?

GawainLynch: > Extension assets are copied to this directory by their composer id, right?

Correct, as they should be.

CarsonF: Ok. I'll see if I can figure out a way to pass this to the extension.

GawainLynch: Thinking about it a bit more, maybe that (lowercase) should be the standard. In that way there is no difference between path patterns for bundles and managed extensions. Make sense?

CarsonF: It does make sense. It would be a bc break though. Also it is something each extension would have to define manually, not that that is a problem.

GawainLynch: Well, bundles are a user driven thing … the rest are Composer installs and IMHO should follow that naming pattern in this case.

GawainLynch:
Nudge @CarsonF … deadline on 3.3.0 is Sunday :wink:

bobdenotter:
Let's bump this one to 3.4. I'd really like to have this in, but we should move forward on 3.3 now. 

@CarsonF Hope you don't mind. Let's make sure we get this one for 3.4! :-)

GawainLynch:
Bumping to 3.5, 'cause @CarsonF and time availability

PR #6762, opened 2017-06-19
[3.3] Allow 'extensions_config' as a prefix for FilePermissions checks

GawainLynch:
"Edit configuration" links won't work on installed extensions in 3.3-HEAD right now

Fixes: #6763


GawainLynch:
Finally a 5 minute fix, that took less than 5 minutes this month! :rofl:

PR #6771, opened 2017-06-21
Update config.yml, adding session settings.

bobdenotter:
As pointed out by @CarsonF : We have a bunch of options available for tweaking the session handlers. Very convenient when trying to fix things like this: 

3qezsratcqrg

Should we add some more details here? And/or add an section in the Docs for this?

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -380,6 +380,14 @@ cookies_lifetime: 1209600
 # to '.example.org'.
 cookies_domain:
 
+# By default Bolt stores the sessions according to what is configured in your
+# PHP settings. If this doesn't work correctly, you can override the settings
+# below. For a list of the available settings, see here:
+# https://github.com/bolt/bolt/blob/release/3.3/src/Provider/SessionServiceProvider.php#L152-L178
+# sessions:
+#     save_handler: filesystem
+#     save_path: cache://.sessions
+
 # The hash_strength determines the amount of iterations for encrypting
 # passwords.
 #


CarsonF:
We also have a way to ignore ini settings, but that's only available as a DI parameter, `session.options.import_from_ini`. We should definitely have a config value for that. Any ideas what name would make the most sense? `merge_php`? `import_php`? `import_php_ini`? `import_from_ini`?

GawainLynch:
As above, (plus comments removed from brevity), this is what I personally think a default config.yml file should look like:

```yaml
database:
    driver: sqlite
    databasename: bolt

sitename: A sample site

theme: base-2016

# For more information on available options see https://docs.bolt.cm
```

CarsonF:
I would take that a step further and say the database section could be removed too.

bobdenotter:
> I would take that a step further and say the database section could be removed too.

Doesn't every single site have database settings?

bobdenotter:
For the record: I very, very much like that the config.yml has a ton of default options.. For example I like scrolling down a bit to tweak the wysywig settings, instead of having to search for it in the docs, copy paste over the code, and only then to be able to change it.

CarsonF:
The problem is keeping that file in sync over updates though.

GawainLynch:
For me, it is the exact opposite, and one of the reason I refuse to use the web UI to edit config … I can't find things any longer, and I only have **one** setting I regularly change, `debug: [true|false]`

But more specifically, I have seen (you've shown them to me, and I own one myself) sites that started life as v1, now running v3, with unneeded/outdated and otherwise default config file parameters that were there from v1. It locks a file's settings down.

bobdenotter:
For the "install section" I've made a small questionnaire with neutral questions, that I put out to a dozen people. The response from that was very interesting, even beyond the actual questions posed. 

Before we make a decision on this, I suggest we do the same thing. It's obvious that we use it differently, so let's get more data on what people want/need/use.

GawainLynch:
Oh, for point of reference … I wasn't saying let's eviscerate `confi.yml` first, just stating what we _should_ be aiming for (and yes things like installers are going to make a huge difference here).

My single objection here is just "target" for this being documented. I 100% agree that it should be. However, changing session handling is not the sort of thing the average person needs to, or should, have to change configuration of, it is too easy to mess things up if you don't know what you're doing. As such, sessions should have a section in the advanced documentation.

> Before we make a decision on this, I suggest we do the same thing

Sure, and adding questions like "Do you know what these relate to?", for things like:

- What does this do
- Why would you use it?
```
debuglog:
```

- Can you explain what a digest hash is?
- Can you explain the implications to all the user's of a site when of changing this?
```
hash_strength:
```

- Do you know what and X-Frame is?
- Do you understand the reason, and security implications, of changing X-Frame handling?
```
headers:
    x_frame_options:
```

- Are you comfortable generating and installing a fresh Composer project on the command line?
- Do you know the how Composer users classmaps?
```
extensions:
    composer:
        minimum-stability:
        prefer-stable:
        prefer-dist:
        prefer-source:
        config:
            optimize-autoloader:
            classmap-authoritative:
```

- How often do you use this setting?
- Do you know what a "trusted proxy" is?
- Do you understanding how to use CIDR notation
- Do you understand IPv4 subnets, routing and gateways?
```
trustProxies:
    - 127.0.0.1
    - 10.0.0.0/8
```

- How often do you use this setting?
```
httpProxy:
    host: my.proxy.server
    user: usr
    password: pwd
```

---

Well, actually to be honest, this one needs to be just removed, and the feature deprecated. it is all doable with filesystem, and this is just burden.
```
theme_path: theme
```


… and wouldn't this and the branding section in a helpful, good looking, updatable, page in the docs on how to customise your site?
```
backend:
    news:
        disable:
    stack:
        disable:
```

GawainLynch:
> I would take that a step further and say the database section could be removed too.

Agreed. I actually meant to fill that in with a fake MySQL connection, as that's what the average user is connecting to and what it would look like after install.

GawainLynch:
> For the record: I very, very much like that the config.yml has a ton of default options

Sure, but we're jamming settings (that change) with documentation (that also changes, regularly!) into a file that is going into a site's git — potentially static for the life of the site — that is just huge and hard to follow in almost every single case that I have seen 

… only two exceptions to that I have ever seen are by people in this conversation already. :wink: 

One of the project's main philosophies is "KI.S.S.", and and overload of information in your face is well known to scare people away.

A few, vital, or added by the installer, settings that make the site work, plus a nice blurb about where to find more information, on a site that's easy to navigate, context search, is regularly updated and looks more presentable to the average user the YAML comments in `monospace` has got to be preferable to the majority of our users.

I mean, settings like `debug:` are hard to access, and rather than being right in your face, it is buried under a tonne of other stuff, and usually forgotten until someone's profiler data on disc gets so big it crashes the entire server.

Seriously, after the years of effort it's taken to pull apart the utter mess that is/was `Storage.php`, `Content.php` & `Backend.php` **alone**, collectively we should be well aware of why it is bad to try and jam everything into one place :disappointed:

GawainLynch:
Another thought that popped into my head to help the groups you're advocating for, @bobdenotter, might be a _simple_ `README.md` installed in the project's `app/config/` that (again, simply) explains what & why files/directories are in `app/config` and where to get detailed information on the **use** and settings of each.

Think of it this way … The closer we get to compact, `# comment` free YAML config files, the closer we get to it being _very_ possible build in features to edit settings in the back-end via a web form UI (with a full set of options available, driven by the config definitions in the loader branch), for novices or _very_ quick edits by a site admin in a hurry. 

A longer term goal sure, but how awesome would that be? :smile:

GawainLynch:
To add context to above, this is an edited version of the (_heavily work in progress_) Nut command in the config loader branch to output all a section's options and their default settings … for the **running** version of Bolt that Nut is invoked on.

… additionally, in the class file where the configuration setting is defined, you can add "info" text that will be added the output below, as a `# comment` that gives a brief explanation of the setting … from PHP objects see what I mean about the ease of form building if you have data given to you by core's base definition of it's own configuration parameters? :wink:

```shell
$ ./app/nut config:dump-reference general
```

```yaml
# ~~~ GENERATED ON A DEVELOPMENT VERSION OF BOLT ~~~
#
# Reference configuration for: config://config.yml config://config_local.yml
#
# Auto-generated on Bolt v3.5.0 alpha 5
#
general:
    sitename:             ~
    payoff:               ~
    theme:                ~
    locale:
        name:                 ~
    timezone:             ~
    maintenance_mode:     false
    maintenance_template: ''
    cron_hour:            ~
    cookies_use_remoteaddr: true
    cookies_use_browseragent: false
    cookies_use_httphost: true
    cookies_lifetime:     1209600
    cookies_domain:       null
    hash_strength:        10
    debug:                true
    developer_notices:    false
    debug_show_loggedoff: false
    debug_permission_audit_mode: false
    debug_error_level:    8181
    debug_error_use_symfony: false
    debug_trace_argument_limit: 4
    production_error_level: null
    strict_variables:     null
    debuglog:
        enabled:              false
        filename:             bolt-debug.log
        level:                DEBUG
    headers:
        x_frame_options:      true
    htmlcleaner:
        allowed_tags:
            # Defaults:
            - div
            - span
            …
        allowed_attributes:
            # Defaults:
            - id
            - class
            - …
    performance:
        http_cache:
            options:              []
        timed_records:
            interval:             3600
            use_cron:             false
    trustProxies:         []
    httpProxy:
        host:                 ~
        user:                 ~
        password:             ~
    upload:
        pattern:              '[^A-Za-z0-9\.]+'
        replacement:          '-'
        lowercase:            true
    accept_file_types:
        # Defaults:
        - twig
        - html
        - js
        - css
        - …
    canonical:            null
    favicon:              ~
    homepage:             ~
    homepage_template:    ~
    notfound:             ~
    recordsperpage:       10
    recordsperdashboardwidget: 5
    record_template:      record.twig
    listing_template:     listing.twig
    listing_records:      6
    listing_sort:         'datepublish DESC'
    taxonomy_sort:        DESC
    search_results_template: search.twig
    search_results_records: 10
    theme_path:           theme
    add_jquery:           false
    changelog:
        enabled:              false
    liveeditor:           false
    enforce_ssl:          false
    backend:
        news:
            disable:              false
        stack:
            disable:              false
    branding:
        name:                 Bolt
        path:                 /bolt
        provided_by:          []
        provided_link:        ''
        news_source:          'http://news.bolt.cm/'
        news_variable:        ~
    caching:
        config:               true
        database:             false
        templates:            true
        request:              false
        duration:             10
        authenticated:        false
        thumbnails:           true
        translations:         true
    database:
        dbname:               null # Required
        prefix:               bolt_
        host:                 localhost
        port:                 null
        user:                 root
        password:             null
        charset:              utf8
        collate:              utf8_unicode_ci
        …
    extensions:
        site:                 'https://extensions.bolt.cm/'
        enabled:              true
        composer:
            minimum-stability:    stable
            prefer-stable:        true
            prefer-dist:          true
            prefer-source:        false
            optimize-autoloader:  false
            classmap-authoritative: false
    mailoptions:
        transport:            mail
        spool:                true
        host:                 localhost
        port:                 25
        username:             ''
        password:             ''
        encryption:           null
        auth_mode:            null
        senderMail:           null
        senderName:           null
    thumbnails:
        default_thumbnail:
            # Defaults:
            - 160
            - 120
        default_image:
            # Defaults:
            - 1000
            - 750
        quality:              80
        cropping:             crop
        notfound_image:       view/img/default_notfound.png
        error_image:          view/img/default_error.png
        save_files:           false
        allow_upscale:        false
        exif_orientation:     true
        only_aliases:         false
        browser_cache_time:   null
    wysiwyg:
        images:               false
        anchor:               false
        tables:               false
        fontcolor:            false
        align:                false
        subsuper:             false
        embed:                false
        underline:            false
        ruler:                false
        strike:               false
        styles:               false
        blockquote:           false
        codesnippet:          false
        specialchar:          false
        ck:
            autoParagraph:        true
            disableNativeSpellChecker: true
            filebrowserWindowHeight: 480
            filebrowserWindowWidth: 640
            allowNbsp:            false

```

SahAssar:
:+1: for keeping this in the docs instead. These settings are specifically for overriding your php.ini settings, so they should not get any use on properly set up servers, right?

As for making config.yml more "slim": I'm all for that. I think people currently expect that every possible option is in the default config.yml since it includes so much, and there is a certain level of information overload currently. The comments are also way to verbose, and should IMO be a short one line explanation and a pointer to the docs.

app/config/config.yml.dist
```diff
@@ -380,6 +380,14 @@ cookies_lifetime: 1209600
 # to '.example.org'.
 cookies_domain:
 
+# By default Bolt stores the sessions according to what is configured in your
+# PHP settings. If this doesn't work correctly, you can override the settings
+# below. For a list of the available settings, see here:
+# https://github.com/bolt/bolt/blob/release/3.3/src/Provider/SessionServiceProvider.php#L152-L178
```
GawainLynch: That link also isn't static, as it points to `release/3.3` HEAD, and it is not unreasonable to expect it to change during the life time of `release/3.3`. It is already heavily rewritten in  both `release/3.4` & `master`. So someone installing a site on 3.3 now, and then a year from now working on 3.22.3, or 4.0.1 is going to be pointed there.

Another reason that this better belongs in the docs, in a more complete, and explained fashion. Not to mention, further demonstrates that there is too much already jammed in here to navigate … for all 3 groups.

GawainLynch:
Back to the original point of this PR, I've PR'ed docs with https://github.com/bolt/docs/pull/758

["There can be only one"](https://www.youtube.com/watch?v=sqcLjcSloXs) :rofl:

bobdenotter:
Obsoleted by https://github.com/bolt/docs/pull/758

PR #6773, opened 2017-06-23
[3.4] Requirement & recommended setting checks on first user

GawainLynch:
This PR adds the results of required & recommended checks from `bolt/requirements` to first user screen.

If everything is good:
first user good

An example of a recommendation failure, i.e. not super critical, but well advised:
first user problem

An example of a required setting failure:
first user failure

---

We obviously need to tweak the heck out of `bolt/requirements` … but that is also dependency of the Nut `setup:run` command, so linked requirements … just puts the motivation in front of a few more eyes :wink: 

Also, I am re-enforcing my disclaimer about Gawain + Visuals … I suck at them … so if we like the "feature", send it in and the UI team can do their magic to make the interface for this kick butt :tada:

### DIFF
--- a/app/view/twig/firstuser/firstuser.twig
+++ b/app/view/twig/firstuser/firstuser.twig
@@ -10,6 +10,42 @@
 
 {% block page_main %}
 
+    <div class="row">
+        {% if context.required %}
+            <div class="panel panel-danger">
+                <div class="panel-heading"><i class="fa fa-close"> </i> Failed System and/or PHP requirements:</div>
+                <div class="panel-body bg-danger">
+                    {% for failed in context.required %}
+                        <p>{{ loop.index }}. {{ failed.testMessage }}</p>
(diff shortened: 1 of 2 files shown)


bobdenotter:
I like the looks of this in general. It'll need a tiny bit of CSS touchups, but that's trivial (like no dropshadow under the notices)

I thought if `app/cache/` wasn't writable, you couldn't even get to the login screen, right?

GawainLynch:
> I thought if `app/cache/` wasn't writable, you couldn't even get to the login screen, right?

You are correct, currently it will fatal. But we should be able to get though that in coming months. It was just low hanging pineapples for quickly making the screen shot, and literally _where_ my IDE opened the source file (last thing I had fixed was the cache paths, and correctly detecting v3 & v4 layouts).

bobdenotter:
Cool, let's merge this in. I suppose there will be some other CSS touch-ups, that I can address in a separate PR.

PR #6775, opened 2017-06-27
[3.4] Refresh config checks page

GawainLynch:
We've had this route, and the aysnc controller that isn't even used, since 3.0-dev, so this PR basically just adds it to the menu, and updates the admin route & templates to use `bolt/requirements`

Eye candy (_normal UI team disclaimer_):

checks-menu

checks-mixed

checks-all

**Edit:** The screenshot for #3602 (original PR) is, with hindsight, pretty funny; "Bolt 2.3.0 alpha 1" … 2 years ago

### DIFF
--- a/app/resources/translations/en_GB/messages.en_GB.yml
+++ b/app/resources/translations/en_GB/messages.en_GB.yml
@@ -481,6 +481,7 @@ logs.system-log: "System Log"
 logs.system-log.cleared: "The system log has been cleared."
 logs.system-log.trimmed: "The system log has been trimmed."
 
+menu.configuration.checks: "Set-up checks"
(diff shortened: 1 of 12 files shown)


GawainLynch:
> Looks good. I think this should be integrated into our :crystal_ball: -questions, for when people have support questions, but didn't provide enough information.

Cool, yeah good idea actually. 

Also, this gives us one more trigger to get someone to review the requirement/recommended list for `bolt/requirements` … I know some are :+1: and some will need :fire:

bobdenotter:
> I know some are 👍 and some will need 🔥

Sure, but that's iterative. ;-)

PR #6787, opened 2017-07-03
Config update

lenvanessen:
Switched out Sendgrid for Sparkpost, because Sendgrid changed their pricing model and now only offer free trails, and Sparkpost is still free up to 100.000 mails

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -331,8 +331,8 @@ liveeditor: false
 # it, your mail might disappear into a black hole, without producing any errors.
 # Generally speaking, using 'smtp' is the better option, so use that if possible.
 #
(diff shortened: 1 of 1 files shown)


SahAssar:
hehe, seems like we will have to do this once a year from now on: #4911

SahAssar:
Hmm, one thing though, could you change your base to the lenvanessen-patch-1 branch? That way we won't get that merge commit :)

GawainLynch:
> Hmm, one thing though, could you change your base to the lenvanessen-patch-1 branch? That way we won't get that merge commit 

Hold my :beer:, I got this.

SahAssar:
@GawainLynch ❤️

PR #6781, opened 2017-06-29
[4.0] Remove fields()

GawainLynch:
Per #6788

PR #6788, opened 2017-07-04
[3.3] Deprecate {{ fields() }} and move to a block call

GawainLynch:
Let's try again shall we :wink: 

This PR:
  - Fully deprecates the Twig function `{{ fields() }}`
  - Cut over to full use of blocks in `base-2016`
  - Derpy derp on copy/pasta

Matching v4 PR is #6781


GawainLynch:
> There are some minor nitpicks I have, but I will put them in a separate PR.

Cool, yeah I figured you might … faster that way too … thanks :+1:

PR #6799, opened 2017-07-07
[BUG/CR] 3.2 - Quick fix for content with large lists of incoming relations

ToBe998:
The current Bolt 3.2 Edit action loads each incoming related record in it's own DB query. If you have more than a couple hundred objects, the edit page will no longer load as you get memory and execution time problems.

After some discussions with @GawainLynch, it seems a "proper" solution using the new storage layers is already available but might not be portable as fast as needed. This quick fix is intended as a temporary solution until the loading is refactored to the new mechanism. (Also keep in mind that rendering 10000 related objects in twig might still not be advisable ;) )

### DIFF
--- a/src/Helpers/Input.php
+++ b/src/Helpers/Input.php
@@ -12,7 +12,7 @@ class Input
      * @param bool  $stripslashes
      * @param bool  $stripControlChars
      *
-     * @return string
+     * @return string|array
      */
     public static function cleanPostedData($var, $stripslashes = true, $stripControlChars = false)
     {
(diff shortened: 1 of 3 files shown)


GawainLynch:
As I mentioned to @ToBe998 in our discussion about this bug, I am _personally_ happy to see this go in … 

Also, neither of us could think of a better config key name … O.C.D. Style Help much appreciated if anyone has ideas :smiley_cat:

ToBe998:
Just noticed, the `{{ content.related("othercontent") }}` twig functions also fetches records one per query. Probably not a problem for most cases, but worth keeping in mind when the real refactoring starts.

bobdenotter:
Looks good to me, too! 👍

PR #6803, opened 2017-07-09
Always reset null templatefields to an empty array instead

### LINKED ISSUE #6790: [BUG] – Repeaters in Templatefields not allowing you to remove last set
Details
-------

 - Relevant Bolt Version: [ 3.2.10 ]
 - PHP version: [ 7.0 ]
 - Used web server: [ Apache ]

Reproduction
------------
 1. **Bug summary**: 
    * Upon trying to delete all sets of a repeater, the last one does not get removed upon save.

 2. **Specifics**:
    * Bolt version 3.2.10, using Chrome
    * Patch has been applied from issue #6518, using this commit: https://github.com/bolt/bolt/pull/6518/commits/12a65d599b56302f32e827673f18f30ba3421ecb

 3. **Steps to reproduce**:

Code for the repeater itself:
```
my_repeater:
            label: My repeater
            type: repeater
            fields:
                name:
                    label: Name
                    type: text
                libk:
                    label: Link
                    error: Please enter a valid URL
                    pattern: url
                    type: text
                logo:
                    label: Logo
                    type: image
```

- Add two sets of content into the repeater, then save
- Remove one of the sets and save, this should work as expected leaving one remaining
- Remove the remaining set, save, then refresh the page
- The last remaining set is not getting removed from the repeater

Have tried re-ordering and removing multiple sets at once, with the same result. If multiple items get removed at once (both items one and two are removed at the same time therefore emptying the repeater), upon refresh they will _both_ still be in the repeater.

 4. **Expected result**: 
    * When all items have been removed, no matter whether I delete one or many, this should save in the database and leave an empty repeater.

 5. **Actual result**: 
    * Repeater seems as if it has saved, but upon refresh the final set that was deleted to leave the repeater empty has not been removed.

rossriley:
Fixes #6790

Because we don't send null values to the database to save, we end up not persisting templatefields if all the values are cleared.

This ensures that a null templatefields is converted to an empty array instead and then will get saved to the db.

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -363,6 +363,9 @@ class Content extends Entity
 
     public function setTemplatefields($value)
     {
+        if ($value === null) {
+            $value = [];
+        }
         $this->templatefields = $value;
     }
 


stueybrock:
`$value` is never null - it's an instance of Bolt\Storage\Entity\TemplateFields or an array of values. The empty value is an array:

```php
[
	0 => ''
]
```
This produces an output of:

```json
{
    "funders_section_2" : "[\"\"]"
}
```
If you then refresh the page, you get the following error:

image

However if you change the value in the DB to:
```json
{
    "funders_section_2" : "[{\"\"}]"
}
```
This works straight away. I presume that repeaters are looking for an subarray which the first instance doesn't produce?

PR #6810, opened 2017-07-11
Adding `theme: skeleton` to core.

bobdenotter:
Initial commit of our new `skeleton` theme to core, as an alternative to the `base-2016` theme.

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -13,11 +13,12 @@ database:
 sitename: A sample site
 payoff: The amazing payoff goes here
 
-# The theme to use. Be sure to copy the {theme_path}/base-2016 folder, and
-# change the name here accordingly.
+# Select a theme to use. Bolt ships with two themes: `base-2016` and `skeleton`.
+# Be sure to copy the `{theme_path}/base-2016` or `{theme_path}/skeleton`
+# folder, and change the name here accordingly, before making modifications.
 #
-# Don't edit the templates in theme/base-2016, because they _will_ get updated
-# in next releases
+# Don't edit the templates in theme/base-2016 or theme/skeleton, because they
+# _will_ get updated in next releases.
 theme: base-2016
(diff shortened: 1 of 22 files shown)


app/config/config.yml.dist
```diff
@@ -13,11 +13,12 @@ database:
...
-# The theme to use. Be sure to copy the {theme_path}/base-2016 folder, and
-# change the name here accordingly.
+# Select a theme to use. Bolt ships with two themes: `base-2016` and `skeleton`.
+# Be sure to copy the `{theme_path}/base-2016` or `{theme_path}/skeleton`
+# folder, and change the name here accordingly, before making modifications.
 #
-# Don't edit the templates in theme/base-2016, because they _will_ get updated
-# in next releases
+# Don't edit the templates in theme/base-2016 or theme/skeleton, because they
+# _will_ get updated in next releases.
```
GawainLynch: Isn't this assertion now anachronistic with both our Composer and distribution installs? i.e. shouldn't we drop that as if it is overridden, then you are using the git install and are self-managing the install as you are an expert user.

theme/skeleton/partials/_sub_field_blocks.twig
```diff
@@ -0,0 +1,85 @@
+{# If 'allowtwig' is true for this field, we'll need to parse it as
```
GawainLynch: So you'll never guess where the blocker I am about to PR is … and now this will perma-pile that, or any other, problems in 3 identically repeating places, `skeleton`, `base-2016`, and `theme_defaults` … the last two are currently out of sync heavily between each other, and between branches, that being the delay in getting my PR in yesterday.

When you have a moment today, can you ping me and lets put heads together and strategise, as I think we're opening up a world of pain here.

bobdenotter:
Made obsolete by https://github.com/bolt/themes

PR #6811, opened 2017-07-12
[3.2+] More refactoring of fields blocks

GawainLynch:
## One good :bug: deserves another

So while tracking down a small bug in 3.3, I found several more all stemming from this area. I double checked and they were present from 3.2 onward, so I thought I'd back-port the "two-liner" to 3.2 … and all hell broke lose, wasting several hours of work nobody really has time to waste. :disappointed: 

The differences between 3.2 & 3.3 were substantial enough to make this a long and painful process, exacerbated by the fact that we have:

  - `theme/base-2016/partials/_sub_field_blocks.twig` supposed to be identical to `app/theme_defaults/_sub_fields.twig`
  - `theme/base-2016/partials/_sub_field_blocks.twig` supposed to be identical to `app/theme_defaults/_sub_field_blocks.twig`

… and #6810 inbound adding a third copy of each, to make this harder to track and keep up-to-date.

Which leave us with the following status:

| Principle  |    |
| ---------- | -- |
| D.R.Y.     | :x:
| S.O.L.I.D. | :x:
| W.E.T.     | :white_check_mark:

A check-mark on W.E.T. is not something we should aim for. It also makes changes to one branch too easily break newer release branches on merge cascade.

## PR Changes

- Adds an `exclude` array variable to the use of the `with` control statement 
- Separate out Markdown logic into own mini sub-block
- Re-sync `theme` & `theme_defaults`
- Remove vestige reference to `fields()` in comment blocks

**Important note:** One of the reasons this PR diff is so large is that :koala:s can't indent :rofl: … Nah, totally my fault.

PR #6812, opened 2017-07-12
[3.3] More Twig deprecations

GawainLynch:
So, funny story, right … I was working on a "ready-to-go" branch for 4.0 when I found a bug I though came from 3.3, so I ran up this and grabbed the deprecations that go with it, and during the monkey-testing of this was when I found more bug(s) that lead to #6811

… a day and a bit later. :tada: 

This should go in _after_ we agree on the approach in #6811, but _before_ I need to do the branch merge, as the differences are substantial and the merge conflicts are just ugly. The life of being a merge :princess: 

When I merge these all down, then I can send in the PR for 4.0.

### DIFF
--- a/theme/base-2016/extrafields.twig
+++ b/theme/base-2016/extrafields.twig
@@ -4,10 +4,9 @@
 
     <h1>{{ record.title }}</h1>
 
-    {# Output all fields, in the order as defined in the contenttype.
-       To change the generated html and configure the options, see:
-       https://docs.bolt.cm/templating #}
+    {% with { 'record': record, 'common': false, 'repeaters': true } %}
     {{ block('sub_fields', 'partials/_sub_fields.twig') }}
+    {% endwith %}
 
     {# Uncomment this if you wish to dump the entire record to the client, for debugging purposes.
     {{ dump(record) }}
@@ -16,4 +15,3 @@
     {% include 'partials/_recordfooter.twig' with { 'record': record, 'extended': true } %}
 
 {% endblock main %}
-
--- a/theme/base-2016/page.twig
+++ b/theme/base-2016/page.twig
@@ -6,7 +6,7 @@
 
     {{ record.teaser }}
 
-    {% if record.image!="" %}
+    {% if record.image != "" %}
         <a href="{{ image(record.image) }}">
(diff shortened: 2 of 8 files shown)


theme/base-2016/extrafields.twig
```diff
@@ -4,10 +4,10 @@
 
     <h1>{{ record.title }}</h1>
 
-    {# Output all fields, in the order as defined in the contenttype.
-       To change the generated html and configure the options, see:
-       https://docs.bolt.cm/templating #}
+    {# Using the 'with' tag with the 'only' keyword, we limit the variable scope inside the block #}
+    {% with { 'record': record, 'common': false, 'repeaters': true } only %}
```
bobdenotter: So, either we pass in `global`, or we drop the `only`, because we need to get the `_route` in the block.

theme/base-2016/page.twig
```diff
@@ -6,19 +6,21 @@
...
     {% endif %}
 
     {{ record.body }}
 
-    {# If there are repeaters or templatefields, we should output them. #}
-    {% with { 'common': false, 'repeaters': true } %}
-        {{ block('sub_fields', 'partials/_sub_fields.twig') }}
+    {# Using the 'with' tag with the 'only' keyword, we limit the variable scope inside the block #}
+    {# If there are repeaters, we should output them. #}
+    {% with { 'record': record, 'common': false, 'repeaters': true, 'exclude': ['teaser', 'body', 'image'] } only %}
```
bobdenotter: Ditto

theme/base-2016/extrafields.twig
GawainLynch: Yeah, totally not fussed either way at this stage … Obviously passing in the global (or the service) is more screen clutter and confusion for the average Twig user, so whatever you think fits that group's needs the best.

bobdenotter: Let's lose the `only`. In record.twig we do: `{{ block('sub_fields', 'partials/_sub_fields.twig') }}` which _also_ passes in global, anyways.

GawainLynch: d---ropped :smile:

bobdenotter:
Also 👍

PR #6817, opened 2017-07-14
Performance in tags load

iskyd:
If you have a lot of tags, like 15/20k the fieldTags takes a lot cause of select2 load. It could take something as 60 seconds and that is not good.


GawainLynch:
Nice work mate, thank you! :+1:

SahAssar:
Looking at the code it seems like this won't work with tags that contain spaces, right? Since the part after the space would be considered a new attribute (in html a attribute that is not quoted may not contain spaces).

GawainLynch:
Thanks @SahAssar … you are correct and it made it past a couple of sets of :eyes: :+1:

PR #6830, opened 2017-07-18
Preparing 3.3.0 RC 1

GawainLynch:
Assets were rebuilt post-branch merge this morning. So this is the CHANGELOG & version bump.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,296 +1,241 @@
 Changelog for Bolt 3.x
 ======================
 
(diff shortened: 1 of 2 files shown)


GawainLynch:
> For a second I couldn't figure out why the delta on this was so large, but then I noticed you combined the changes for the betas

Yeah, I was proactively trying to save you some work. I know you usually do it on actual stable, but went for the "we can tweak once next week" option :smile:

Issue #6832, opened 2017-07-18
[BUG] Sortable select field type causes twig runtime error when the values are from a content type


smgdesign:
When creating a field of type `select` and setting `sortable` to `true` on a field where the values are from another content type it causes an undefined string offset error in the twig rendering.
The actual error is:
`An exception has been thrown during the rendering of a template ("Notice: Uninitialized string offset: 0") in "@bolt/editcontent/fields/_select.twig" at line 48.`

This is caused in the method to create a unique superset of items from the `leadingvalues` and `values` variables. Commenting this line out `{% set values = unique(leadingvalues, values) %}` on line 48 of `vendor/bolt/bolt/app/view/twig/editcontent/fields/_select.twig` allows for the page to display. Setting the `values` variable just `leadingvalues` allows the page to display but obviously only returns the selected items.

The unique function seems to be unable to handle key based arrays and causes this runtime error, I haven't been able to find where the unique function is defined yet but will keep looking and provide a suitable update to the method when I find it (maybe just detecting if the key is unique on non-indexed arrays could solve this). A temporary fix of commenting out line 48 in the _select file obviously isn't ideal when deploying to multiple environments but should at least get anyone else around this problem for now

Details
-------

 - Relevant Bolt Version: >3.2.11 and 3.3
 - Install type: Composer install
 - PHP version: 7.0
 - Used web server: Nginx version 1.10.2

GawainLynch:
Great report … Thank you! :+1:

GawainLynch:
Fixed in #6841 and tagged in 3.2.16 & 3.3.0-RC2

PR #6834, opened 2017-07-19
Sortable select fields using content as the values

smgdesign:
This is a more comprehensive fix to complement the PR for issue #6638.
Twig filter for associative array merging `array|assocmerge(array)`

Twig function to check if an array is a sequential indexed array or an associative array `isseq(array)`

`_select.twig` now detects if a list of sortable values is indexed or associative and merges selected values with the full list accordingly

Fixes: #6832
Related to: #6638


bobdenotter:
As discussed on Slack, alternative fix here: #6835

Issue #6561, opened 2017-04-10
Custom sorting doesn't work for type: select


jarek-gilewski:
I have block of elements with contacts and I would like to show them in specific order on multiple pages in page footer - different contacts based on different pages. Contacts are visible on page and also in admin panel but custom sorting works only once - I mean when I change order of contacts and save all is fine. When I refresh admin panel and save (even different field) contacts are sorted in alphabetical order.

```
selectentry:
            label: Contact
            type: select
            values: contact/title
            group: Contacts
            autocomplete: true
            multiple: true
            sortable: true
```

Details
-------

 - Relevant Bolt Version: [3.2.9]
 - Install type: [Zip/tarball install]
 - PHP version: [5.6.16]
 - Used web server: [ Apache ] version [ Apache/2 ]
 - For UX/UI issues: [ all browsers ]

bobdenotter:
For reference, a complete contenttype to reproduce this is: 

```
bars:
    name: Bars
    singular_name: Bar
    fields:
        title:
            type: text
            class: large
        slug:
            type: slug
            uses: [ title ]
        selectentry:
            type: select
            values: entries/id,title
            postfix: "Select an entry"
            autocomplete: true
            sortable: true
            multiple: true
```

This is related to #6355

bobdenotter:
Fixed for Bolt 3.2.x in #6610. We're looking into Bolt 3.3.x.

bobdenotter:
Did some testing, currently broken for 3.3.. @rossriley, could you look into this? 

![http://g.recordit.co/v4gpRiRWP3.gif](http://g.recordit.co/v4gpRiRWP3.gif)

bobdenotter:
Fixed in #6610 and #6638

PR #6835, opened 2017-07-19
[WIP] Select fields should keep all items, even when using a "string" as value.

bobdenotter:
Currently, `type: select` will not work correctly, if you use `values: entries/title`, instead of a compound value like `values: entries/id,title`. This will cause items that start with the same letter to "vanish".

We fix this by making sure the intermediate array of values always consists of an array, keeping the keys unique. 

Works correctly for these cases: 

```
        selectentry:
            type: select
            values: entries/id,title
            postfix: "Select an entry"
            autocomplete: true
            sort: title
            multiple: true
            sortable: true
        selectentry2:
            type: select
            values: entries/title
            autocomplete: true
            sort: title
            multiple: true
            sortable: true
        selectentry:
            type: select
            values: entries/id,title
            autocomplete: true
            sort: title
            multiple: true
            sortable: true
        selectentry3:
            type: select
            values: entries/title
            autocomplete: false
        selectentry4:
            type: select
            values: entries/id,title
            autocomplete: false
```

Screenshot: 

screen shot 2017-07-19 at 14 05 42

Marked as `[WIP]`, because this area of the code has been changed often, and is very brittle. Seemingly "simple" fixes in this part have so far always caused side effects and introduced new bugs.

### DIFF
--- a/app/view/twig/_buic/_select.twig
+++ b/app/view/twig/_buic/_select.twig
@@ -84,7 +84,7 @@
                             {% endif %}
 
                             {# option #}
-                            {% set opttext = option.text|default(option.value)|striptags %}
+                            {% set opttext = option.text|default(option.value)|join()|striptags %}
                             {# for details on why 'disabled' is in here, see https://github.com/bolt/bolt/pull/6590 #}
                             {% set optattr = {
                                 'value!':   option.value,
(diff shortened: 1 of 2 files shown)


bobdenotter:
To illustrate how deep the rabbit hole goes, see: #6834, #6638, #6834, #6561, #6355, #6610, #6332, #6291

GawainLynch:
… and https://travis-ci.org/bolt/bolt/jobs/255232635

GawainLynch:
the rabbit hole

Issue #6836, opened 2017-07-19
[RFC] Better Exceptions for non-excisting Bundled extension classes


lenvanessen:
Hi,

When using version 3.3. of Bolt, I ran into an issue autoload issue. The class I created for my bundled extension was not included, and did not exist.

After adding the class to bolt.yml to activate it, Bolt threw an exception.

`LogicException in Bootstrap.php line 163
Bolt\Extension\XXX needs to implement Bolt\Extension\ExtensionInterface`

This got me searching for an issue that wasn't really there. I was extending SimpleExtension, so I shouldn't have to implement the Interface (if i understood the docs correctly)

After a searching for the issue in the wrong places, I realized that there was something wrong in the composer.json and that my class wasn't really included. 

## Proposed solution
Check if the class exists and throw a more descriptive exception if it doesn't.

GawainLynch:
Fixed in #6839

PR #6839, opened 2017-07-20
More descriptive exception when the bundled-extension class does not exist 

lenvanessen:
A class thatt doesn't exist can't implement an interface, so this should be checked first.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -169,12 +169,12 @@ class Bootstrap
                                 throw new LogicException(sprintf('Extension class name "%s" is defined in .bolt.yml or .bolt.php, but the class name is misspelled or not loadable by Composer.', $extensionClass));
                             }
                             if (!is_a($extensionClass, ExtensionInterface::class, true)) {
-                                throw new LogicException("$extensionClass needs to implement " . ExtensionInterface::class);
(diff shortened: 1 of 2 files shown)


lenvanessen:
fixes #6836

src/Bootstrap.php
```diff
@@ -158,6 +158,9 @@ public static function run($rootPath)
                 'extensions',
                 function ($extensions) use ($config) {
                     foreach ((array) $config['extensions'] as $extensionClass) {
+                        if(! class_exists($extensionClass) ) {
+                            throw new LogicException("$extensionClass is defined in bolt.yml, but class is misspelled or is not loaded by composer");
+                        }
```
GawainLynch: This will throw exceptions for instantiated classed.

Bundles are expert level, from the beginning by design, this is out of scope

GawainLynch:
Reopening, as now I have more time to reply (plus what I mentioned in Slack), lets get this over the line. My reply was too terse :laughing: 

I can do some magic for you to get this where it needs to be, or grab me again in Slack and lets go over it with more detail :+1:

GawainLynch:
As mentioned in Slack, I've rebased the branch to pull in the tests from #6844 and tweaked the location of the check, plus added a test for it.

Catch you when you're back :+1:

PR #6845, opened 2017-07-22
[3.2] Fix for TemplateField select options

GawainLynch:
Can someone on the team test & review urgently please … ~~This needs to be agreed on and merged in the next hour or so.~~

Follow up to #6841

I'd like to dedicated the work on this to Chester Bennington … For a lot of us, you got us though some hard times. R.I.P.


bobdenotter:
Monkey-testing this, it seems to fix the problem. They show up in the backend, persist, and get retrieved as expected. They are also retrievable in the frontend.

PR #6849, opened 2017-07-24
[3.2] Lock Doctrine version as DBAL 2.6 & Cache 1.7 require PHP 7.1

### LINKED ISSUE #6846: Doctrine 2.6.0 is out, and it gives the middle finger to semver.
Doctrine DBAL 2.6.0 is out, and it was just now pulled in by `composer update`. It looks great, but it supports only PHP 7.1. https://github.com/doctrine/dbal/releases/tag/v2.6.0

Will we need to lock our version at 2.5.x?

GawainLynch:
Closes #6846


GawainLynch:
FWIW, I don't think this is the right approach, and is reactionary:
* If someone is installing via Composer, then Composer will keep their versions set to what is supported 
* If we're building distribution packages on anything greater than PHP 5.5 then we're creating autoloaders that are not supported on 5.5+
* If people have a valid need (like several of us who have been running DBAL 2.6.x-dev in production for a while now) to run DBAL 2.6.0+ then we screw the pooch for them forcing people like me to fork Bolt, and set up a new Packagist project to keep their sites running.

bobdenotter:
> If we're building distribution packages on anything greater than PHP 5.5 then we're creating autoloaders that are not supported on 5.5+

Is that true? I've been packaging the releases using PHP 7.1 for a while, and as far as I know the package can be used just fine on PHP 5.5. 

That said, locking should be a last resort, as long as we can still provide a working solution for the people on PHP < 7.1.

GawainLynch:
Ah, the autoloader generation code seem to have been updated at some point this year(ish) … they were generated using a null coalescing operator for PHP 7.0+ but that doesn't seem to be the case.

Nonetheless, there is nothing stopping other packages doing the same and they should only ever be built on the lowest PHP version we support … I've asserting this for a few years now.

GawainLynch:
Point made, agreement reached … Common sense prevails … Nothing to see here.

PR #6859, opened 2017-07-26
[3.2] Keeping `keys:` intact for select fields. 

### LINKED ISSUE #6858: [BUG] keys are not stored in database for select fields
I'm using a select fieldtype , populating it with values from another contenttype. What I want to use is the "keys" setting in order to store the name instead of the id of this contenttype, as mentioned in documentation :
> If you wish to store another field or value from the original ContentType in your database, use the keys setting. If you do this, it will not store the 'id', but the value of the field you specify. For example:

But it seems that only the ID is stored in the database

Here is my contenttype : 
```
events:
    fields:
        beginning:
            type: date
        end:
            type: date
        location:
            type: select
            values: locations/city
            keys: city
        slug:
            type: slug
            uses: [ beginning, end, location ]
    title_format: [ beginning, end, location ]
```
It seems that neither slug, nor title format are able to use the location city and just use the location ID. When looking in the database, city name is not stored and I'm just able to see city ID.
Moreover, I'm not able to search by city in my events.

Details
-------

 - Relevant Bolt Version: [ 3.3@RC3 ]
 - PHP version: [ 7.0 ]
 - Used web server: [ Apache]

bobdenotter:
Fixes #6858

### DIFF
--- a/src/Form/Resolver/Choice.php
+++ b/src/Form/Resolver/Choice.php
@@ -80,10 +80,11 @@ final class Choice
         $limit = $field['limit'];
         $sortable = $field['sortable'];
         $filter = $field['filter'];
+        $key = isset($field['keys']) ? $field['keys'] : 'id';
 
         // Get the appropriate values
         return is_string($values)
-            ? $this->getEntityValues($values, $limit, $sortable, $filter)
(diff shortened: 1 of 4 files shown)


GawainLynch:
are you the keymaster
keymaster

Issue #6864, opened 2017-07-27
[BUG] Select fields inside repeaters are not displaying results


zomars:
When using a select field that populates from a  content type inside a repeater it shows no results.

Details
-------

 - Relevant Bolt Version: 3.2.17
 - Install type: Composer install
 - PHP version:  7.1
 - Used web server: Apache 2.4 
 - For UX/UI issues: Chrome 59.0.3071.115

Reproduction
------------

```yaml
normalselect:
    type: select
    values: menus/id,title
menus:
    type: repeater
    fields:
        repeaterselect:
            type: select
            values: menus/id,title
```

Result
------------
![untitled](https://user-images.githubusercontent.com/3504472/28696469-fb3af882-72f2-11e7-80bc-c44ae9a54707.gif)

GawainLynch:
Good news, @zomars! This is fixed in #6845 and is included in 3.3.0 and will be in 3.2.18 when tagged shortly (we held off on 3.2.18 as we wanted to get 3.3.0 out, and to give it a chance to see if we missed anything else).

Time travelling :koala: strikes again :wink:

bobdenotter:
@GawainLynch I'm afraid not. 6845 was included in 3.2.17: https://github.com/bolt/bolt/blob/3.2/changelog.md

This might be an additional area that needs a tweak.

GawainLynch:
Ugh… reopening then … and going on the hunt for that :koala: first!

zomars:
I can confirm it works!

bobdenotter:
Closed by #6865

PR #6865, opened 2017-07-28
[3.2] Fix for select fields in repeaters

GawainLynch:
Fixes: #6864
Related: #6841 & #6845 
Happiness: @zomars


zomars:
It's working!
it-working1

PR #6867, opened 2017-07-30
Include Slack channel count badge

GawainLynch:
Tagging as an RFC as I have no great opinion on this one, just having fun and seeing what we can do.

The currently implemented options are:
  *  — Active users
  *  — Active/total ratio 
  *  — Total user count

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ page.
 
 ---
 
-[![Build Status][travis-badge]][travis] [![Scrutinizer Continuous Inspections][codeclimate-badge]][codeclimate] [![SensioLabsInsight][sensio-badge]][sensio-insight]
+[![Build Status][travis-badge]][travis] [![Scrutinizer Continuous Inspections][codeclimate-badge]][codeclimate] [![SensioLabsInsight][sensio-badge]][sensio-insight] [![Slack][slack-badge]](https://slack.bolt.cm)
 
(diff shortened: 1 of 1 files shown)


rossriley:
1-millionth-visitor

GawainLynch:
I like Ross' suggestion better :stuck_out_tongue_winking_eye:

bobdenotter:
I like options 1 and 3. 2 is nice 'n all, but the numbers don't really fit in the lozenge. ;-)

GawainLynch:
Let me know which one, if at all, you want and I'll update or close the PR :wink:

bobdenotter:
Can we make 2 a bit wider so it fits the text? If so, i'd like that one. If not, number 3.

GawainLynch:
Yeah, it is just an SVG in `template/badge.svg.twig` in the Slack invites site repo, but I have no idea what I am doing with images, let alone SVGs. However, it should be a 10 second job for someone who knows what they are doing.

I even tried to use the Slack branding ones, but quickly found out that I was completely out of my depth versus the amount of time I had.

bobdenotter:
Like so? https://github.com/bolt/bolt-slack-invites/pull/2

GawainLynch:
Like that … deployed, so if you refresh this page you'll see the updated versions in the OP :smile: 

Thanks mate :+1:

PR #6872, opened 2017-08-01
[3.4][Tests] Setcontent

GawainLynch:
Two :koala: derps that were my fault while helping out. 

Follow up to me being "helpful" in #6622

### DIFF
--- a/src/Twig/SetcontentTokenParser.php
+++ b/src/Twig/SetcontentTokenParser.php
@@ -18,9 +18,7 @@ class SetcontentTokenParser extends AbstractTokenParser
     private $legacy;
 
     /**
-     * @param bool $legacy
-     *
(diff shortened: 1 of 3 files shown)


src/Twig/SetcontentTokenParser.php
```diff
@@ -18,9 +18,7 @@ class SetcontentTokenParser extends AbstractTokenParser
     private $legacy;
 
     /**
-     * @param bool $legacy
-     *
-     * @deprecated Deprecated since 3.4, to be remove in v4.
+     * @param bool $legacy @deprecated Deprecated since 3.4, to be remove in v4.
```
CarsonF: Tags aren't supposed to be inline unless they are wrapped with `{}`

PR #6876, opened 2017-08-04
Show singular name instead of contenttype slug.

lenvanessen:
'New pages' becomes 'New page', which is grammatically correct


GawainLynch:
Looks good, and the Travis fail on 7.1 is due to me getting you to rebased before changing the branch here so unrelated :-)

If nobody gets to this before me, I'll go over it (monkey test) tomorrow and merge in :+1:

GawainLynch:
@SahAssar if you have a moment, can you sanity check this for me? You have a much clearer picture of translation behaviour, and expected outcomes than I do. Obviously this is correct, but my worry is an negative unexpected change of behaviour that I am not seeing.

I'm over here heating up the :fire: thrower if you need me …

GawainLynch:
@SahAssar agrees with me in Slack that this won't cause a #1555 … sending in!

PR #6878, opened 2017-08-04
Restore support for sorting in Contenttype Selects

### LINKED ISSUE #6877: [BUG] Sorting on ContentType select fields does not work
I'm using this in my ContentType:

        contentlink:
            label: Link to
            type: select
            values: pages/title
            sort: title

But when I go to add/edit a item, the `sort` parameter has no effect. The items in the dropdown are always ordered by id. I've tried `title`, `-title`, `-datepublish`, `-id` and more but the order in the dropdown is always the same. I also tried `order` as the key (as that is used for relations) but no difference.

Bolt 3.3.1, PHP 5.5.38, Apache 2.4

rossriley:
Fixes #6877

I think this one must have got lost in the re-org.

When using the following syntax:

```yml

        entrylink:
            label: Link to Entries
            type: select
            values: entries/title
            sort: -title
            multiple: true
```

The selects should be ordered by descending title, default is to sort by ascending column in the values field.

### DIFF
--- a/src/Form/Resolver/Choice.php
+++ b/src/Form/Resolver/Choice.php
@@ -90,16 +90,18 @@ final class Choice
             'limit'    => null,
             'sortable' => false,
             'filter'   => null,
+            'sort'     => null,
         ];
         $values = $field['values'];
         $limit = $field['limit'];
         $sortable = $field['sortable'];
         $filter = $field['filter'];
         $key = isset($field['keys']) ? $field['keys'] : 'id';
+        $orderBy = $field['sort'];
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
Thanks mate, I'm going to add unit test to this before sending in, and I'm also going to bump it down to 3.2 as it will affect there too.

GawainLynch:
My week has been too hectic, and I CBF wresting with PHPUnit on 3.2 … it's all fixed in 3.3 and I'll revisit the test(s) later … unless someone wants to make me a happy :princess:

PR #6892, opened 2017-08-12
[3.4] Implement use of bolt/common

GawainLynch:


src/Configuration/LowlevelChecks.php
```diff
@@ -45,7 +46,7 @@ public function __construct($config = null)
 
         $this->config = $config;
         $this->magicQuotes = get_magic_quotes_gpc();
-        $this->safeMode = ini_get('safe_mode');
```
CarsonF: You can just hardcode `false` here. `safe_mode` was removed in PHP 5.4.

src/Provider/DebugServiceProvider.php
```diff
@@ -196,7 +197,7 @@ public function boot(Application $app)
                 // 2. Converted to an exception and thrown to the user that way.
                 // Both of these are regardless of HTML or CLI output.
                 // This makes the built-in display_errors redundant.
-                ini_set('display_errors', 0);
+                Ini::set('display_errors', 0);
```
CarsonF: `false` would be better here instead of `0`

src/Storage/Entity/LogChange.php
```diff
@@ -359,8 +360,8 @@ private function fieldSelect($key, $value, array $fields)
      */
     private function fieldVideo($key, $value, array $fields)
     {
-        $before = json_decode($value[0], true);
-        $after  = json_decode($value[1], true);
+        $before = Json::parse($value[0], true);
+        $after  = Json::parse($value[1], true);
```
CarsonF: Second parameter for all of these

src/Storage/Field/Type/FieldTypeBase.php
```diff
@@ -248,17 +249,6 @@ public function getTemplate()
...
     protected function isJson($value)
     {
-        if (!is_string($value)) {
-            return false;
-        }
-
-        // This handles an inconsistency in the result from the JSON parser across 5.x and 7.x of PHP
-        if ($value === '') {
-            return false;
-        }
```
CarsonF: I don't think `Json::test` includes this check, maybe we need to add it?

tests/phpunit/unit/Controller/Backend/ExtendTest.php
```diff
@@ -114,9 +116,9 @@ public function testInstallInfo()
         $response = $this->controller()->installInfo($this->getRequest());
 
         $this->assertEquals(Response::HTTP_OK, $response->getStatusCode());
-        $parsedOutput = json_decode($response->getContent());
-        $this->assertNotEmpty($parsedOutput->dev);
-        $this->assertNotEmpty($parsedOutput->stable);
+        $parsedOutput = Json::parse($response->getContent(), 0);
```
CarsonF: `0` is default.

src/Helpers/Str.php
```diff
@@ -38,48 +41,6 @@ public static function makeSafe($str, $strict = false, $extrachars = '')
...
-     */
-    public static function replaceLast($search, $replace, $subject, $caseSensitive = true)
-    {
-        $pos = $caseSensitive ? strrpos($subject, $search) : strripos($subject, $search);
-        if ($pos !== false) {
-            $subject = substr_replace($subject, $replace, $pos, strlen($search));
-        }
-
-        return $subject;
-    }
```
CarsonF: These are BC breaks. The methods should be kept here and flip the parameters before calling parent.

tests/phpunit/unit/Controller/Backend/ExtendTest.php
GawainLynch: Typo `<insert facepalm emoji>`

src/Storage/Field/Type/FieldTypeBase.php
GawainLynch: Ah, good point!

src/Configuration/LowlevelChecks.php
GawainLynch: Typo `<insert facepalm emoji>`

src/Helpers/Str.php
GawainLynch: Derp `<insert facepalm emoji>`

GawainLynch:
> There's lots of places Json::parse() is called with a second parameter that is not needed.

> There are places Json::dump() is called with 0, but probably could be fine with some options. I understand some places don't make sense for pretty print, but I don't see why we need to escape slashes or unicode.

I knew this would come up, was waiting to see what/how :laughing:  Basically BC on the pretty print, and I would've expected the same on the output escaping, no?

CarsonF:
> I knew this would come up, was waiting to see what/how 😆 Basically BC on the pretty print, and I would've expected the same on the output escaping, no?

I mean JSON shouldn't be programmatically inspected, it should just be parsed. The only exception, I think, would be test assertions. So regardless of the dump flags, the output should be valid JSON which can be parsed anywhere. Now with that said, I know pretty printed can make things less readable for small data and increases size which should only matter when it is being transported over the wire. HTML _might_ have a problem with unescaped unicode, that should be double checked...I'm still drinking ☕️ . I've never seen the unescaped slashes hurt anything either.

src/Configuration/LowlevelChecks.php
GawainLynch: Change is below … caught me out for a minute why this wasn't refreshing

src/Controller/Backend/FileManager.php
```diff
@@ -330,12 +331,13 @@ private function getRelatedFiles(FileInterface $file)
...
-        if (Str::endsWith($base, '.dist')) {
+        Path::getExtension($base);
+        if (Path::getExtension($base) === '.dist') {
             $base = substr($base, 0, -5);
         }
         $ext = pathinfo($base, PATHINFO_EXTENSION);
-        $base = Str::replaceLast(".$ext", '', $base);
-        $base = Str::replaceLast('_local', '', $base);
+        $base = Str::replaceLast($base, ".$ext", '');
+        $base = Str::replaceLast($base, '_local', '');
```
CarsonF:
```php
        // Match foo(_local).*(.dist)
        $base = $file->getFilename('.dist');
        $base = Path::getFilenameWithoutExtension($base);
        $base = Str::replaceLast($base, '_local', '');
```

src/Storage/Field/Type/ListTypeBase.php
```diff
@@ -22,7 +23,7 @@ public function persist(QuerySet $queries, $entity)
         $value = $entity->get($key);
 
         if ($value !== null && !empty($value)) {
-            $value = $this->isJson($value) ? json_decode($value, true) : $value;
+            $value = $this->isJson($value) ? Json::parse($value) : $value;
```
CarsonF: `isJson` method could be deprecated and replaced with `Json::test`

src/Storage/Migration/Output/JsonFile.php
```diff
@@ -47,7 +48,7 @@ public function addRecord(array $data, $last)
         }
 
         // Generate the JSON string
-        $json = json_encode($data, JSON_PRETTY_PRINT);
+        $json = Json::dump($data, JSON_PRETTY_PRINT);
```
CarsonF: This needs work. There's a false check below that needs to be converted to try/catch. Still don't know why you don't want escaped slashes and unicode here.

src/Twig/Runtime/TextRuntime.php
```diff
@@ -40,7 +41,7 @@ public function __construct(LoggerInterface $systemLogger, Slugify $slugify)
      */
     public function jsonDecode($string)
     {
-        return json_decode($string, true);
+        return Json::parse($string);
```
CarsonF: Is throwing an exception ok here? Maybe people are using false returns here too...

tests/phpunit/unit/Controller/Backend/UploadTest.php
```diff
@@ -81,10 +81,11 @@ public function testInvalidFiletype()
...
         $this->assertEquals(Response::HTTP_OK, $response->getStatusCode());
 
-        $content = json_decode($response->getContent());
+        $content = Json::parse($response->getContent());
         $file = $content[0];
-        $this->assertAttributeNotEmpty('error', $file);
-        $this->assertRegExp('/extension/i', $file->error);
+        $this->assertArrayHasKey('error', $file);
+        $this->assertNotEmpty($file['error']);
+        $this->assertRegExp('/extension/i', $file['error']);
```
CarsonF: If you're asserting regex you don't need the empty assertion

tests/phpunit/unit/Helper/StrTest.php
```diff
@@ -79,10 +79,4 @@ public function testSwedishMakeSafe()
         $input = 'Skämt åsido satan vilket uruselt tillvägagångsätt';
         $this->assertEquals('skaemt-aasido-satan-vilket-uruselt-tillvaegagaangsaett', Str::makeSafe($input, true, ';'));
     }
-
-    public function testReplaceFirst()
-    {
-        $input = 'this is a test string this is a test string';
-        $this->assertEquals('one is a test string this is a test string', Str::replaceFirst('this', 'one', $input));
-    }
```
CarsonF: Probably should keep this and mark as legacy.

src/Storage/Field/Type/ListTypeBase.php
GawainLynch: Done

src/Controller/Backend/FileManager.php
GawainLynch: Done but tweaked as the `.dist` doesn't make sense here, they're not installed as `.dist` in end-user installs, don't list the other two (normal & _local) when being edited, and if they're available in the editor why would they be able to be saved? (they're distribution files, not end-user ones)

src/Storage/Migration/Output/JsonFile.php
GawainLynch: That … is … "Coding while sober". 

Fixed and second parameter is G.O.N.E. … gone!

src/Twig/Runtime/TextRuntime.php
GawainLynch: There is an invisible `try/catch` here, it just isn't showing up in your IDE … yet :stuck_out_tongue_winking_eye: 

Oh, and as this will come up,  the native [`json_decode`](http://php.net/manual/en/function.json-decode.php) returns a `null` if decoding fails.

tests/phpunit/unit/Helper/StrTest.php
GawainLynch: Done

tests/phpunit/unit/Controller/Backend/UploadTest.php
GawainLynch: Done

src/Storage/Field/Type/FieldTypeBase.php
GawainLynch: I'll update the constraint in this PR when we tag that branch on `common` :+1:

src/Controller/Backend/FileManager.php
CarsonF: If users don't have dist files anymore my first thought is :fire: but I'm also wondering if there are other use cases for this. There could be other places dist files are used, not just config files and not core files. I think since we have the functionality implemented it is fine to keep it.

src/Storage/Field/Type/FieldTypeBase.php
GawainLynch: This is included now in `bolt/common` 1.1.0+

GawainLynch:
OK! This should be good to go, and our combined hashing out of the JSON parameters should have both our O.C.D. triggers covered :+1: 

Unless you have derps to point out, lets :shipit: as by the time you're online today I'll be about to go offline, and we lose you over the weekend. I want to tie up the last of the beta work over the weekend & Monday EU time as @bobdenotter will be back as well.

src/Storage/Migration/Output/JsonFile.php
```diff
@@ -47,17 +49,16 @@ public function addRecord(array $data, $last)
...
+        } catch (DumpException $e) {
             $this->export
                 ->setError(true)
                 ->setErrorMessage('Unable to generate valid JSON data!');
 
             return false;
-        } elseif (!$last) {
-            $json .= ",\n";
         }
+        $json .= ",\n";
```
GawainLynch: Oh … my … `<insert double facepalm emoji>`

Issue #1694, opened 2014-09-04
[RFC] Extension added/generated menus


GwendolenLynch:
From previous discussion in Bolt 1 days, I know that we're keen to keep extension generated menu items grouped together...  But a couple of things about this that are annoying me, but it might be just me...
![extensions](https://cloud.githubusercontent.com/assets/1427081/4151502/9d03d1d2-3446-11e4-8801-68422f888688.png)

From an end-user perspective:
1. There is now an '**Extensions**' menu and an '**Extend**' menu.  Which is supposed to do what?
2. "What do you mean BoltBB is an 'extension', I just want to edit someone's reply"
3. What if the extension wants to have a '**Do Foo**' _and_ an '**Edit Bar**' menu:
- Are they configuration changes and belong as a 'Configuration' submenu?
- Are they 'Extending Bolt' options and belong there?
- Are they something else?

PR #6895, opened 2017-08-14
[3.4] Menus … 

GawainLynch:
* Fixes: #6896 — _"(No content.)"_ in recently edited menus
* Finishes: Admin menu generation in PHP instead of Twig 

So a bit more than I expected at this stage, but let's talk …

## Service definition

Is now overloaded and ugly. The menu construction itself needs a class of its own, but there are a few things to consider. Obviously happy for ideas, but the timer has started :wink: 

## Extension driven menus

This is a complete _CharlieFoxtrot_ currently, and we're not going back! :stuck_out_tongue_winking_eye: So we need to make some calls here. They're logically broken, let's get this right :smile: 

Extension driven menus require the `extensions` permission as a default in order for the parent menu to persist, and that makes no sense that to use MenuEditor a person should also have the level of access to add/remove extensions. 

Suggestions from me, either:
  - Create a new top-level menu group that requires the `settings` permission at the top-level, and then the extensions just define the menu's permission 
  - Stop kidding ourselves that extensions are going to flood the menu (#1694 & #3461), when it's easy enough to do with ContentTypes, e.g.

image

### DIFF
--- a/app/resources/translations/cs_CZ/messages.cs_CZ.yml
+++ b/app/resources/translations/cs_CZ/messages.cs_CZ.yml
@@ -10,8 +10,8 @@
 #   0 Unused messages
 #  10 Legacy untranslated messages
 # 116 Untranslated messages
-#  35 Legacy translation messages
-# 435 Translation messages
+#  34 Legacy translation messages
+# 436 Translation messages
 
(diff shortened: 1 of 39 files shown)


GawainLynch:
Given that people are more interested in lines of F.U.D. and gitignore files …

Issue #6758, opened 2017-06-16
[RFC] Seach in File Management


inadsan:
Stack functionality is very useful for writing new pages, news, etc.
But when you have many images that you reuse in several posts it costs a lot to find the images to add to the stack.

Proposal
It would be very good to be able to search within the file management.
And it would also be very useful if the administration could establish the properties of the images as "Alternative Text"

Thank you

GawainLynch:
Fixed in #6898

Issue #2263, opened 2014-12-03
In "File Management / Uploaded files" the "Upload file" procedure is confusing


GawainLynch:
I was just asked how to upload a file as the "Upload file" is greyed out.  It uses the old-fashioned uploader and needs to be updated inline uploader.

![uploader-buttons](https://cloud.githubusercontent.com/assets/1427081/5281170/4a0a6cd4-7ac8-11e4-92ee-7095ee68d3a3.png)

SahAssar:
Personally I'd like for us to prioritize to fix this for 3.1 (or 3.0 if possible), since it deviates from all the other file upload patterns we use, so I'm tagging this for 3.1.

GawainLynch:
So we're on the verge of releasing 3.3.0-stable and this is still a visual, and logical mess. Adding as a 3.4 blocker.

GawainLynch:
Fixed in #6898

PR #6898, opened 2017-08-15
[3.4] File manager improvements

lenvanessen:
Fixes:
#6758, #2263

Currently:
- Transforms body into dropzone for file uploads + adds dropzone hint on drop-event
- Moves event-handling handling outside of twig templates
- Improves the lay-out by moving the buttons to the right, instead of the bottom
- Lazy loads images, instead of pulls every thumbnail in when loading the page
- Adds search to the filemanager page

Test away!


app/view/twig/files/_files.twig
```diff
@@ -12,7 +12,7 @@
 
     <tbody>
         {% for file in context.files %}
-            <tr>
+            <tr class="entry" data-bolt-browse-name="{{ file.filename|excerpt(80)|shy | trim }}">
```
CarsonF: I don't think any of these filters are needed.
Why limit the filename?
The `|shy` filter isn't needed as this is data not display text.
The `|trim` filter isn't needed as filenames cannot contain spaces.

GawainLynch: > Why limit the filename?

I believe this came down to a the way BUIC handles things currently.

lenvanessen: Did not see this earlier, sorry about that.
Currently the filename itself is trimmed down in the display, for BUIC as wel as the filemanager. 

It comes down to this, if the filename in the data-attribute would be the original (if longer than 80), than you'd get an inconsitancy in the display, the marking, and the search.

If you'd copy paste a filename in that is longer than 80 chars it would be found, but not marked.

If we'd remove the trim, we should also remove it for the display, so they are consistent. Simply left it there because I don't know why it was put in in the first place.

PR #6903, opened 2017-08-16
[3.3] Performance: Replace slow |excerpt use with |slice

GawainLynch:
So while debugging some performance regressions that need to be addressed, the first one to pop up was in `Bolt\Helpers\Excerpt` … it's really slow :disappointed: 

The simple test below shows it to consistently be **5 - 7 times slower** that the native Twig function (+20ms on listing pages with only 5 records).

This PR just replaces its use where it make sense. More PRs to come :wink: 

```php
#!/usr/bin/env php
<?php
include_once __DIR__ . '/vendor/autoload.php';

function excerpt($content, $length = 200, $focus = null)
{
    $excerpter = new Excerpt($content);
    $excerpt = $excerpter->getExcerpt($length, false, $focus);

    return $excerpt;
}

$env = new Twig_Environment();
$subject = 'Bolt is an open source Content Management Tool, which strives to be as simple and straightforward as possible. It is quick to set up, easy to configure, uses elegant templates, and above all: It’s a joy to use.';
$length = 20;

$start = microtime(true);
for ($i = 0; $i <= 1000000; ++$i) {
    $result = twig_slice($env, $subject, 0, $length);
}
$run = microtime(true) - $start;
echo "twig_slice took $run seconds\n";

$start = microtime(true);
for ($i = 0; $i <= 1000000; ++$i) {
    $result = excerpt($subject, $length);
}
$run = microtime(true) - $start;
echo "Excerpt took $run seconds\n";
```

### DIFF
--- a/app/view/twig/_base/_listing.twig
+++ b/app/view/twig/_base/_listing.twig
@@ -186,9 +186,9 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
             <td class="username hidden-sm hidden-xs">
                 <i class="fa fa-user fa-fw"></i>
                 {% if content.user.displayname is defined %}
-                {{ content.user.displayname|excerpt(15) }}
+                {{ content.user.displayname|slice(0, 15) }}
             {% else %}
-                <s>{{ content.values.ownerid|excerpt(15) }}</s>
+                <s>{{ content.values.ownerid }}</s>
                 {% endif %}<br>
                 {% if content.status == 'timed' %}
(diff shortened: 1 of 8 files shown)


SahAssar:
Since we allow html elements in for example titlefields this needs to be `|striptags|slice`, right?

GawainLynch:
Ah, for the one in `app/view/twig/_macro/_macro.twig` yes, I can't read.

GawainLynch:
Updated and force pushed

PR #6907, opened 2017-08-17
[3.3] Fix incorrect date formatting

### LINKED ISSUE #6891: [BUG] Incorrect date format for locale set in config.yml
Bug
----

When choosing a locale in the config.yml file, it seems to be ignored for the purposes of setting the date format, possibly in the filter for Bolt's `localedatetime()` function. When setting it to `locale: 'en_GB'` the date format is returned on date fields for content types as 'm/d/y' and it should be 'd/m/y' for the 'en_GB' locale.

Details
-------

 - Bolt 3.3.1
 - Composer install
 - PHP 5.6 and 7.0
 - Apache 2.4

Reproduction
------------

Set the locale to 'en_GB' and then add a content type with a date field. The resulting date field will use the date format 'm/d/y' instead of the correct 'd/m/y' format.

Screenshots
-------------

screen shot 2017-08-11 at 12 51 01

GawainLynch:
Fixes #6891

So this one has been laying dormant since #5836 back in September 2016 (c.f. https://github.com/bolt/bolt/commit/fb5a7c65beead0092b24189c2b67cc7fe647e458) … and to think I spent ages swearing at JavaScript, when it was just the locale files not loading :rofl: 

This is turning into a day of "bugs due to a single missing character" :champagne:

### DIFF
--- a/app/view/twig/_base/_page.twig
+++ b/app/view/twig/_base/_page.twig
@@ -28,10 +28,10 @@
 
 {# Add library locale files #}
 {% for lib in ['datepicker', 'moment', 'select2'] %}
-    {% if app.filesystem.has("bolt_assets://js/locale/#{lib}#{page_locale_long}.min.js") %}
-        {% set page_scripts_inc = page_scripts_inc|merge([asset("js/locale/#{lib}#{page_locale_long}.min.js", 'bolt')]) %}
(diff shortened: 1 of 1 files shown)


bobdenotter:
Github mentioned an "out of date branch", but seeing how this is just one file, it seems harmless to merge in as-is. :-)

GawainLynch:
> Github mentioned an "out of date branch",

Yeah, still trying that out. I think it makes sense for our other projects, but not sure right now about core.

PR #6910, opened 2017-08-17
Allow definition lists

Skateside:
Looks like a typo - the `<dt>` element wasn't allowed but a `<dh>` element was - as far as I can tell, the `<dh>` element doesn't exist


bobdenotter:
@Skateside Thanks, good catch! 👍

GawainLynch:
Awesomesauce! :+1:

Skateside:
That was much easier than I thought it'd be - thanks :)

PR #6918, opened 2017-08-20
Remove `.idea`, `.project`, `.sublime` from our git ignores, as they should be in the global ignore instead.

bobdenotter:


.gitignore
```diff
@@ -46,6 +46,7 @@ npm-debug.log
 # File-system cruft and temporary files
 .DS_Store
 .idea
+.vscode
```
GawainLynch: Both should be in the global ignore

bobdenotter: Strictly speaking, yes. But we both know that a lot of people don't have them, so we should put them in our `.gitignore`.

GawainLynch: Then they should not be using git

GawainLynch: The agreement struck — while getting quite heated at each other **at your desk** — was that we'd move these things to the composer install target, and that a core developer should know what they are doing

bobdenotter:
Right, it _should_ be in there.. Will PR it. 

If we do that, then we should _remove_ `.idea`, `*.sublime-*` and `.project` from the `bolt/bolt` repo's gitignore for consistency's sake.

GawainLynch:
> If we do that, then we should remove .idea, *.sublime-* and .project from the bolt/bolt repo's gitignore for consistency's sake.

Yeah, that was the original discussion and both of us were/are just too overloaded to keep up … hence one of the other one of us ending up in a grumpy mood too often.

bobdenotter:
Ok, on the same page now.. I'll change the PR to _remove_ those instead of _adding_ the `.vscode` one.

bobdenotter:
Fixed.

bobdenotter:
Ok, fixed now.

GawainLynch:
For reference, not that it matters right now, we should clean up the rest of the cruft in there.

PR #6919, opened 2017-08-20
[3.4] Allow for consecutive prefills. 

### LINKED ISSUE #6917: [3.4] Can't add more "lorem ipsum" records, if there are already records.
I get this result: 

screen shot 2017-08-20 at 12 10 25

Instead, it should just add some more records to the contenttype.

bobdenotter:
- Allow for consecutive prefills, with the Lorem Ipsum generator
 - Fix logic error in `$form->get('contenttypes')->has('contenttypes')`
 - Small UI change: show the "prefill with Lorem Ipsum" button not only after finished DB check.

Fixes #6917.

### DIFF
--- a/src/Controller/Backend/General.php
+++ b/src/Controller/Backend/General.php
@@ -194,15 +194,17 @@ class General extends BackendBase
 
         if ($request->isMethod('POST') || $request->query->getBoolean('force')) {
             $form->handleRequest($request);
-            if ($form->get('contenttypes')->has('contenttypes')) {
+            if (!empty($form->get('contenttypes')->getData())) {
                 $contentTypeNames = (array) $form->get('contenttypes')->getData();
+                $skipNonEmpty = false;
             } else {
                 $contentTypes = $this->app['config']->get('contenttypes');
                 $contentTypeNames = array_keys($contentTypes);
+                $skipNonEmpty = true;
             }
 
             $builder = $this->app['prefill.builder'];
-            $results = $builder->build($contentTypeNames, 5);
+            $results = $builder->build($contentTypeNames, 5, $skipNonEmpty);
             $this->session()->set('prefill_result', $results);
 
             return $this->redirectToRoute('prefill');
--- a/src/Controller/Backend/General.php
+++ b/src/Controller/Backend/General.php
@@ -194,17 +194,20 @@ class General extends BackendBase
 
         if ($request->isMethod('POST') || $request->query->getBoolean('force')) {
             $form->handleRequest($request);
-            if (!empty($form->get('contenttypes')->getData())) {
-                $contentTypeNames = (array) $form->get('contenttypes')->getData();
-                $skipNonEmpty = false;
-            } else {
-                $contentTypes = $this->app['config']->get('contenttypes');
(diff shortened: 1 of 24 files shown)


src/Controller/Backend/General.php
```diff
@@ -197,15 +197,17 @@ public function prefill(Request $request)
 
         if ($request->isMethod('POST') || $request->query->getBoolean('force')) {
             $form->handleRequest($request);
-            if ($form->get('contenttypes')->has('contenttypes')) {
+            if (!empty($form->get('contenttypes')->getData())) {
```
GawainLynch: Friends don't let friends use `empty()`

GawainLynch:
So we are or we are not allowing this? That is why the force was there.

src/Controller/Backend/General.php
bobdenotter: It returns `[]` when nothing is selected. Open to suggestions on how to do it properly.

bobdenotter:
The 'force' is for a slightly different usecase. If you're on a fresh install with no content, it shows the blue bar: 

screen shot 2017-08-20 at 14 54 51

This link has the `?force=1` attached to it. What it does is not "overriding" the number of items to prefill, but rather _skip_ the screen where it asks which contenttypes to prefill, but instead going straight ahead to prefilling all the empty ones.

src/Controller/Backend/General.php
GawainLynch: The negation will work for an empty array, however after running though xdebug it returns an `[0]` which won't pass negation, but [checking Symfony docs](http://symfony.com/doc/current/reference/forms/types/choice.html#empty-data) when the field type is `ChoiceType` **and** `multiple` is true, **and** `expand is true it should return an empty array … that isn't empty to me … so for now your `empty()` wins as a solution to getting this over the line. Just unsure as to what that line is.

src/Storage/Database/Prefill/Builder.php
```diff
@@ -40,10 +40,11 @@ public function __construct(EntityManager $storage, callable $generatorFactory,
      *
      * @param array $contentTypeNames
      * @param int   $count
+     * @param bool  $skipNonEmpty
      *
      * @return null
      */
-    public function build(array $contentTypeNames, $count)
+    public function build(array $contentTypeNames, $count, $skipNonEmpty)
```
GawainLynch: This is a BC break (as shown by the tests)

GawainLynch:
Given your coming back online from a couple of weeks away, RL work is going to want all of your attention, I'll pick this up and get it over the line for you.

bobdenotter:
I though about cheating, and changing it to: 

```
public function build(array $contentTypeNames, $count, $skipNonEmpty = true)
```

, but that would obviously not be the right way to go.

If you have time, fixing it would be much appreciated. If not, i'll make time in the next few days. :-)

GawainLynch:
Yeah, I am making time … there is a legitimate problem to be fixed here, this PR goes most of the way to fixing it, and I really don't want to delay this beta any longer thank needed.

GawainLynch:
There we go … passes tests, and no BC breaks.

PR #6931, opened 2017-08-22
[3.4] More menus

GawainLynch:
- Follow-up to #6895
- Fixes _Set-up Checks_ menu item showing when when `files:check` permission is given, but the route will deny access
- Fixes recently edited icons
- Allows extension driven menus to appear for the correct permission. Failing continued requests for a decision on this, none came … I'll give it the evening and then :shipit:


GawainLynch:
> As far as the "Top level menus from extensions" go, I'm not a proponent of it,

Yeah, I still don't know how I feel about it. On the one hand it make sense, on the other your objection has had long standing grounds and the way that certain other product behaves … well.

The next part I might take up in Slack, as it is better suited to there :wink:

bobdenotter:
I'd say: Let's roll with it, and if it turns out that it's becoming chaos, we can always add a limit for it later. Something like "collapse extension-added menu's into one item, if there's more than 6" or something.

PR #6934, opened 2017-08-23
Use followLinks() on Finder

ntomka:
Use [followLinks()](http://api.symfony.com/2.3/Symfony/Component/Finder/Finder.html#method_followLinks) on Finder, because on Linux dev environments composer creates symlinks instead copying source.

### DIFF
--- a/src/Composer/EventListener/PackageEventListener.php
+++ b/src/Composer/EventListener/PackageEventListener.php
@@ -120,6 +120,7 @@ class PackageEventListener
             ->name('composer.json')
             ->notPath('vendor/composer')
             ->depth(2)
+            ->followLinks()
         ;
         try {
             $finder->in(['vendor']);


GawainLynch:
> on Linux dev environments composer creates symlinks instead copying source.

Not quite. The only place/time symlinks are created by Composer is in the [BinaryInstaller](https://github.com/composer/composer/blob/master/src/Composer/Installer/BinaryInstaller.php#L160) (calling [Filesystem](https://github.com/composer/composer/blob/1.5/src/Composer/Util/Filesystem.php#L537-L548)) to set up links from `vendor/bin/` to the package install location.

Also, Bolt's filesystem uses Flysystem and doesn't support symlinks (we hope to enhance that soon) so they're currently not supported.

What is the problem you're hitting?

ntomka:
I started to develop a new extension locally and set the repository path in extensions/composer.json as path type. Composer didn't copy it just symlinked.

Composer will create symlink by default if possible: [composer doc](https://getcomposer.org/doc/05-repositories.md#path)

> The local package will be symlinked if possible, in which case the output in the console will read Symlinked from ../../packages/my-package. If symlinking is not possible the package will be copied. In that case, the console will output Mirrored from ../../packages/my-package.

I read it now there's an option to force disable symlinking, I may try that later.

GawainLynch:
Ah right … Let me check the Composer source again then. I did just look before I responded to make sure I wasn't doing a :koala: … but I have been known to make mistakes (like daily :rofl:).

Tell me though, when you installed the extension with your fix, did everything work normally with the extension or was it too early in the development of it to tell?

ntomka:
It works as expected, extension loaded and used. Before Bolt didn't even loaded the extension because Finder didn't listed it. :)

GawainLynch:
Ah yeah, I should have used `ag` instead of relying on PhpStorm :imp: … it is indeed in the [PathDownloader](https://github.com/composer/composer/blob/master/src/Composer/Downloader/PathDownloader.php#L105) as well … which is a worry as the PHP code should be fine, but people are going to get mighty confused when paths to extension assets don't work.

Let me just hash it over quickly with the team, I think we can merge this and update the docs as for the common use-case it would be fine … but for the poor person trying to load a JavaScript file from their extension it's going to be a bad day :smile_cat:

GawainLynch:
OK, @bobdenotter agrees with me … Let's do this! :+1:

ntomka:
Great! Thank you! :+1:

PR #6935, opened 2017-08-23
[3.3] Catch exceptions better, don't break layout.

bobdenotter:
Sometimes the Exception pages would throw an exception, because we tried to output objects as if they were strings:

```
Catchable Fatal Error: Object of class Bolt\Extension\Bobdenotter\Chmodinator\ChmodinatorExtension could not be converted to string"
```

This would give an obscure, hard to read error, instead of pointing to the correct place where it should be fixed. 

This PR handles that slightly better, as well as makes sure that the output stays within bounds of the screen.

Before: 

screen shot 2017-08-23 at 17 23 33

After:

screen shot 2017-08-23 at 17 28 35

Also before: 

screen shot 2017-08-23 at 17 26 26

Also after: 

screen shot 2017-08-23 at 17 29 47

### DIFF
--- a/app/view/css/bolt_exception.css
+++ b/app/view/css/bolt_exception.css
@@ -12,3 +12,15 @@ pre[data-line]{position:relative;padding:1em 0 1em 3em}.line-highlight{position:
     font-style: italic;
     cursor: pointer;
 }
+
+th {
+    width: 22%;
+}
+
+th.narrow {
+    width: 8%;
+}
+
+td {
+    word-break: break-all;
+}
--- a/app/view/twig/exception/_request.twig
+++ b/app/view/twig/exception/_request.twig
@@ -1,58 +1,52 @@
-{% macro printvalue(value) %}
-    {% if value is iterable and value is not empty %}
-        {% for k, v in value %}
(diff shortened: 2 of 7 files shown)


src/Twig/Extension/ArrayExtension.php
```diff
@@ -188,4 +203,28 @@ public function unique(array $arr1, array $arr2)
...
     }
+
+    /**
+     * Tests if a given value is an array or not, returns the result.
+     *
+     * @param mixed $value
+     *
+     * @return bool
+     */
+    public function array($value)
```
GawainLynch: These can't be used as function names in PHP 5. Plus as discussed in Slack :thinking: over what we do here

bobdenotter: Changing the function name will be easy, at least.

GawainLynch: Dare I say it, that would have shown up if you were using PhpStorm and not that other one :stuck_out_tongue_winking_eye:

GawainLynch:
OK, open for debate … I've pushed some extra commits to discussion approach on.

Whatever we agree on, I'll clean up the commits in the branch pre-merge :smile:

PR #6936, opened 2017-08-23
[4.0] Slugify 3.0

GawainLynch:


bobdenotter:
Any nice improvents in Slugify 3.0, or just a regular update?

GawainLynch:
Just an update … but might as well stay on target as we're a while off release :smile:

PR #6940, opened 2017-08-25
[3.3] Support DBAL 2.6

### LINKED ISSUE #6856: DBAL 2.6 Problems
Apart from the rest of the dramas a default install on 3.2 & 3.3 (probably all versions) give the following per platform:

Sqlite
```
Table bolt_pages is not the correct schema: invalid column template
Table bolt_showcases is not the correct schema: invalid column file
```

PostgreSQL:
```
Table bolt_blocks is not the correct schema: invalid column templatefields, invalid column image
Table bolt_entries is not the correct schema: invalid column templatefields, invalid column image, invalid column video
Table bolt_field_value is not the correct schema: invalid column value_json_array
Table bolt_log_change is not the correct schema: invalid column diff
Table bolt_log_system is not the correct schema: invalid column source
Table bolt_pages is not the correct schema: invalid column templatefields, invalid column image
Table bolt_showcases is not the correct schema: invalid column templatefields, invalid column geolocation, invalid column video, invalid column image, invalid column imagelist, invalid column  filelist, invalid column multiselect
Table bolt_users is not the correct schema: invalid column stack, invalid column roles
```

GawainLynch:
Fixes #6856. This is currently WiP and for testing on approach 

**Updated:** This should now cover PostgreSQL 9.1 again, as well as MySQL 5.7.


bobdenotter:
The problem this fixes was in 3.2 and up, while this is targeted at 3.3. 

I'm perfectly fine with that, but after this lands, we need to keep in mind to urge people who bump in this to upgrade to at least 3.3.latest. :-)

GawainLynch:
3.2 went End of Life yesterday :wink:

bobdenotter:
Hence the "perfectly fine with that" :-p

bobdenotter:
I've tested this locally on sqlite 3.20, 3.16 and mysql 5.7.. I've used the databases from a few production websites on both sqlite and mysql from bolt 2.2 and 3.2, and so far no problems.. 

I have the feeling that the updates suggests changing _more_ fields than before, but the results seem fine.

GawainLynch:
OK, someone with coverage of MySQL 5.5 (other than me) would be good too.

The fields are more numerous as one of the changes that has to be made is dropping the defaults on nullable fields. It never made sense to have those anyway, but there was some reason during 2.3-dev that we kept them, I just can't remember what that was and it is part of what I am nervous about.

That said, it could have been changes in DBAL 2.4 -> 2.5 that resolved whatever we were hitting as well.

src/EventListener/DoctrineListener.php
```diff
@@ -82,12 +84,52 @@ public function postConnect(ConnectionEventArgs $args)
...
-        } elseif ($platform === 'postgresql') {
-            /**
-             * @see https://github.com/doctrine/dbal/pull/828
-             */
+        } elseif ($platformName === 'postgresql') {
+            /** @see https://github.com/doctrine/dbal/pull/828 */
             $db->executeQuery("SET NAMES 'utf8'");
         }
+
+        if (DBAL\Version::compare('2.6.0' >= 0)) {
```
CarsonF: Parenthesis mistake.

src/Storage/Database/Schema/Types/JsonType.php
```diff
@@ -0,0 +1,86 @@
...
+use Bolt\Common\Exception\DumpException;
+use Bolt\Common\Exception\ParseException;
+use Bolt\Common\Json;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types;
+
+/**
+ * Type generating JSON object values.
+ *
+ * @internal Provided to bridge DBAL 2.5 & 2.6 support.
```
CarsonF: Maybe note to drop when 2.5 and PHP 7.0 is dropped?

src/Provider/DatabaseServiceProvider.php
```diff
@@ -81,8 +85,14 @@ function ($app) {
             $app->extend(
                 'db',
                 function ($db) use ($app) {
+                    /** @var \Bolt\Storage\Database\Connection|\Bolt\Storage\Database\MasterSlaveConnection $db */
                     $db->setQueryCacheProfile($app['db.query_cache_profile']);
 
+                    // Currently, this can get called early, meaning the connection listener is too late
+                    if (DBAL\Version::compare('2.6.0') >= 0 && !Type::hasType('json')) {
```
CarsonF: I'd drop the version comparison here. The feature detection is sufficient.

src/EventListener/DoctrineListener.php
```diff
@@ -82,12 +84,52 @@ public function postConnect(ConnectionEventArgs $args)
...
+            }
+        } elseif ($platformName === 'postgresql') {
+            if ($platform instanceof Platforms\PostgreSQL92Platform) {
+                /** @see \Doctrine\DBAL\Platforms\PostgreSQL92Platform::getJsonTypeDeclarationSQL */
+                $platform->registerDoctrineTypeMapping('JSON', Schema\Types\JsonType::JSON);
+            } else {
+                /** @see \Doctrine\DBAL\Platforms\PostgreSqlPlatform::getClobTypeDeclarationSQL */
+                $platform->registerDoctrineTypeMapping('TEXT', Schema\Types\JsonType::JSON);
+            }
+        }
```
CarsonF: This should be moved to `JsonType::getMappedDatabaseTypes`.

src/EventListener/DoctrineListener.php
GawainLynch: `<insert facepalm>`

src/EventListener/DoctrineListener.php
GawainLynch: Moved

src/Provider/DatabaseServiceProvider.php
GawainLynch: I can't logic some days! Done :+1:

src/Storage/Database/Schema/Types/JsonType.php
GawainLynch: Yep, merged the docblock from the listener method when I moved it over

src/Storage/Database/Schema/Types/JsonType.php
```diff
@@ -0,0 +1,126 @@
...
+
+
+    /**
+     * Register platform specific JSON type mappings for DBAL prior to 2.6.0.
+     *
+     * @param AbstractPlatform $platform
+     *
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public static function registerTypeMapping(AbstractPlatform $platform)
```
CarsonF: Implement this as `getMappedDatabaseTypes`. You don't need to call it either, DBAL will handle registering it.

GawainLynch: :+1:

src/Storage/Database/Schema/Types/JsonType.php
```diff
@@ -0,0 +1,127 @@
...
+            if ($platform instanceof Platforms\PostgreSQL92Platform) {
+                /** @see \Doctrine\DBAL\Platforms\PostgreSQL92Platform::getJsonTypeDeclarationSQL */
+                $type = 'json';
+            } else {
+                /** @see \Doctrine\DBAL\Platforms\PostgreSqlPlatform::getClobTypeDeclarationSQL */
+                $type = 'text';
+            }
+        }
+
+        $platform->registerDoctrineTypeMapping($type, static::JSON);
```
CarsonF: You don't need to call this. Doctrine does it for you [here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L247-L249).

src/Provider/DatabaseServiceProvider.php
```diff
@@ -81,8 +84,16 @@ function ($app) {
             $app->extend(
                 'db',
                 function ($db) use ($app) {
+                    /** @var \Bolt\Storage\Database\Connection|\Bolt\Storage\Database\MasterSlaveConnection $db */
                     $db->setQueryCacheProfile($app['db.query_cache_profile']);
 
+                    // Currently, this can get called early, meaning the connection listener is too late
+                    if (!Type::hasType('json')) {
+                        Type::addType('json', Schema\Types\JsonType::class);
+                    }
```
CarsonF: Maybe we should override it for 2.6+ so it is consistent? Our Type is different than the one doctrine provides.

src/Storage/Database/Schema/Types/JsonType.php
GawainLynch: Then something weird is happening, I originally dropped it but things went :boom: 

I'll dig now.

src/Provider/DatabaseServiceProvider.php
GawainLynch: In that case we might as well lock on DBAL 2.5, and I'd just have to maintain a fork Bolt for PHP 7.1+. But I don't think the differences for 2.5 users are enough of a difference to warrant it.

src/Storage/Database/Schema/Types/JsonType.php
GawainLynch: I don't know what you're talking about, that line isn't really there … this is your imagination :stuck_out_tongue_winking_eye: 

Oh look … :smile_cat:

GawainLynch:
> Do we really need to keep BC for the CarbonTypes? When would users interact with those directly?

Yeah, we **keep** getting caught out by those assumptions and I'm pretty much over having to explain to people why the project dropped the ball on a BC change. 

If we move it or replace it, and it **did not** have an `@internal` on it, we maintain BC until 4.0.

> If you insist, then the new classes are missing class_alias calls.

Ugh, I was taking it from how you did it elsewhere … but if _you_ insist, sure :stuck_out_tongue_winking_eye:

GawainLynch:
>  Also I don't think we need to keep IDE completion for these deprecated classes, so I would skip the non-used class subclass.

Yeah, looking at it, that does make sense. I was going for completion, and to not have you ask "Why did you not… ?" … but I managed to trigger that anyway :rofl:

src/Provider/DatabaseServiceProvider.php
CarsonF: How does that follow? I am saying ours now matches what dbal 3.0 will look like. So we need to apply it to all of dbal 2.x

GawainLynch: First thing that comes to mind is the exception and JSON processing … there were others, but at this point if that's a breaker I am just deleting the branch

src/Storage/Mapping/Type/CarbonDateType.php
```diff
@@ -2,25 +2,11 @@
...
 
-        if (empty($value)) {
-            return null;
-        }
-
-        return Carbon::instance(parent::convertToPHPValue($value, $platform));
-    }
-}
+// Load CarbonDateType.php which defines class alias
+class_exists(Types\CarbonDateType::class);
```
CarsonF: The class exists call and comment are not needed and wrong. The alias is defined here not in the new class.

GawainLynch:
> The new Carbon Type classes should be marked as internal. Don't you think?

Well, I wanted to … but is that justified given they are supporting an old use, albeit a generally invalid one?

src/Storage/Mapping/Type/CarbonDateType.php
GawainLynch: I just hadn't pushed the update from our conversation, I was still waiting on an answer in Slack :disappointed:

src/Storage/Database/Schema/Types/JsonType.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return 'json';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return !$platform->hasNativeJsonType();
```
CarsonF: I'm pretty sure that this always needs to return true (matching upstream) that way DBAL doesn't try to use JsonArrayType when mapping back to PHP value.

GawainLynch: OK, I *think* I know part of what is confusing us here. Running it though xdebug on freshly dropped/recreated PostgreSQL 9.6 tables (which definitely has JSON support :wink:) 

### ContentType
```
koalas:
    name: Koalas
    singular_name: Koala
    fields:
        title:
            type: text
            class: large
            group: content
        slug:
            type: slug
            uses: title
        images:
            type: imagelist
        body:
            type: html
            height: 300px
```
### Schema
```
bolt_testing=# \d+ bolt_koalas
                                                              Table "public.bolt_koalas"
     Column     |              Type              |                        Modifiers                         | Storage  | Stats target |  Description   
----------------+--------------------------------+----------------------------------------------------------+----------+--------------+----------------
 id             | integer                        | not null default nextval('bolt_koalas_id_seq'::regclass) | plain    |              | 
 slug           | character varying(128)         | not null                                                 | extended |              | 
 datecreated    | timestamp(0) without time zone | not null                                                 | plain    |              | 
 datechanged    | timestamp(0) without time zone | not null                                                 | plain    |              | 
 datepublish    | timestamp(0) without time zone | default NULL::timestamp without time zone                | plain    |              | 
 datedepublish  | timestamp(0) without time zone | default NULL::timestamp without time zone                | plain    |              | 
 username       | character varying(32)          | default NULL::character varying                          | extended |              | 
 ownerid        | integer                        |                                                          | plain    |              | 
 status         | character varying(32)          | not null                                                 | extended |              | 
 templatefields | json                           |                                                          | extended |              | (DC2Type:json)
 title          | character varying(256)         | default NULL::character varying                          | extended |              | 
 images         | json                           |                                                          | extended |              | (DC2Type:json)
 body           | text                           |                                                          | extended |              | 
Indexes:
    "bolt_koalas_pkey" PRIMARY KEY, btree (id)
    "idx_aef722b7b00651c" btree (status)
    "idx_aef722b989d9b62" btree (slug)
    "idx_aef722ba5131421" btree (datepublish)
    "idx_aef722bafba6fd8" btree (datecreated)
    "idx_aef722bb7805520" btree (datedepublish)
    "idx_aef722bbe74e59a" btree (datechanged)
```

### Breakpoints
screenshot from 2017-08-31 06-46-27

### Record
screenshot from 2017-08-31 06-48-47

### Trace
image

### Overridden JSON class
image

Note the break point is still on the deprecated DBAL class.

GawainLynch: It is worth noting that not **all** of the JSON fields go to the legacy class (even when repeated with the hinting off), it seems to be just some of the CT ones. I am guessing this is where we are crossing wires?

GawainLynch: :rofl: I know why … commit(s) incoming

GawainLynch:
@CarsonF: Regards your comment on _"Simpler and stricter null checks on CarbonTypes"_, you are obviously correct. The empty string was/is for BC. I have a data migration branch underway that we can use in v4 to clean up entries like this, but for now ¯\_(ツ)_/¯

src/Storage/Database/Schema/Types/JsonType.php
GawainLynch: OK, so now with the "storage type mapping" commit, I can now start with a fresh DB on above set up, create first user, and create content without hitting the legacy method … Including starting on a fresh 3.2 install.

GawainLynch: Dear Internet historians … OK, so here is how this turned out.

So in February 2011 Benjamin [added this commit](https://github.com/doctrine/dbal/commit/28c7e09ddec068f0d3abc759858dcba8e05c1067) that enabled Dcotrine types to have comments to determine "Is it necessary for the platform to add a parsable type comment to allow reverse engineering the given type?" via the `$this->doctrineTypeComments` class property.

Now the only places this is used are 
- `AbstractPlatform::isCommentedDoctrineType()`
- `AbstractPlatform::markDoctrineTypeCommented()`
- `AbstractPlatformgetDoctrineTypeComment()` 

… and the only place any of those is called in the Doctrine libraries Bolt includes is in `DB2Platform` 

… and while I am still a closet fan of DB2, we don't use or support it.

But wait, there's more!

Fast forward to 2015 and https://github.com/doctrine/dbal/pull/891 introduced the first attempt to enhance and fix some of the original design shortcomings in `JsonArrayType`, which was correctly pushed back on due to the BC break it would have created.

The PR author then created an updated PR, https://github.com/doctrine/dbal/pull/892, to introduce a new type … hence where we are right here.

In https://github.com/doctrine/dbal/pull/892 there is [this key comment from Christophe](https://github.com/doctrine/dbal/pull/892#r36311026):

> But the native type can only be mapped to a single doctrine type, and all other doctrine types reusing the same native type must use a comment hint to make the schema retrieval work properly.
> Otherwise the native field will always be considered as a json_array and any of your json field will trigger an ALTER TABLE query in each schema update run or migration generation. this is precisely why we introduced comment hints

In other words, as I read it, the comment(s) are only for migrations … and we don't use, or support, migrations.

GawainLynch:
Thanks for your attention to detail on this review @CarsonF :+1:

PR #6951, opened 2017-08-29
Refactor fields trait to extend the type service with new fields

rossriley:


ToBe998:
Warning. This does not work.

`$app['storage.typemap']` needs an array with key > **name of class**
`$manager->addFieldType($fieldName, $fieldClass);` needs a key and **class object**

PR #6964, opened 2017-09-02
Added CSS to known extensions for files directory

znegva:
CSS files should be provided with the correct Content-Type from the files directory.
This can be useful e.g. for :
* theme creators (to allow users to upload/edit a small custom CSS file in the backend to manipulate some basic settings such as anchor color)
* authors can provide CSS-files to be used in other sites (e.g. [blog.fefe.de](https://blog.fefe.de/faq.html) offers the option to include a external css)

The change is necessary b.c. if the Content-Type does not match browsers will not use the CSS-file, see [here](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Configuring_server_MIME_types#Why_are_correct_MIME_types_important.3F)


bobdenotter:
Thanks, @znegva! 👍

PR #6970, opened 2017-09-03
[3.4] Don't show files & directories that start with a dot

### LINKED ISSUE #6899: Hidden files & directories show in file server file selector UI
image

GawainLynch:
Fixes #6899

Could we hang off merging this until after the 11th when Carson is back from vacation. I have a feeling he'll trigger on this, but before we release the next major of `bolt/filesystem` we can't modify the interface, and my OCD (potentially) can't handle the overhead of a method check … so … let's see :crossed_swords: :wink:

### DIFF
--- a/app/config/permissions.yml.dist
+++ b/app/config/permissions.yml.dist
@@ -72,6 +72,7 @@ global:
     extensions:config: [ developer ]
     fileedit: [ admin, developer ]
     files:config: [ developer ]
+    files:hidden: [ developer ]
     files:theme: [ developer ]
     files:uploads: [ admin, developer, chief-editor, editor ]
     files: [ admin, developer, chief-editor, editor ]
--- a/src/Controller/Async/FilesystemManager.php
+++ b/src/Controller/Async/FilesystemManager.php
@@ -6,7 +6,7 @@ use Bolt\Filesystem\Exception\ExceptionInterface;
 use Bolt\Filesystem\Exception\FileExistsException;
 use Bolt\Filesystem\Exception\FileNotFoundException;
 use Bolt\Filesystem\Exception\IOException;
-use Bolt\Filesystem\Handler\HandlerInterface;
+use Bolt\Filesystem\Listing;
 use Bolt\Translation\Translator as Trans;
 use Silex\ControllerCollection;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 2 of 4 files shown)


GawainLynch:
OK, pushing this now to "RFC" status, if we can get general consensus I'll have to bump to 3.4 though due to it crossing over into feature territory with the new class.

This now allows someone to enable the showing of hidden files, should they chose that option.
```yaml
file_manager:
    show_hidden: true
```
The PR also adds a service to manage this in one place, so we get some degree of consistency.

bobdenotter:
Looks good at first glance. I'd still be in favour though of a distinction between `.htaccess` and other `.foo` files and folders. (I wonder if the apache foundation would accept a PR to rename `.htaccess` to `apache_local.conf`, which would make more sense to me, but that's besides the point.. ;-) )

Anyhow, how about this?

```yaml
file_manager:
    show_hidden: false
    show_htaccess: true
```

GawainLynch:
Well, I know you're making a joke but there is no shortage of people, myself included, that think `.htaccess` was a huge mistake. Igor's ["famous" little rant](https://www.nginx.com/resources/wiki/start/topics/examples/likeapache-htaccess/) gives just one good reason why.

Another is security … and oh my, don't start me on this one. Honestly, making a distinction here just promotes the utter worse behaviour imaginable and it is bad enough we're compromising on showing dot files :disappointed:

GawainLynch:
To put it more bluntly, if you're editing `.htaccess` files in a web editor. Get out of web development.

GawainLynch:
> For the record, I think we should make an exception for .htaccess, but this is an improvement over what we have. If people complain about missing that specific file, we can always re-consider.

Just tabbed past this and something struck me that I might have missed when I read this earlier today `<insert faceplam emjoi>`

You might be getting this, and I might just be misunderstanding, but `show_hidden: true` will allow you to edit the Apache files. 

While I feel pretty strongly that adding access to dot files is very bad full stop, it is a compromise I think can be made and this PR makes that compromise. What I strongly object to is singling out `.htaccess` files in config, as it send the message that we both support & agree upon the approach, and that people not well versed in the risk will see that as a green light on good approach … it is literally the sort of re-enforcement the broader security community has been working hard to combat for years.

That said, given this is designed to be called from a controller, another option is to drop the config options all together and give `root` users the option to turn on/off hidden files via a button & query parameter backed by a security check in the route. The thinking there is in our security model, the root user(s) are analogous to a UNIX root user (just not at the OS level in this case), and have full control over the site.

TBH, having this wrapped in a security check eases the pain in my mind, and I am already preferring the approach … talk to me :wink:

bobdenotter:
>You might be getting this, and I might just be misunderstanding, but show_hidden: true will allow you to edit the Apache files.

Yes, I got that. I was just making a distinction between `.htaccess` and "others", because in my experience, people _do_ want to sometimes edit the `.htaccess`, but I've never heard of someone wanting to edit a different .dot file through the interface. 

But yeah, this option will allow them to do so, so let's get it in. Do you want to retarget for 3.4, or leave as it?

It _does_ need to get documented somewhere, somehow, though.

GawainLynch:
Bumped to 3.4, changed it to depend on the `files:hidden` permission.

Lets run it past @CarsonF tomorrow.

lenvanessen:
FWIW, workes like a charm here:)

PR #6972, opened 2017-09-04
New Storage Layer - Add Compatibility features to allow both systems to work concurrently

rossriley:
This is another set of changes that allow themes to render correctly in TWIG no matter which storage layer provides the objects.

Main additions are moving the field type information into a Trait that can be shared by both objects, adding a Twig filter to post process new-style Taxonomy objects and adding some extra checks to the Frontend controller to handle potential differences between the two object types.

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -150,7 +150,7 @@ class Frontend extends ConfigurableBase
         // First, try to get it by slug.
         $content = $this->getContent($contenttype['slug'], ['slug' => $slug, 'returnsingle' => true, 'log_not_found' => !is_numeric($slug)]);
 
-        if (!$content && is_numeric($slug)) {
+        if (is_numeric($slug) && (!$content || count($content) === 0)) {
             // And otherwise try getting it by ID
             $content = $this->getContent($contenttype['slug'], ['id' => $slug, 'returnsingle' => true]);
         }
@@ -164,16 +164,6 @@ class Frontend extends ConfigurableBase
 
         // Then, select which template to use, based on our 'cascading templates rules'
         $template = $this->templateChooser()->record($content);
-
-        // Update the route attributes to change the canonical URL generated.
-        if ($content->isHome() && ($template === $this->getOption('general/homepage_template'))) {
-            $request->attributes->add(['_route' => 'homepage', '_route_params' => []]);
-        } else {
-            list($routeName, $routeParams) = $content->getRouteNameAndParams();
-            if ($routeName) {
-                $request->attributes->add(['_route' => $routeName, '_route_params' => $routeParams]);
-            }
-        }
 
         // Setting the editlink
         $this->app['editlink'] = $this->generateUrl('editcontent', ['contenttypeslug' => $contenttype['slug'], 'id' => $content->id]);
--- a/app/theme_defaults/_sub_taxonomylinks.twig
+++ b/app/theme_defaults/_sub_taxonomylinks.twig
@@ -1,5 +1,5 @@
 {% if record.taxonomy is defined %}
-    {% for type, values in record.taxonomy %}
+    {% for type, values in record|taxonomy %}
         <em>
         {% if values|length < 2 %}
             {{ config.get('taxonomy')[type].singular_name }}:
--- a/src/Controller/Base.php
+++ b/src/Controller/Base.php
@@ -88,6 +88,8 @@ abstract class Base implements ControllerProviderInterface
         }
         $context += $globals;
 
+        $this->addResolvedRoute($context, $template->getTemplateName());
+
         if ($this->getOption('general/compatibility/template_view', false)) {
             return new TemplateView($template->getTemplateName(), $context);
         }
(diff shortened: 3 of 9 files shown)


src/Twig/Runtime/CompatRuntime.php
```diff
@@ -0,0 +1,53 @@
...
+            $compiled[$el->getTaxonomytype()]['/' . $el->getTaxonomytype() . '/' . $el->getSlug()] = $el->getSlug();
+        }
+
+        return $compiled;
+    }
+
+    protected function isLegacy()
+    {
+        return $this->appConfig->get('general/compatibility/setcontent_legacy', true);
+    }
```
GawainLynch: Wouldn't it cover more bases to drop this and instead check the type in `taxonomyCompat()`

src/Twig/Extension/CompatExtension.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getFilters()
+    {
+        $safe = ['is_safe' => ['html']];
+        $deprecated = ['deprecated' => true];
+
+        return [
+            new TwigFilter('taxonomy_compat', [Runtime\CompatRuntime::class, 'taxonomyCompat'], $safe + $deprecated)
```
GawainLynch: OK, so digging I have a couple of question:

1. Why not just `|taxonomy`
2. I know with 1. you've marked it up as deprecated, but is it, and what is the way forward?
3. Shouldn't this just go into the `RecordRuntime`

src/Controller/Frontend.php
```diff
@@ -443,6 +443,16 @@ public function template($template)
         return $this->render($template);
     }
 
+    public function isHome($content)
```
GawainLynch: In before @CarsonF :wink: … But shouldn't this either be private or somewhere else. It just feels like more of the historical disasters we're trying to get away from.

src/Twig/Extension/CompatExtension.php
rossriley: I might still have a rethink about how to do this better, but yes, the problem it solves is that calling `record.taxonomy` on the legacy content objects gives a very specific key/value array structure that it's impossible to patch in without ruining the concept of having proper collections for taxonomies.

But at the same time, the taxonomy_links twig templates depend on them being this way, so I can't make it BC, so the next best option as far as I can think is to provide a filter to standardise until a time the theme author can rewrite the Twig.

So for this reason, it's new, but immediately deprecated so that we can remove in 4.0 by which time themes can have been tweaked to use the new structure.

Finally, I just thought it was better not to add the legacy config stuff into existing extensions but I don't mind either way, I started doing it in RecordRuntime and just thought it might be more isolated as a separate extension that we can throw away as soon as it's time has passed.

src/Controller/Frontend.php
rossriley: Where do you reckon? It was in the actual content record before, which is a bit of a nightmare because we would have to inject all the config into the content entities.

That info is all available in the controller and it seems like the sort of thing that can/should be done there because you're using that info to decide which template to show etc.

src/Twig/Extension/CompatExtension.php
GawainLynch: Oh totally in favour of a filter as opposed to trying to make it work as a method on the object, and that was the closest consensus we got to in previous meetings. Deity classes :fire: :+1: 

I was more thinking the naming of the filter, as people are getting increasingly narky at changes we make … that single point is my worry here. I know we are approaching this from "It is deprecated and will change", but I am just wondering after a few weeks of various people getting upset, it might be good to think about how we give them the way it will be to transition to, rather than one change now, and another at the end of the year … I don't have a solution either by the way, just trying to get the conversation happening to take some of this off you and I.

But if I am totally missing the point, please throw a wet rag at me.

src/Controller/Frontend.php
GawainLynch: Yeah, of everything I commented on here, this was the one that made my stomach churn as it is really hard … as above, my single worry is just public API (ab)use and the lack of flexibility in changing if we need to. If it is a helper it I'd guess a protected method in `Base`? … Please see my comment as looking to achieve a consensus with you too

rossriley: cool, that sounds fair enough. 

Only other strategy I can think of since the goal of this is to add some metadata to the request object is to do something with an event that can have the selected record passed in along with the `$app['config']` and `$app['request']` services passed in as dependencies.

If we did that though it would probably have to be up to the controller to trigger the event, unless we already have an event we could hook into.

GawainLynch: >  since the goal of this is to add some metadata to the request object is to do something with an event 

Actually, that's a bloody good idea! 

I'm going to get slapped for the suggestion of location, but `Bolt\EventListener\GeneralListener` has to have the container passed due to needing to not construct `Config` early, so adding `KernelEvents::REQUEST => 'onRequest',` to the `getSubscribedEvents` would be totally feasible

rossriley:
So with a nod to #6985 should I drop the idea of the compatibility step and just start porting any necessary filters and functions into the Records twig extension?

GawainLynch:
> So with a nod to #6985 should I drop the idea of the compatibility step and just start porting any necessary filters and functions into the Records twig extension?

WFM :+1: … That was kinda where I was heading with my review and discussion in Slack. Mainly so we don't have multiple changes that people have to deal with, we can give them something new to grab a hold of and use in 3.4+ and they can have confidence that their hard work won't be lost when they update to 4.x.

rossriley:
ok, so this should be good to go. 

I'm not going to stuff all those filters into this PR, just keep this focused on fixing taxonomy and isHome functionality, but this PR is enough to get a record page rendering identically on both sides of the toggle.

GawainLynch:
Zero stress … I'll go over it in the morning as I think we all deserve to chill the "you know what" out and drink some :beers: right now… 'cause Friday.

Translation: Thanks mate, have one on me! :+1:

PR #6977, opened 2017-09-05
[3.4] Nut command to optimise the configured database

GawainLynch:
Just going to leave this one here for RFC.

I can see it being both useful and dangerous. But I'll leave the call on that for discussion. Also just happy to delete the branch, there is virtually zero work in this one.

Oh, and a note on why the logic is in the Nut class. This is not something we'd otherwise want in the API and there is **zero reason** to ever do this from the UI as we make database calls all over the place and this should only be performed when mutation requests are not being made.

### DIFF
--- a/src/Nut/DatabaseOptimize.php
+++ b/src/Nut/DatabaseOptimize.php
@@ -1,0 +1,139 @@
+<?php
+
+namespace Bolt\Nut;
+
+use Bolt\Collection\MutableBag;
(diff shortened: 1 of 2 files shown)


bobdenotter:
Where is the "vacuum" emoji when you need it? ;-)

Anyhow, I like this. 👍

bobdenotter:
Since there are no objections, I'm merging in. 👍

PR #6978, opened 2017-09-05
Group Concat settings in config.yml

iskyd:
If you have to work with a lot of relations you can't increase max group_concat_max_len cause DoctrineLister overrides mysql value:
`$db->executeQuery('SET SESSION group_concat_max_len = 100000');`
I came up with that problem so I think is useful to put it in config file if someone wants override it.

With love, Shayla <3


bobdenotter:
Until we re-think the config file, shouldn't this be added to `config.yml.dist`?

iskyd:
@bobdenotter, I think it shouldn't because it's a very custom option. If no one has ever noticed that probably the only things to do is document that part.

bobdenotter:
Ok, cool. Let's leave it out, then. :-)

SahAssar:
Hmm, shouldn't we have docs for it though?

GawainLynch:
No objections from me

PR #6982, opened 2017-09-06
[4.0] Phase II: Symfony 4/Flex & Silex 2 project filesystem layout

GawainLynch:
This is Phase II of the changes started in #6739 to remodel the project's internal layout, which looks something like:

```
%site%/
├─ app/
│  ├─ database/
│  └─ translations/
├─ assets/       # Formerly app/src/
├─ bin
│  └─ nut
├─ config/
│  └─ extensions/
├─ public/
│  ├─ bolt/      # Formerly in app/view/
│  │  ├─ css/
│  │  ├─ fonts/
│  │  ├─ img/
│  │  └─ js/
│  ├─ files/
│  ├─ theme/
│  └─ index.php
├─ src/
├─ templates/
│  ├─ bolt/      # Formerly app/view/twig/
│  ├─ defaults/  # Formerly app/theme_defaults/
│  └─ toolbar/   # Formerly app/view/toolbar/
├─ var/
│  ├─ cache/
│  ├─ logs/
│  └─ sessions/
├─ tests/
└─ vendor/
```

For reference this is based around various discussions in #development, and the following upstream:
- [Symfony 4: Directory Structure Updates](http://fabien.potencier.org/symfony4-directory-structure-updates.html)
- [Using Symfony Flex to Manage Symfony Applications: Upgrading Existing Applications to Flex](https://symfony.com/doc/master/setup/flex.html#upgrading-existing-applications-to-flex)

The thinking with respect to applying this the to repository version is centred around consistency with both our installed product, and application structure moving forward.

Let the discussion begin. :smile:


bobdenotter:
screen shot 2017-09-06 at 20 53 02

🙄 

Kidding aside, looks good to me. Happy that under SF 4 we can keep `theme/` with the templates as a whole under `public/`. This will keep frontend devs happy.

GawainLynch:
> Happy that under SF 4 we can keep theme/ with the templates as a whole under public/. This will keep frontend devs happy.

Well, yes, but that is a separate point here. This is just structure for the core layout, all this will still end up installed in `vendor/bolt/bolt/` and we copy what we need, e.g. assets, to where we need.

That said, yeah, while I think it is better to have your Twig in `%site%/templates/` and your web assets in `%site%/public/`, I don't see a point in forcing that on people, rather I think giving them a choice is the fairest solution :smile:

PR #6983, opened 2017-09-06
Add inline variant to textarea field

zomars:
I know this is not a huge deal and it's okay if is not accepted but I like my backend nice and tidy.

captura de pantalla 2017-09-06 a las 16 27 33

### DIFF
--- a/app/view/twig/editcontent/fields/_textarea.twig
+++ b/app/view/twig/editcontent/fields/_textarea.twig
@@ -8,6 +8,7 @@
     required:    field.required|default(false),
     errortext:   field.error|default(''),
     placeholder: field.placeholder|default(''),
+    variant:     field.variant,
 } %}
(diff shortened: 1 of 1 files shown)


GawainLynch:
Should the `postfix` also be inlined then? 
image

I don't actually know, btw, just testing and giving feedback :rofl: Yes, it is true, I **actually** commented on a visual change :scream:

zomars:
From what I saw `prefix` and `postfix` are located outside the `fieldset` element so I don't think those should be affected by this. (Also I just based this PR on how it's made on the `_text.twig` file)

GawainLynch:
Oh zero worries on my part. UI is something I leave to people like @bobdenotter or @SahAssar to approve/merge as I still can't even figure out how to add a `<blink>` tag

PR #6988, opened 2017-09-10
[3.3] Pass boolean string as boolean to YamlUpdater

GawainLynch:
As requested by @ichag :smile: 

This just makes `./app/nut config:set parameter false|true` save a YAML Boolean, not a string

### DIFF
--- a/src/Nut/ConfigSet.php
+++ b/src/Nut/ConfigSet.php
@@ -36,10 +36,11 @@ class ConfigSet extends AbstractConfig
         $value = $input->getArgument('value');
         $backup = $input->getOption('backup');
 
-        $updater->change($key, $value, $backup, true);
-        if (is_bool($value)) {
-            $value = $value ? 'true' : 'false';
-        }
+        $newValue = $value === 'true' || $value === 'false'
+            ? filter_var($value, FILTER_VALIDATE_BOOLEAN)
+            : $value
+        ;
+        $updater->change($key, $newValue, $backup, true);
 
         $this->io->title(sprintf('Updating configuration setting in file %s', $this->file->getFullPath()));
         $this->io->success([


src/Nut/ConfigSet.php
```diff
@@ -36,10 +36,11 @@ protected function doExecute(YamlUpdater $updater, InputInterface $input, Output
...
         $backup = $input->getOption('backup');
 
-        $updater->change($key, $value, $backup, true);
-        if (is_bool($value)) {
-            $value = $value ? 'true' : 'false';
-        }
+        $newValue = $value === 'true' || $value === 'false'
+            ? filter_var($value, FILTER_VALIDATE_BOOLEAN)
+            : $value
+        ;
```
bobdenotter: I don't like this. Assignment, comparisons and a ternary operator in one statement.. Let's keep it separate: 

```
if ($value === 'true' || $value === 'false') {
    $value = filter_var($value, FILTER_VALIDATE_BOOLEAN);
}
$updater->change($key, $value, $backup, true);
```

GawainLynch: You mean like in a large chunk of the rest of the code base.

The thing here is that then it has to become an if/else, or alternatively have an additional if/else on return to switch the boolean back to text for the output

bobdenotter: Right, I missed the part about needing to keep the original `$value`. Yeah, let's not do that. :-)

GawainLynch: Funnily enough it was what took me so long today to get this one to PR … it just seemed the least evil option :disappointed:

Ichag:
Thanks guys! Something like that was it what I had in mind.
I appreciate your support!

PR #6995, opened 2017-09-12
Fixes: Taxonomies get saved as 'slugs'

### LINKED ISSUE #6973: Taxonomies get saved as 'slugs', but not checked as such in the editor.
Details
-------

 - Relevant Bolt Version: 3.2 and up
 - Install type: N/A
 - PHP version: N/A
 - Used web server: N/A
 - For UX/UI issues: N/A

Reproduction
------------

As discovered by Nestor today. If you have a taxonomy like: 

```
groups:
    slug: groups
    singular_slug: group
    behaves_like: grouping
    options:
        main: "Main"
        the_meta_group: "Meta group"
        other: "Other"
    has_sortorder: true
```

And the user selects "Meta group", the value in the editor will be `the_meta_group`, but this will be saved slugified as `the-meta-group`. 
When the page gets edited again, the code to mark items as selected, will compare the taxonomy 
 value `the_meta_group` to the saved `the-meta-group`. This happens in `app/view/twig/editcontent/fields/_grouping.twig`: 

```twig
{% set attr_opt = {
    value:     slug,
    selected:  context.content.taxonomy.containsKeyValue(taxonomy.slug, slug),
} %}
```

and `src/Storage/Collection/Taxonomy.php`

```php
    public function containsKeyValue($field, $value)
    {
        foreach ($this->getField($field) as $element) {
            if ($element->getSlug() == $value) {
                return true;
            }
        }

        return false;
    }
```

Since `element->getSlug()` will always return a slugified version, I propose we change it to: 

```
if ($element->getSlug() == Slugify::create()->slugify($value)) {
```

Either that, or slugify the keys when the taxonomies get read and initialised.

bobdenotter:
Fixes #6973

Fixes: Taxonomies get saved as 'slugs', should also be checked as such in the editor when determining the 'selected' one(s).

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -349,11 +349,9 @@ class Config
             // Make sure the options are $key => $value pairs, and not have implied integers for keys.
             if (!empty($taxonomy['options']) && is_array($taxonomy['options'])) {
                 $options = [];
-                foreach ($taxonomy['options'] as $optionkey => $optionvalue) {
-                    if (is_numeric($optionkey)) {
-                        $optionkey = Slugify::create()->slugify($optionvalue);
-                    }
-                    $options[$optionkey] = $optionvalue;
+                foreach ($taxonomy['options'] as $optionKey => $optionValue) {
+                    $optionKey = Slugify::create()->slugify($optionKey);
+                    $options[$optionKey] = $optionValue;
                 }
                 $taxonomy['options'] = $options;
             }
--- a/src/Storage/Collection/Taxonomy.php
+++ b/src/Storage/Collection/Taxonomy.php
@@ -147,8 +147,8 @@ class Taxonomy extends ArrayCollection
         foreach ($this as $k => $existing) {
             if (
                 $existing->getContent_id() == $entity->getContent_id() &&
-                $existing->getTaxonomytype() == $entity->getTaxonomytype() &&
-                $existing->getSlug() == $entity->getSlug()
+                $existing->getTaxonomytype() === $entity->getTaxonomytype() &&
+                $existing->getSlug() === $entity->getSlug()
             ) {
                 return $existing;
             }
@@ -167,7 +167,7 @@ class Taxonomy extends ArrayCollection
     public function getField($fieldname)
     {
         return $this->filter(function ($el) use ($fieldname) {
-            return $el->getTaxonomytype() == $fieldname;
+            return $el->getTaxonomytype() === $fieldname;
         });
     }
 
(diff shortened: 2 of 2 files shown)


lenvanessen:
Pulled it in and works here:)

bobdenotter:
> This function never gets called,

It's used twice: 

https://github.com/bolt/bolt/blob/3.3/app/view/twig/editcontent/fields/_categories.twig#L9
https://github.com/bolt/bolt/blob/3.3/app/view/twig/editcontent/fields/_grouping.twig#L35

> It is handled above in setFromPost() on save. 

As explained in the issue, that is the exact problem. It's handled there, causing it to be inconsistent. 

> If existing entries are broken they should be fixed themselves.

It's not for _existing_ records only. It's reproducible on a fresh install. 

> … and OMG 🔥 the loose comparison

I did notice that, but it was consistent with the rest of that file. Will fix them all.

GawainLynch:
So we introduce **another** random change "somewhere" in the execution cycle … See where I am going?

If we normalise on config load it is consistent everywhere, no? :wink:

bobdenotter:
> If we normalise on config load it is consistent everywhere, no? 😉

Yes, but the impact of that is much larger. Do you want us to make those changes in a 0.0.1 release, knowing that it's probably going to change in 4.0 regardless?

GawainLynch:
> Yes, but the impact of that is much larger.

Not at all, it just doesn't introduce behaviour in one place that doesn't occur in another. It is a "key" value that gets persisted as a slugified value, retrieved as persisted, used as persisted. This PR would mean that it is then only compared in **one location** as a slugified value. BC breaking **and** inconsistent to how it will work anywhere else.

> Do you want us to make those changes in a 0.0.1 release, 

This already is making behaviour modifying and BC breaking changes in a 0.0.1 (patch) release :stuck_out_tongue_winking_eye:

GawainLynch:
> Let's change `extend` to `extensions`, we're the only one's that use it!

How well did that work our for us as an example?

bobdenotter:
> This already is making behaviour modifying and BC breaking changes in a 0.0.1 (patch) release 😜

In that case you can consider _any_ bugfix to be a breaking change.

GawainLynch:
> In that case you can consider any bugfix to be a breaking change.

No, and that is semantics. This approach might "fix" a bug, but it immediately introduces a known regression bug to replace it.

bobdenotter:
Ok, i give up. just tell me what you want changed, and i'll work on that.

GawainLynch:
Remove the conditional here https://github.com/bolt/bolt/blob/v3.3.3/src/Config.php#L352-L354

What is even more interesting, there is a second corner-case regression that slugifying this anywhere triggers … I was testing with `the_meta_group` which gets slugified to `meta-group` as Slugify regexes out functionals & articles from a sentence. Two regressions for the price of one :wink:

PR #6996, opened 2017-09-12
[3.4] Don't check for changes in dynamically inserted elements

### LINKED ISSUE #6991: [BUG] Browser nags about loosing form data when saving page
Details
-------

 - Relevant Bolt Version: [ 3.5 ]
 - PHP version: [ 7.1 ]
 - Used web server: [ Internal PHP server ]
 - For UX/UI issues: [ Chrome 60 ]

Reproduction
------------
Install Bolt 3.5 alpha6 through composer and try to save a new page or a existing one.

### Bug summary
When saving a new page the browser will prompt you for the following confirmation.

screen shot 2017-09-12 at 10 44 35

lenvanessen:
Fixes #6991

An input element is inserted on save, causing the form to think it's changed, and thus preventing the save. Needs sanity check from @GawainLynch since the editcontent.js is long and hard and takes more brainpower than I got left

### DIFF
--- a/app/src/js/modules/editcontent.js
+++ b/app/src/js/modules/editcontent.js
@@ -74,14 +74,16 @@
         var changes = 0;
 
         $('form[name="content_edit"]').find('input, textarea, select').each(function () {
-            if (this.type === 'textarea' && $(this).hasClass('ckeditor')) {
-                if (ckeditor.instances[this.id].checkDirty()) {
-                    changes++;
-                }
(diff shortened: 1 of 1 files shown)


GawainLynch:
> Needs sanity check from @GawainLynch since the editcontent.js is long and hard and takes more brainpower than I got left

Hahaha that file needs more brain power than we've all got left, combined :rofl:

PR #6998, opened 2017-09-13
Added: Add support for "Singleton" ContentTypes

lenvanessen:
Working off of the proposed solution here https://github.com/bolt/bolt/pull/6309
The twig logic kinda got me off track a bit, but here is what i've come up with so far. Would love some feedback.

Update:
So moved away the logic from the Twig templates, and added some extra features like:
- Don't go back to overview when you delete the singleton, but to create new. You don't need the overview for a singleton
- Still just a PR about aligning the UI with the singleton principle
- Removed the singleton from the recently edited section

Lots of love for koalakiller @GawainLynch for the help

### DIFF
--- a/app/view/twig/_nav/_secondary-content.twig
+++ b/app/view/twig/_nav/_secondary-content.twig
@@ -16,6 +16,8 @@
         {# Recently edited records for this ContentType #}
         {% if menu.has('recent') %}
             {% for record_menu in menu.get('recent').children %}{% set sub = sub|merge([record_menu]) %}{% endfor %}
+        {% elseif menu.has('singleton') %}
+            {% set sub = sub|merge([menu.children|first]) %}
         {% endif %}
 
         {{ nav.submenu(menu.icon, menu.label, sub, active, true) }}
--- a/src/Controller/Backend/Records.php
+++ b/src/Controller/Backend/Records.php
@@ -50,6 +50,10 @@ class Records extends BackendBase
         $duplicate = $request->query->getBoolean('duplicate');
         $new = $duplicate ? false : empty($id);
 
+        // Override return redirect for singletons
+        $isSingleton = $this->getOption('contenttypes/' . $contentTypeKey . '/singleton');
+        $deleteRoute = $isSingleton ? 'editcontent' : 'overview';
+
         // Test the access control
         if ($response = $this->checkEditAccess($contentTypeKey, $id)) {
             return $response;
@@ -76,7 +80,7 @@ class Records extends BackendBase
             if ($button->getName() === 'delete') {
                 $this->app['storage.request.modify']->action($contentTypeKey, [$id => ['delete' => true]]);
 
-                return $this->redirectToRoute('overview', ['contenttypeslug' => $contentTypeKey]);
+                return $this->redirectToRoute($deleteRoute, ['contenttypeslug' => $contentTypeKey]);
             } else {
                 $response = $this->recordSave()->action($formValues, $contentType, $id, $new || $duplicate, $returnTo, $editReferrer);
             }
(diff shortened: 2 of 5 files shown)


SahAssar:
Is the idea here to make singleton only affect the UI or to actually make it be a part of the permission structure that would allow/disallow anyone to create a record?

bobdenotter:
Looks like it's UI only for now. I wonder if it's really necessary to limit it in code. If not, it'll allow non-destructive toggling on and off. (ie, existing records wouldn't get deleted, just hidden)

lenvanessen:
@bobdenotter for now only UI yes, I updated the TODO with the permissions question. If more people feel that we need to add a bit more checks (e.g. permissions, limit the total amount of accepted entries etc). I'd be more than happy to look into that! 

I'm leaning a bit more to Svante's side on this, if you define a contenttype to be a singleton i'd expect that really means one single entry to be allowed (Besides the UI changes that make it more intiutive)

app/view/twig/_nav/_macros.twig
```diff
@@ -35,13 +34,20 @@
         <li{% if class %} class="{{ class }}"{% endif %}>
             {% if allowedamount == 1 and not force_submenu|default(false) %}
                 {% set item = allowedsingle %}
+
                 <a href="{{ item.uri }}">
-                    {{ icon(item.icon, "icon") }}{{ item.label|default("<em>(" ~ __('general.phrase.no-content') ~ ")</em>")|raw }}
+                    {{ icon(item.icon, "icon") }}{{ item.label|default("<em>(" ~ __('general.phrase.no-content') ~ ")</em>")|raw }}1
```
GawainLynch: Is that `1` a stray?

app/view/twig/_nav/_macros.twig
```diff
@@ -8,8 +8,7 @@
  # @param bool    wide
  # @param array   subitems
  #}
-{% macro submenu(icon, label, popoveritems, active, wide, subitems, force_submenu) %}
-
+{% macro submenu(icon, label, popoveritems, active, wide, subitems, force_submenu, singleton) %}
```
GawainLynch: Don't move _more_ logic into Twig, this whole macro needs desperately to :boom: :koala: :fire: 

_**Hint:**_ Have a look at what `subitems` contains

app/view/twig/_nav/_secondary-content.twig
```diff
@@ -7,18 +7,38 @@
...
+            {%  if records|length == 1 %}
+                {% set path =  path('editcontent', {contenttypeslug: name, id: records|first.id}) %}
+                {#{% set has_records = true %}#}
+            {% else %}
+                {% set path = menu.get('new') %}
+            {% endif %}
+
+            {% set singleton = { 'uri': path, 'icon': menu.icon, 'label': menu.label } %}
+        {% endif %}
+
```
GawainLynch: As above, this entire file's logic is better handled in PHP … `Bolt\Menu\MenuEntry` & `Bolt\Menu\Builder\AdminContent` might give you more hints

src/Controller/Backend/Records.php
```diff
@@ -50,6 +50,14 @@ public function edit(Request $request, $contenttypeslug, $id)
         $duplicate = $request->query->getBoolean('duplicate');
         $new = $duplicate ? false : empty($id);
 
+        // Override return redirect for singletons
+        $deleteRoute = 'overview';
+        $isSingleton = $this->app['config']->get('contenttypes')[$contentTypeKey]['singleton'];
```
GawainLynch:
```
$this->getOption('contenttypes/' . $contentTypeKey . '/singleton');
```

GawainLynch:
```
$deleteRoute = $isSingleton ? 'editcontent' : 'overview';
```

GawainLynch:
Awesome work mate!

bobdenotter:
Pretty cool, @lenvanessen! 👍

PR #7001, opened 2017-09-14
Migrate pre-fill builder to use Bags

GawainLynch:
Nut's pre-fill command was throwing an exception on PHP 7.2 … so while I was there :wink:


bobdenotter:
Whoop! :shipit:

PR #7010, opened 2017-09-19
Add `is_numerical` check - See #6995

xiaohutai:
See #6995 .

The numerical check is still required because you can define taxonomies as:
```
- foo
- bar
- baz
```
and
```
foo: Foo
bar: Bar
baz: Baz
```

The code before this pull, would actually save it as:
```
0: foo
1: bar
2: baz
```

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -332,13 +332,13 @@ class Config
 
         foreach ($taxonomies as $key => $taxonomy) {
             if (!isset($taxonomy['name'])) {
-                $taxonomy['name'] = ucwords($taxonomy['slug']);
+                $taxonomy['name'] = ucwords(str_replace('-', ' ', Common\Str::humanize($taxonomy['slug'])));
             }
             if (!isset($taxonomy['singular_name'])) {
                 if (isset($taxonomy['singular_slug'])) {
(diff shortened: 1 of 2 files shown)


src/Config.php
```diff
@@ -356,6 +356,9 @@ protected function parseTaxonomy()
             if (!empty($taxonomy['options']) && is_array($taxonomy['options'])) {
                 $options = [];
                 foreach ($taxonomy['options'] as $optionKey => $optionValue) {
+                    if (is_numeric($optionKey)) {
+                        $optionKey = $optionValue;
+                    }
```
GawainLynch: Oh wait, that will need to go below the line, um, below this … as the value should be slugged.

xiaohutai:
In the table, you save a slug and a value. The value should not be slugified, right?
The slug should be either the (slugified) slug or the slugified value (if no key is set explicitly)

src/Config.php
GawainLynch: Tests say you were right :wink:

GawainLynch:
> In the table, you save a slug and a value. The value should not be slugified, right?

Yeah, I was getting lost in the sea of logic in that method … hence the tests :wink:

PR #7011, opened 2017-09-19
Pass in the default orderby to a filter on the select fieldtype

### LINKED ISSUE #7006: [BUG] Select populated with contenttype loses sorting when filtering
Details
-------

 - Relevant Bolt Version: 3.3.3
 - PHP version: 7.0
 - Used web server: Apache version [2.x, my hosting doesn't expose this, though I could probably retrieve it if it is crucial]

Reproduction
------------

### Bug summary

When adding a select field to a contenttype, populating it with another contenttype's data, filtering it, and then setting sort to something, the sorting isn't (apparently) done.

### Steps to reproduce
Add these two contenttypes:

```
events:
    fields:
        title:
            type: text
            required: true
        location:
            type: text
        slug:
            type: slug
            uses: title
        date:
            type: date
        time:
            type: text
        description:
            type: html
        ext_url:
            type: text

newsletters:
    fields:
        subject:
            type: text
            required: true
        slug:
            type: slug
            uses: subject
        body:
            type: html
            required: true
        senddate:
            type: date
            required: false
        events:
            type: select
            values: events/date,title
            filter: { date: '>yesterday' }
            sort: date
            multiple: true
```

### Expected result

Find a select field for Newsletters, which is populated with Events with a 'date'-valuedated in the future, ordered by the same date field.

### Actual result
image
The field isn't ordered by the date-field. However it does get sorted when you comment out the filter.

rossriley:
Fixes #7006

This just passes along the default value of order through to a filter unless you set the order your self in the filter.

### DIFF
--- a/src/Form/Resolver/Choice.php
+++ b/src/Form/Resolver/Choice.php
@@ -152,21 +152,25 @@ final class Choice
             }
         }
 
-        if ($orderBy !== null) {
-            if (substr($orderBy, 0, 1) === '-') {
-                $orderBy = [substr($orderBy, 1), 'DESC'];
-            } else {
-                $orderBy = [$orderBy, 'ASC'];
-            }
-        } else {
-            $orderBy = [$queryFields[0], 'ASC'];
-        }
-
         $values = [];
         if ($filter === null) {
+
+            if ($orderBy !== null) {
+                if (substr($orderBy, 0, 1) === '-') {
+                    $orderBy = [substr($orderBy, 1), 'DESC'];
+                } else {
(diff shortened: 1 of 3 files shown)


GawainLynch:
@bobdenotter Sorry to throw this one over the fence, but when testing I hit a subtle bug that sent me around in circles, then I realised that I had done things the way I did  'cause 3.2 and the intent was to clean this up for 3.3 … so I did.

Coverage is now complete again, and sorting is covered too. Given that both Ross and I have had our hands on this, I pass the baton to you :wink:

PR #7012, opened 2017-09-20
[3.4] Refactor Import / Export

bobdenotter:
This PR makes some changes to the import/export mechanism:

 - Adds a "header" section with some meta information
 - Doesn't break if we're importing a ContentType without a field named 'title'
 - removes the "This is experimental" warning. 

The meta information looks like this: 

```yaml
-
    __bolt_export_meta:
        date: '2017-09-20T14:56:03+02:00'
        version: '3.5.0'
        platform: mysql
```


SahAssar:
Just to be clear: this basically makes `bolt_meta_information` a reserved CT slug, right?

bobdenotter:
@SahAssar Yes. I didn't want to change the entire format of the file, and `bolt_meta_information` seemed like a safe bet.

GawainLynch:
Might be worth prefixing the name with a double underscore `__` like we do with the YAML repeated nodes, as that is far less likely to conflict with long tail use.

bobdenotter:
> Might be worth prefixing the name with a double underscore __

👍

PR #7030, opened 2017-09-27
Singletons in submenus need to skip "overview" too.

bobdenotter:
This is the last bit to make Singletons fully functional in a grouped menu. They were basically working, except the linked to the "overview" page, instead of the first/new record. See Screenshot:

screen_shot_2017-09-27_at_16_17_07


GawainLynch:
image

PR #7032, opened 2017-09-28
Adding font sizes to headings in ckeditor

peterboorsma:
Ckeditor gives no styling to headings in the WYSIWYG editor and leaves that to the users' browser. This causes headings in some browsers to be displayed smaller than paragraph text. 
I added font sizes for h1, h2, h3, h4, h5, h6 tags. This improves readability and visual hierarchy in WYSIWYG fields.

### DIFF
--- a/app/view/css/ckeditor-contents.css
+++ b/app/view/css/ckeditor-contents.css
@@ -61,10 +61,45 @@ ol,ul,dl
 	padding: 0 40px;
 }
 
+p, li
+{
+	font-size: 14px;
+}
+
 h1,h2,h3,h4,h5,h6
 {
 	font-weight: normal;
 	line-height: 1.2;
+}
+
+h1
+{
+	font-size: 26px;
+}
+
+h2
+{
+	font-size: 24px;
+}
+
+h3
+{
+	font-size: 22px;
+}
+
+h4
+{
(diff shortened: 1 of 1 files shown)


GawainLynch:
List numbers match heading tags:
image

:+1: :+1: :+1: :+1: :+1:

peterboorsma:
Oh. Maybe the p and li tags also need a font-size.

SahAssar:
Shouldn't this be set using em/rem instead?

bobdenotter:
The other settings in that file were already in PX, so i'd think it's fine. Maybe someday we'll convert all of it. Probably CKeditor 5 will come with a new version of that entirely. :-)

app/view/css/ckeditor-contents.css
```diff
@@ -61,12 +61,47 @@ ol,ul,dl
 	padding: 0 40px;
 }
 
+p, li
+{
+	font-size: 12px;
```
bobdenotter: Should be `14px`. ;-)

("inspected" it in the before situation. It was 14px, and 12 would make it too small)

peterboorsma:
@SahAssar I used the units already used in the ckeditor scss. But I do agree with you :)

PR #7033, opened 2017-09-28
[3.3] Add Legacy Content Type to Entity

### LINKED ISSUE #7029: Argument 2 passed to Edit::action()
Type error: Argument 2 passed to Bolt\Storage\ContentRequest\Edit::action() must be an instance of Bolt\Storage\Mapping\ContentType, string given, called in vendor/bolt/bolt/src/Controller/Backend/Records.php on line 94

Content::getContenttype() may return a string but Edit::action() requires ContentType object.

Simple solution is to create this object and pass it through, but it is marked as @derecated so probably all usages of this object have to be converted to array (which might be simple since ContentType extends ArrayAccess).

JarJak:
fixes #7029

### DIFF
--- a/src/Controller/Backend/Records.php
+++ b/src/Controller/Backend/Records.php
@@ -91,6 +91,10 @@ class Records extends BackendBase
 
         // We're doing a GET
         $duplicate = $request->query->get('duplicate', false);
+
+        // Ensure custom entities have the legacy ContentType set
+        $this->app['storage.legacy_service']->setupContenttype($content);
+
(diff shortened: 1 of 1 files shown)


GawainLynch:
Legendary! Thanks mate :+1:

GwendolenLynch (approved):

JarJak:
Would you release 3.3.6 with this patch?

GawainLynch:
3.3.6 will include this yes. But please bare in mind that there is a long process to a release, and a single fix is rarely cause for one. That said, we are doing triage for a release soon, probably before Tuesday.

Issue #5676, opened 2016-08-12
Select2 (omnisearch, tags, multiselects) broken in IE10 and 11


SahAssar:
The omnisearch is currently broken in IE10 and IE11, probably because of upstream issues in select2.

I didn't get too far in my testing yet, but reverting select2 to 4.0.0 fixes the issue, while 4.0.1, 4.0.2 and 4.0.3 all seem broken in IE.

Edge seems to still work though.

SahAssar:
The tag fields and searching in categoryfields with multiple: true are also broken... so basically anything that does anything even mildly complex with select2.

bobdenotter:
Should we perhaps revert to Select2 4.0.0 for 3.1 and 3.0.x for now?

SahAssar:
@bobdenotter we talked about that in IRC, but that might introduce other issues since a large number of issues were fixed between 4.0.0 and 4.0.3... In my testing it didn't, but we'd need a more serious cross-browser test to know for sure.

SahAssar:
In my shallow testing everything seems to work when using 4.0.0 in IE 10 and 11. If we (ping @rarila) are okay with that solution as long as it does not break anything in other browsers I can test the rest and if it works perhaps we'll want to revert to 4.0.0 for now.

IE11 is about 4.2% usage, so it's not a disaster, but still something worth fixing.

SahAssar:
Oh, and thanks to @Largo for finding this (he found it in bacbos/bolt-menu-editor#33)

rarila:
We only have to follow upstream so that we can bump again when it's fixed or when IE explodes and vanishes forever under heavy lighning. — whatever happens earlier.

bobdenotter:
Any progress on this?

SahAssar:
@bobdenotter Well, no... Either we revert and might face bugs because of that (since it's used in quite a few places that use it differently), or we keep the current version and it's broken on IE10/11.

I'm not sure which one is more preferable, so I haven't made a PR to revert or closed this...

GawainLynch:
OK… "*hard*" question … what do we support, browser-wise? … EoS

SahAssar:
I think that when I asked that question a year or so ago the answer was IE10+... I think we can drop IE10 (0.3% global usage), but IE11 (4.3% global usage) might be harder to drop.

GawainLynch:
Yeah, figured. Thanks @SahAssar

rossriley:
[...]

PR #7048, opened 2017-10-03
[3.4] Select2 4.0.4

GawainLynch:
Select2 4.0.4 was released late last month, and given nobody else got to this **and** it really irks me to be using a 30 month old release of anything with JavaScrap in it … :coffee: task! 

Given we're about to minor, and we're in beta, now seems like a time to take advantage and all.

@bobdenotter I *think* this doesn't have the problem we were seeing, but if it does please just close this and we'll `<insert rant here>`


bobdenotter:
We had it locked on 4.0.0, because 4.0.3 didn't work in IE. It was utterly broken. I'll spin up the ol' Internet Explorer today, and see if they've gotten their heads out of their asses and fixed it.

GawainLynch:
> IE ~. It was~ is utterly broken

Fixed the text for you :stuck_out_tongue_winking_eye:

bobdenotter:
TBH, IE 11 isn't nearly as bad as what came before, and Edge is fairly decent. Don't care to much about the browser's UI, but no real problems with the rendering engine.

GawainLynch:
Yeah, in my mind Edge !== IE … and regardless, the whole thing is just low hanging pineapples to make fun of irrespective of fact.

bobdenotter:
For reference: #5676

bobdenotter:
Well, that was a morning well spent. 

Unfortunately, the 4.0.4 version does not resolve the problem we have on IE 10 and IE 11 with the omnisearch. 

screen shot 2017-10-03 at 12 32 08

GawainLynch:
C'est la vie

bobdenotter:
Let's make sure we get rid of it in V4. ;-)

PR #7054, opened 2017-10-04
Add logs to profiler

JarJak:
This PR fulfills empty Logs tab in Profiler, using default Symfony LoggerCollector. It should have no impact on performance since logging is disabled by default and then NullLogger is used.

### DIFF
--- a/src/Provider/ProfilerServiceProvider.php
+++ b/src/Provider/ProfilerServiceProvider.php
@@ -8,6 +8,7 @@ use Bolt\Profiler\DebugToolbarEnabler;
 use Doctrine\DBAL\Logging\DebugStack;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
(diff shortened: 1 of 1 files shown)


GawainLynch:
This is already added by Silex's [`WebProfilerServiceProvider`](https://github.com/silexphp/Silex-WebProfiler/blob/1.0/WebProfilerServiceProvider.php#L84)

JarJak:
Interesting, it's not working for me... maybe because it is 'logs' instead of 'logger'?

GawainLynch:
Hmm … what are you seeing as the difference?

JarJak:
Now I think it was just my bad, I've deleted changes and it still works :) But I will come back if the problem will occur again.

GawainLynch:
Nah, it wasn't you … it was the naughty :koala: sneaking in and changing things when you weren't looking :wink:

PR #7057, opened 2017-10-06
Docker container for Bolt

BerkhanBerkdemir:
I  wanted help, and I made a container, for your team. Just you need docker server (If you don't how you will use docker server, I can explain easiest way)

This container build with Ubuntu16.04, PHP7, Apache2, Composer, Git.

Actually I added row for jetbrains in .gitignore.

### Usage

`cd bolt`
Build the container
`docker build -t <containername> .`
Run the container
`docker run -d -p 80:80 <containername>`

This container for development, and I didn't try this with SQLite 3, but I have problem with routing.

If I run container everthing is OK, but If I check in browser apache return 404 error. PHP redirect to me <hostname>/bolt/userfirst URL. If I change URL <hostname>/index.php/bolt/userfirst everthing fine.

So, What should I do this status. Should I delete .htaccess when I import conf file for v-host?

### TO-DO:
- [ ] Seprate container for Production, Development and Testing
- [ ] Will add PostgreSQL and MySQL support
- [ ] You can use local SQLite file or you can save SQLite file your local machine.
- [ ] You can choice RDBMS type.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,8 +61,9 @@ RUN mv /home/bolt /var/www/html && \
     chown www-data:www-data -R /var/www/html && \
     chmod 755 -R /var/www/html
 
-# Restart Apache2 HTTP Web Server
-RUN apache2ctl restart
+# Rewrite module enabled
+RUN a2enmod rewrite && \
+    apache2ctl restart
 
 # EDIT VIRTUALHOST OR MOVE TO /VAR/WWW/HTML FILES FROM BOLT FILE
 


.gitignore
```diff
@@ -54,3 +54,5 @@ Vagrantfile
 *.lock
 .buildpath
 
+# JetBrains .idea folder
+.idea/*
```
GawainLynch: This should be in a global ignore, not the repo's

GawainLynch:
I'm not convinced we want this in core TBH. This approach is only valid for those people working on the core of Bolt itself … building web sites this way, it is specifically not supported.

bobdenotter:
There are a bunch of people who use Docker to build sites _*with*_ Bolt, who would benefit from something like this. But, i don't think it should be in this repo, which is exclusively for _building Bolt_ itself. 

Maybe `bolt/composer-install` would be a better place? That way people could use `composer create project` and get a basic setup, together with a Docker file. I don't know enough about Docker to judge if that's how it usually goes, or if it's "better practice" to get Docker first, and then get a Bolt setup in that.

GawainLynch:
> There are a bunch of people who use Docker to build sites with Bolt, who would benefit from something like this
> `[snip]`
> Maybe bolt/composer-install would be a better place?

Yes, like the one that @rossriley has maintained for the last 3+ years — https://github.com/rossriley/docker-bolt

bobdenotter:
Perhaps we should have a page in the official docs for that one? Maybe even with additional functionality from this PR, and then elevate that to an official way to use Bolt with Docker?

GawainLynch:
> Perhaps we should have a page in the official docs for that one? Maybe even with additional functionality from this PR, and then elevate that to an official way to use Bolt with Docker?

There is less functionality in this PR, and I don't know we want to add another supported install method

bobdenotter:
Perhaps not as a "install method" but as a "HOWTO" page on this workflow?

GawainLynch:
> but as a "HOWTO" page on this workflow?

That still makes it our problem, and we made a very deliberate team decisions earlier this year to cast off all that stuff as it just accumulated dust eventually.

JarJak:
Is this docker image https://github.com/rossriley/docker-bolt official? Is it mentioned in official docs as another way to install (even simpler than with composer)?

SahAssar:
@JarJak It's as official as it gets for a docker image for bolt. Looking at the conversation above we will not provide an "proper" official docker image.

This seems like a good middle ground for me, anyone against closing this?

GawainLynch:
> This seems like a good middle ground for me, anyone against closing this?

No objections.

Thanks @BerkhanBerkdemir for the work nonetheless :smile:

PR #7067, opened 2017-10-10
Fixes for `title_format`

### LINKED ISSUE #6870: Exception thrown when title_format uses multiple select field
Well, actually I don't know what would be the normal behaviour, but sure I didn't expect it to be an error.

To sum up : thanks to recent improvements, I was able to use a select field in title_format... until I change "multiple" attribute of my select field.

Suggestion : instead of throwing an error, perhaps just don't use select field in title if ther is more thant one item selected.

Details
-------

 - Relevant Bolt Version:  3.3

Reproduction
------------

 1. **Bug summary**: 
    * title_format can use select field type
    * but if multiple attribute of this field type is set to true, an error is thrown in backend listing
 2. **Specifics**:
    * Here is the error message :
` Twig_Error_Runtime in Template.php line 447:
An exception has been thrown during the rendering of a template ("Warning: strip_tags() expects parameter 1 to be string, array given") in "@bolt/_nav/_secondary-content.twig" at line 28. `
 3. **Steps to reproduce**:
    * Create a new Contenttype
    * Add a field with type "select"
    * Add attribute "multiple: true" to this field
    * Add attribute "title_format: [selectfieldname] to contenttype
    * Update database ad go to backend listing of your new contenttype
 4. **Expected result**: 
    * If multiple is not set to true, listing throws no error and title format uses select field normally
 5. **Actual result**: 
    * When multiple is set to true, an error is thrown

### LINKED ISSUE #7064: [3.4] `title_format` doesn't work.
The `title_format` doesn't work anymore on 3.4.

Details
-------

 - Relevant Bolt Version: [  3.4  ]

Reproduction
------------
Set up a content type like: 

```yml
dummies:
    name: Dummies
    singular_name: Dummy
    fields:
        title:
            type: text
            class: large
            group: content
        slug:
            type: slug
            uses: title
        foo:
            type: text
        selector:
            type: select
            multiple: false
            postfix: "Select your favourite turtle(s)."
    title_format: [ foo, selector ]
```

On 3.3 it shows: 

screen shot 2017-10-09 at 19 47 07

(with 'foo' set to 'Bar', and selector set to 'Donatello')

Bumping it to 3.4 gives: 

screen shot 2017-10-09 at 19 46 18

bobdenotter:
Fixes #6870. Fixes #7064.

This PR makes `title_format` work for the new storage layer, as it did for the old one. The code in `ContentTypeTitleTrait::getTitleColumnNames` is mostly adapted from `ContentValuesTrait::getTitleColumnName`.

I also modified the old implementation in `ContentValuesTrait`, to fix #6870 for places where the old storage layer is used. (like in the "Last modified" block in the right aside.

### DIFF
--- a/app/view/twig/_macro/_macro.twig
+++ b/app/view/twig/_macro/_macro.twig
@@ -28,7 +28,7 @@
  #}
 {% macro contentlink(contenttype, content) %}
     {% import _self as crosslinks %}
-    {% set title = content.getTitle(true) %}
+    {% set title = content.getTitle() %}
     {{ crosslinks.contentlink_by_id(contenttype, title, content.id) }}
 {% endmacro %}
 
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -334,13 +334,21 @@ class Content extends Entity
      */
     public function getTitle()
     {
-        if (array_key_exists('title', $this->_fields)) {
-            return $this->_fields['title'];
+        $fields = $this->_fields;
+
+        if (array_key_exists('title', $fields)) {
+            return $fields['title'];
         }
 
-        $fieldName = $this->getTitleColumnName($this->contenttype);
+        $fieldNames = $this->getTitleColumnNames($this->contenttype);
 
-        return $this->$fieldName;
+        $title = [];
+        foreach ($fieldNames as $fieldName) {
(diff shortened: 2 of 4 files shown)


GawainLynch:
> like in the "Last modified" block in the right aside.

Only if you haven't got legacy storage disabled :wink:

PR #7068, opened 2017-10-10
Use 'singular_name' instead of slug.

### LINKED ISSUE #7066: [3.4] Bottom save button on records is different to the one on the side
In the editing page for a record, the save button at the bottom has different UX to that on the side widget. This is especially prevalent with singleton records.

Details
-------

 - Relevant Bolt Version: 3.4-beta2
 - Install type: Composer Install
 - PHP version: 7.1
 - Used web server: Apache

Reproduction
------------

### Steps to reproduce

Create a new Bolt site, add any content type (make it a singleton for dramatic effect). For example

```
home_pages:
    name: Home Page
    singular_name: Home Page
    fields:
        about_heading:
            label: About Heading
            type: text
    singleton: true
    default_status: published
    sort: -id
    icon_many: "fa:home"
```

### Expected result

The save button on the bottom behaves exactly the same as the one on the widget on the right.

### Actual result

The one on the bottom is different, it allows me to create new singleton records with the dropdown arrow, and its label is the `singular_slug` for the contenttype, not its `singular_name`.

image

bobdenotter:
Fixes #7066

### DIFF
--- a/src/Controller/Backend/Records.php
+++ b/src/Controller/Backend/Records.php
@@ -63,7 +63,7 @@ class Records extends BackendBase
         $contentType = $this->getContentType($contentTypeKey);
 
         $data = null;
-        $options = ['contenttype_name' => $contentTypeKey];
+        $options = ['contenttype_name' => $contentType['singular_name']];
         /** @var Form $form */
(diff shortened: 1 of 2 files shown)


bobdenotter:
Ha, will update tests in the Morning. 😇

PR #7071, opened 2017-10-12
[3.4] Composer script logic for default site bundle

GawainLynch:
Discussion: https://github.com/bolt/composer-install/pull/28

### DIFF
--- a/src/Nut/Init.php
+++ b/src/Nut/Init.php
@@ -5,10 +5,11 @@ namespace Bolt\Nut;
 use Bolt\Translation\Translator as Trans;
 use Bolt\Version;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\Output;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
- * Nut command to output phpinfo().
+ * Nut command to perform initial setup tasks.
  */
 class Init extends BaseCommand
 {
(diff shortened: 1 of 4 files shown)


bobdenotter:
I like it.. From what I glean by looking at it, this would fit well with how I personally see people using bolt most (getting the distro, running composer for updates or just using `composer create-project`).

src/Nut/Init.php
```diff
@@ -38,4 +48,21 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+    /**
+     * Archive distributions have the site bundle files installed as .dist
+     * files to prevent the real files being overridden. If the .dist file
+     * exists, but the original doesn't then we should rename them.
+     */
+    private function updateDistBundles()
+    {
+        $fs = $this->app['filesystem']->getFilesystem('root');
+        $paths = ['.bolt.yml', 'composer.json', 'composer.lock', 'src/Site/SiteBundleLoader.php'];
```
xiaohutai: Don't you prefer:
```
$paths = [
    '.bolt.yml',
    'composer.json',
    'composer.lock',
    'src/Site/SiteBundleLoader.php',
];
```

GawainLynch: Haha … Payback is the proverbial … yeah, probably. I don't know that this is **exactly** what we want either way, just trying to keep the conversation moving right now :wink:

PR #7084, opened 2017-10-14
Fixed: Handle images that aren't found a bit more gracefully.

### LINKED ISSUE #7080: `{{ image('foo.bar') }}` gives an error, should not do that.
For example: 

```
{{ image('foo.bar', 100, 100) }}
```

Outputs: `thumbs/100x100c/foo.bar`

```
{{ image('') }}
```

outputs nothing. 

Currently `{{ image('foo.bar') }}` gives an error like below. It should do either of the above two. 

screen shot 2017-10-13 at 16 08 59

bobdenotter:
Fixes #7080.

Before, if debug was disabled, it would just show the "whoops something went wrong", which is not very user friendly. Now it falls back to the pleasant 404 image, while maintaining the exception for when you're in development (debug) mode.

### DIFF
--- a/src/Twig/Runtime/ImageRuntime.php
+++ b/src/Twig/Runtime/ImageRuntime.php
@@ -3,6 +3,7 @@
 namespace Bolt\Twig\Runtime;
 
 use Bolt\Config;
+use Bolt\Filesystem\Exception\FileNotFoundException;
 use Bolt\Filesystem\Handler\ImageInterface;
 use Bolt\Filesystem\Handler\NullableImage;
 use Bolt\Filesystem\Manager;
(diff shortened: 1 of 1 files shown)


src/Twig/Runtime/ImageRuntime.php
```diff
@@ -74,13 +75,22 @@ public function image($fileName = null, $width = null, $height = null, $crop = n
...
-        $file = $this->filesystemMatcher->getFile($fileName);
-
-        $url = $file->url();
+        try {
+            $file = $this->filesystemMatcher->getFile($fileName);
+            $url = $file->url();
+        } catch (FileNotFoundException $e) {
+            // If a non-existing filename is given, we either re-throw the Exception (in DEV) or defer to the
+            // Thumbnailer, so the user sees the "404 image".
+            if ($this->config->get('general/debug')) {
```
CarsonF: It would be better to use `$twigEnv->isDebug()` instead here.

GawainLynch: See #7122

PR #7092, opened 2017-10-16
Only pass user's ID into Users::addRole()

kioan000:
Resolved small bug that occurs when first user is created via app/nut command without Root permissions.
At line 458, `addRole()` is aimed to accept a user id (int or string), the whole user object as array was passed instead causing a crash at line 234 when userId is used in `array_key_exists`.
Sorry for IDE autoindent


kioan000:
Sorry for the messy request :(

GawainLynch:
Meh, mistakes are how we learn in life, and:
image

Oh, and more importantly around here nobody makes mistakes, it was always the sneaky :koala:

PR #7093, opened 2017-10-16
Remove |shy

### LINKED ISSUE #7090: 'Deprecated' causes FatalErrorException on dashboard
What happens: An error is displayed and the CSS rules for the layout are broken.
Bolt version: Bolt 3.4.0 beta 3
URL: On the dashboard page
Where: In the system log area

How does it look:
image

The call to `use Bolt\Common\Deprecated;` causes and error in the system log in the
```
FatalErrorException in Str.php line 5: 
Compile Error: Cannot use Bolt\Common\Deprecated as Deprecated because the name is already in use
```
Removing the line makes the error message go away
https://github.com/bolt/bolt/blob/9ccb3a01de8f2fec97b2ac6ab98d26b9751a2397/src/Helpers/Str.php#L5

jadwigo:
Fixes #7090

### DIFF
--- a/app/view/twig/components/panel-activity-system.twig
+++ b/app/view/twig/components/panel-activity-system.twig
@@ -27,7 +27,7 @@
                 {% elseif log.route == 'logout' %}
                     <em>{{ macro.userlink(log.ownerid|default('')) }}</em> {{ __('panel.latest-activity.logged-out') }}
                 {% else %}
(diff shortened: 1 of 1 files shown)


jadwigo:
@bobdenotter and me will look into this further tomorrow

GawainLynch:
Going to NAK this one, as the Twig function itself isn't problematic and is orders of magnitude faster than `excerpt`.

PR #7094, opened 2017-10-17
Remove import for Bolt\Common\Deprecated

### LINKED ISSUE #7090: 'Deprecated' causes FatalErrorException on dashboard

GawainLynch:
Fixes #7090
Related to PHP bug [66773](https://bugs.php.net/bug.php?id=66773), [66862](https://bugs.php.net/bug.php?id=66862) and [71079](https://bugs.php.net/bug.php?id=71079) (commit https://github.com/php/php-src/commit/2a75f5026a47099f585e29c5a9d8a2989dab42af)

### DIFF
--- a/src/Helpers/Str.php
+++ b/src/Helpers/Str.php
@@ -2,7 +2,6 @@
 
 namespace Bolt\Helpers;
 
-use Bolt\Common\Deprecated;
 use Cocur\Slugify\Slugify;
 
 class Str extends \Bolt\Common\Str


jadwigo:
@GawainLynch @bobdenotter This weird case pobably needs to be fixed for https://github.com/bolt/bolt/blob/38a79aacb48046633a3b4d873db1c22fb65af394/src/Helpers/Arr.php#L6 too

GawainLynch:
It would! Go for it! … my line of questioning yesterday was only to get to the bottom of the "why".

PR #7097, opened 2017-10-17
Remove import for Deprecated in Arr.php too

jadwigo:
Same change as https://github.com/bolt/bolt/pull/7094 in a different place


GawainLynch:
Ha! You did my trick, @bobdenotter of beating the other person by a few seconds :wink:

PR #7099, opened 2017-10-17
Allow `group` on named repeater's sets

xiaohutai:
Nothing fancy, but this is a nice PoC/snippet.

It's not as advanced as the `group` for fields, because this one goes from top to bottom and doesn't check for existing groups.

The button isn't split with a button and a caret anymore.

From:

named-repeaters

To:

grouped-named-repeaters


GawainLynch:
I think it need more :cow: :bell:

PR #7108, opened 2017-10-19
Fixed: Edge-case in `title_format` for CT's with no textual fields

bobdenotter:
giphy-downsized-large

### DIFF
--- a/src/Storage/Mapping/ContentTypeTitleTrait.php
+++ b/src/Storage/Mapping/ContentTypeTitleTrait.php
@@ -30,7 +30,7 @@ trait ContentTypeTitleTrait
     }
 
     /**
-     * Get an array of the columnname(s) of the title.
+     * Get an array of the column name(s) of the title.
      *
      * @param ContentType|array $contentType
      *
@@ -40,7 +40,7 @@ trait ContentTypeTitleTrait
     {
         Deprecated::method();
 
-        // If we specified a specific fieldname or array of fieldnames as 'title'.
+        // If we specified a specific field name or array of field names as 'title'.
         if (!empty($contentType['title_format'])) {
             return (array) $contentType['title_format'];
         }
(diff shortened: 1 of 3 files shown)


src/Storage/Mapping/ContentTypeTitleTrait.php
```diff
@@ -69,8 +69,6 @@ protected function getTitleColumnNames($contentType)
         }
 
         // If this is a ContentType without any textfields
-        $keys = array_keys($fields);
-
-        return reset($keys);
+        return array_keys($contentType['fields']);
```
GawainLynch: Why is returning all fields then?

bobdenotter: The method itself returns an array of fieldnames that are considered to cobble together the title. If _no_ suitable column is found at all, it'll either display `(no title)` or something put together from fields. For example: 

```
dummy:
    name: Dummies
    singular_name: Dummy
    fields:
        turtlesallthewaydown:
            type: select
        image:
            type: image
        datum:
            type: date
```

No `title` field, no `title_format` and no `type: text` fields. This CT would show "Donatello" or "Koopa" or the image filename in the listing overview, and "latest changes". It's not ideal, but IMHO better than before. 

Before: 

screen shot 2017-10-20 at 14 10 29


After: 

screen shot 2017-10-20 at 14 12 31

GawainLynch: I can't begin to tell you how wrong I find this approach :cry:

GawainLynch: Also, the change introduced in #7067 tries to return a non-existing variable … This needs tests!

bobdenotter: > Also, the change introduced in #7067 tries to return a non-existing variable

Yes, that is exactly what this PR is fixing: that piece of code that came from elsewehere wasn't adapted properly. 

> This needs tests!

When we discussed this, I thought we decided that we'd just fix it up to fix the bugs we're seeing, since it is marked "deprecated" and will get replaced sooner or later regardless.

bobdenotter:
Should be GTG now.

PR #7117, opened 2017-10-24
add an empty hidden field to ensure the field is saved

### LINKED ISSUE #7116: [3.4b4] Named repeaters are not saved even if empty
Named repeaters don't get saved if the fields are empty. I think assigning a block is sufficient to have it saved, even if empty.
  - For example, a block with (only) an unchecked checkbox would never get saved, or in case of bunch of optional fields (but you still want the info that there is a block of a certain type)
  - The `default` setting does not work (e.g. to pre-fill a type field), but that might be a separate issue/feature.

rossriley:
Fixes #7116

Just a port of the same issue with normal repeaters. Adding a hidden empty form value makes sure that the field is saved even if the content is empty.

### DIFF
--- a/app/view/twig/editcontent/fields/_block.twig
+++ b/app/view/twig/editcontent/fields/_block.twig
@@ -19,6 +19,9 @@
         </div>
     {% endif %}
 
+    {# This ensures that an empty value is always submitted even if there are no subsequent repeater sets #}
+    <input type="hidden" name="{{ name }}[]">
+
     <div class="block-slot">
         {% set blockfieldvals = context.content.get(contentkey) %}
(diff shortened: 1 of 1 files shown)


xiaohutai:
@rossriley I was looking at this just now, but this still doesn't allow "empty" block items to be saved.
So if I add a new item, and leave the title field empty, it will not get saved.

Say I have a block, with a block item that has 1 field

- 1 block items with empty field --> no blocks
- 2 block items, both fields empty --> no blocks
- 2 blocks items, one field non-empty --> all blocks saved

It seems it needs to have some value inside before saving any of the items

On a different note, the value field is not respected when outputting fields (because then I could add a hidden field).

PR #7119, opened 2017-10-25
[3.3] Handle extension asset packages using paths based on web root

### LINKED ISSUE #7077: Bundled extension displaying numerous error messages in system log for missing assets
After converting to a bundled extension from a local extension, the system log is displaying numerous error messages for missing assets - in this case a CSS file. The CSS file is loaded correctly and everything works as it should, it only includes the file on the back-end as specified (setZone) and in the correct position at the bottom of the template (setLate).

Details
-------

 - Bolt Version:3.3.6
 - PHP version: 7.0
 - Web server: Apache/2.4.18

Reproduction
------------

### Bug summary

So there's a bundled extension that is used on the back-end of the website and it includes a single CSS file to style up this back-end page. This is registered using the following function:

```php
  protected function registerAssets()
  {
    $asset = new Stylesheet();
    $asset->setFileName('extensions/local/tompato/contentreview/contentreview.css')->setLate(true)->setZone(Zone::BACKEND);
    return [ $asset, ];
  }
```

This adds the asset to the bottom of the template on the back-end and correctly links up and the included CSS is parsed.

```html
<link rel="stylesheet" href="/extensions/local/tompato/contentreview/contentreview.css?f3f43731e4" media="screen">
```

However, when looking through the system log, I get the following error message every few seconds (possibly overtime the front-end or back-end of the website is hit but I can't confirm that)

```
Couldn't add file asset 'extensions/local/tompato/contentreview/contentreview.css': File does not exist in either web://Tompato/ContentReview or theme://extensions/local/tompato/contentreview/contentreview.css directories. Make sure the file exists in either of these locations, by placing the file there manually (for Bundled Extensions) or by uninstalling / installing the extension again (for Managed Extensions).
```

Now, the file is definitely in `public/extensions/local/tompato/contentreview/contentreview.css` and the attached screenshot will show that.

screen shot 2017-10-13 at 12 45 47

Please note that the extension and everything works fine, it's just the spamming of the system log that is the issue.

GawainLynch:
Fixes #7077

The controversial part here is the addition of the web root as an asset package. We've avoided it up until now for several reasons, but I personally think it valid. 

If we're behind this enough to merge, then I'll follow up with some changes in the docs around this.

### DIFF
--- a/src/Extension/AssetTrait.php
+++ b/src/Extension/AssetTrait.php
@@ -192,8 +192,16 @@ trait AssetTrait
         }
 
         $app = $this->getContainer();
+        $filesystem = $app['filesystem'];
 
-        $themeFile = $app['filesystem']->getFile(sprintf('theme://%s', $path));
+        $publicFile = $filesystem->getFile("web://$path");
+        if ($publicFile->exists()) {
+            $asset->setPackageName('web');
+
+            return;
+        }
+
+        $themeFile = $filesystem->getFile("theme://$path");
         if ($themeFile->exists()) {
             $asset->setPackageName('theme');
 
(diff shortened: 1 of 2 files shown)


rossriley:
I can't think of a better way to handle this because whilst we don't test for it, by letting the path pass through we implicitly allow it.. 

so if the asset is `/path/to/public.jpg` we were throwing an error to say it doesn't exist but still spitting out the  `/path/to/public.jpg` url which the web server was happy to serve.

The only other way would be to throw an exception if it wasn't found in the theme or extension webdir which I don't think we want to do.

GawainLynch:
Yeah, the decision shift for me on this was thinking from the perspective of what a Bundle is & does. Mainly that we don't have logic for Bundles to install assets into a given/known location, and we're saying to people "put your things in the web root", so why in turn should we make it hard to reference the webroot. :wink:

PR #7120, opened 2017-10-25
Add 'status' to `querySelectList()`, so it can be used in relations.

bobdenotter:
This change will allow you to do this, for example:

```yml
    relations:
        entries:
            label: "Choose an entry"
            format: "{{ item.status != 'published' ? '(Not published) ' : '' }} {{ item.title|escape }} (№ {{ item.id }})"
```

### DIFF
--- a/src/Storage/ContentRequest/Edit.php
+++ b/src/Storage/ContentRequest/Edit.php
@@ -244,10 +244,13 @@ class Edit
     {
         $fields = [];
 
-        // Regex the 'format' for things that look like 'item.foo', and intersect with the actual fields in the contenttype.
+        // Get the field names from the ContentType, including 'status'.
+        $fieldNames = array_merge(array_keys($relationConfig['fields']), ['status']);
+
+        // Regex the 'format' for things that look like 'item.foo', and intersect with the field names.
         if (!empty($relationValues['format'])) {
             preg_match_all('/\bitem\.([a-z0-9_]+)\b/i', $relationValues['format'], $matches);
-            $fields = array_intersect($matches[1], array_keys($relationConfig['fields']));
+            $fields = array_intersect($matches[1], $fieldNames);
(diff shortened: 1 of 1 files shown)


JarJak:
Still not compatible with docs https://docs.bolt.cm/3.3/contenttypes/relationships

> For example if you have two fields for firstname and lastname you might put '{{item.firstname}} {{item.lastname}}' here.

bobdenotter:
@JarJak That should work.. Check method `neededFields()` in `bolt/bolt/src/Storage/ContentRequest/Edit.php`, where it compiles a list of fields it needs to get.. `status` needed to be added separately, because it's not a "real" field.

JarJak:
Ok :)

GawainLynch:
@bobdenotter This is "feature" and should go in 3.4 :wink:

bobdenotter:
@GawainLynch I was thinking about that, but I feel it's a "bug", instead of a "feature". Currently, the docs say: 

> How to show the titles for each record that can be selected. This takes a twig string where item is the record that can be selected. For example if you have two fields for firstname and lastname you might put '{{item.firstname}} {{item.lastname}}' here. The default is '{{ item.title|escape }} (№ {{ item.id }})'

I think most people would reasonably assume that they can use fields like they do in a regular twig template, including the `.status`. That doesn't work, so it's a bug. 

(i wouldn't _really_ mind if we bumped it to 3.4, though)

GawainLynch:
Ah, right.

On approach though, this is [better here](https://github.com/bolt/bolt/blob/v3.3.6/src/Storage/ContentRequest/Edit.php#L227) with the addition of:

```
$neededFields[] = 'status';
```

That will keep the logic out of the repo and in the hands of the caller.

bobdenotter:
Will do.

PR #7121, opened 2017-10-25
Confusing PHPDOC and method name StorageEvent->getContentType()

JarJak:
Added an alias and deprecated old name.

### DIFF
--- a/src/Events/StorageEvent.php
+++ b/src/Events/StorageEvent.php
@@ -67,8 +67,19 @@ class StorageEvent extends GenericEvent
      * Return the record's ContentType name.
      *
      * @return string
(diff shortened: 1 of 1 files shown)


GwendolenLynch (changes requested):
We're trying to move away from "slug"

JarJak:
@GawainLynch feel free to propose any other name

GawainLynch:
That is still under heavy debate. I am not seeing what problem this solves though.

JarJak:
It improves Developer Expierience

GawainLynch:
> It improves Developer Expierience

That is highly subjective

PR #7125, opened 2017-10-26
[3.4] Hyphenated ContentTypes & validation checks

### LINKED ISSUE #7123: [3.4] Using a contenttype with dash throws exception, instead of the notice to fix it on the dashboard.
If you have a contenttype like this: 

```yml
dummies-foo:
    slug: dummies-foo
    name: Dummies
    singular_name: Dummy
    singular-slug: dummy-foo
    fields:
        title:
            type: text
            class: large
            group: content
        slug:
            type: slug
            uses: title
        teaser:
            type: html
            height: 150px
```

The dashboard throws an exception: 

screen shot 2017-10-26 at 12 22 19

In 3.3 it showed the notice how to fix it: 

screen shot 2017-10-26 at 12 25 12

bobdenotter:
When testing out this fix, initially i was confused, because the "notice" showed only once, and the cached config was saved to cache, despite setting `$passed = false;`. 

Turns out, `checkConfig()` is run thrice on each request. If it fails on a wrong CT, it unsets the offending contenttype and 'fails', but then the second and third time it runs, it passes. And, because it passes, the cache gets saved. 

I'm not sure on approach, hence the `[WIP]`. 

Fixes #7123


src/Config.php
```diff
@@ -74,6 +74,8 @@ class Config
     /** @var JsonFile */
     private $cacheFile;
 
+    private $passed = null;
```
GawainLynch: Doesn't need the assignment :wink

Issue #7126, opened 2017-10-26
No Search in "Repeater" FieldType


develth:
Would be nice if the search also shows results for the content in the "repeater" field type

bobdenotter:
This is similar to #5829. Sometime, somewhere, we'll implement a module to properly index, cache and make searchable all content. Even the text in linked tables. 

So, we hear you, and it's on our list, but this is not "low hanging fruit". :-)

develth:
Thanks for the quick reply. 

Not really nice to hear that giving the enduser a good experience is not a "low hanging fruit".

bobdenotter:
Hi @develth, 

English is not my first language, so I may not have expressed myself properly. 

["Low hanging fruit"](https://www.urbandictionary.com/define.php?term=low-hanging%20fruit) is a "quick win" or "easy fix". I meant to say that this is a complicated thing to fix. It's important, and we _do_ try to give users a good experience, but it's not simple to implement.

develth:
Hey @bobdenotter 

if my PR solves that, it wasn´t as complicated ;)

Cheers, 
Thomas

PR #7127, opened 2017-10-26
[3.3] Searchable repeater

develth:
Added the possibility to search through a set of the field type 'repeater'

Also made the select query creation a little bit more readable & reusable.

Fixes: #7126

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -525,6 +525,7 @@ class Storage
         // (see also Content->getFieldWeights)
         $searchableTypes = ['text', 'textarea', 'html', 'markdown'];
         $table = $this->getContenttypeTablename($contenttype);
+        $fieldValueTable = $this->getTablename('field-value');
 
         if ($implode) {
             $query['words'] = [ implode(' ', $query['words']) ];
@@ -537,6 +538,18 @@ class Storage
                 foreach ($query['words'] as $word) {
                     // Build the LIKE, lowering the searched field to cover case-sensitive database systems
                     $fieldsWhere[] = sprintf('LOWER(%s.%s) LIKE LOWER(%s)', $table, $field, $this->app['db']->quote('%' . $word . '%'));
+                }
+            } elseif ($fieldconfig['type'] === 'repeater') {
+                foreach ($query['words'] as $word) {
+                    // Build the LIKE, lowering the searched field to cover case-sensitive database systems
+                    $word = $this->app['db']->quote('%' . $word . '%');
+                    $fieldsWhere[] = sprintf(
+                        "((LOWER(%s.value_text) LIKE LOWER(%s)) OR (LOWER(%s.value_string) LIKE LOWER(%s)))",
+                        $fieldValueTable,
+                        $word,
+                        $fieldValueTable,
+                        $word
+                    );
                 }
             }
         }
@@ -571,6 +584,20 @@ class Storage
             }
         }
 
+        // make repeaters work
+        $taxonomytable = $this->getTablename('taxonomy');
+        $taxonomies    = $this->getContentTypeTaxonomy($contenttype);
+        $tagsWhere     = [];
+        $tagsQuery     = '';
+
+        foreach ($taxonomies as $taxonomy) {
+            if ($taxonomy['behaves_like'] == 'tags') {
+                foreach ($query['words'] as $word) {
+                    $tagsWhere[] = sprintf('%s.slug LIKE %s', $taxonomytable, $this->app['db']->quote('%' . $word . '%'));
+                }
+            }
+        }
+
         // Build actual where
         $where = [];
         $where[] = sprintf("%s.status = 'published'", $table);
@@ -579,12 +606,15 @@ class Storage
 
         // Build SQL query
         $select = sprintf(
-            'SELECT %s.id FROM %s LEFT JOIN %s ON %s.id = %s.content_id WHERE %s GROUP BY %s.id',
+            'SELECT %s.id FROM %s LEFT JOIN %s ON %s.id = %s.content_id LEFT JOIN %s ON %s.id = %s.content_id WHERE %s GROUP BY %s.id',
             $table,
             $table,
             $taxonomytable,
             $table,
             $taxonomytable,
+            $fieldValueTable,
+            $table,
+            $fieldValueTable,
             implode(' AND ', $where),
             $table
         );
--- a/src/Storage/Entity/ContentValuesTrait.php
+++ b/src/Storage/Entity/ContentValuesTrait.php
@@ -97,6 +97,20 @@ trait ContentValuesTrait
                 if ($field['type'] === 'markdown') {
                     $excerptParts[] = $this->app['markdown']->text($this->values[$key]);
                 }
+                // add 'repeater' field
+                if ($field['type'] === 'repeater') {
+                    /** @var RepeatingFieldCollection $repeater */
+                    $repeater = $this->values[$key];
+                    /** @var FieldValue $repeatField */
+                    foreach ($repeater->flatten() as $repeatField) {
+                        if (in_array($repeatField->getFieldType(), ['text', 'html', 'textarea'])) {
+                            $excerptParts[] = $repeatField->getValue();
+                        }
+                        if ($repeatField->getFieldType() === 'markdown') {
+                            $excerptParts[] = $this->app['markdown']->text($repeatField->getValue());
+                        }
+                    }
+                }
             }
         }
 


src/Legacy/Storage.php
```diff
@@ -717,6 +717,12 @@ private function searchSingleContentType($query, $contenttype, $fields, array $f
                     // Build the LIKE, lowering the searched field to cover case-sensitive database systems
                     $fieldsWhere[] = sprintf('LOWER(%s.%s) LIKE LOWER(%s)', $table, $field, $this->app['db']->quote('%' . $word . '%'));
                 }
+            } elseif ($fieldconfig['type'] === 'repeater') {
+                foreach ($query['words'] as $word) {
+                    // Build the LIKE, lowering the searched field to cover case-sensitive database systems
+                    $word = '%'.$word.'%';
+                    $fieldsWhere[] = "((LOWER(bolt_field_value.value_text) LIKE LOWER('$word')) OR (LOWER(bolt_field_value.value_string) LIKE LOWER('$word')))";
+                }
```
GawainLynch: The `Bolt\Legacy` classes have been closed to additions since 2.2.

src/Legacy/Storage.php
```diff
@@ -785,6 +796,13 @@ private function searchSingleContentType($query, $contenttype, $fields, array $f
         return $results;
     }
 
+    protected function buildSearchQuery($table, $taxonomytable, $where)
+    {
+        $where = implode(' AND ', $where);
+
+        return "SELECT $table.id FROM $table LEFT JOIN $taxonomytable ON $table.id = $taxonomytable.content_id LEFT JOIN bolt_field_value ON $table.id = bolt_field_value.content_id WHERE $where GROUP BY $table.id";
```
GawainLynch: Um, 215 characters … This is much less readable, breaks coding standard (PSR-2), expands the API unnecessarily.

src/Legacy/Storage.php
GawainLynch: The `Bolt\Legacy` classes have been closed to additions since 2.2.

GawainLynch:
Also, this is a feature, so needs to be shifted to 3.x

src/Legacy/Storage.php
develth: But are currently in use for search, or i missed something?

src/Legacy/Storage.php
develth: Ok - fixed.

develth:
@GawainLynch 

thanks for review.

> Also, this is a feature, so needs to be shifted to 3.x

What does this mean? 
No release in 3.3?

Imho its not a new "feature", it extends the search functionality by search a missing field type.

develth:
Hi @GawainLynch, 

why changed to 3.4? 
Wouldn´t it fit into a 3.3 release?

Thanks, 
Thomas

GawainLynch:
@develth the thing is that it adds functionality, albeit to an in-progress feature, making it not a bug fix as such in terms of managing the codebase, and associated risks to others.

We've discussed internally, and we're going to have a closer look and will probably accept this into 3.4, which is due out very soon. That will effectively EoL 3.3, and as this isn't a bug fix it won't be accepted into a stable branch.

GawainLynch:
@develth we've talked more, and then some more again, then yet more again about this change.

Ultimately the decision we've come to is to not allow the changes to legacy. The `Bolt\Legacy` classes have been frozen by policy since the beginning of 2016. While some small things have been agreed upon, they have consistently caused a tonne more work to maintain and support for an already overstretched team :smiley: 

One of the harder parts of this decision was that we didn't want to look as if we're discouraging contribution, or stifling progress. But we're a small team, with a lot of users, doing all of this as volunteer work. So we need to make some hard calls sometimes when it comes to keeping things moving, and this has been one of those hard calls.

If you're up to the task, we'd certainly appreciate help & testing on the new storage layer :+1:

develth:
Thats pretty sad. 
You close / freeze a whole set of classes that are still in use and important parts without provide new possibilities. Search Content is a important part on a website / webapplication. And if i can´t search content with the tools and classes you provide me (or allow me to make things work), you hinder me using a CMS / Project - and that should never happen. 

This decision will force me to drift off this project.

> If you're up to the task, we'd certainly appreciate help & testing on the new storage layer 👍

I don´t think this new storage layer is in use in 3.3

GawainLynch:
We put in countless hours to provide you with something for free … we only have so much time and resources. If that is your attitude towards sharing and collaboration … good riddance :wave:

bobdenotter:
> This decision will force me to drift off this project.

I hope you realise how incredibly ungrateful this sounds? 

We do see there's an issue, but we ask of you to understand that there's a process involved? If we merge/accept all changes as they come in, it _will_ hurt the project in the long run. We need to make sure we keep the long-term vision in mind, and sometimes that means we can't go along with a change that benefits a specific usecase in the short term only.

develth:
Please don´t take the words out of my mouth.

If its to heavy for you to press the merge button - ok. If this press will resolve into more work for you to make a lot of changes - ok. If you just don´t want to accept merges into deprecated code - ok. If you decide to close important parts in your Project for changes and fixes and nothing new is here for replacement - ok. 

I accept those decisions, because its your project. 
But please also accept that there are people that put also a lot of free time into your project while contributing that do not follow those decisions. And then you should accept that those decisions are not the way they want to rely on for collaboration / business / future projects because you don´t let them. 

@bobdenotter this shouldn´t sound ungrateful - its just a decision. I was searching for a new tool to work with and to collaborate to. But something like this is just not easy to understand for me and i will see future problems with another PRs. So i drift off this project and have to search something new. 
Your Project is great - a nice and powerful tool. 

And i don´t said you should accept every PR that come in. But the changes i provided don´t crash anything and fit into the 3.3. version. And i also don´t get why this blocks your long term vision. 

But as i wrote above - those are your decisions. I´m fine with that. But i don´t have to go with them.

SahAssar:
@develth 

> But please also accept that there are people that put also a lot of free time into your project while contributing that do not follow those decisions.

We very much do. There have been times where deprecating or removing features or making changes have been delayed or the implementation has been changed due to this. But when it comes to adding features that behavior will need to be maintained for a significant amount of time. There are features in bolt that have stagnated because the commiter is no longer around and there is no stakeholder for it. I hope you understand the dilemma here.

> And then you should accept that those decisions are not the way they want to rely on for collaboration / business / future projects because you don´t let them.

IMO if someone wants to rely on bolt based on it's current capabilities or features that's perfectly fine, but no one should rely on it based on an assumption that a future PR will get merged. There will always be discussions around PRs, and some will not get in, but usually that is because of the implementation, not the idea (unless it is out of scope). In this case the feature was only implemented in deprecated code, not in the code we will use going forward. 

In my experience all the people involved with bolt try their best to help with issues and to allow for collaboration both in core and in extensions. Sometimes the ideas are out of scope, sometimes they will not be possible in bolt core, but they/we always try to help out.

> But the changes i provided don´t crash anything and fit into the 3.3. version. And i also don´t get why this blocks your long term vision.

If this was in both the new and the old storage layer I think it'd have a good chance to get merged (in the next feature branch), but only adding features to deprecated code is a hazard since someone else will have to write that code for the new layer at a later point and it also causes an inconsistency between the new and old storage.

I hope you understand the reasoning behind this

develth:
Just upgraded to 3.4.6 - is there still no possibility to search through the bolt "supported" content types? 
Still no progress in deprecated Storage search / replacement?

bobdenotter:
Seeing how this issue is still open, there is _**still**_ is no way to do it. We are just too lazy to get it implemented. 

Thankfully, i've heard of someone, who is less lazy than us, that there will be an extension available soon, that basically aggregates all text from repeaters and blocks into an hidden field, effectively making them searchable.

rossriley:
New storage engine implemented in #7311.

develth:
@rossriley seems like your PR makes the repeater searchable with the QueryParser, but not yet via FrontendController´s SearchAction, or am i wrong?

rossriley:
@develth yes you're right.. 3.5 will allow the switch over entirely to the new storage engine and that will cover all the use in the controllers/templates.

If you want to use it in the 3.4.x series you'll need to do it in your own controller methods.

bobdenotter:
Oooh, I missed that particular detail of your PR, @ross. 

@develth If it's in that PR, then it'll be available in Bolt 3.5. :-)

xiaohutai:
Searching on the front-end through both **repeaters** and **blocks** would be (a lot) more work, right? As you need to check for all the searchable content (text, textarea, html, markdown) defined per block.

PR #7134, opened 2017-10-31
[3.4] More Twig namespaces

GawainLynch:
Just a few more that have had aliases assigned/fixed in the last 6 or so months … otherwise "nothing to see here"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
         "symfony/web-profiler-bundle": "^2.8",
         "symfony/yaml": "^2.8",
         "tdammers/htmlmaid": "~0.7",
-        "twig/twig": "^1.34.2",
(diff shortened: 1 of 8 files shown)


bobdenotter:
It seems some of these changes landed in Twig 1.34. Should we bump the requirement, to be sure another dependency doesn't pull in an older one? 

https://github.com/bolt/bolt/blob/3.3/composer.json#L74

GawainLynch:
Might want to check the correct branch https://github.com/bolt/bolt/blob/3.4/composer.json#L77 … that was bumped in #6742 (June 6th) :wink: 

That said, this bumps it to `^1.35`

bobdenotter:
Ah, right! Nothing to see here, then! 👍

PR #7155, opened 2017-11-06
Only set a group when the taxonomy is relevant

### LINKED ISSUE #7149: [3.4] Using more than one taxonomy makes overview lose grouping..
In a default setup, pages has: `taxonomy: [ groups ]`. Looks like this: 

screen shot 2017-11-05 at 13 39 54

If you change that to `taxonomy: [ groups, tags ]`, the overview will look like this: 

screen shot 2017-11-05 at 13 40 18

Note: The _order_ is correct, just the headings are missing.. At first glance, it seems `$content['multiplecontent']` gets set, but is missing the `group` in each record.

rossriley:
Fixes #7149

This was a slight logic error, we only want to set the group if the taxonomy concerned is a grouping one, otherwise we just skip over it.


GawainLynch:
:koala:

PR #7158, opened 2017-11-06
Allocate the other results to an ungrouped array

### LINKED ISSUE #7150: [3.4] Records "Disappear" when adding "groups".
If you add `taxonomy: [ groups ]` to a contenttype that didn't previously have it, and you then add a group to _one_ record, that will cause all other records to disappear from the overview.. 

Steps to reproduce: 

1) Change `taxonomy: [ groups ]` in the default 'Entries' contenttype.

2) See it work OK in the backend. (except for the missing "(no grouping)" subheading).

screen shot 2017-11-05 at 13 50 51

3) Then, when you edit any record, and assign a group to it, all other records will seemingly disappear.

rossriley:
Fixes #7150

In instances where no taxonomy is set on a record some were slipping through the net, this ensures that they get passed on to the results array.

### DIFF
--- a/src/Storage/ContentRequest/Listing.php
+++ b/src/Storage/ContentRequest/Listing.php
@@ -154,6 +154,8 @@ class Listing
             }
             if ($taxGroup !== null) {
                 $grouped[$taxGroup][] = $result;
+            } else {
+                $grouped['ungrouped'][] = $result;
             }
         }
 


rossriley:
in before :koala: :koala: :koala:

rossriley:
damn too late

PR #7169, opened 2017-11-10
Re-introduce `implements SanitiserAwareInterface` for `type: text` fields.

bobdenotter:


bobdenotter:
Upcoming: A PR for docs, detailing which fields do and which fields don't allow "random HTML and javascript"

PR #7172, opened 2017-11-12
[Tests] Coverage for Bolt\Asset

GawainLynch:


GawainLynch:
Changing to 3.4 'cause PHPUnit :wink:

PR #7196, opened 2017-11-22
documentation for general/caching/duration updated for timed records

ToBe998:
The documentation for the `general/caching/duration` parameter was not including the fact that this also controls the interval for checks on timed records.

Took me a while to find out why records wouldn't be published on time. Hope this helps others.

**Bolt 3.2** - This seems to have changed in 3.3+


GawainLynch:
3.2 is now "end of life" and the branch is closed :wink:

bobdenotter:
Should be an "easy pick" for someone.. Leaving it open for now.

Issue #6985, opened 2017-09-08
[RFC] Thoughts on deprecating all methods on the Content object


rossriley:
This may not be desirable or doing it now for deprecation in 4.0 might be deemed to be too hasty so I'm just interested in getting some general feedback before I start working on anything.

As you know I'm working on the compatibility between old and new storage layers and one of the biggest jobs is making sure that a new content object behaves as similarly as possible to an old one.

One of the downsides of this is that these content objects need to know too much and do too much and if we want to improve things in the future I'm suggesting that we standardise our public Twig API on functions and filters rather than having core functionality mixed into the content objects themselves.

The other part is that it's very hard to determine what is and what isn't the public API and ideally when 4.0 hits we want to reduce the number of surprises as people migrate their themes over.

So the main methods to address are:

```
record.link()
record.previous()
record.next()
record.excerpt()
record.fieldtype()
```

Also perhaps have a pass through the classes and traits and mark unnecessarily public methods as protected.

The idea would be to provide this functionality via functions or filters where applicable and these can then have the correct dependencies injected into them.. for example something like...

```
record|link or link(record)
record|link('next') or link(record, 'next')
record|link('previous') or link(record, 'previous')
record|excerpt or excerpt(record)
fieldtype(record, 'fieldname')
```

Obviously these are just initial suggestions and details can be hammered out in implementation phase.

Thoughts / alternative ideas welcome.

GawainLynch:
Related #6774 and dev meeting discussion from [11 July](https://github.com/bolt/bolt/wiki/Dev-meeting-2017-07-11)

GawainLynch:
As I raised in #6774 and have become pretty firm on, I think your suggestions here reflect a good way to go. 

I am not fussed if they are filters, functions, or both. But there absolutely needs to be a way forward now so, as you said, there is less work/surprise in the future for people.

xiaohutai:
Also related #6798. In combination with Twig.

SahAssar:
@bobdenotter @GawainLynch @rossriley @CarsonF did we land on the `|explain` idea and deprecate/remove here or are there any objections?

rossriley:
I am going to be working on this for 3.5, have a branch in progress now so happy to close this. Can't remember what was discussed about the explain filter what did we decide?

SahAssar:
IIRC the compromise plan that everyone was okay with was to have a `|explain` filter in twig which would show what could be accessed from the content object (like relations, links and so on).

[...]

PR #7197, opened 2017-11-22
[3.4] Use "first_image" to return the first image, for display in overview listings.

### LINKED ISSUE #7194: [3.4] The "magic" image doesn't work anymore.
If you have a contenttype like this: 

```
dummies:
    name: Dummies
    singular_name: Dummy
    fields:
        title:
            type: text
            class: large
            group: content
        picture:
            type: image
        teaser:
            type: html
            height: 150px
```

The 'picture' is not used as default thumbnail, see screenshot: 

screen shot 2017-11-21 at 17 27 15

Up until 3.4, the `content.getImage` was "magic", as it would return the first found image. Now it only returns a field named actually 'image'. 

This is the exact same issue as what happened here: #7067 (only there it was for the magic `getTitle`. 

I see two ways to fix it: 

 - We implement the `getImage` as a trait, like the deprecated `getTitle`
 - We get a head start on #6985, and implement a `findImage()` filter, so we can use `{{ content|findImage() }}`. 

Either way would get the listing to show "an image" again. 

Suggestions?

bobdenotter:
Fixes #7194

Note: Once Ross' work for #6985 lands, this will be integrated into that, and this specific implementation will be replaced / removed.

### DIFF
--- a/app/view/twig/_base/_listing.twig
+++ b/app/view/twig/_base/_listing.twig
@@ -166,7 +166,7 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
 
         {# COLUMN: Thumbnail #}
         <td class="listthumb">
-            {% set listimage = content.getImage %}
+            {% set listimage = content|first_image %}
             {% if listimage is not empty %}
                 {% if listimage is iterable and listimage['file'] is not defined %}
                     {% set listimage = listimage|first %}
--- a/src/Twig/Extension/RecordExtension.php
+++ b/src/Twig/Extension/RecordExtension.php
@@ -53,6 +53,7 @@ class RecordExtension extends AbstractExtension
             new TwigFilter('selectfield', [Runtime\RecordRuntime::class, 'selectField']),
             new TwigFilter('trimtext',    [Runtime\RecordRuntime::class, 'excerpt'], $safe + $deprecated + ['alternative' => 'excerpt']),
             new TwigFilter('taxonomy',    [Runtime\RecordRuntime::class, 'taxonomy']),
+            new TwigFilter('first_image', [Runtime\RecordRuntime::class, 'getFirstImage'], $safe + $deprecated),
             // @codingStandardsIgnoreEnd
         ];
     }
--- a/src/Twig/Runtime/RecordRuntime.php
+++ b/src/Twig/Runtime/RecordRuntime.php
@@ -135,6 +135,35 @@ class RecordRuntime
     }
 
     /**
+     * Gets the first image from the given content record.
+     *
+     * This method is deprecated. In Bolt 3.5 it'll be replaced / removed to
+     * fall in line with the work done for #6985
+     *
+     * @param \Bolt\Storage\Entity\Content $content
+     *
+     * @return array|null image
+     */
+    public function getFirstImage($content)
+    {
+        Deprecated::method('3.4');
+
+        if (!is_array($content->contenttype['fields'])) {
+            return null;
+        }
+
+        // Grab the first field of type 'image', and return that.
+        foreach ($content->contenttype['fields'] as $key => $field) {
+            if ($field['type'] === 'image' && is_array($content->get($key))) {
+                return $content->get($key);
+            }
+        }
+
+        // otherwise, no image.
+        return null;
+    }
+
+    /**
      * Output all (relevant) fields to the browser. Convenient for dumping the
      * content in order in, say, a `record.twig` template, without having to
      * iterate over them in the browser.
@@ -149,7 +178,7 @@ class RecordRuntime
      * @param string|array         $exclude
      * @param bool                 $skip_uses
      *
-     * @return string
+     * @return string|null
      */
     public function fields(
         Environment $env,


app/view/twig/_base/_listing.twig
```diff
@@ -166,7 +166,7 @@ row_header:     prop.extended and (prop.first or (prop.has_groupname and prop.ne
 
         {# COLUMN: Thumbnail #}
         <td class="listthumb">
-            {% set listimage = content.getImage %}
+            {% set listimage = content|findimage() %}
```
GawainLynch: Empty filters don't need `()`

src/Twig/Extension/RecordExtension.php
```diff
@@ -53,6 +53,7 @@ public function getFilters()
             new TwigFilter('selectfield', [Runtime\RecordRuntime::class, 'selectField']),
             new TwigFilter('trimtext',    [Runtime\RecordRuntime::class, 'excerpt'], $safe + $deprecated + ['alternative' => 'excerpt']),
             new TwigFilter('taxonomy',    [Runtime\RecordRuntime::class, 'taxonomy']),
+            new TwigFilter('findimage',   [Runtime\RecordRuntime::class, 'findimage'], $safe + $deprecated),
```
GawainLynch: It isn't "finding an image" it is "returning the first image"

src/Twig/Runtime/RecordRuntime.php
```diff
@@ -134,6 +134,35 @@ public function excerpt($content, $length = 200, $focus = null)
...
+     * Find the first image in the given content record.
+     *
+     * This method is deprecated. In Bolt 3.5 it'll be replaced / removed to
+     * fall in line with the work done for #6985
+     *
+     * @param \Bolt\Storage\Entity\Content $content
+     *
+     * @return array image
+     */
+    public function findimage($content)
```
GawainLynch: PSR-2 & type hint

src/Twig/Runtime/RecordRuntime.php
```diff
@@ -134,6 +134,35 @@ public function excerpt($content, $length = 200, $focus = null)
...
+    {
+        Deprecated::method('3.4');
```
GawainLynch: Replaced with?

src/Twig/Runtime/RecordRuntime.php
```diff
@@ -134,6 +134,35 @@ public function excerpt($content, $length = 200, $focus = null)
...
 
+    /**
+     * Find the first image in the given content record.
+     *
+     * This method is deprecated. In Bolt 3.5 it'll be replaced / removed to
+     * fall in line with the work done for #6985
+     *
+     * @param \Bolt\Storage\Entity\Content $content
+     *
+     * @return array image
```
GawainLynch: `@return array|null`

bobdenotter:
> I do not see the point in this at all. 

The "point" is that it fixes a bug and a regression.

> It just adds to the API for the lifetime of v3 (at least 1.5 more years) with a bad approach 👎

I asked you for advice on approach, and you literally said "i couldn't care less".

src/Twig/Runtime/RecordRuntime.php
bobdenotter: I was just following what other methods in that class did, like https://github.com/bolt/bolt/blob/3.4/src/Twig/Runtime/RecordRuntime.php#L69 and https://github.com/bolt/bolt/blob/3.4/src/Twig/Runtime/RecordRuntime.php#L152. But, i'll happily change it.

src/Twig/Runtime/RecordRuntime.php
bobdenotter: It will be replaced (or integrated with) Ross' work for #6985. It doesn't exist quite yet, so i can't refer to it.

src/Twig/Extension/RecordExtension.php
bobdenotter: I thought about naming, and this wat the best i came up with. "getImage" would be confusing, as it is too close to the records current magic getter, and "getfirstimage" is a bit long. I don't mind changing it, because it is literally used in one location now.

src/Twig/Runtime/RecordRuntime.php
GawainLynch: Yeah, there is a huge lack of consistency … it wasn't one I was terribly worried about, just threw it in with the other comments.

src/Twig/Extension/RecordExtension.php
GawainLynch: What about just `first_image` (Twig) and `getFirstImage()` (PHP)? Both speak to what it does.

GawainLynch:
> I asked you for advice on approach, and you literally said "i couldn't care less".

I seem to have missed the point where you were adding another Twig function … Can we at the very least tag it `@internal` so that removing it doesn't raise ire from people.

bobdenotter:
> Can we at the very least tag it @internal so that removing it doesn't raise ire from people.

Sure, no problem.. 

I'm not sure how to fix the tests, though.. It's [saying](https://travis-ci.org/bolt/bolt/jobs/306170550):

```
---------
1) BackendAuthorCest: Search for the 'about' page as the 'author' user
 Test  tests/codeception/acceptance/104-BackendAuthor/BackendAuthorCest.php:omnisearchTest
                                                                                                                                                                                                                                                                                                                                                                   
  [TypeError] Argument 1 passed to Bolt\Twig\Runtime\RecordRuntime::findimage() must be an instance of Bolt\Storage\Entity\Content, instance of Bolt\Legacy\Content given, called in /home/travis/build/bolt/bolt/app/cache/development/twig/266d677…/7f/7f12f181d15aa….php on line 458  
                                                                                                                                                                                                                                                                                                                                                                   
``` 

If I debug `_listing.twig`, it's most certainly a `Bolt\Storage\Entity\Content` that's getting passed in. 🤔

src/Twig/Extension/RecordExtension.php
bobdenotter: Fair enough. 👍

PR #7202, opened 2017-11-24
Added "create new" button in contentedit

Boorj:
When list of records is empty, it's very **friendly** to show a button "create new" button.  That really helps for some of users.

image
(sorry for colors, i use themes for my clients)


GawainLynch:
Thanks mate

PR #7210, opened 2017-11-27
Update Save.php

nroopesh:
quick PR


GawainLynch:
Thanks for your PR @nroopesh

What are you trying to achieve with this change though?

nroopesh:
Hi @GawainLynch 
there are two if conditions for "save_return" i discussed with @bobdenotter as per his advice i removed the unreached code. But i still doubt there is no condition for "save" check on $returnTo

GawainLynch:
Ah right, yeah looks to be a bad copy/pasta on the last updates I did in there.

> But i still doubt there is no condition for "save" check on $returnTo

There doesn't need to be, it will just fall through and stay on the same page. :smile_cat:

Issue #7198, opened 2017-11-24
[RFC] Improved file upload sanitizer


Boorj:
### problem 
when you upload a file with non-alphanumeric name, bolt is currently sanitizing it, [according to config](https://github.com/bolt/bolt/blob/3.x/src/Provider/UploadServiceProvider.php#L30).
So, for example, cyrillic filename "новый-файл 1.jpg" becomes some kind of `_____-_____1.jpg`. Imagine the problem with collisions, on massive file upload.

### solution 
For my purposes i created sanitizer provider: [improved sanitizer gist](https://gist.github.com/Boorj/e602f95faf139a09f14cc8a802532993), which carefully converts `'ě' → 'e'`, and after that applies replacement pattern.
The code is lightweight, so it won't heavily affect performance.

### profit
This will impact all languages with upper and lower accents, and also Cyrillic symbols. People will get less mess in their folders. 

### rfc
Maybe it should be implemented in core, 'out of the box'?

bobdenotter:
Closing, because this is fixed in #7198. Thanks, @Boorj and @GawainLynch 👍

Boorj:
@bobdenotter @GawainLynch Problem came back again!

PR #7217, opened 2017-11-28
Improved uploaded file sanitizer

Boorj:
translates non-ascii European symbols to ascii-latin + transliteration
Related to #7198

### DIFF
--- a/src/Provider/UploadServiceProvider.php
+++ b/src/Provider/UploadServiceProvider.php
@@ -4,6 +4,7 @@ namespace Bolt\Provider;
 
 use Bolt\Filesystem\Filesystem;
 use Bolt\Filesystem\UploadContainer;
+use Cocur\Slugify\Slugify;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
 use Sirius\Upload\Handler as UploadHandler;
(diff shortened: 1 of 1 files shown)


Boorj:
@GawainLynch , 😆 really. 
will `$app['slugify']->slugify($filename)` be a solution?

GawainLynch:
https://github.com/cocur/slugify/blob/1.4/src/Slugify.php#L54-L753 :wink:

Boorj:
yes, this is the file that i researched, before posting previous post :wink:
So i'll assume that you meant yes for my question).
I'll do that.

Boorj:
@GawainLynch 
here's gist with that stuff: https://gist.github.com/Boorj/7e28a9604d250929fcd047e2c89d6fce

GawainLynch:
Better. On quick first look though, don't use the `$app['config']` object outside of the closure(s)

Boorj:
@GawainLynch i updated gist, is it ok now?

GawainLynch:
@Boorj 'cause we're mates and both really busy, I implemented the correct parts of your gist and squashed the commits. Now `$app['upload.sanitizer']` is an overridable service using Slugify and everyone's needs should be able to be met :+1:

GawainLynch:
Also as this adds features, not directly fixes bugs, I've moved the branch to 3.5 … which is going to be released a.s.a.p.

src/Provider/UploadServiceProvider.php
```diff
@@ -18,6 +19,13 @@ class UploadServiceProvider implements ServiceProviderInterface
 {
     public function register(Application $app)
     {
+        $app['upload.sanitizer'] = $app->share(function ($app) {
+            $pattern = $app['config']->get('general/upload/pattern', '[^A-Za-z0-9\.]+');
+            $lowercase = $app['config']->get('general/upload/lowercase', true);
+
+            return new Slugify("/$pattern/", ['lowercase' => $lowercase]);
```
Boorj: @GawainLynch i wasn't sure, how much memory does Slugify take and is it only 1 instance per PHP response called. I mean, if multiple files and one php session, and heavy Slugify. So that's why i created distinct variables. This solution was my first one. And thanks for using that.

GawainLynch: Yeah, Slugify is tiny … there are much bigger maps in use. But also this is the DI service model, they only get instantiated once per request as you mention :smile:

PR #7226, opened 2017-12-01
Fix NL translation

bobvandevijver:
Error in the translation resulting in a non-functional link.


bobdenotter:
Good catch, thank you @bobvandevijver 👍

PR #7227, opened 2017-12-02
[WIP] Feature/hierarchical content

mrenigma:
So here is the long-awaited PR!

This PR implements a hierarchy system in Bolt which allows individual content types to be defined as hierarchical or not. If a content type is hierarchical, it will be reflected in the routing layer, it will be given a `parent` field to define the record parent (restricted to the same content type) and it will add additional methods to the record to retrieve its hierarchy, its children and it's root parent record.

This implementation was originally built for a client site, so there may be methods or processes which don't need to happen in core Bolt! 

### How to make it work
To make hierarchical work you have to add the following to a content type in `contenttypes.yml`:

```
hierarchical: true
```

And add/edit the following in the `router.yml` :

```
contentlink:
    path: /{contenttypeslug}/{slug}
    requirements:
        contenttypeslug: controller.requirement:anyContentType
        slug: '[a-z0-9-_\/]+'
```

This is to allow `/` in the routing structure (may be possible to automate?).

### Still to do/issues
- [ ] Move `parent-change.js` into the Bolt core JS files. I've started work on this but I'm definitely no expert on the Bolt backend assets pipeline, so any help would be greatly appreciated! 4557cf7.
- [ ] Once that is done, resolve this [TODO](https://github.com/mrenigma/bolt/blob/feature/hierarchical-content/src/Controller/Backend/Records.php#L135) and remove.
- [ ] Consider whether using the JSON approach from the TODO above is worth replacing with an AJAX call instead.
- [ ] Style the listing pages nicely, especially when several levels deep or in groups.
- [ ] Decide whether to ignore pagination when displaying hierarchical content (i.e. show more than 10) or to try and make that work too.
- [ ] Update "View saved version" link on update of the parent field (this actually doesn't work in non-hierarchical Bolt either, changing the slug doesn't update that link!).
- [ ] Make sure the URL field on records is updating on change of parent and the full hierarchical URL is visible on record page load.
- [ ] Check and clean up any code style issues (FYI an `editor.config` in Bolt would be amazingly helpful in future)
- [ ] Add to tests to all new methods and changes to existing ones


jkazimir:
Wow! :-)

Having a look, and noticed there's a ton of cache files here adding to the already long diff. Probably shouldn't have been committed?

mrenigma:
@jkazimir thanks, didn't even notice those! I'll remove them from git

bobdenotter:
Awesome! Will check it out soon to monkey-test. I do notice quite a bit of the changes are whitespacing and code-style things. But, we can fix those. :-)

mrenigma:
Yeah I know. My PHPStorm isn't set up in Bolt style. I tried setting up the code style stuff but it didn't work initially. I haven't had time to have another go since.
I've done my best to not save and format to try and avoid the style differences but things have probably slipped through the net.

.gitignore
```diff
@@ -54,4 +54,5 @@ Vagrantfile
 .swp
 *.lock
 .buildpath
+.idea
```
GawainLynch: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

src/Storage/Entity/ContentRouteTrait.php
```diff
@@ -31,6 +32,10 @@ public function editlink()
     {
         $perm = 'contenttype:' . $this->contenttype['slug'] . ':edit:' . $this->id;
 
+        if (empty($this->app)) {
+            $this->app = AppSingleton::get();
```
GawainLynch: Must not be used, ever :wink:

src/Storage/Entity/ContentRouteTrait.php
```diff
@@ -48,10 +53,38 @@ public function editlink()
     public function link($referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
     {
         list($name, $params) = $this->getRouteNameAndParams();
+
+        if (empty($this->app)) {
+            $this->app = AppSingleton::get();
```
GawainLynch: As above

src/Storage/Entity/ContentRouteTrait.php
```diff
@@ -61,6 +94,10 @@ public function link($referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
      */
     public function isHome()
     {
+        if (empty($this->app)) {
+            $this->app = AppSingleton::get();
```
GawainLynch: Same here

src/Storage/Entity/ContentRouteTrait.php
```diff
@@ -133,10 +170,26 @@ public function getRouteNameAndParams()
      */
     protected function getRouteConfig()
     {
+        if (empty($this->app)) {
+            $this->app = AppSingleton::get();
```
GawainLynch: Here also

src/Storage/Hierarchy.php
```diff
@@ -0,0 +1,555 @@
...
+    private $isLegacy = true;
+    private $pathPieces = [];
+    private $parentTree = [];
+    private $lastResult;
+    private $routePrefix = '/';
+
+    public function __construct($app)
+    {
+
+        $this->app = $app;
```
GawainLynch: Don't inject the application, pass the services you need instead

GwendolenLynch (changes requested):
Also, we'll need to rebase out the cache commits, revert some of the removals and the spacing

.gitignore
mrenigma: Would you like me to remove this? I saw you had quite a few IDE specific ignores but not PHPStorm so I added it but I can remove

GawainLynch: Yeah, I'll catch this and a few other things in the rebase we talked about in Slack just now :smile:

src/Storage/Entity/ContentRouteTrait.php
mrenigma: Hmmm I swear I copied this from somewhere else in Bolt :thinking:. Doesn't matter, I'll update all of these to remove the `empty`!

GawainLynch: Ah, it is not the `empty()` but the use of the application object in the entity … we'll chat about it in Slack in the coming week when I am back in NL

mrenigma: Not the `empty()`?! :astonished: I never thought I'd hear you say that :stuck_out_tongue_winking_eye:. Okay sure, let's chat once your back from NL.

graham73may:
@mrenigma, anythings I can do on this PR to get the ball rolling again?  If you ELI5 I might get somewhere 😉

mrenigma:
@graham73may please do! I've been swamped (as you know). All the to-dos are on this PR so take your pick. If you see something you think you could help with, let me know and I'll do my best to try and explain it.

mrenigma:
@graham73may please do! I've been swamped (as you know). All the to-dos are on this PR so take your pick. If you see something you think you could help with, let me know and I'll do my best to try and explain it.

graham73may:
I can replace the app things if you've got an idea of what it needs to be?
Or an example from another file or something?

mrenigma:
`ContentRouteTrait` is tied to the entity changes being done. Not sure what the core teams plans are on this one.

On the `Hierarchy.php` the `app` needs to be changed to inject the services from app instead. So rather than `$this->app['service']` it should be passed so it's used as `$this->service`.

There are several services being initiated around the hierarchy initiation so plenty of examples of how to do it there :smile:

bobdenotter:
@graham73may @mrenigma Hey guys, making progress on this one? :-)

mrenigma:
@bobenotter honestly I haven't touched it. I'm currently switching jobs and working overtime during the transition. I don't know if @graham73may has had a chance?

graham73may:
No news from me. 

I think it would be good to get some help from the Bolt Core team on how the tasks in the PR get implemented: 

## JS 
- Move parent-change.js into the Bolt core JS files. I've started work on this but I'm definitely no expert on the Bolt backend assets pipeline, so any help would be greatly appreciated! 4557cf7.
- Once that is done, resolve this TODO and remove.
- Consider whether using the JSON approach from the TODO above is worth replacing with an AJAX call instead.

## Backend Styling
- Style the listing pages nicely, especially when several levels deep or in groups.
- Decide whether to ignore pagination when displaying hierarchical content (i.e. show more than 10) or to try and make that work too.

## Backend issues / bugs
- Update "View saved version" link on update of the parent field (this actually doesn't work in non-hierarchical Bolt either, changing the slug doesn't update that link!).
- Make sure the URL field on records is updating on change of parent and the full hierarchical URL is visible on record page load.

## Housework
- Check and clean up any code style issues (FYI an editor.config in Bolt would be amazingly helpful in future)
- Add to tests to all new methods and changes to existing ones

rossriley:
In terms of approach I'd suggest that we get the new storage updates merged in soon and start work on the 3.5 release process. Then we get this PR rebased onto 3.6 and get it merged in. That means I can jump in and do some work on it and means we'll have a month or two to get everything solid before the 3.6 release.

bobdenotter:
Hi, @mrenigma and @graham73may,

Has there been any recent progress on this issue?

graham73may:
Hi @bobdenotter 

This has been sat on the shelf whilst we were waiting for the new storage layer changes to be merged in (from Ross's last comment). 

To be honest, we'd feel a lot better if someone could help us with the backend Javascript and CSS stuff, we're not familiar with any of this at all and Grunt was being a bit of a pig the last time we tried it. 

We are VERY keen to get this in to core as soon as possible, is anyone of the team available to help push this forward from your end?

rossriley:
I'm happy to look after getting this into core, just needed to get the storage PR in first, since this can hopefully be one of the cool features that persuades people to switch the old layer off.

@bob I might need some input on the UI side but we can iron those things out at the testing phase.

One suggestion for discussion, I wondered whether in choosing the field names that end users set in contenttypes, we might pick something simpler to spell than `hierarchical` and `parentid` I was thinking perhaps we could go just with `type: parent` and `type: child` or maybe `tree` and `child` .. just thought that  might save some typos in people's day to day use but any other thoughts?

bobdenotter:
I'll happily help out on the UI side of things! Let's make proper work of this, when the Storage PR lands! 👍

graham73may:
The Storage PR is tagged as 3.6.... Any chance this would fall under 3.6 too? 👼

JarJak:
@bobdenotter @rossriley how about implementing this one in v4?

PR #7230, opened 2017-12-03
Only use `news_variable` from `config.yml`, if it's actually present in the news feed.

### LINKED ISSUE #7229: [3.4] backend breaks on 'dashboardnews' if people are not paying attention very well to what they're doing.
If you're not paying attention, it's pretty easy to break the bolt backend. 

To reproduce: 

1) Set up a default Bolt instance, using your preferred method. 

2) Think to yourself "I'd like to set the branding options!"

3) Edit config.yml, and uncomment the following section from the default configuration: 

```yaml
branding:
    name: SuperCMS
    path: /admin
    provided_by: [ supercool@example.org, "Supercool Webdesign Co." ]
    news_source: http://news.bolt.cm
    news_variable: news
```

4) Visit the bolt backend: 

screen shot 2017-12-03 at 17 28 59

bobdenotter:
Fixes #7229.

Note: I'm using `json_decode($fetchedNewsData, true)` because it seems that it's dependant on the format of the feed whether or it's an object or an array. Using `true` makes sure we get an array, and the check for `$newsVariable` doesn't break.

### DIFF
--- a/src/Controller/Async/General.php
+++ b/src/Controller/Async/General.php
@@ -3,6 +3,8 @@
 namespace Bolt\Controller\Async;
 
 use Bolt;
+use Bolt\Common\Exception\ParseException;
+use Bolt\Common\Json;
 use Bolt\Extension\ExtensionInterface;
 use Bolt\Filesystem;
 use Bolt\Storage\Entity;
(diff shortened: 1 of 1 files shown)


src/Controller/Async/General.php
```diff
@@ -393,9 +393,11 @@ private function fetchNews($hostname)
             ];
         }
 
-        $fetchedNewsItems = json_decode($fetchedNewsData);
-        if ($newsVariable = $this->getOption('general/branding/news_variable')) {
-            $fetchedNewsItems = $fetchedNewsItems->$newsVariable;
+        $fetchedNewsItems = json_decode($fetchedNewsData, true);
```
GawainLynch: Oh dear … it seems I missed this one in the `bolt/common` cutover … or I derped a merge :disappointed: 

`$fetchedNewsItems = \Bolt\Common\Json::parse($fetchedNewsData)` will do the same job, plus it will throw a catchable exception if there is _any_ problem in the decoding … at which point (the catch) you can return something friendly to the end user.

bobdenotter: Ah, derp! I knew that.. 

I'll work on this PR some more.

GawainLynch: Oh, it is totally not your fault here … mine if anyone's.

bobdenotter: Regardless, it's a good pointer to improve the current PR, so i'll work on it. ;-)

bobdenotter:
Updated.

PR #7242, opened 2017-12-07
add extra check to test block exists and matches

### LINKED ISSUE #7232: [3.4.2] Blocks in back-end don't respect grouping order
If you have a bunch of blocks, then re-order and save... and then refresh, the order in the back-end is not the same as in the database. It seems it _is_ correct when they are iterated (for-loop) in Twig / front-end.

Details
-------

 - PHP version: [ 5.6 ]
 - Used web server: [ Apache ] version [ Apache/2.2.22 (Debian) ]

Reproduction
------------

1. Use Contenttype + Blocks
2. Create a record of that Contenttype
3. Shuffle some blocks up and down
4. Save and refresh

See this image:

blocks-order

rossriley:
Fixes #7232 

As described in the issue, block types where the fields had the same names as other blocks were being cross written to. This was a bug in the code that selected the correct entity to update.

This fix checks to see if the block exists and if so ensures that the new value matches the old one.

### DIFF
--- a/src/Storage/Field/Collection/RepeatingFieldCollection.php
+++ b/src/Storage/Field/Collection/RepeatingFieldCollection.php
@@ -122,7 +122,8 @@ class RepeatingFieldCollection extends ArrayCollection
             if (
                 $existing->getName() == $entity->getName() &&
                 $existing->getGrouping() == $entity->getGrouping() &&
-                $existing->getFieldname() == $entity->getFieldname()
+                $existing->getFieldname() == $entity->getFieldname() &&
+                (!$existing->getBlock() || $existing->getBlock() == $entity->getBlock())
             ) {
                 return $existing;
             }


xiaohutai:
@rossriley 

This does not work for me. It does keep the values in the correct block type, but the ordering is a bit off.

My scenario, with pages contenttype from #7232

- New Page
- Add blocks "One", "Two", "Three", "Four", "Five"
- Fill in "1" in fields of "One", "2" in "Two", etc.
- Save, refresh, confirm: 1,2,3,4,5 ordering
- Now, re-sort by moving up and down, so that it becomes 5,4,3,2,1
- Save, refresh, I get: 3,5,4,2,1


But in MySQL:

ordering-in-mysql

It seems it takes the ordering of the rows, instead of the grouping value?

xiaohutai:
To clarify, this pull request fixes the data loss issue. But in my earlier comment, I noticed a _different_ issue (with the ordering), should I open a new issue for that so that this can be merged?

GawainLynch:
Oh wow, this didn't go in … totally off our game … is it December and snowing or something?

GawainLynch:
Ugh … and yes please, @xiaohutai

Issue #7241, opened 2017-12-06
[BUG][3.4] HTML field's data inside a repeater field is not getting duplicated


JarJak:
Details
-------

 - Relevant Bolt Version: 3.4 and up (it worked in 3.3)
 - Browser: Chrome

Reproduction
------------

### Bug summary

In repeater field when I have HTML field and click 'duplicate' then this field's data is not being copied.

### Specifics

You can check it in http://bolt-au46k8.bolt.dockerfly.com/bolt/editcontent/showcases/9#repeater

### Steps to reproduce

1. fill HTML field in repeater with some text
2. click duplicate

### Expected result

Field's data is getting copied.

### Actual result

Duplicated repeater has empty HTML field.

GawainLynch:
Confirmed. It seems to be the CK event derping, as a saved HTML field will duplicate.

PR #7261, opened 2017-12-17
Fix for 7241: Check for CKE changes on repeater duplication

jkazimir:
Check for CKE changes on repeater duplication.

Fixes: #7241 

Details
-------

CKEditor is "fun" in that it doesn't update the value of the textarea that spawns it, however in some cases may load instances later, where the data in the textarea can be expected to be up to date. This switches from simply grabbing the input value, to first checking if it's a CKE field and if so looking for an instance with the potentially updated data. If there's no instance yet or if it's simply not a CKE field, it then can safely just use the value.

### DIFF
--- a/app/src/js/widgets/fields/fieldRepeater.js
+++ b/app/src/js/widgets/fields/fieldRepeater.js
@@ -91,8 +91,15 @@
 
                 // Copy values from source to new set.
                 $(setToDuplicate).find(":input").each(function (index) {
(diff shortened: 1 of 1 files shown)


JarJak:
@GawainLynch can you merge it to 3.5 branch too?

GawainLynch:
There is a regular cascade on Mondays

Issue #7262, opened 2017-12-18
[BUG][3.4.4] Not all records are shown in the back-end


xiaohutai:
Records are missing in the Bolt back-end UI.

Details
-------

 - Relevant Bolt Version: [ 3.4.4 ] (or 3.4 branch as of writing)
 - Install type: [ GitHub checkout ]
 - PHP version: [ 5.6 ]
 - Used web server: [ Apache ]

Reproduction
------------

- Install Bolt
- Add some sample [Records wtih Loripsum text] for the `pages` contenttype (perhaps a few times)
  - Make sure to check how many records you have added
- Navigate to `/bolt/overview/pages` and see a few records missing
  - Note: the records are still in the database
  - Note: If you go directly to `/bolt/editcontent/pages/{id}`, you can still edit it

(Might be related to #7257)

xiaohutai:
If there are multiple sortorders with the same number (e.g. `0`), then it will only show one of those items (is that because of `ksort`?)

PR #7263, opened 2017-12-18
Fix proper sortorders for groupings

xiaohutai:
Fixes: #7262

### DIFF
--- a/src/Storage/ContentRequest/Listing.php
+++ b/src/Storage/ContentRequest/Listing.php
@@ -153,16 +153,18 @@ class Listing
             return $results;
         }
         $grouped = [];
+        $resultTaxOrders = [];
         foreach ($results as $result) {
             $taxGroup = null;
             foreach ($result->getTaxonomy() as $taxonomy) {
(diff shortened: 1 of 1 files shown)


xiaohutai:
This works for me, but please check if my train of thought is good :)

src/Storage/ContentRequest/Listing.php
```diff
@@ -153,16 +153,18 @@ protected function runGroupSort($results)
...
         }
         $grouped = [];
+        $resultTaxOrders = [];
         foreach ($results as $result) {
             $taxGroup = null;
             foreach ($result->getTaxonomy() as $taxonomy) {
                 if ($taxonomy->getTaxonomytype() == $result->getTaxonomy()->getGroupingTaxonomy()) {
                     $taxGroup = $taxonomy->getSlug();
                     $taxOrder = $taxonomy->getSortorder();
+                    $resultTaxOrders[$result->getId()] = $taxOrder;
```
xiaohutai: Not sure if we can get varying contenttypes here, then this part would be troublesome for getting the correct sortorder.

i.e. `pages/1` and `entries/1` in the results

Issue #7247, opened 2017-12-11
Extension without ReadMe, gives json error on 'readme' button


anketwokings:
On the extension page:
When clicking the 'View Readme' button on an extension that does not have a `ReadMe`, it gives a json error:

```
An unknown error occurred. This was the error message:
JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

Could this be changed into a friendly error, such as 'No ReadMe available' ? Or even better, hide the ReadMe button?

(And yes, I will ping / make a pull request for the developer of the extension to include a ReadMe, but there should be a good empty-state)

Thx!
Okbye!

Details
-------

 - Relevant Bolt Version: 3.4

bobdenotter:
"Hiding the button" is really tricky, but i _have_ improved the message shown in PR #7267.

Closing this for now. Feel free to re-open, if it needs more tweaking.

PR #7267, opened 2017-12-20
Improve upon the dreaded `unexpected token <` in Extensions screen.

bobdenotter:
If you've used the Extensions screen, you've undoubtedly seen this: 

screen shot 2017-12-20 at 12 08 07

There are two main reasons why this happens: 

 - The result is an exception, and we should return proper JSON
 - The result is an exception / error we couldn't catch or have no control over.

This PR won't fix it for all cases, but it improves on this behaviour for the end user in two distinct ways: 

1) For async requests to our own Bolt code, return a JSON-formatted error. 
(note, i've changed the structure, to match [what's used elsewhere](https://github.com/bolt/bolt/blob/3.4/src/Controller/Backend/Extend.php#L496-L503).)

screen shot 2017-12-20 at 13 01 24

2) When the result is _not_ JSON, it displays it as plain text, instead of the JSON parse error.

screen shot 2017-12-20 at 12 33 42

### DIFF
--- a/app/src/js/modules/extend.js
+++ b/app/src/js/modules/extend.js
@@ -26,6 +26,12 @@
         return $('.extend-bolt-container').find(selector);
     }
 
+    function stripHTML(string) {
+        var tempElement = $(string);
+        var result = tempElement.find('h1, h2, h3, p').text() || tempElement.text();
+        return result.trim().replace(/ +/g, " ").replace(/\s+\n/g, "\n");
+    }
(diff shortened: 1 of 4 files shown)


bobdenotter:
Kinda, sorta, fixes #7247, by giving a better message. 

And thanks to @GawainLynch for the initial fixes in `ExceptionToJsonListener.php`.

PR #7272, opened 2017-12-23
Update sortable plugin to work with latest version of select2

rossriley:
Couple of small tweaks to the select2sortable plugin to work with the slightly altered dom layout of the latest version.

### DIFF
--- a/app/src/lib/select2-ext/select2.sortable.js
+++ b/app/src/lib/select2-ext/select2.sortable.js
@@ -8,13 +8,13 @@
                     return undefined;
                 }
             });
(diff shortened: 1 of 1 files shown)


bobdenotter:
For future reference, if you have a field like this: 

```
        multiselect:
            type: select
            values: [ A-tuin, Donatello, Rafael, Leonardo, Michelangelo, Koopa, Squirtle, Foo, Bar ]
            sortable: true
```

Re-ordering the items gives errors in the console:

screen shot 2017-12-23 at 16 09 13

Items _seem_ sorted in the backend UI, but those changes aren't persisted on save. 

This PR fixes the issue, and makes them behave as they should.

PR #7278, opened 2017-12-31
[3.5] Add "embed" field type

### LINKED ISSUE #6292: [RFC] Possible to add youtube parameters through oEmbed?
Hi all, 

I use the video-field to embed a video. Since I do not want youtube to serve me related video's (rel=0) or the title (showinfo=0) I added the parameters to the video.
`https://www.youtube.com/watch?v=GgO2HkVGQmw&showinfo=0&rel=0`

This is, as @bobdenotter explained, ignored by embedly, that just serves a video with that ID.

On the embedly docs I found that it might support custom parameters: http://support.embed.ly/hc/en-us/articles/204266045-Custom-Video-Parameters

Could someone look into this and see if this is something that could be added to the bolt-embedly implementation? I think it would really be appreciated by a lot of people.

TLDR:
Working on a government site and I don't want to show the related video's after the video of the secretary of health has played. ;)

​​---------
 
- Relevant Bolt Version: [ 3.2 ]

### LINKED ISSUE #6928: [feature request] - oEmbed field type
Hi, 

Video field type is great but that's a shame it has been limited to video only. 

oEmbed isn't only meant for videos. oEmbed can also provide some data and HTML pieces for tweets, bandcamp players, google maps, and so many providers. 

Could a more global oEmbed field be available in core?

GawainLynch:
Someone … somewhere … wanted this :wink: 

Closes #6928
Closes #6292


ToBe998:
Just a small note as i stumbled over this too. The values a oEmbed service returns are completely proprietary. The only thing that "most" services share is the `html` field, but even this isnt available in general.
So, a field like this will only work if there is some configurable/extendable per-service logic to generate the html embed block. One size-fit's-all is sadly not possible with oEmbed.

GawainLynch:
Yes, that is correct. But what the `embed/embed` library does give us is an interface that we can use that gives us some fields, so that is what we ran with.

PR #7291, opened 2018-01-11
[3.6] More improvements to New Storage Layer Compatibility

rossriley:
This is targeted at installs where legacy `setcontent` is turned off so the Content objects are filled using the new storage layer.

As discussed in #6985 methods on the content object itself are moved to twig functions and filters with a compatibility patch used to avoid things breaking until 4.0.

There's also some ongoing work on performance here, targeted at closing some performance problems with content object hydration.

Existing systems using the legacy setcontent system will continue to work as before.

### DIFF
--- a/src/Controller/Base.php
+++ b/src/Controller/Base.php
@@ -424,8 +424,11 @@ abstract class Base implements ControllerProviderInterface
             return $this->storage()->getContent($textQuery, $parameters, $pager, $whereParameters);
         }
         $params = array_merge($parameters, $whereParameters);
+        unset($params['log_not_found']); // New storage system removes this functionality from the query engine
 
-        return  $this->app['query']->getContent($textQuery, $params);
+        return $this->app['twig.records.view']->createView(
+            $this->app['query']->getContentByScope('frontend', $textQuery, $params)
+        );
     }
 
     /**
--- a/src/Storage/Query/Query.php
+++ b/src/Storage/Query/Query.php
@@ -3,6 +3,7 @@
 namespace Bolt\Storage\Query;
 
 use Bolt\Storage\Entity\Content;
+use Bolt\Twig\TwigRecordsView;
 
 class Query
 {
@@ -10,15 +11,18 @@ class Query
     protected $parser;
     /** @var array */
     protected $scopes;
+    /** @var TwigRecordsView */
+    protected $recordsView;
 
     /**
      * Constructor.
      *
      * @param ContentQueryParser $parser
      */
-    public function __construct(ContentQueryParser $parser)
+    public function __construct(ContentQueryParser $parser, TwigRecordsView $recordsView)
     {
         $this->parser = $parser;
+        $this->recordsView = $recordsView;
     }
 
     /**
@@ -81,5 +85,20 @@ class Query
         }
 
         return null;
+    }
+
+    /**
+     * Helper to be called from Twig that is passed via a TwigRecordsView rather than the raw records.
+     *
+     * @param $textquery
+     * @param array $parameters
+     *
+     * @return QueryResultset|null
+     */
+    public function getContentForTwig($textquery, $parameters = [])
+    {
+        return $this->recordsView->createView(
+            $this->getContentByScope('frontend', $textquery, $parameters)
+        );
     }
 }
--- a/src/Twig/Extension/ArrayExtension.php
+++ b/src/Twig/Extension/ArrayExtension.php
@@ -58,8 +58,12 @@ class ArrayExtension extends AbstractExtension
      *
      * @return array
      */
-    public function order(array $array, $on, $onSecondary = null)
+    public function order($array, $on, $onSecondary = null)
     {
+        // We only want to work with real arrays, anything else is ignored
+        if (!is_array($array)) {
+            return $array;
+        }
         // If we don't get a string, we can't determine a sort order.
         if (!is_string($on)) {
             throw new \InvalidArgumentException(sprintf('Second parameter passed to %s must be a string, %s given', __METHOD__, gettype($on)));
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -8,6 +8,8 @@ use Bolt\Asset\Snippet\Snippet;
 use Bolt\Asset\Target;
 use Bolt\Helpers\Input;
 use Bolt\Response\TemplateResponse;
+use Bolt\Storage\Entity\Taxonomy;
+use Bolt\Storage\Repository\TaxonomyRepository;
 use Bolt\Translation\Translator as Trans;
 use Silex\ControllerCollection;
 use Symfony\Component\HttpFoundation\RedirectResponse;
@@ -186,6 +188,8 @@ class Frontend extends ConfigurableBase
      * @param Request $request         The Symfony Request
      * @param string  $contenttypeslug The content type slug
      *
+     * @throws \Exception
+     *
      * @return TemplateResponse
      */
     public function preview(Request $request, $contenttypeslug)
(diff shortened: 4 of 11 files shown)


rossriley:
OK, good to go.

For testing this remember you need to make the below addition to config.yml

```
compatibility:
    setcontent_legacy: false
```

bobdenotter:
I've noticed that "previewing" with Blocks / Repeaters is broken: 

screen shot 2018-01-22 at 20 59 56

This is something I can(should) fix in the templates themselves, though. Will do that after this is merged in. 

Other than that, i didn't bump into any quirks yet, so that's great. I'll update a non-trivial site tomorrow, and see if it works smoothly, too. 👍

bobdenotter (dismissed):

rossriley:
@bob I've had a bit of an attempt at fixing this in the storage layer but because we need to get it to work in both legacy and new storage I'm thinking that you were probably correct in suggesting fixing this at the template level.

We have a method in the `storage.metadata` service `getFieldTypeFor($contentType, $column, $field = null)` which pulls back the field type info for any ct/field combination and optionally with the third parameter for the sub-field of a repeater.

What I would suggest is that we expose this method via a Twig function and then rather than calling: `{% set fieldtype = repeaterfield.fieldtype() %}` as we currently do in the _sub_fields template (which fails when repeaterfield is an array) we call the new function using the contenttype/field/repeatername and avoid this error.

I can either add this to this PR or we can handle separately, whatever works best.

bobdenotter:
> What I would suggest is that we expose this method via a Twig function

Sounds good to me. If you add it to this PR, i'll update the themes accordingly.

rossriley:
ok, the new filter/function is included here and I've tested with the base-2016 theme and all seems to be ok. Here's what needs to be modified...

Firstly in the outer loop: `{% for key, value in record.values if (key not in omittedkeys) %}`
any call to `record.fieldtype(key)` simpley gets changed to `record|fieldtype(key)`

Secondly in the inner loop for repeaters and blocks we need to keep access to the original key variable so one minor change to the for loop: `{% for repeatkey, repeaterfield in repeater %}` that is using repeatkey instead of key.

Then final changes, to get the fieldtype now in this repeater loop: 
`{% set fieldtype = record|fieldtype(key, repeatkey) %}` same as above except we also pass in the sub-field name.

Finally to get the value we need to check whether we have an object or array so the final block looks like:
```
                {% for repeater in value %}
                    {% for repeatkey, repeaterfield in repeater %}

                        {% set fieldtype = record|fieldtype(key, repeatkey) %}
                        {% if attribute(repeaterfield, 'getValue') is defined %}
                            {% set value = repeaterfield.value() %}
                        {% else %}
                            {% set value = repeaterfield %}
                        {% endif %}
                        {{ block('common_fields') }}
                        {{ block('extended_fields') }}

                    {% endfor %}
                {% endfor %}
```

JarJak:
I get:
```
Exception in Storage.php line 78:
\MyExtension\CustomContentClass does not extend \Bolt\Legacy\Content.
```
Looks like custom content classes still need to extend legacy :(

It happens because new EntityManager uses old Storage, which does callback to itself in line 1627 (callback is defined in line 1055).

JarJak:
And if I extend \Bolt\Legacy\Content then I get:
`An exception has been thrown during the rendering of a template ("Notice: Undefined variable: pager")`
in line:
```
{% setcontent url = ct.slug where {ID: record.relation.urls|first} returnsingle %}
```

rossriley:
@JarJak Once this is merged into 3.6 I'll revisit, there's been some changes to the default theme since and I'm guessing a few things have broken since.

JarJak:
@rossriley I don't think this can be merged, it is still buggy. And I don't think these bugs have anything in common with theme updates.

GwendolenLynch:
> I don't think this can be merged,

Yeah, we're still in review phase and have talked about this PR at pretty much every dev meeting this year :laughing:

JarJak:
I know :D But I wanted to say that it needs to be revisited before merge,
not after.

src/Storage/Query/Query.php
```diff
@@ -82,4 +86,19 @@ public function getContentByScope($scopeName, $textquery, $parameters = [])
...
+
+    /**
+     * Helper to be called from Twig that is passed via a TwigRecordsView rather than the raw records.
+     *
+     * @param $textquery
+     * @param array $parameters
+     *
+     * @return QueryResultset|null
+     */
+    public function getContentForTwig($textquery, $parameters = [])
```
JarJak, line 98: why view and model layers are mixed here?

src/Twig/Extension/ArrayExtension.php
```diff
@@ -58,8 +58,12 @@ public function getFilters()
      *
      * @return array
      */
-    public function order(array $array, $on, $onSecondary = null)
+    public function order($array, $on, $onSecondary = null)
     {
+        // We only want to work with real arrays, anything else is ignored
+        if (!is_array($array)) {
+            return $array;
```
JarJak, line 65: silencing invalid argument? what's the purpose of this?

src/Controller/Base.php
```diff
@@ -424,8 +424,11 @@ protected function getContent($textQuery, $parameters = [], &$pager = [], $where
             return $this->storage()->getContent($textQuery, $parameters, $pager, $whereParameters);
         }
         $params = array_merge($parameters, $whereParameters);
+        unset($params['log_not_found']); // New storage system removes this functionality from the query engine
```
JarJak, line 427: 1. Is this a BC Break?
2. Docs and config samples should be updated too.
3. I'm 99% sure that this param is also used somewhere else.

JarJak (changes requested):

src/Controller/Base.php
GwendolenLynch: > 3. I'm 99% sure that this param is also used somewhere else.

```
$ ag log_not_found
src/Controller/Frontend.php
152:        $content = $this->getContent($contenttype['slug'], ['slug' => $slug, 'returnsingle' => true, 'log_not_found' => !is_numeric($slug)]);

src/Legacy/Storage.php
1602:        if (isset($parameters['log_not_found'])) {
1603:            $logNotFound = $parameters['log_not_found'];
1604:            unset($parameters['log_not_found']);
```

In the case of the controller, that is only going to legacy, and legacy is legacy.

rossriley: Yes, this is only a legacy feature and I don't think the query parser is the right place for it. We have HTTP status exceptions and they can look after logging 404 errors if so needed. As with everything in this PR legacy stays exactly as is and we can decide for 4.0 onwards what we want to keep and drop. If in this testing phase a feature is found to be missing, then we will happily add it in, but there's no need to automatically include existing features if we think they don't fit.

JarJak: Oh ok, now I can see that this is unset only for new Storage, legacy one still receives this parameter. 👍

JarJak (approved):

JarJak (changes requested):
just reverting approval

GwendolenLynch (approved):
Let's get this in and under test :+1:

JarJak:
@GawainLynch I see this got merged without changes that I've requested and without any explanation, why is that?

GwendolenLynch:
Your opinions/reviews are **very much** welcomed, and considered, but you are not a core team member, so not in possession of a blocking vote. :wink: 

Further, the weekly meetings are a much better place to go over finer details. While I agree with some of your review comments, they also partially miss some of what needs to be balanced out to get this across the line :smile: 

That said, PRs to RFC additional changes are also welcome. But the core team members are heavily stretched in RL this year, so "keeping things moving" is important.

JarJak:
Thanks for explanation. I'm fine with that I'm not a decider ;) just wanted to know why it got merged without replies.

I will test it in real project and report issues if any of them will pop out :)

GwendolenLynch:
> I will test it in real project and report issues if any of them will pop out :)

Please! All help & bug reports/PRs appreciated.

bobdenotter:
> Further, the weekly meetings are a much better place to go over finer details. While I agree with some of your review comments, they also partially miss some of what needs to be balanced out to get this across the line 😄

If you want to, you're free to pop in tomorrow (tuesdays, 19:30 CET), to discuss additional changes with us! :+1:

Issue #6073, opened 2016-11-23
[RFC] More advanced getContent queries


rossriley:
As discussed in the dev meeting yesterday we need to finalize the syntax spec for doing more complex searches, primarily at this stage on relations to content and also on repeater fields but this will also apply to any of the new fields we have planned like named blocks etc.

There's so far been two proposals for syntax, but this is an ideal time to propose any more or tweaks to these as set out.

## Option One

As we do in the contenttype part of the query use a slash as a separator eg:

`$records = $app['query']->getContent('pages', ['relations' => 'entries/1' ])`

The equivalent in Twig when it moves to the new engine:

`{% setcontent records = 'pages' where {'relations': 'entries/1'}`

## Option Two

Allow arrays to be passed as arguments instead of using a string eg:

`$records = $app['query']->getContent('pages', ['relations': ['entries' => '1']])`

The equivalent in Twig when it moves to the new engine:

`{% setcontent records = 'pages' where {'relations': {'entries':'1'}}`

## Option Three

Is to implement both syntaxes and allow users to choose their preferred method.

## Edge Cases and Considerations

Initial comments were that option one seems most intuitive to users of the current syntax. One concern of using it is that without a workaround we may lose the ability to include the `/` character in a search and although that is currently the case with the supported operators, they are a little less common than the `/` character (eg: `'!1'`, `'%searchterm%'` etc). This could perhaps be mitigated a little by doing the transform at a field level, so searches on a text field would receive a single string whereas the field types that want to implement nested search can do the splitting.

Option two is basically what option one will be compiled to, so there are no clashes, although it might give the impression of two parallel syntaxes since we would be supporting strings on some types of query and arrays on others.

graham73may:
From my experience with this recently, I would prefer option two. Of course I would be happy with option three, but that's because it allows me to use option two. 

I do have some questions (or maybe just things to take into consideration)... Would passing arrays allow for multiple values, e.g.:

```
$records = $app['query']->getContent('pages', ['relations': ['entries' => ['1', '2']]]);
```

How would you go about looking up a specific contenttype and id? For example, find me any page, news or event which has a relation to event/1?

```
$records = $app['query']->getContent(('pages', 'news', 'events'), ['relations': ['entries' => 'event/1']])
```

The use case in my head is for featured content, related content, or something such as finding all contenttypes where author = 'people/John Smith'

graham73may:
[...]

PR #7311, opened 2018-01-23
Add missing repeater search to query parser

rossriley:
This adds a missing part of the query syntax which was originally talked about in #6073.

The brief summary is that you can now add subfield parameters to `getContent` searches assuming you are using the new storage engine.

The syntax is as follows:

```php
$res = $app['query']->getContent('showcases', ['repeater' => ['repeattitle'=> '%test%', 'repeatcontent'=>'%testing%']]);
```

By way of explanation. `repeater` is the main field name and `repeattitle` is the sub-field name. The result is that showcase records will only be selected if they have a repeater/repeattitle field that contains `test` _AND_ a repeater/repeatcontent field that contains `testing`.

There are a few limitations in this initial version all of which can probably be improved upon with a bit more time and creativity. 

1. There is no support for the multiple key/value filters (eg: `field1 || field2 => 'val1 || val2'`)
2. By extension there's no way to do OR queries on multiple repeater fields
3. Only string and text fields are supported initially, this is purely a case of time to test and others can be added shortly although it's almost certain that json_array fields will not be supported until we safely start introducing native engine json queries.

### DIFF
--- a/src/Storage/Field/Type/RepeaterType.php
+++ b/src/Storage/Field/Type/RepeaterType.php
@@ -7,6 +7,10 @@ use Bolt\Exception\FieldConfigurationException;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Field\Collection\RepeatingFieldCollection;
 use Bolt\Storage\Mapping\ClassMetadata;
+use Bolt\Storage\Query\Filter;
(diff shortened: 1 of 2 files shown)


bobdenotter:
FWIW, looks good to me. 👍

PR #7314, opened 2018-01-24
Allow grouping for other taxonomies than only the first one that has `behaves_like: grouping`.

### LINKED ISSUE #7313: Grouping by a "grouping" taxonomy only works if the selected taxonomy is the _first_ one that has `behaves_like: grouping`
See screenshot. 

screen_shot_2018-01-24_at_15_46_13

Fix incoming.

bobdenotter:
Fixes #7313

### DIFF
--- a/src/Storage/Collection/Taxonomy.php
+++ b/src/Storage/Collection/Taxonomy.php
@@ -2,6 +2,7 @@
 
 namespace Bolt\Storage\Collection;
 
+use Bolt\Common\Deprecated;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Mapping\MetadataDriver;
 use Closure;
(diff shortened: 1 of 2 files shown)


src/Storage/Collection/Taxonomy.php
```diff
@@ -256,6 +257,8 @@ public function serialize()
      */
     public function getGroupingTaxonomy()
     {
+        Deprecated::method(3.4);
```
GawainLynch: `Deprecated::method(3.4, 'getGroupingTaxonomies');`

src/Storage/Collection/Taxonomy.php
```diff
@@ -264,4 +267,20 @@ public function getGroupingTaxonomy()
 
         return null;
     }
+
+    /**
+     * @return null|string
```
GawainLynch: `@return array|dropbear`

PR #7315, opened 2018-01-28
[3.4] Update handling of user form data

### LINKED ISSUE #7309: user level is changed when admin-level user edits own password
## Issue
If a user edits their own password, their user level changed. Thus making them no longer admin, but only editor. 

## Steps to reproduce

- create a new user (let's call him Frank)
- make this user an admin, chief-editor and editor 
- logout
- login as user Frank
- go to config > users and let Frank edit his own password. NOTE: the admin user-level is greyed out, with a disabled checkbox (see screenshot)
- Don't do anything with the user levels, and save the user.
- Tadaaa! no longer Admin.  (insert muffled cries)

screenshot

### what goes wrong? 
The disabled admin level is not saved

## Expected result
- When Frank edits his own password, the users levels should stay exactly as they were.

( He shouldn't have to call me to ask why he suddenly can't see the settings anymore, and make me have to file a bugreport :) )

## specs
Bolt 3.4.6

GawainLynch:
Fixes #7309

There are two key things here, that I really wanted to tackle ages ago, but thought we could work around in the meantime … Namely, add a list mutator, and use a DTO on the user forms

This mutator is constructed with a list of available possible items, and a second sub-list of items that can be mutated.

The object can then be invoked with two parameters, first is the original list contents, the second being proposed changes.

The returned list will:
- contain only items that are in the 'available' list
- contain any items that appear in both original and proposed lists
- gain only available items that are also mutatable
- lose only items that are mutatable

### DIFF
--- a/src/Controller/Backend/Users.php
+++ b/src/Controller/Backend/Users.php
@@ -5,6 +5,7 @@ namespace Bolt\Controller\Backend;
 use Bolt\AccessControl\Permissions;
 use Bolt\Events\AccessControlEvent;
 use Bolt\Form\FormType;
+use Bolt\Helpers\ListMutator;
 use Bolt\Storage\Entity;
 use Bolt\Translation\Translator as Trans;
 use Silex\ControllerCollection;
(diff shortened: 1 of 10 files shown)


bobdenotter:
screen shot 2018-01-29 at 08 34 46

Works exactly as expected. 👍

PR #7321, opened 2018-02-01
[3.4] Add return early for Postgres sequence look up failures

GawainLynch:
Something … something … :zipper_mouth_face:


bobdenotter:
👍

PR #7336, opened 2018-02-12
[3.6][WIP] Changelog fixes / touch-ups

bobdenotter:
WIP: I'll need to make sure changes in repeaters/blocks are logged.

### DIFF
--- a/app/view/twig/changelog/_blocks.twig
+++ b/app/view/twig/changelog/_blocks.twig
@@ -9,7 +9,7 @@
     <h3>
         "<a href="{{ path('editcontent', {'contenttypeslug': context.entry.contenttype, 'id': context.entry.contentid}) }}">{{ context.entry.title }}</a>"
         <em><span class="time">{{ changelog_mutation(context.entry.mutation_type) }} {{ buic_moment(context.entry.date) }}</span>
-            {{ __('general.phrase.by') }}</em> <strong>{{ userlink(entry.ownerid|default('')) }}</strong>
+            {{ __('general.phrase.by') }}</em> <strong>{{ userlink(context.entry.ownerid|default('')) }}</strong>
     </h3>
 
     <div class="changelog-details">
@@ -38,16 +38,16 @@
                     <tr>
                         <td><strong>{{ field.label }}</strong></td>
 
-                        {% if field.type == 'html' %}
+                        {% if field.type in ['html', 'text', 'integer', 'float', 'slug', 'file' ] %}
                             <td class="change-log-item">
-                                {{ field.before.render|raw }}
+                                {{ field.before|raw }}
                             </td>
                             <td class="change-log-item">
-                                {{ field.after.render|raw }}
+                                {{ field.after|raw }}
                             </td>
                         {% elseif field.type == 'filelist' %}
                             <td class="change-log-item">
-                                {% for file in field.before.render %}
+                                {% for file in field.before %}
                                     <strong>File:</strong> {{ file.filename }}<br>
                                     <strong>Title:</strong> {{ file.title }}
 
@@ -57,7 +57,7 @@
                                 {% endfor %}
                             </td>
                             <td class="change-log-item">
-                                {% for file in field.after.render %}
+                                {% for file in field.after %}
                                     <strong>File:</strong> {{ file.filename }}<br>
                                     <strong>Title:</strong> {{ file.title }}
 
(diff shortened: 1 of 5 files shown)


app/view/twig/changelog/_blocks.twig
```diff
@@ -188,10 +188,12 @@
...
                         {% else %}
                             <td class="change-log-item">
-                                {{ field.before.render }}
+                                {# {{ field.before.render }} #}
+                                {{ dump(field.before.render)}}
                             </td>
                             <td class="change-log-item">
-                                {{ field.after.render }}
+                                {# {{ field.after.render }} #}
+                                {{ dump(field.after.render)}}
```
GwendolenLynch: Erm … you sure about that?

src/Controller/Async/General.php
```diff
@@ -28,8 +28,8 @@ protected function addRoutes(ControllerCollection $c)
         $c->get('/', 'async')
             ->bind('async');
 
-        $c->get('/changelog/{contenttype}/{contentid}', 'changeLogRecord')
-            ->value('contenttype', '')
+        $c->get('/changelog/{contenttypeslug}/{contentid}', 'changeLogRecord')
+            ->value('contenttypeslug', '')
```
GwendolenLynch: BC break

src/Storage/Entity/LogChange.php
```diff
@@ -270,25 +270,25 @@ private function fieldList($key, $value, array $fields)
...
             'before' => [
                 'render' => [
-                    'address'           => $before['address'],
-                    'latitude'          => $before['latitude'],
-                    'longitude'         => $before['longitude'],
-                    'formatted_address' => $before['formatted_address'],
+                    'address'           => isset($before['address']) ? : '',
+                    'latitude'          => isset($before['latitude']) ? $before['latitude'] : '',
+                    'longitude'         => isset($before['longitude']) ? $before['longitude'] : '',
+                    'formatted_address' => isset($before['formatted_address']) ? $before['formatted_address'] : '',
```
GwendolenLynch: All this should be in a `Bag`

bobdenotter: Consider it done.

src/Controller/Async/General.php
bobdenotter: How come? It's only used in one place, and i'm not actually changing what's passed in, just what's done with it.

app/view/twig/changelog/_blocks.twig
bobdenotter: Nope, that's part of the WIP. It's there so it won't break on Repeaters/Blocks. I need to work on that.

src/Controller/Async/General.php
GwendolenLynch: It changes the router signature

src/Storage/Entity/LogChange.php
```diff
@@ -270,25 +270,25 @@ private function fieldList($key, $value, array $fields)
...
 
         return [
             'type'   => $fields[$key]['type'],
             'before' => [
                 'render' => [
-                    'address'           => $before['address'],
-                    'latitude'          => $before['latitude'],
-                    'longitude'         => $before['longitude'],
-                    'formatted_address' => $before['formatted_address'],
+                    'address'           => isset($before['address']) ? : '',
```
andrewtheone: shouldnt it be isset($before['address']) ?$before['address'] : '', ?

bobdenotter:
It's not perfect, but it's much better than it was. For one, it doesn't crash, and it handles repeaters now too.

screen shot 2018-03-07 at 16 54 25

bobdenotter:
Oh, and the target is changed from `3.4` to `3.5`.

xiaohutai:
Somehow it doesn't show the proper before and after changes for blocks and repeaters; I always get the same values left and right:

changelog--blocks-repeaters

**Edit: 2018-03-12**

It's really weird what's going on. Sometimes it shows both the old values, and somtimes both the new ones.

With repeaters/blocks, you always get to see those items, regardless of whether they have changed or not. Also taxonomy and relation are always showing up.
Try clicking save while not having fields modified. This may be related to abovementioned issue.

Note: It is because there is actually only 1 RepeatingFieldCollection, if it exists, it will start (lazily) fetch the same values from the database.

xiaohutai:
Still needs more investigation [and still a useful feature to have].

Could be fixed by serializing the before record/values (iterate over all values in lazy items)?

bobdenotter:
Yes, should absolutely keep this.. I've put it on hold, to fix up after #7291 has landed, because then we can fix it properly :-)

PR #7346, opened 2018-02-16
Update exception handling in NotFoundListener

### LINKED ISSUE #7345: if config notfound is incorrect - no explaining message on error
1. make incorrect `notfound: ...` field in config
2. log out and (or) turn off debug.
3. navigate to not existing page OR put `throw new NotFoundHttpException();` somewhere in controller

You'll get an error that doesn't tell you to fix notfound field.

Here's the code for that:
```php
        try {
            $html = $this->twig->resolveTemplate($template)->render($context);
            $event->setResponse(new Response($html, Response::HTTP_NOT_FOUND));
        } catch (RuntimeError $e) {
            throw new RuntimeException('Unable to render 404 page!', null, $e);
        }
```
but if notfound field is set to the template that is not found by twig (maybe twig path is not registered) `RuntimeError` will not fire.

Boorj:
fixes #7345

Now it catches the error in `->resolveTemplate()` and any in `->render()`

I can use `use` statement for that. Also i suppose that we can check if template exists and throw more explanatory exception, before rendering it in chain. 
```php
$template = $this->twig->resolveTemplate($template);

# checking if template is resolved here

$html = $template->render($context);
```

### DIFF
--- a/src/EventListener/NotFoundListener.php
+++ b/src/EventListener/NotFoundListener.php
@@ -11,6 +11,7 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Twig\Environment;
 use Twig\Error\LoaderError;
@@ -79,6 +80,9 @@ class NotFoundListener implements EventSubscriberInterface
 
         $content = $this->storage->getContent($this->notFoundPage, ['returnsingle' => true]);
         if (!$content instanceof Content || empty($content->id)) {
+            $msg = sprintf('No page could be shown, because the "notfound" setting "%s" in config.yml or theme.yml is not valid.', $this->notFoundPage);
+            $event->setException(new NotFoundHttpException($msg, $e));
+
             return;
         }
 


rossriley:
I'm just going to look at the test failures, but one minor code style issue, can you add use statements at the top for the Runtime / Exception classes and not use the `\` absolute reference.

GawainLynch:
This is very much not correct … what are you trying to achieve exactly @Boorj ?

rossriley:
I can potentially see the problem, the current code only catches a RuntimeError and at least according to PHPStorm that Twig resolveTemplate call throws other types of exceptions too.

GawainLynch:
Yes, it throws:
-  `Twig\Error\Loader` when the template can't be found (should not change or be caught here)
-  `Twig\Error\Syntax` when the template has a Twig syntax error (should not change or be caught here)

GawainLynch:
Catching an `\Exception` here will literally revert us to the shitfight we had in v2 with not knowing WTF was going on when there was a developer error

GawainLynch:
OK, so here is what is happening …
https://github.com/bolt/bolt/blob/d80d45053f7cf35c84ceba6b2829a19c54bdc663/src/EventListener/NotFoundListener.php#L71-L86

The call to `renderNotFound()` on line 72 throws a `LoaderError` as the template can't be found. This is 100% correct and desired behaviour.

The exception is caught and ignored, as per the comment:
```
             // Template not found, fall though to see if we can render a 
             // record, failing that let the exception handler take over 
```

A call is then made to `getContent()` that (obviously) can't find a record match, so it returns control to the exception handler, eventually re-throwing the `NotFoundHttpException`.

bobdenotter:
I agree, this is working as it should.. I can see how "fixing" it would make it "easier" for developers at first sight, but in the long run, such a "fix" would only obfuscate and hide problems that become less obvious. 

I propose we leave this as is.

GawainLynch:
Is it Monday or something :woman_facepalming: 

@bobdenotter & @Boorj I have updated the logic flow now so that a separate exception is added to the stack, and set as the response's exception object, you'll now get something along the lines of 

> The "notfound" template or record setting "foobar" is not valid. Correct the value set in your site config.yml file.

bobdenotter:
Will test! 👍

PR #7356, opened 2018-02-22
[3.4] Be consistent with session & IP address handling

### LINKED ISSUE #7355: IP address still used with cookies_use_remoteaddr: false
The documentation states that setting **cookies_use_remoteaddr** to **false** will disable session IP address checks. This is only true for first 600 seconds of the session.

https://github.com/bolt/bolt/blob/ab43c6cf4384da1a1370477ee0c0f6334c7e3f03/src/AccessControl/AccessChecker.php#L193

After that time period, session is re-validated against database data. If at that time IP address doesn't match the initially stored one, database query won't return any result and session will no longer be valid.

https://github.com/bolt/bolt/blob/ab43c6cf4384da1a1370477ee0c0f6334c7e3f03/src/AccessControl/AccessChecker.php#L213

This situation can very easily force the log out in the middle of page editing.

Possible resolutions:
1. Do not use IP address in `getToken()` method if **cookies_use_remoteaddr** is set to **false**
2. Change the 600 seconds expiration to 600 seconds of inactivity.
3. Add config option to adjust the 600 seconds expiration

Details
-------

 - Relevant Bolt Version: 3.4.7
 - Install type: Composer
 - PHP version: 7.2
 - Used web server: Nginx

GawainLynch:
Fixes #7355

:egg: :koala: :beers:

### DIFF
--- a/src/AccessControl/AccessChecker.php
+++ b/src/AccessControl/AccessChecker.php
@@ -174,7 +174,7 @@ class AccessChecker
         }
 
         $this->session->remove('authentication');
-        $this->session->migrate(true);
+        $this->session->migrate(true, 0);
 
         return false;
     }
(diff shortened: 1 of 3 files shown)


bobdenotter:
cny-8-giphy

PR #5366, opened 2016-05-24
Lock paragonie/random_compat at 1.x, to keep it from breaking shit.

bobdenotter:
Fixes #5361 🍻

PR #7359, opened 2018-02-28
Remove v1 version-lock on paragonie/random_compat

### LINKED ISSUE #7358: "Composer update" not working, because of security advisories
After [this](https://github.com/Roave/SecurityAdvisories/commit/1bd5e61029a7a7fbbfc709181f63ced00eea34cd) change in https://github.com/Roave/SecurityAdvisories/ 

We get an error on `composer update`: 

```
$ composer update                                                                                  (3.4)
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - roave/security-advisories dev-master conflicts with paragonie/random_compat[v1.4.0].
    - roave/security-advisories dev-master conflicts with paragonie/random_compat[v1.4.1].
    - roave/security-advisories dev-master conflicts with paragonie/random_compat[v1.4.2].
    - roave/security-advisories dev-master conflicts with paragonie/random_compat[v1.4.2].
    - Installation request for roave/security-advisories dev-master@dev -> satisfiable by roave/security-advisories[dev-master].
    - Installation request for paragonie/random_compat ^1.4 -> satisfiable by paragonie/random_compat[v1.4.0, v1.4.1, v1.4.2].
```

PR with fix incoming.

bobdenotter:
Fixes #7358.

🐵  Tested and found to be working on PHP 5.5.38, 5.6.31, 7.0.22 and 7.1.6 and 7.21.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,6 @@
         "ircmaxell/random-lib": "^1.1",
         "jdorn/sql-formatter": "^1.2",
         "monolog/monolog": "^1.12",
-        "paragonie/random_compat": "^1.4",
         "passwordlib/passwordlib": "^1.0@beta",
         "php": "^5.5.9 || ^7.0",
         "silex/silex": "^1.3",


SvanteRichter:
We should probably be clear in the docs that for windows we will require php-mcrypt from now on (since random-compat will fail otherwise.)

composer.json
```diff
@@ -40,7 +40,7 @@
         "ircmaxell/random-lib": "^1.1",
         "jdorn/sql-formatter": "^1.2",
         "monolog/monolog": "^1.12",
-        "paragonie/random_compat": "^1.4",
+        "paragonie/random_compat": "^2.0",
```
GwendolenLynch: If we go this way, this line can just be dropped … You added it in #5366 to version lock.

GwendolenLynch (changes requested):

GwendolenLynch:
OK, so we don't have a choice in this one, it is a security win IMHO. But the losers are PHP5 users on shared hosts, as it breaks Bolt 100% of the time if `open_basedir` is in effect … so people using incompetent shared hosts lose support when this is merged:

```
file_exists(): open_basedir restriction in effect. File(/dev/urandom) is not within the allowed path(s)
```

bobdenotter:
Hmm.. We should of course make changes, if they're dictated by security recommendations, but I'm not too keen on getting (perhaps) an increase on us having to explain to people "Yes, it was in the notes. No, it's not our fault, it's better. Yes, you should contact your host, even if they live under a rock." 

🤔 💭

GwendolenLynch (approved):

PR #7361, opened 2018-03-04
Add unicode word support for searches

### LINKED ISSUE #7246: [BUG][3.4]filtering content type in overview page utf-8 problem
Details
-------

 - PHP version: [ 7.0  ]
 - Used web server: [ Apache | Nginx ] version [ Apache/2.2.22 | nginx/1.11.10]

Reproduction
------------

### Bug summary

Trying to filtering contenttype in backend on  overview page cause problem when use utf-8 symbols such as `üöäß` or  `абырвалг`

### Specifics

there is an error

```
Uncaught Exception: QueryParseException.

QueryParseException in QueryParameterParser.php line 276: 
No matching value found for "%üöäß%"
```

### Steps to reproduce

- go to overview listing of contenttype in a backend
- type a search string (utf-8 chars,ex. üöäß) in a filter field
- press filter button

### Expected result

may be  to use **/u**  modifier for regexp pattern in `QueryParameterParser.php` `parseValue `method

rossriley:
Fixes #7246

Thanks to @Raxta for this, it seems as though the classification of characters that match `\w` can vary based on locale and possibly version of the pcre library, so this solution is much more solid.

### DIFF
--- a/src/Storage/Query/QueryParameterParser.php
+++ b/src/Storage/Query/QueryParameterParser.php
@@ -40,17 +40,19 @@ class QueryParameterParser
 
     public function setupDefaults()
     {
+        $word = "[\p{L]\p{N}_]+";
+
         // @codingStandardsIgnoreStart
-        $this->addValueMatcher('<(\w+)',            ['value' => '$1', 'operator' => 'lt']);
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Looks good to me! 👍

PR #7364, opened 2018-03-06
[3.5] Update ctrl+s, meta+s keybindings for newer version of plugin

bobdenotter:
Update keybindings, fixes issue where page saves on every click.


GwendolenLynch (approved):
WFM

PR #7368, opened 2018-03-08
[3.5] Add feature to fetch placeholder images from remote URLs

bobdenotter:
Tired of the same random placeholder images? Too few kittens? This PR will allow you to add new placeholder images to your contenttypes, when using the "add loripsum content": 

screen shot 2018-03-08 at 19 27 36

```
    fields:
        title:
            type: text
            class: large
            group: content
        …
        image:
            type: image
            group: media
            placeholder: https://source.unsplash.com/1600x900/?nature,water/__random__
```

or `https://source.unsplash.com/1600x900/?kitten/__random__` for more kittens. 

Should be trivial to make work with other placeholder services, like pexels, visualhunt or placekitten. Or we could even write our own, if we don't want to be reliant on a 3rd party service.

### DIFF
--- a/src/Storage/Database/Prefill/ApiClient.php
+++ b/src/Storage/Database/Prefill/ApiClient.php
@@ -3,6 +3,7 @@
 namespace Bolt\Storage\Database\Prefill;
 
 use GuzzleHttp\Client;
+use Psr\Http\Message\StreamInterface;
 
 /**
  * Handles fetching prefill text content from an API service.
@@ -35,5 +36,19 @@ class ApiClient
         $uri = $base . ltrim($request, '/');
 
         return $this->client->get($uri, ['timeout' => 10])->getBody();
+    }
+
+    /**
+     * Fetches an image from a remote source.
+     *
+     * @param string $url URL of the remote image to fetch
+     *
+     * @return StreamInterface
+     */
+    public function getImage($url)
+    {
+        $url = str_replace('__random__', rand(100000, 999999), $url);
+
+        return $this->client->get($url, ['timeout' => 10])->getBody();
     }
 }
--- a/src/Storage/Database/Prefill/RecordContentGenerator.php
+++ b/src/Storage/Database/Prefill/RecordContentGenerator.php
@@ -3,11 +3,13 @@
 namespace Bolt\Storage\Database\Prefill;
 
 use Bolt\Collection\ImmutableBag;
+use Bolt\Filesystem\Exception\IOException;
 use Bolt\Filesystem\FilesystemInterface;
 use Bolt\Storage\Collection;
 use Bolt\Storage\Entity;
 use Bolt\Storage\Repository\ContentRepository;
 use Cocur\Slugify\Slugify;
+use GuzzleHttp\Exception\RequestException;
 
 /**
  * Create a generated set of pre-filled records for a ContentType.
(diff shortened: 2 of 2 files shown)


src/Storage/Database/Prefill/RecordContentGenerator.php
```diff
@@ -453,9 +456,21 @@ private function getRandomTags($count = 5)
...
     {
         $pathKey = $type === 'image' ? 'file' : 'filename';
+
+        if ($placeholder) {
+            try {
+                $image = $this->apiClient->getImage($placeholder);
+                $filename = sprintf('placeholder_%s.jpg', substr(md5(microtime()), 0, 12));
+                $this->filesystem->write('files://' . $filename, $image);
+                return [$pathKey => $filename, 'title' => 'placeholder', 'alt' => 'placeholder'];
+            } catch (\Exception $e) {
```
GwendolenLynch: This is not a good idea, you don't want to catch everything silently, it makes it *extremely* hard to debug.

So …
- `getImage()`  can throw a `GuzzleHttp\Exception\RequestException` if there is a problem handling the call
- `write()` will throw a `Bolt\Filesystem\Exception\FileExistsException` that can be avoided by calling `put()`

If other exceptions are being thrown, we want to know about them as something else is fundamentally wrong.

src/Storage/Database/Prefill/RecordContentGenerator.php
```diff
@@ -453,9 +456,21 @@ private function getRandomTags($count = 5)
...
+    private function getRandomImage($type, $placeholder = null)
     {
         $pathKey = $type === 'image' ? 'file' : 'filename';
+
+        if ($placeholder) {
+            try {
+                $image = $this->apiClient->getImage($placeholder);
+                $filename = sprintf('placeholder_%s.jpg', substr(md5(microtime()), 0, 12));
+                $this->filesystem->write('files://' . $filename, $image);
+                return [$pathKey => $filename, 'title' => 'placeholder', 'alt' => 'placeholder'];
```
GwendolenLynch: PSR-2 requires a newline before then `return`

src/Storage/Database/Prefill/ApiClient.php
```diff
@@ -36,4 +36,17 @@ public function get($request, $base = 'http://loripsum.net/api/')
...
+     * Fetches an image from a remote source.
+     *
+     * @param string $url URL of the remote image to fetch
+     *
+     * @return StreamInterface
+     */
+    public function getImage($url)
+    {
+        $url = str_replace('__random__', rand(100000, 999999), $url);
+        return $this->client->get($url, ['timeout' => 10])->getBody();
```
GwendolenLynch: PSR-2 requires a newline before then return

src/Storage/Database/Prefill/RecordContentGenerator.php
```diff
@@ -278,12 +278,15 @@ private function addText(Entity\Content $contentEntity, $fieldName, $type, array
      */
     private function addJson(Entity\Content $contentEntity, $fieldName, $type)
     {
+        $contentType = $this->repository->getEntityManager()->getContentType($contentEntity->getContenttype());
+        $placeholder = isset($contentType['fields'][$fieldName]['placeholder']) ? $contentType['fields'][$fieldName]['placeholder'] : null;
+
         $value = null;
```
GwendolenLynch: While I am nagging, drop the newline before this one :stuck_out_tongue_winking_eye:

src/Storage/Database/Prefill/ApiClient.php
```diff
@@ -36,4 +36,17 @@ public function get($request, $base = 'http://loripsum.net/api/')
 
         return $this->client->get($uri, ['timeout' => 10])->getBody();
     }
+
+    /**
+     * Fetches an image from a remote source.
+     *
+     * @param string $url URL of the remote image to fetch
+     *
+     * @return StreamInterface
```
GwendolenLynch, line 46: Needs a `use Psr\Http\Message\StreamInterface;` added

src/Storage/Database/Prefill/RecordContentGenerator.php
```diff
@@ -453,9 +456,21 @@ private function getRandomTags($count = 5)
      *
      * @return array|null
      */
-    private function getRandomImage($type)
+    private function getRandomImage($type, $placeholder = null)
```
GwendolenLynch: `$placeholder` doesn't need a default, and does need a `@param string|null $placeholder` in the docblock

GwendolenLynch (changes requested):

src/Storage/Database/Prefill/ApiClient.php
bobdenotter: Added.

GwendolenLynch (approved):

PR #7369, opened 2018-03-08
[3.5] Add a Homepage singleton to default contenttypes, remove old images

bobdenotter:
This PR: 

 - Adds a Homepage singleton to the defaults. 
 - It removes the old default sample images.
 - It adds new placeholder images from unsplash. 🐨 🐱 🐶 🐮 

Depends on #7368, and the upcoming PR for base-2018.


eduardomart:
May I suggest that when editing a singleton we remove the word "New" from the very top? since it's the *only* record.

GwendolenLynch:
Added to the branch. Let me know if that's not OK with you, @bobdenotter and I'll drop before I push the test fixes

bobdenotter:
That's a good point, @eduardomart. Let's do that!

GwendolenLynch (approved):

PR #7379, opened 2018-03-20
[3.4] Skip test on Travis

GwendolenLynch:
Travis uses a version of GnuTLS that doesn't properly handle TLS 1.2+ and have been promising for two years to update the images to 16.04 … now 18.04 is out, and we've only just made 14.04.

Ubuntu … :zipper_mouth_face: 

P.S. I am going to follow up with a better way of handling all this today in 3.5, but it would be feature territory, so not aiming it at 3.4


bobdenotter (approved):
:shipit:

PR #7380, opened 2018-03-20
[3.5] Better handling of Marketplace connections

GwendolenLynch:
* Re-enable Nut's extension view test
* Move pinging of the Marketplace connection to a service
* Enable Guzzle to output cURL debug data in Nut `extensions:install` & `extensions:update` commands when `-vvv` is used

nut - fail

nut - pass


bobdenotter (approved):
👍

PR #7381, opened 2018-03-20
[3.6] Allow fields to be unsearchable

xiaohutai:
This is a new feature in the _legacy_ part of Bolt. The fix was simpler than I thought.

### Use Case

For some clients, we add a `textarea`-type field for _notes_; however, the search still searches this and highlights the keywords in that field. My initial thought was to set the `searchweight` to something low, but a low `searchweight` does not mean it's _filtered_.

### Usage

Add `searchable: false` to your field definition:

```diff
 field_notes: &field_notes
     notes:
         label: Notes
         type: html
         height: 300px
+        searchable: false
```

search-bois

---

Update with related Docs PR: https://github.com/bolt/docs/pull/888

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -533,7 +533,8 @@ class Storage
         // Build fields 'WHERE'
         $fieldsWhere = [];
         foreach ($fields as $field => $fieldconfig) {
-            if (in_array($fieldconfig['type'], $searchableTypes)) {
+            if (in_array($fieldconfig['type'], $searchableTypes)
+                && (!isset($fieldconfig['searchable']) || $fieldconfig['searchable'] !== false)) {
                 foreach ($query['words'] as $word) {
                     // Build the LIKE, lowering the searched field to cover case-sensitive database systems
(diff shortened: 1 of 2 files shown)


bobdenotter (dismissed):
Implemented both for Storage as well as legacy! 👍

xiaohutai:
Oi, it's the Storage of the Legacy. I think (only) @rossriley knows how it really works with the new one.

GwendolenLynch:
As discussed in #development as this is a feature, we're pushing to `unstable` … plus we've pushed back on legacy changes for a long while now, so we'll all collaborate on getting this into 3.6 as an update to the new layer.

GwendolenLynch (changes requested):
Migrate to new layer

JarJak:
I personally think setting `searchweight: 0` should do the same trick.

rossriley:
actually haven't tried that, does it work?

bobdenotter:
Setting it to `searchweight: 0` means that found terms in the field don't affect the search results. However, these fields _will_ show up in the list of results.

PR #7386, opened 2018-03-21
[3.4] Ensure search result array is valid

### LINKED ISSUE #7384: Bolt search page exception
Bolt automatically triggers search when visiting search page, even if there's no `?search=` parameter in query string.
The issue lies with https://github.com/bolt/bolt/blob/3.4/src/Controller/Frontend.php#L393 returning `false`, but code below expecting search result object/structure returned.

 

Details
-------

 - Relevant Bolt Version: 3.4.9
 - PHP version: 7.2
 - Used web server: Nginx

### Bug summary

image

### Steps to reproduce

In Slack discussion it was mentioned that the issue is specific to PHP version 7.2

GwendolenLynch:
Fixes #7384

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -391,22 +391,27 @@ class Frontend extends ConfigurableBase
         }
 
         $result = $this->storage()->searchContent($q, $contenttypes, $filters, $limit, $offset);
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
👍

PR #7394, opened 2018-03-24
Ensure array paramters are set before accessing in getContentByTaxonomy

Boorj:
that's still php, not js ;) if `['limit']`/`['page']` is not set - it throws exception.

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -800,8 +800,8 @@ class Storage
             }
         }
 
-        $limit = !empty($parameters['limit']) ? $parameters['limit'] : 9999;
-        $page = !empty($parameters['page']) ? $parameters['page'] : 1;
+        $limit = isset($parameters['limit']) && $parameters['limit'] ? $parameters['limit'] : 9999;
+        $page = isset($parameters['page']) && $parameters['page'] ? $parameters['page'] : 1;
 
         // If we're allowed to use pagination, use the 'page' parameter.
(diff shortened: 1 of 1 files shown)


src/Legacy/Storage.php
```diff
@@ -888,8 +888,9 @@ public function getContentByTaxonomy($taxonomyslug, $name, $parameters = [])
 
         $slug = $this->app['slugify']->slugify($name);
 
-        $limit = $parameters['limit'] ?: 9999;
-        $page = $parameters['page'] ?: 1;
+        $limit = (!empty($parameters['limit']) && $parameters['limit']) ? $parameters['limit'] : 9999;
+        $page = (!empty($parameters['page']) && $parameters['page']) ? $parameters['page'] : 1;
```
SvanteRichter: I think `empty` is pretty much a codesmell, could this not be changed to `$page = isset($parameters['page']) ? $parameters['page'] : 1;` and pretty much the same for the above line?

When we move to php7 only we can also switch this to `$page = $parameters['page'] ?? 1;`

Also, the newline below is unnecessary.

Boorj: I'm not sure if we should allow limit === 0 pass. Because the original code was using the second ternary block in case of zero limit.  
So, probably we should change `!empty` to `isset`, but leave the second condition?

p.s. love that `??` php7 feature :(

GwendolenLynch (approved):

PR #7404, opened 2018-03-27
Don't Run a group sort if user has ordered via the UI

### LINKED ISSUE #7383: Sorting on a specific field in overview not working for ContentTypes with a 'grouping'.
See screenhot: 

screen_shot_2018-03-21_at_11_40_05

rossriley:
Fixes #7383

Only passes the content result to the groupsort method if we don't have one provided by the user.


bobdenotter (approved):
Like a charm ✨

PR #7412, opened 2018-03-28
Only filter null and empty strings, leave zeros untouched

### LINKED ISSUE #7411: [bug] Incorrect saving of select field
Steps
---
(44 sec video - below)
- i have a contenttype field called "works". Here's how it's set up:
    ```yml
    works:
    type: select
    multiple: true
    values: 
        '1'  : 'Motorhomes upholstery'
        '2'  : 'Caravans upholstery'
        '3'  : 'RV upholstery'
        '4'  : 'Camper van upholstery'
        '5'  : 'Static Caravans upholstery'
        '6'  : 'Furniture reupholstery'
    ```
- in content edit i select 2nd and 4th options, press save
- what gets to database? **`["1","3"]`**. Whaa? Ookay, we assume that we're saving array indices of my options, but in _string format_. Okay, looks like we always save strings array. Not asking, why we ignore original keys.
- delete record in DB (field is not required, so it's ok)
- reload content edit in browser
- Now we select 1st and 4th options. Press save
- What comes to db? **`{"1":"3"}`** . Whaaaa💥 Is it an object with a _`key`_ of my first selected option and _`"index"`_ of the second one?

Video
---

## Upd1: What is sent in request ✈️ 

In request plain order indices are sent. So if you select options: 1st and 4th, you'll see in request form data:
**works[]** 0
**works[]** 3

## Upd2: Where the problem is 💩 

• Here's that little naughty piece of  code, that magically converts `["0","3"] to ["1"=>"3"]`:
https://github.com/bolt/bolt/blob/3.4/src/Storage/Entity/Content.php#L60
image

• so, if you don't pass a callback, all *falsy* values will be rejected
```php
$arr = [0, 1, 2, '0', '', null];
print_r(array_filter($arr));  // Will output: Array ( [1] => 1, [2] => 2 )
```

• then, if frontend sends `[0, 3]` in form, only `[1 => 3]` passes.

• And after all it's dumped as `{"1":"3"}` to database.

Resume
---
1. **Array_filter** is a weak part. It rejects 0 option.
2. Frontend sends incorrect keys. So it's a big bug, because if you rely on documentation ([defining-values-as-a-hash](https://docs.bolt.cm/3.4/fields/select#defining-values-as-a-hash)) people will assume that hash `{1: 'option1', 'anotheroption': 'option2'}` will work. But in fact - it won't.

##### Another reproduction:
- create a select key with values: 
```yml
values:
    1: opt1
    2 : opt2
    foo: opt3
```
- create a record, save it
- change the order (1 ←→ 2).
- 💥 you'll get another values.

Boorj:
Partially fixes #7411

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -57,7 +57,9 @@ class Content extends Entity
         $setter = 'set' . ucfirst($key);
         if (is_array($value)) {
             // Filter empty values from the array
-            $value = array_filter($value);
+            $value = array_filter($value, function ($item) {
+                return $item !== null && $item !== '';
+            });
         }
         $this->$setter($value);
     }


bobdenotter:
There's been quite a few fixes in the past half a year or so, with the "select" field type. It's turned out to be a particularly tricky part of the code to get right. 

I'll go over your fix, to see if it doesn't interfere with some other edge case. :-)

src/Storage/Entity/Content.php
```diff
@@ -57,7 +57,9 @@ public function set($key, $value)
         $setter = 'set' . ucfirst($key);
         if (is_array($value)) {
             // Filter empty values from the array
-            $value = array_filter($value);
+            $value = array_filter($value, function ($item) {
+                return ($item !== null) && ($item !== '');
+            });
```
GwendolenLynch: This doesn't do anything. See the [PHP docs](http://php.net/manual/en/function.array-filter.php):

> If no callback is supplied, all entries of array equal to FALSE (see [converting to boolean](http://php.net/manual/en/language.types.boolean.php#language.types.boolean.casting)) will be removed.

Easy enough to see:
```
$value = [null, false, ''];

$value = array_filter($value);
dump($value); 
// Outputs []

$value = array_filter($value, function ($item) {
    return ($item !== null) && ($item !== '');
});
dump($value);
// Outputs []
```

GwendolenLynch (changes requested):

src/Storage/Entity/Content.php
Boorj: Actually, as i told in issue,  i was fixing `[..., 0, ...]` case. I didn't know exactly, what was the purpose of filtering at all - so i fixed my actual case.
@GawainLynch can you take a look at investigation in #7411 ?

GwendolenLynch: Derp! Yeah, that makes sense.

Although what I would say is that I don't think the bug as reported is valid due to a number of reasons. But I haven't had time to look yet :innocent:

Boorj: Ookay, so please take your time) and notify me on update)

Boorj: Maybe.. you could just merge this PR )?

bobdenotter: @Boorj We should never, ever "just merge" a PR, if we have an inkling that we don't fully grasp the consequences. If we'd do that, there's much potential for opening ourselves up to other inexplainable bugs down the road. 

This might 'resolve' some of the wrong result, but it doesn't fix the _cause_. :-)

GwendolenLynch (approved):
A bit nervous about this one, but we've agreed to give it a try

Boorj:
yay 🐼

PR #7416, opened 2018-03-28
[3.4 backport] Fix for non-latin file names

Boorj:
I was just repeating this PR again #7217 (Issue #7198)

### DIFF
--- a/src/Provider/UploadServiceProvider.php
+++ b/src/Provider/UploadServiceProvider.php
@@ -4,6 +4,7 @@ namespace Bolt\Provider;
 
 use Bolt\Filesystem\Filesystem;
 use Bolt\Filesystem\UploadContainer;
+use Cocur\Slugify\Slugify;
 use Silex\Application;
 use Silex\ServiceProviderInterface;
 use Sirius\Upload\Handler as UploadHandler;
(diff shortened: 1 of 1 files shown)


bobdenotter:
This has been fixed for 3.5, which will be out soon. I don't think we should backport this to 3.4.

GwendolenLynch:
I moved branches on #7217 for a reason. :wink:

Boorj:
By the way, many people just don't update bolt. I'm using 3.2 somewhere. Probably, someone's using `"^3.4"`. Maybe it still needs to be fixed in 3.4.11? no ? But you decide, of course)

GwendolenLynch:
> By the way, many people just don't update bolt. I'm using 3.2 somewhere

I still have a 1.4 site, and I am not backporting this that far :stuck_out_tongue: 

There are limited resources, and everyone is doing this on their free time out of love.

> Maybe it still needs to be fixed in 3.4.11? no ? But you decide, of course)

It adds functionality, therefore *must* go into a minor release according to SemVer.

bobdenotter:
> Probably, someone's using "^3.4"

That same reasoning can be applied to _any_ new feature, then.. "I'm still using Bolt 1.6, but I'd like to have Singletons.. Can you backport them?" 

Our goal is to "Keep stable releases stable", and if someone _needs_ a new feature, we'll kindly suggest them to upgrade. ;-)

Boorj:
> It adds functionality,

looks like it was removed in 3.4 or somewhere else, and it WAS in 3.3 as i remember. But, again. No problem, i'll update my sites)

GwendolenLynch:
This is an addition of funcitonaltiy, end of story
```php
+        $app['upload.sanitizer'] = $app->share(function ($app) {
+            $pattern = $app['config']->get('general/upload/pattern', '[^A-Za-z0-9\.]+');
+            $lowercase = $app['config']->get('general/upload/lowercase', true);
+            return new Slugify("/$pattern/", ['lowercase' => $lowercase]);
+        });
```

PR #7417, opened 2018-03-28
fixing FileManager to slugify uploaded filename

### LINKED ISSUE #7409: [bug] Incorrect uploaded file name
Problem # 1 🔔 incorrect filename in notification
---
### Repro
1. goto file manager
2. upload any file with non-latin name.
3. you'll get incorrect notification:
    

### Where is the problem
It's here: ([/src/Controller/Backend/FileManager.php#L250](https://github.com/bolt/bolt/blob/3.4/src/Controller/Backend/FileManager.php#L250))

```php
   // ...
   private function handleUpload(FormInterface $form, DirectoryInterface $directory)    {
        $files = $form->get('select')->getData();
        foreach ($files as $fileToProcess) {
            // ...
            $originalFilename = $fileToProcess['name'];
            $filename = preg_replace('/[^a-zA-Z0-9_\\.]/', '_', basename($originalFilename));
            // ....
            $this->processUpload($directory, $filename, $fileToProcess);
            // ....
        }
    }

    private function processUpload(DirectoryInterface $directory, $filename, array $fileToProcess)    {
        // ....
        if ($result->isValid()) {
            $this->flashes()->info(
                Trans::__('page.file-management.message.upload-success', ['%file%' => $filename])
            );
        // ....
    }
```

Problem # 2 🤖  uncontrolled filename transformation 
---
Shouldn't SiriusPHP use predictable filename processing? And shouldn't it be the same in all places?

Problem # 3 ⛸  possible thin ice 
---
Btw, also i found a place in code, which could possible produce incorrect filename transformation.
I fixed that once [here](https://github.com/bolt/bolt/pull/7217/commits/6a237426227928fdbe1c986ee6ee9dc548215ac6), but then we moved to Sirius PHP for validating uploads. So now we [rolled back](https://github.com/bolt/bolt/blob/3.4/src/Provider/UploadServiceProvider.php#L34-L42) to old sanitation. 
@GawainLynch Shouldn't it be fixed again? I thought that there _shouldn't_ be any filename transformation, which uses _only_ **lowercase** over **slugify** ?

---
Solution 🎆 🙌 🍾 
---
Maybe we will use single unified filename transform function?

upd1
---
It came, that i had extension for sanitizer provider. So bolt now doesn't have this support.
image
💥 💥 💥 **All non-latin symbols are replaced**💥 💥 💥

Boorj:
Fixes #7409

### DIFF
--- a/src/Controller/Backend/FileManager.php
+++ b/src/Controller/Backend/FileManager.php
@@ -247,7 +247,8 @@ class FileManager extends BackendBase
             ];
 
             $originalFilename = $fileToProcess['name'];
-            $filename = preg_replace('/[^a-zA-Z0-9_\\.]/', '_', basename($originalFilename));
+            $filename = basename($originalFilename);
+            $filename = $this->app['upload.sanitizer']->slugify($filename, $this->app['config']->get('general/upload/replacement', '-'));
 
             try {
                 $isAllowed = $permissions->allowedUpload($filename);


bobdenotter:
I don't think this is correct.. I'm getting the following, because there's no `upload.sanitiser`. 

screen shot 2018-03-29 at 10 33 02

Boorj:
@bobdenotter that worked together with #7416 . How is it going in 3.5 ?

Boorj:
In 3.5 there **is** upload sanitizer. 
https://github.com/bolt/bolt/blob/3.5/src/Provider/UploadServiceProvider.php

And in 3.4 it works with #7416. 
So looks like you've been testing this on 3.4, non patched with 7416  ☺️ 

Can you move that PR to 3.5 branch?

bobdenotter:
Works like a charm in 3.5. Thanks, @Boorj. 👍

bobdenotter (approved):

PR #7441, opened 2018-04-13
Improve fieldtype block.

binarykiwi:
My first PR. :)

Toggle panel-body for a better overview. This might be very handy if you have more content inside a block.

Capsuled with new setting (`toggleblocks: true` | `toggleblocks: all`)

image

`true`: add a small toggle button to the panel-heading to hide/show the panel-body

image

`all`: same as true plus adds buttons to hide/show all panel-body elements with just one click

image


jadwigo:
This is nice.

I would want the default to be `toggleblocks: all` because I think everyone wants this.
And people can the always set `toggleblocks: false` if they want to hide it.

GwendolenLynch:
Thanks @binarykiwi :+1: 

As this is a feature, I've rebased this against `3.5` but we'll need to decide is it is a small enough change to include in an RC.

I also dropped the built assets.

GwendolenLynch:
Oh, and I just noticed that you PRed from your fork's `3.4` branch. After the merge, you'll need to either reset it, or delete the remote branch here on GH.

binarykiwi:
Ah, I just forked the default branch and thought this would be the right way ;-) 
So I have to change my branch to `3.5` and _repull_ it again?

GwendolenLynch:
If you wanted to have a feature branch to work with locally, you can just do something like this in your repository directory:

```
git remote update --prune
git checkout feature/improved-blocks origin/3.4
```
That will give you a local branch called `feature/improved-blocks` that matches the contents of this one.

**After** this branch is merged, you can just go to https://github.com/binarykiwi/bolt/settings/branches and press the delete button next to `3.4` … and the others, you only need a default branch plus whatever PR branches are currently unmerged.

bobdenotter:
Thanks, @binarykiwi 

This time, I think I'm going to be the party-pooper here, and vote against stuffing this into 3.5 at this stage. 3.5 is done as far as i'm concerned, and this PR will need some work. 

 - It should act the same for "Blocks" and "Repeaters" alike. 
 - I don't like the chevron in the circle as the indicator. We shouldn't introduce a _new_ element to the UI, when we already have tons of different things already. It doesn't feel like a good fit.

If it's up to me, we'll either merge this after the 3.5.0 release into `3.5` or into`3.6`. The plan is to have a short cycle after 3.5, with mostly Ross' PR for Storage. 

@eduardomart Do you have your unfinished branch for this lying around, still?

GwendolenLynch:
@bobdenotter as it is a feature, we can either sneak it into `3.5` prior to release, or it has to be bumped to `3.6` which should be short-cycle as you mentioned.

bobdenotter:
Ok, suggestion time: We merge the [PR for RC2](https://github.com/bolt/bolt/pull/7435), tag, package, ship it.. Then, if on monday this PR is good-to-go, we'll merge it in, and do another RC. 

If it's found that it'll take some more time, we'll move ahead with 3.5.0, and push this to 3.6.0. Sounds like plan?

GwendolenLynch:
I'm easy :innocent: … just trying to stick to a reasonable, fair and consistent process.

I'm about to go offline for the day though, so if you need something from me you have approximately 11 minutes :wink:

bobdenotter:
Ok, then you have 10 minutes left to merge #7435 and tag it. I'll handle the rest. :-p

bobdenotter:
Oh, and @binarykiwi If you have any questions about what i'd like to see changed: I'll be online for the larger part of the weekend, so i'll gladly provide feedback on stuff. ☺️

binarykiwi:
@bobdenotter thanks for your comment. I fully understand your two points and will work on them. Is there an icon to collapse elements somewhere inside Bolt?

bobdenotter:
@binarykiwi I think i'd pick `fa-expand` and `fa-compress`. Perhaps integrate them with the already present buttons, like this: 

screen shot 2018-04-13 at 16 21 19

I think we should also truncate the labels in these buttons a bit, because it's becoming too wordy for my tastes. As in, we should make it say `Up` instead of `Move Up`.

bobdenotter:
Oh, and I agree with @jadwigo: `all` should be the default, if nothing's set. I also think maybe `collapsible: ` might be a better name than `toggleblocks: `.

rossriley:
:+1:  this is very nice, I'd also second Bob's idea to use collapsible as the setting name, since that allows us to use the same on other types of fields too.

bobdenotter:
Did some work on this PR in #7443.

eduardomart:
@bobdenotter Yes I have it. I'll see what I can do today to make it available to you guys.

binarykiwi:
Many thanks to all of you guys. I planned to continue working on this improvement today on Monday. But you rocked it over the weekend.

GwendolenLynch:
Commit merged in #7443

PR #7443, opened 2018-04-15
New Feature: Collapsible Blocks and Repeaters

bobdenotter:
Builds on / replaces PR #7441 by @binarykiwi. 

Changes: 

 - Clean up the UI a bit. 
 - Apply same logic to Repeaters, not just Blocks
 - Simplify logic a bit: Only use `collapsible: true` (default) and `collapsible: false`.

screen shot 2018-04-15 at 18 30 14


jadwigo:
I would move the delete button to a separate group..  the delete button is an irreversible action and should be visually different and harder to access than reversible actions in the UI

jadwigo (changes requested):
Separate the delete button so reversible and irreversible actions are split

bobdenotter:
Updated. 

screen shot 2018-04-16 at 08 13 47

jadwigo (approved):

PR #7449, opened 2018-04-25
Add an 'if' statement for missing Block field

ankedsgn:
If you try to access a page that uses a block field that is no longer used, you get an error, when using strict variables. (Since the field does not exist anymore)

This might happen after a refactor, when the block field is still in the DB, but no longer exists in contenttypes.yml.

Adding a `is defined` fixes this!

Error:
image


GwendolenLynch (approved):

GwendolenLynch:
Thanks @anketwokings :+1:

Issue #7442, opened 2018-04-15
XSS protection disabling is broken


marcingajda:
When you have a field with HTML and you open the preview page Bolt sends that code in POST parameters. Chrome has a feature that detects HTML code in POST and blocks displaying the response to protect from XSS. It can be disabled however by the server by sending a header `X-XSS-Protection` set `0`. Bolt tries to use that trick but fails with it. 

Details
-------

 - Relevant Bolt Version: 3.4.8 and probably later
 - Install type: Composer install
 - PHP version: 7.1
 - Used web server: Nginx version 1.13.9
 - Browser: Chrome 65.0.3325.181

Reproduction
------------

### Bug summary

Preview button doesn't work on Chrome if you have an field with HTML in the content type editor. It can be a field for embeding an iframe or just a Tweet added with the build-in embeding tool.

### Specifics

The [`DisableXssProtectionListener`](https://github.com/bolt/bolt/blob/14b3ec1019fd3a6af91ff92e246ec9081e50e49f/src/EventListener/DisableXssProtectionListener.php) which adds that header is runed when the current route name is in the array defined in the [`disable_xss_protection_routes`](https://github.com/bolt/bolt/blob/180a3a4c21e37fe55a7c681657f53e1ada682c81/src/Provider/EventListenerServiceProvider.php#L30). The problem is when you click the Preview button Bolt is using `contentlink` route, not the `preview` one. I checked it by inserting `dump($route)` in the mentioned listener below 44 line.

### Steps to reproduce

* Add to any content type a text field
* Go to record editor
* Put in that field `<iframe src="http://google.com"></iframe>`
* Click the preview button

### Expected result

Preview of that record should be displayed

### Actual result

![image](https://user-images.githubusercontent.com/6128754/38773669-cbf18c98-4052-11e8-900e-45cda3b65b5d.png)

How to fix it (temporarly)?
-------------

If you have a bundled extension you can overwrite there the list of not protected routes
```
$app['disable_xss_protection_routes'] = [
    'preview',
    'fileedit',
    'contentlink' // <= that helps
];
```
However this is probably a workaround and just the route used for previews is wrong. That is also why I didn't create a PR.

bobdenotter:
Pinging @JarJak, @marcingajda. Hi, any progress on this one, yet? ;-)

marcingajda:
@bobdenotter I will prepare a PR in free time ;)

GwendolenLynch:
Closed by #7458

PR #7456, opened 2018-04-30
[3.4] Add 'contentlink' to routes with disabled XSS protection

marcingajda:
Add the `contentlink` route to the list of routes for which XSS protection, a Google Chrome feature, is disabled. It fixes the preview screen for records with iframe codes or embedded tweets.

Fixes: #7442

Details
-------

Chrome has a feature that detects HTML code in POST and blocks displaying the response to protect from XSS. It can be disabled however by the server by sending a header `X-XSS-Protection` set to `0`. Bolt needs to do it when displaying for example a record preview, however it doesn't do it anymore because the route used for preview was changed to `contentlink`. This pull request adds that route to the list of routes for which XSS protection is disabled.

### DIFF
--- a/src/Provider/EventListenerServiceProvider.php
+++ b/src/Provider/EventListenerServiceProvider.php
@@ -30,6 +30,7 @@ class EventListenerServiceProvider implements ServiceProviderInterface
         $app['disable_xss_protection_routes'] = [
             'preview',
             'fileedit',
+            'contentlink',
         ];
         $app['listener.disable_xss_protection'] = $app->share(
             function ($app) {


SvanteRichter:
Wouldn't this add the `X-XSS-Protection: 0` header to all contentlink routes? Which would mean that we disable it even on non-logged-in production views of any single record.

Also, this isn't a Chrome feature. It was originally introduced in IE and then adopted by chrome, firefox and safari.

marcingajda:
> Which would mean that we disable it even on non-logged-in production views of any single record.

No because this header is never added to requests send with a safe method (GET, HEAD, etc...). See [here](https://github.com/bolt/bolt/blob/v3.4.10/src/EventListener/DisableXssProtectionListener.php#L40) and then [here](https://github.com/symfony/http-foundation/blob/master/Request.php#L1423).

> It was originally introduced in IE and then adopted by chrome, firefox and safari.

You are right with the except of Firefox. It was never adopted there. See the table [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection#Browser_compatibility).

bobdenotter:
While this PR seems to fix the issue, and I agree that it _shouldn't_ introduce any glaring security issues, I would prefer to fix the underlying issue, instead of patching a workaround for it, without addressing the actual problem. 

In my opinion, the problem is this: "The 'preview' button shows a preview of the record the user is editing, but it's showing as 'contentlink'." 

The link in the page is correct: 

screen shot 2018-05-02 at 09 09 26

When you go to the preview, the profiler indicates a correct match too: 

screen shot 2018-05-02 at 09 10 57

screen shot 2018-05-02 at 09 11 03

Yet, the actual preview page says this: 

screen_shot_2018-05-02_at_09_11_40

So, ideally we'd fix the routing, because if we do, it'll automatically behave correct again, without this  workaround.

Thoughts, @GawainLynch, @rossriley ?

SvanteRichter:
> No because this header is never added to requests send with a safe method (GET, HEAD, etc...). See here and then here.

RIght, but there are quite a few situations where we have post requests on the frontend, like with extensions like boltforms.

Another way to fix this (besides the proper way mentioned by @bobdenotter ) would be to only add the header when we have a correct CSRF token, since those will always get added to the editcontent forms anyway.

marcingajda:
I agree with that solution. As far as I remember someone suggested on Slack that I should implement the workaround as PR but maybe I misunderstand that someone.

bobdenotter:
@marcingajda Might've been me, who suggested that. :-)

It sometimes makes a lot of sense to see something, and then you think to yourself "Yes, looks good", or "Hmm, there's probably more to it…". In this case it's the latter, but only after I saw your PR. So, it's not wasted effort at all. 👍

GwendolenLynch:
Closed by #7458

PR #7457, opened 2018-05-02
Only allow POST on `/preview`, because there's no need to GET it.

bobdenotter:
Part of the fix for #7442 -> We should disable the "XSS protection header" on the "preview" route, but that also means we should _not_ accept GET requests on that route, in order to prevent potential XSS.

### DIFF
--- a/app/config/routing.yml.dist
+++ b/app/config/routing.yml.dist
@@ -28,6 +28,7 @@ preview:
         _controller: controller.frontend:preview
     requirements:
         contenttypeslug: controller.requirement:anyContentType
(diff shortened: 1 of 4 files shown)


bobdenotter:
Ok, failing tests seems to indicate this needs some work.

GwendolenLynch:
In case I come back to this sometime in the future wondering why there is a `MethodNotAllowedHttpException` being thrown in the controller itself; "It is for legacy BC, Gawain" :smile:

GwendolenLynch (approved):

PR #7477, opened 2018-05-10
[3.6][WIP] Ignore default value alteration for MySQL TEXT and BLOB types

JarJak:
**Steps to reproduce:**
1. Create field with textarea type
2. Update database
3. Add default value to this type
4. Run database check and update

**Expected result:**
Database is up to date

**Actual result:**
Even after database update there is message "Invalid column X".
"Show proposed alterations" does not generate any SQL.

See https://github.com/doctrine/dbal/pull/514


src/Storage/Database/Schema/Comparison/MySql.php
```diff
@@ -37,5 +37,18 @@ protected function removeIgnoredChanges(TableDiff $diff)
...
+            /* @var $columnDiff \Doctrine\DBAL\Schema\ColumnDiff */
+            $column = $columnDiff->column;
+
+            if ($columnDiff->hasChanged('default') &&
+                count($columnDiff->changedProperties) === 1 &&
+                ($column->getType() instanceof TextType || $column->getType() instanceof BlobType)
+            ) {
+                unset($diff->changedColumns[$key]);
+            }
+        }
```
GwendolenLynch, line 54: All this is not needed and would be bi-directional.

Assuming that this is valid to ignore, and I don't think it is for all versions of that crap platform you'd just want:
```php
    protected function setIgnoredChanges()
    {
        $this->ignoredChanges[] = new IgnoredChange('changedColumns', 'type', 'text', 'blob');
    }
```

Also, it needs to be considered if this affects DBAL 2.5, 2.6 & 2.7, MySQL 5.5, 5.6, 5.7, 8.0, 9.0, and MariaDB =< 10.1, 10.2, 10.3.

GwendolenLynch (changes requested):

src/Storage/Database/Schema/Comparison/MySql.php
JarJak: This is not about directions. Bug is not caused by column type change, just both TEXT and BLOB types are affected.

It affects DBAL >= 2.5 and probably every version of MySQL / MariaDB.

GwendolenLynch: >  Bug is not caused by column type change

Then the PR isn't valid :wink: 

> It affects DBAL >= 2.5 and probably every version of MySQL / MariaDB.

Have you tested this on all combinations? As I am willing to bet a beer that you're wrong.

JarJak: Starting from MariaDB 10.2.1, BLOB and TEXT columns can have a DEFAULT value. However DBAL seems to ignore this feature: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L598
Mysql is affected from 5.5 up to 8.0. I can't check it for 9.0 as this version is not yet released.

GwendolenLynch: >  I can't check it for 9.0 as this version is not yet released.

Ah, no stress on that one. I thought it had gone GA recently but happily, I am wrong. :smile: 

> Starting from MariaDB 10.2.1, BLOB and TEXT columns can have a DEFAULT value. However DBAL has separate Driver for this version of MariaDB: 

Yeah, and this is part of the living disaster that probably makes it unworkable.

Would a Bundle on a port of this work for you needs though? https://github.com/bolt/bolt-extension-compat-mariadb

JarJak: @GawainLynch the extension you provided disables all default checks, this is not needed nor good solution. Here's why:
- this is not only about MariaDB 10.2.1+ but for MySQL 5.5+ too
- only TEXT column type is affected (I don't think Bolt uses BLOB anywhere)

JarJak:
keep

bobdenotter:
@jarjak Is this still valid for current 3.6?

JarJak:
Yes

bobdenotter (approved):
👍

PR #7478, opened 2018-05-11
Portuguese pt-PT revision

horus68:
Translation update for pt-PT (Portuguese - Portugal)

Details
-------
Portuguese pt-PT revision
Translation update for pt-PT (Portuguese - Portugal)
Bolt - v3.5.1


GwendolenLynch:
So much easier to review, thank you so very much! :+1:

horus68:
Note: the problems with my Bolt install (quoting unused strings) are not solved yet, so I can't use it to process translation files.

GwendolenLynch:
I still can not reproduce what you are saying occurs, so if I can't there is nothing I/we can do

GwendolenLynch (approved):

PR #7486, opened 2018-05-16
[WIP][3.5] PackageManager error when Descriptor is null

JarJak:
On management/extensions page I was getting error:
```
Call to a member function getConstraint() on null
File: vendor/bolt/bolt/src/Composer/PackageManager.php::270
```
So this PR fixes it.


JarJak:
You can see that one of Travis CI tests is failing because tested extension is not managed.
But why is it not managed?

screen shot

I can update this PR to make it pass, but I'm not sure if this failure is a bug or feature.

JarJak:
keep

PR #7489, opened 2018-05-19
[3.6] Provide database data type update processing

GwendolenLynch:
Just now …

**Bob:** What should we do about handing invalid types in upgraded Bolt databases?
**Gawain:** I have a branch …
**Bob:** Great, when did you do this?
**Gawain:** 12th of May … 
**Bob:** Oh, so very recently then!
**Gawain:** 12th of May 2017

---

As above, this is something I've had baking for a while and in our hackday discussion's we've decided to table it as-is for the time being to get some testing and feedback. _HINT! HINT! HINT!_

~~I literally haven't looked at this in over a year, so it will need some love prior to merging.~~

### Using

```
php app/nut database:update-data
```

### DIFF
--- a/app/resources/translations/nl_NL/messages.nl_NL.yml
+++ b/app/resources/translations/nl_NL/messages.nl_NL.yml
@@ -7,15 +7,11 @@
 #          process and your PR so that it will eventually be necessary to
 #          re-map your translations.
 
-#   1 Unused messages
+#   0 Unused messages
 #   1 Legacy untranslated messages
 #   0 Untranslated messages
 #  38 Legacy translation messages
 # 572 Translation messages
-
-#--- Unused messages ----------------------------------------------------------
-
-"contenttypes.generic.no-proper-type": "In het contenttype voor '%contenttype%' heeft het veld '%field%' type: %type%, wat geen geldig veld type is. Pas contenttypes.yml aan en corrigeer dit."
 
 #--- Legacy untranslated messages ---------------------------------------------
 
@@ -334,7 +330,6 @@ general.phrase.of: "van"
 general.phrase.omnisearch: "Omnisearch"
 general.phrase.only-root-change-file-extensions: "Alleen root kan bestandsextensies wijzigen."
 general.phrase.options: "Opties"
-general.phrase.or: "of"
 general.phrase.order: "Volgorde"
 general.phrase.order-colon-sort: "Volgorde: %sort%"
 general.phrase.other-content: "Andere content"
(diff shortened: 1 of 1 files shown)


src/Storage/Migration/DataType/FieldDataTransformer.php
```diff
@@ -0,0 +1,70 @@
...
+    public static function image($data)
+    {
+        if (empty($data)) {
+            return null;
+        }
+
+        $isJson = static::isJson($data);
+        if ($isJson) {
+            try {
+                $response = \GuzzleHttp\json_decode($data, true);
```
bobdenotter: Should't this use \Bolt\Common\Json instead?

GwendolenLynch: Yes, as with the test above

bobdenotter (dismissed):
Seems to work fine, in my test cases.  :+1:

bobdenotter:
Just a note to track this, so I won’t forget: The other day we bumped into an issue, when a `type: select` field was changed by adding `multiple: true`. Bolt then crashed, because existing data in the DB (strings) were trying to get parsed as arrays.

Will need to check if this code fixes that too, or else if it can be added.

GwendolenLynch:
Yes, each type can be handled by a transformer class, and remind me next week after Tuesday and I'll happily address this and if we can brainstorm a couple of others, those too!

I have a push coming up that is uncommitted here that does a few more things, including an attempt at adding this to the web UI. Although I am going to state for the record I am not (currently) in favour of adding this … doing it in a web request is only asking to destroy data.

bobdenotter:
Did some more testing on this feature. This one is still an issue: "when a type: select field was changed by adding multiple: true."

To reproduce, add a field to a contenttype, like this: 

```yaml
        multiselect:
            type: select
```

Update DB, add a record and select a "Turtle". Save it, and then change the field to:

```yaml
        multiselect:
            type: select
            multiple: true
```

Now, the Bolt backend will break:

screen shot 2018-07-14 at 13 33 11

GwendolenLynch:
> This one is still an issue: "when a type: select field was changed by adding multiple: true."

Updated the code to cope, adding a new transformer in the process … oh, and don't mention the logic bug that magically disappeared :wink: 

The only thing we **may** want to think about it going the other way … but that would, by definition, be a data-loss scenario and I am not in favour of anything that deliberately does something like that.

GwendolenLynch:
FTR, I still don't like `:type-migrate` … suggestions still welcome from those who are good at nomenclature.

bobdenotter:
How about `database:sanitize` ?

GwendolenLynch:
How about `database:update-types`? 

As it is a related task to `database:update`, and keeping the names close.

bobdenotter:
Also fine with me.. I'd prefer `update-types` over `type-migrate` by a large margin.

Or perhaps `update-columns`?

bobdenotter (approved):

bobdenotter:
Did some more monkey testing, still works well. :shipit:

GwendolenLynch:
No worries. I have a related PR that I will send in soon, then tweak this one … can discuss tonight.

PR #7502, opened 2018-05-31
Allow custom slug field label

### LINKED ISSUE #7424: field.slug.label:'foo' doesn't work for slug
If i make this config 
```yml
# contenttypes.yml
fieldname:
    slug:
        label  : 'Постоянная ссылка'
        type   : slug
        uses   : title
        postfix : "Постоянная ссылка"
```
postfix works, label - doesn't.

image

Maybe we should take label from contenttype config, and fallback to default only if it's missing?

Boorj:
Fixes #7424 . Now it looks up for `field.label` , then to previous options.


bobdenotter (approved):
Works like a charm! 👍

PR #7505, opened 2018-06-02
[Hotfix 3.5] Ensure select choices are built and accessed via multi-level field names

### LINKED ISSUE #7496: [BUG] Field with same name as other one inside repeater overwrites config of the first one
Bug summary
---

If field in repeater has the same name as other field outside of this repeater, then the first one takes configuration from the second. Configuration of field in repeater is overwritten.

Details
-------

 - Relevant Bolt Version: 3.5.1

Reproduction
------------

Content type config:
```
articles:
    name: Articles
    fields:
        details:
            type: repeater
            label: Details
            fields:
                value:
                    type: html
                    label: Details information
                type:
                    type: select
                    label: Detail type
                    values:
                        important: Important!
                        normal: Normal one
        type:
            label: Article type
            type: select
            values: 
                news: News
                other: Other article
```
With this config, enter Article edit page and you can notice that `Detail type` select has values taken from `Article type`.

Expected result
----

`Detail type` select should have options: `Important!`, `Normal one`.

Actual result
----

`Detail type` select has options: `News`, `Other article`.

rossriley:
Fixes #7496 

This one actually affected all repeaters and block fields. The current version builds a flat array of field names to select values which means that any values in base fields are overwritten by any other fieldname with the same key.

This PR adjusts the repeater and block builds to add to a multi-dimensional array so that the only constraint is unique field name per repeater or block set.

### DIFF
--- a/app/view/twig/editcontent/fields/_block-group.twig
+++ b/app/view/twig/editcontent/fields/_block-group.twig
@@ -48,7 +48,7 @@
         {% for rkey, rfield in field.fields[block].fields %}
             {% set rfield = defaults|merge(rfield) %}
             {% if rfield.type == 'select' %}
-                {% set values = context.values.select_choices[rkey] %}
+                {% set values = context.values.select_choices[name][block][rkey] %}
             {% endif %}
             {% set rcontext = {
                 'key':        name ~ '_' ~ index ~ '_' ~ block ~ '_' ~ rkey,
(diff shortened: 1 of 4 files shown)


bobdenotter (dismissed):
Change looks good to me, but I think the tests need to be updated to match.

GwendolenLynch (approved):

Issue #7497, opened 2018-05-24
Unable to add a relation with setRelation()


jadwigo:
When I import records in a custom bundle, I can't add a relation between entities.

There seems to be a mismatch between `setRelation()` in `Storage/Entity/Content` 
https://github.com/bolt/bolt/blob/3.5/src/Storage/Entity/Content.php#L256-L259
and  `setRelation()` in `Storage/Entity/ContentRelationTrait`
https://github.com/bolt/bolt/blob/3.5/src/Storage/Entity/ContentRelationTrait.php#L104-L109

As usual I don't even know if I'm doing it the right way. Is there a normal way to add entity relations that I can use?

Details
-------

 - Relevant Bolt Version:  3.5
 - Install type: Composer install | GitHub checkout

Reproduction
-------
The page 15 exists and the persons [5016, 5017, 5004, 4903] also exist.
There is a relation set in `contenttypes.yml`
```yml
pages:
  name: Pages
  singular_name: Page
  fields:
    [.. some fields ..]
  relation_persons: &relation_persons
    persons:
      label: "Contacts"
      multiple: true
      order: last_name
      format: "{{ item.last_name }}, {{ item.first_name }}"

persons:
  name: Persons
  singular_name: Person
  fields:
    [.. some fields ..]
```
When you run the following code:
```php
        $parent_type = 'pages';
        $target_type = 'persons';
        $repo = $this->app['storage']->getRepository($parent_type);

        $currentrecord = $repo->find(15);

        $currentrecord->clearRelation($target_type);

        foreach ([5016, 5017, 5004, 4903] as $target_record_id) {
          $currentrecord->setRelation($target_type, $target_record_id);
        }
        $repo->save($currentrecord);

        dump($currentrecord);

```
Result
-------
```
FatalThrowableError in Content.php line 256:
Type error: Argument 1 passed to Bolt\Storage\Entity\Content::setRelation() must be an instance of Bolt\Storage\Collection\Relations, string given, called in /[..]/extensions/bundles/myvendor/mybundle/src/Controller/Backend.php on line 102
```

rossriley:
OK, first the simple answer, the way we do it in core is like this:

```
$currentrecord = $repo->find(15);
$currentrecord->getRelation()->setFromPost(['relation' => ['person' => [5016, 5017, 4903] ] ], $currentrecord);
$repo->save($currentrecord);

```

[...]

PR #7506, opened 2018-06-03
[3.6] Improving relation get set interface

rossriley:
This is based on feedback from #7497 that the interface for interacting with relations has become a bit complicated.

This is a small PR that adds a helper method to make it easier to associate content entities with an owner.

The public-facing API is straightforward, given either an entity, a collection or a string/id combination we can easily associate relations.

```php
$repo = $app['storage']->getRepository('entries');
$entry = $app['query']->getContent('entries/1');
// Or direct from the repo
$entry = $repo->find(1);

// Option 1: Associate relations with this entity by name and ids
$entry->relation->associate('pages', [1,2,3]);

// Option 2: Associate relations via passing a single Entity
$page = $app['query']->getContent('pages/1');
$entry->relation->associate($page);

// Option 3: Associate a set of relations by passing a collection
$pages = $app['query']->getContent('pages', ['id'=> '1 || 2 | |3]']);
$entry->relation->associate($pages);


// Save
$repo->save($entry);
```

```

### DIFF
--- a/src/Storage/Collection/Relations.php
+++ b/src/Storage/Collection/Relations.php
@@ -7,6 +7,7 @@ use Bolt\Storage\Entity;
 use Bolt\Storage\EntityManager;
 use Bolt\Storage\EntityProxy;
 use Doctrine\Common\Collections\ArrayCollection;
+use DomainException;
 
 /**
  * This class stores an array collection of Relations Entities.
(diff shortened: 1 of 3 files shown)


bobdenotter (dismissed):
I like it! 👍

GwendolenLynch (approved):

PR #7510, opened 2018-06-06
Make UploadContainer agnostic of invalid file content type

JarJak:
More info: https://github.com/siriusphp/upload/pull/28

This could be a temporary fix.

### DIFF
--- a/src/Filesystem/UploadContainer.php
+++ b/src/Filesystem/UploadContainer.php
@@ -41,7 +41,7 @@ class UploadContainer implements ContainerInterface
      */
     public function save($file, $content)
     {
-        $this->filesystem->put($file, $content);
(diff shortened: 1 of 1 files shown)


GwendolenLynch:
See #6754

JarJak:
So the problem exists for almost a year, wow... At least I've found where to fix it (in siriusphp/upload#28) :D

GwendolenLynch:
> So the problem exists for almost a year, wow

Yes, and the reason the last one was rejected is it is an upstream bug that we shouldn't workaround here

JarJak:
This can be fixed by adding autocinfirm setting as an option in #7511

GwendolenLynch:
Rejecting for the same reason as #6754 … but will probably merge #7511 on green :wink:

PR #7513, opened 2018-06-08
[3.6] Improve the iterator to swap out proxies when using Lazy Collections

rossriley:
Part of a series relocating smaller feature changes out of #7291

### DIFF
--- a/src/Storage/Collection/LazyCollection.php
+++ b/src/Storage/Collection/LazyCollection.php
@@ -3,28 +3,16 @@
 namespace Bolt\Storage\Collection;
 
 use Bolt\Storage\EntityProxy;
+use CallbackFilterIterator;
 use Doctrine\Common\Collections\ArrayCollection;
 
 /**
  * LazyCollection Class - Allows a set of references to Entities to be loaded on demand.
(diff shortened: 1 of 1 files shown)


GwendolenLynch (approved):

src/Storage/Collection/LazyCollection.php
```diff
@@ -38,4 +26,23 @@ public function serialize()
...
+     * This overrides the base getIterator method and on access swaps out the EntityProxy objects
+     * for real entities.
+     *
+     * @return CallbackFilterIterator
+     */
+    public function getIterator()
+    {
+        $els = parent::getIterator();
+
+        return new CallbackFilterIterator($els, function (&$current) {
```
JarJak, line 40: ugh, filter is for filtering, not mapping... i'm sure there is a better way to do this

JarJak: like http://php.net/manual/en/function.iterator-apply.php

JarJak: @rossriley could you improve this?

PR #7528, opened 2018-06-12
[3.5] Ensure a content editing form CSRF token exists at logon

### LINKED ISSUE #7512: Save button does not work sometimes
Hi everyone!

It seems the Save button sometimes does not work when saving a record.

It is not easy to reproduce but I have managed several times. The way I have done it is: 

- Install a fresh Bolt v3.5.3.
- Prefill Database with 10 record of each default Contenttype.
- Open the prefilled record in several Tabs and wait for some time.
- After a minute or 2 start pressing the Save button of each opened Record. Most of the times the save action goes through but sometimes it does fail and the "Saved on" datetime message is not updated anymore.
capture

I have noticed that when the Record is not saved the `Content-Type` Response Header is: `text/html` and when it works the Response Header is `Content-Type: application/json`.

Here the preview when it doesn't work:

capture2

Here the preview when it works:

capture3

Details
-------

 - Relevant Bolt Version: 3.5.3
 - PHP version: 5.6
 - Used web server: Apache

GwendolenLynch:
Workaround the `content_edit` CSRF token race by creating the token at login. :disappointed: 

Closes #7512

### DIFF
--- a/src/EventListener/AccessControlListener.php
+++ b/src/EventListener/AccessControlListener.php
@@ -4,8 +4,9 @@ namespace Bolt\EventListener;
 
 use Bolt\AccessControl\Token\Token;
 use Bolt\Common\Serialization;
+use Bolt\Events\AccessControlEvent;
+use Bolt\Events\AccessControlEvents;
 use Bolt\Events\StorageEvent;
-use Bolt\Events\StorageEvents;
 use Bolt\Filesystem\Exception\FileNotFoundException;
 use Bolt\Filesystem\FilesystemInterface;
 use Bolt\Session\SessionStorage;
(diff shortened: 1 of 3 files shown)


src/EventListener/AccessControlListener.php
```diff
@@ -127,8 +145,9 @@ private function deleteSessions(Entity\Users $user)
     public static function getSubscribedEvents()
     {
         return [
-            StorageEvents::PRE_DELETE => 'onStorageEventPostSave',
-            StorageEvents::PRE_DELETE => 'onStorageEventPreDelete',
+            //\Bolt\Events\StorageEvents::PRE_DELETE => 'onStorageEventPostSave',
+            //\Bolt\Events\StorageEvents::PRE_DELETE => 'onStorageEventPreDelete',
```
bobdenotter, line 149: Did you comment these out instead of removing them on purpose?

GwendolenLynch: On purpose. Below you'll see this PR adds the listener. It hasn't been hooked up in a very very long time, so investigation of that isn't in scope for this one :smile_cat: 

A later PR can remove related logic if needed.

GwendolenLynch:
I much prefer the approach in #7529 :smile:

PR #7543, opened 2018-06-29
Exception RequestSanitiser.php crash fixed - support for nested arrays

### LINKED ISSUE #7540: Nested arrays in RequestSanitiser cause ContextErrorException
When you suddenly delete any required route from `routing.yml` (like `'homepage'`), on the login page you get a non-informative error:
image
When you put routes back - it works ok. This refers to #7175 . 

1. Can we detect those missing required routes and show more informative error?
2. Can we make a fallback for `homepage`, if those pages are missing? And, probably, show warning. Actually, those routes are necessary only for generating urls. It even doesn't need for controller action to exist. Only mapping `routeName => path`. Sooooo... we can make fallbacks to allow people to exist without those unused routes.

Boorj:
Fixes #7540

If a list item is an array - we `json_encode` it.

### DIFF
--- a/src/Helpers/RequestSanitiser.php
+++ b/src/Helpers/RequestSanitiser.php
@@ -46,7 +46,18 @@ final class RequestSanitiser
         foreach ($bag as $k => $v) {
             /** @var Bag $v */
             if (is_array($v) && !is_callable($v)) {
-                $bag[$k] = Bag::from($v)->join(' ');
+                // Flattening arrays, json encoding nested items
+                $flattenV = [];
+                foreach ($v as $item) {
+                    if (is_array($item)) {
+                        $flattenV[] = json_encode($item);
+                    }
+                    else {
+                        $flattenV[] = $item;
+                    }
+                }
+                // imploding to string
+                $bag[$k] = Bag::from($flattenV)->join(' ');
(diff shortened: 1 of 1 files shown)


bobdenotter:
I've commented on the other issue, since I can't yet verify this one. :-)

Boorj:
don't know what's wrong with that test. Something is deprecated or whatever

Boorj:
- when exception happens, it tries to renderException:
image

- it requires to get Context

image


- To return context, Exception listener runs request sanitiser's `filter`
image

- Filter gets session parameters with `$request->getSession()->all()` and passes them to  `getValues`
image


- one of session parameters is `retreat`, which has deeply nested structure:

image


```yml
retreat:
    route: dashboard
    params: 
        zone: backend
__csrf/user_login: tNYOyu8gp....
```

- but loop in getValues doesn't have any recursive calls, it only works with 1 level nesting:
image


### Conclusion
in my PR i solved that. Goal of `renderException` is to show parameters. If it has deeply (> 1lvl) nested structure, we can just `json_encode` nesting branches. 
Here's code for that
https://github.com/bolt/bolt/pull/7543/files#diff-78e469545651857cf6b133156561849aR49

PR #7544, opened 2018-06-29
increasing input width in _panel-actions_files

Boorj:
In some languages "keyword ..." is longer. So i increased width of that field for 20px to avoid placeholder clipping.


bobdenotter (approved):
👍

Boorj:
thanks) Comparison:
```
keyword
ключевое слово
```

PR #7565, opened 2018-07-15
[3.5][Tests] Set database:repair command call to non-interactive

GwendolenLynch:
This prevents blocking test execution when the test database schema can't be updated.


bobdenotter (approved):
👍

Issue #7558, opened 2018-07-12
("Notice: Undefined property: Bolt\Storage\Field\Collection\FieldCollection::$collection")


stryxx:
I found the bug during serialization of FieldCollection.
https://github.com/bolt/bolt/blob/3.6/src/Storage/Field/Collection/FieldCollection.php#L148
There you have $this->collection, which is non-existing property of class. There should be

`foreach ($this as $field) {` instead of `foreach ($this->collection as $field) {`

 

Details
-------

 - Relevant Bolt Version: 3.6
 - Install type: composer install
 - PHP version: 7.1 | 7.2
 - Used web server: Nginx 1.15.1

Reproduction
------------

Try to serialize fieldCollection.

### Bug summary

https://github.com/bolt/bolt/blob/3.6/src/Storage/Field/Collection/FieldCollection.php#L148
There you have $this->collection, which is non-existing property of class. There should be

`foreach ($this as $field) {` instead of `foreach ($this->collection as $field) {`

### Specifics

I'm using the bold version 3.6

### Steps to reproduce

Try to serialize FieldCollection.

### Expected result

- Serialization of FieldCollection works.

### Actual result

An error:
`("Notice: Undefined property: Bolt\Storage\Field\Collection\FieldCollection::$collection")`

PR #7566, opened 2018-07-16
[3.5] Fix for undefined property in FieldCollection::$collection

stryxx:
Fix serialization for field collection - remove reference to non-existing property.

When I want to serialize RepaterCollection with FieldCollection objects, I cannot do this because of reference to non-existing property and serialization cannot be done.

Fixes: #7558


GwendolenLynch (approved):

GwendolenLynch:
Thanks, @stryxx and welcome to the contributor's board :beers: 

I've pushed this down to 3.5 (it will arrive in 3.6+ via merge soon), and tweaked slightly.

Issue #7328, opened 2018-02-06
Preview with contentblocks shows no contentblocks


anketwokings:
## How to recreate
- Preview a page that has content stored in contentblocks.

- When previewing a page that has content stored in `contentblocks`, the contentblocks do not show.

## Expected behaviour
The contentblocks showing exactly as they do on the saved webpage

## Error
With `strict_variables: true`, the error that is given is 
```
Key "block" for array with keys "textsection" does not exist in
 "_sub_blocks.twig" at line 4.
```
The contentblocks are made exactly like the [docs](https://docs.bolt.cm/3.4/fields/block), even with the exact same fieldnames.

As I understand, the issue is that the contentblocks do not 'exist' in the way they are used in the template yet. 

Could this be fixed, because we had multiple complaints from editors using these contenttypes, or is there maybe a workaround for the preview? 

Also, if this is fixed in a newer version that is not available yet, feel free to close! 
Thanks all!

- Relevant Bolt Version: [ 3.4 ]

xiaohutai:
### Discrepancies between Preview and Saved Page

In abovementioned, `contentblocks` are of `type: block` (i.e. named repeaters). The difference between preview and saved page:

The preview results in:
```
array:3 [
  0 => array:1 [
    "block_1" => [...]
  ]
  1 => array:1 [
    "block_2" => [...]
  ]
  2 => array:1 [
    "block_3" => [...]
  ]
]
```

while a saved page has a `RepeatingFieldCollection` where each item's `block` name in the `RepeatingFieldCollection` can be retrieved as `{{ item.block }}`. From Twig's perspective, such a (hydrated) collection would look like:

```
RepeatingFieldCollection => [
    0 => [
        "block" => "block_1",
        ...
    ],
    1 => [
        "block" => "block_2",
        ...
    ],
    2 => [
        "block" => "block_3",
        ...
    ],
]
```

In preview, we get a (key, value)-pair, where the `key` is the `block` name.

---

#### Discrepancy with regards to a Taxonomy of Grouping type

Another difference I noticed was that a grouping taxonomy does not exist under `record.taxonomy.mygrouping`, but only under `record.group`, so I refactored that to `record.group.slug`.

---

### Temporary Quickfix via a Twig Filter

I think a simple Twig function to always populate each item with a field named `block` with the block's name could help a lot, but that's not something I'd put into Bolt; because it would override a field that actually had the name `block` and potentially break other features I am not aware of:

```
[...]

PR #7567, opened 2018-07-17
Fix data compatibility in preview page

stryxx:
Fix for data compatibility in preview page by creating empty contentObject and fill post data to it.

Fixes: #7328 

Details
-------

Repeater type field in preview page has wrong type during serialization, because of this field was array instead of RepeatingFieldCollection

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -195,13 +195,7 @@ class Frontend extends ConfigurableBase
         }
 
         $contenttype = $this->getContentType($contenttypeslug);
-
-        $id = $request->request->get('id');
-        if ($id) {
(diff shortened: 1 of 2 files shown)


bobdenotter:
Did some more testing, but I don't think this PR fixes the problem adequately.

JarJak:
@bobdenotter why?

bobdenotter:
@JarJak Thanks for pushing a bit. I tried this earlier, but I thought it didn't work properly. It doesn't solve the issue when previewing using the default 'record.twig' `{{ block('sub_fields') }}`, but it _does_ work better, with named repeaters / blocks, like `{% for group in record.contentblocks %}`.

👍

bobdenotter (approved):

PR #7570, opened 2018-07-19
[3.5] Hack around DBAL Sqlite schema diff bug

GwendolenLynch:
~Work~Hack around for upstream bug https://github.com/doctrine/dbal/pull/3220

This doesn't make me comfortable but is the best we can do until this lands upstream: https://github.com/doctrine/dbal/pull/3221


bobdenotter (approved):
👍

PR #7571, opened 2018-07-19
[3.6] Add class support to new storage and where applicable disable all legacy polyfills

rossriley:
OK, I think this is the last PR that makes 3.6 feature complete when using new storage instead of the old one.

This is all assuming you have:

```yml
compatibility:
    setcontent_legacy: false
```

in your config and you can now add something like this to your contenttype definition:

```yml
pages:
    name: Pages
    singular_name: Page
    fields:
        title:
            type: text
            class: large
            group: content
        slug:
            type: slug
            uses: title
        image:
            type: image
        teaser:
            type: html
            height: 150px
        body:
            type: html
            height: 300px
        template:
            type: templateselect
            label: Choose the template
    class: Bundle\Site\RossEntity
```

The only proviso here is that your entity needs to extend the base Content entity although this will hopefully change in the future when config and contenttypes have more precise classes with interfaces.

There's a few other minor clean ups in here, mainly code quality and when the flag is on this now allows us to disable all the legacy service injection into content entities.

### DIFF
--- a/src/Menu/MenuBuilder.php
+++ b/src/Menu/MenuBuilder.php
@@ -13,12 +13,17 @@ class MenuBuilder
     /** @var Application */
     private $app;
 
+    /**@var string @internal */
+    private $storageAccessor;
+
     /**
      * @param Application $app
+     * @param string      $storageAccessor
      */
-    public function __construct(Application $app)
+    public function __construct(Application $app, $storageAccessor = 'storage')
     {
         $this->app = $app;
+        $this->storageAccessor = $storageAccessor;
     }
 
     public function menu($identifier = null, $resolved = true)
@@ -203,8 +208,9 @@ class MenuBuilder
      */
     private function populateItemFromRecord(array $item, $path)
     {
-        /** @var \Bolt\Legacy\Content $content */
-        $content = $this->app['storage']->getContent($path, ['hydrate' => false]);
+        /** @var \Bolt\Legacy\Storage $engine */
+        $engine = $this->app[$this->storageAccessor];
+        $content = $engine->getContent($path, ['hydrate' => false]);
 
         if ($content) {
             if (empty($item['label'])) {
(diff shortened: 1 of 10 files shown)


bobdenotter:
🎉 🎊

src/Menu/MenuBuilder.php
```diff
@@ -56,6 +58,17 @@ public function resolve(array $menu)
         return $this->menuBuilder($menu);
     }
 
+    /**
+     * Here to allow transition between old and new query engines. For internal use only.
+     *
+     * @internal
+     * @param $accessor
+     */
+    public function setQueryEngine($accessor)
```
GwendolenLynch: I really don't think we need this, as we can just modify the constructor signature to add `$accessor = 'storage'`

GwendolenLynch (dismissed):

src/Menu/MenuBuilder.php
rossriley: Updated

bobdenotter:
I can't get this to work yet.. You mentioned: 

> The only proviso here is that your entity needs to extend the base Content entity

I assumed that'd be `Bolt\Storage\Entity\Content`. So, what I tried is this: 

In my contenttype I added: `class: Bundle\Site\BobEntry`. Made a `BobEntry.php` like this: 

```
<?php

namespace Bundle\Site;

use Bolt\Storage\Entity\Content;

class BobEntry extends Content
{

}
```

Gave This error: 

screen shot 2018-07-31 at 21 24 24

Tried this next: 

```
<?php

namespace Bundle\Site;

use Bolt\Legacy\Content;

class BobEntry extends Content
{

}
```

And got this error: 

screen shot 2018-07-31 at 21 26 02

Am I doing it wrong, or is there a quirk in this, still?

rossriley:
did you have the setcontent legacy flag set to false?

bobdenotter (dismissed):
> did you have the setcontent legacy flag set to false?

I was quite certain I did.. However, i checked out a fresh setup, and now it works like a charm! Awesome, @rossriley

GwendolenLynch (approved):

ToBe998:
Many thanks guys. That's a great feature!

Issue #7575, opened 2018-07-29
Search functionality breaks if there isn't at least one text field


eduardomart:
If you define a content type that doesn't have at least a single text input field, the front end functionality of your site will break.

Details
-------

 - Relevant Bolt Version: [ 3.5.5 | master ]
 - PHP version: [ 7.0 ]

Reproduction
------------

- Create a content type such as:
```
homepage:
    name: Homepage
    singular_name: Homepage
    singleton: true
    show_in_menu: Pages
    fields:
        images:
            type: repeater
            label: Images
            fields:
                image:
                    type: image
                    attrib: [ title ]
```

- Go to your front page and use the search box

### Specifics

- This is the error that you will see:
![image](https://user-images.githubusercontent.com/1833020/43370838-43a65bfc-933b-11e8-95f2-e731760d51cd.png)

- It appears the issue is that there isn't any text fields in the content type
- The error goes away if you add a text field to your content type
```
homepage:
    name: Homepage
    singular_name: Homepage
    singleton: true
    show_in_menu: Pages
    fields:
        name:
            type: text
            label: Title
            class: large
        images:
            type: repeater
            label: Images
            fields:
                image:
                    type: image
                    attrib: [ title ]
```

bobdenotter:
This bug happens in both 3.5 as well as in 3.6. 

In 3.6 the error is different though, and in fact it's fixed by #7584

PR #7584, opened 2018-08-04
Don't use `getContent` on non-existing legacy items.

bobdenotter:
Prevents breakage on new storage layer, if the template tries to fetch a non-existing Record. 

See screenshot:

screen shot 2018-08-04 at 11 50 40


bobdenotter:
After the latest changes, i can't reproduce this anymore.. Let's let it sit for a while, and if it doesn't pop up again, we can close this PR.

bobdenotter:
Found another reproducable case.. This PR fixes #7575 

@rossriley, could you take a look at this one too?

rossriley (approved):

PR #7593, opened 2018-08-10
Get news through a proxy w/o authentication

bmfp:
This patch allows bolt to look for news using a proxy without user/password fields, if general/httpProxy/user is empty

Details
-------

This PR may be tested using a simple squid proxy

Choosing a target branch
------------------------

Change made on branch 3.5


bobdenotter (approved):
Thanks, @bmfp ! 👍

PR #7599, opened 2018-08-11
[Hotfix 3.6] Allow Methods To work with array or ContentType

### LINKED ISSUE #7590: [3.6] Saving with New Storage gives error
Editing a Record with `setcontent_legacy: false` throws an error, when trying to save: 

screen shot 2018-08-08 at 16 18 13

Looks like this is because `$contenttype` is an Object now, instead of an Array. @rossriley, could you look into this?

rossriley:
Fixes #7590

Type hint error when using new storage level.


bobdenotter (approved):
Thanks, @rossriley 👍

Issue #7568, opened 2018-07-17
Slug numbers will not increase？


ziselive:
Details
-------

 - Relevant Bolt Version: [  3.5  ]
 - Install type: [ Zip/tarball install ]
 - PHP version: [ 7.1.15 ]
 - Used web server: [ Apache  ] version [ 2.4 ]
 - For UX/UI issues: [ browser name and version ]

Reproduction
------------
Slug numbers will not increase
Sending several articles in a row, Slug numbers has always been 1
/entry/
/entry/1
/entry/1
/entry/1
/entry/1

bobdenotter:
Seems like this is correct.. If you leave the title (or other field to fill the slug) empty, it doesn't properly generate a good slug.. If you input even a single character in the title field, it will work correctly:

![screen shot 2018-07-29 at 15 08 45](https://user-images.githubusercontent.com/1833361/43366638-77f6af0a-9341-11e8-83d3-f154992e50ad.png)

ziselive:
@bobdenotter  Slugs can be generated automatically if the title is in English or numbers.However, if the title text is a multi-character font, it will not automatically generate slug.

xiaohutai:
It seems it doesn't slugify those characters by default (probably assumes English).

You could make your own extension, or something like https://github.com/SahAssar/internationalslugs 
that handles your language/locale, and for example, for Chinese: https://github.com/cocur/slugify/pull/195

Edit: It seems we're still on v1.4.1 for `cocur/slugify`. Maybe that can be updated.

bobdenotter:
Seems like `cocur/slugify` is at 3.1 already, but it seems like it's still compatible with PHP `5.5.9`. So yes, let's see if we can update this

PR #7532, opened 2018-06-16
Slugifying folder names, created from file manager [tests update required]

Boorj:
**tests update required (!)**

### closes #7425 

When user creates a folder, if the folder name contains non-latin symbols, name is slugified. And if it's changed, we a show warning:
`Directory "%DIR%" was created with another name: "%SLUGDIR%".`

PR #7600, opened 2018-08-12
Update `cocur/slugify` to latest `^3.1.0`

bobdenotter:
First step in fixing #7568 and #7532


xiaohutai:
Do you think this may break extensions if they use and depend on the old Slugify class?

bobdenotter:
@xiaohutai If you used `Slugify::create()` with default settings, it'll be fine. If you used it to pass in a regexp, it'd fail.

xiaohutai (approved):

GwendolenLynch:
Thanks for breaking builds and years of trying to get SemVer on track. 

This is now one less site running Bolt today, more to follow.

Issue #7550, opened 2018-07-05
Content changes are not saved in backend


otrelin:
Sometimes I observe the situation that changes in content through the backend are not saved.

Details
-------

 - Relevant Bolt Version: [ 3.5 ]
 - PHP version: [ 7.2 ]
 - Used web server: [ Apache ]

Reproduction
------------

### Bug summary

I make changes to the contenttype item, click the Save button and I see a message under the button that the last save was a few seconds ago.
![1530781608411](https://user-images.githubusercontent.com/15085327/42313774-34007a0e-804c-11e8-8956-4edad9433015.jpg)
But when I reload the page, I see that there are no my changes and the message under the button, that last change was a month ago.
![1530781316450](https://user-images.githubusercontent.com/15085327/42313802-43b4d558-804c-11e8-8b97-b6fc35aaf13b.jpg)

### Specifics

It is difficult to describe the specifics, I did not notice them. This happens arbitrarily. Under the same conditions, it is sometimes saved, and sometimes not saved. Sometimes to save changes, I have to save three or four times with reload the page after every save for checking.

### Steps to reproduce

I just make changes to the content of the pages in the backend.

### Expected result

Changes must be saved each time I click the Save button.

If you need any more details, please tell me.

GwendolenLynch:
Probably a duplicate of #7512 in which case it should be covered in #7529 and released in v3.5.4 … could you confirm for us, please? :+1:

otrelin:
Yes, my version is 3.5.3.
I'll try to upgrade to version 3.5.4 and watch for this.

bobdenotter:
Most likely a duplicate of #7575, yes.

PR #7603, opened 2018-08-14
Ensure we have a CSRF login on session start and resume

bobdenotter:
Additional fix for issues #7512, #7575, #7550


SvanteRichter (approved):
Still not 100% on this, but I hate the bug!

PR #7604, opened 2018-08-14
Fix excerpts in new Storage

### LINKED ISSUE #7602: [3.6] `record.excerpt` doesn't work in new storage.
To reproduce, go to `/showcases` or `/search?search=ipsum` on a stock Bolt 3.6 setup. The `listing.twig` outputs content using `{{ record.excerpt }}`, only nothing shows. 

Output on 3.6: 

screen shot 2018-08-12 at 18 41 10

Output on 3.5: 

screen shot 2018-08-12 at 18 41 24

bobdenotter:
Fixes #7602.

Additionally, prevents excerpts from looking 'ugly' with serialised JSON in them.


src/Helpers/Excerpt.php
```diff
@@ -73,8 +74,14 @@ public function getExcerpt($length = 200, $includeTitle = false, $focus = null)
...
             array_walk($this->body, function ($value, $key) use (&$excerpt, $stripKeys) {
-                if (is_string($value) && !in_array($key, $stripKeys)) {
-                    $excerpt .= $value . ' ';
+                if (in_array($key, $stripKeys)) {
+                    return;
+                }
+                // We need non-empty strings that don't look like serialized JSON.
+                // Otherwise, Twig Markup is also OK.
+                if ((is_string($value) && !empty($value) && ($value[0]!='{')) ||
+                    (is_object($value) && ($value instanceof Markup))) {
```
SvanteRichter: is_object is not needed:

> instanceof does not throw any error if the variable being tested is not an object, it simply returns FALSE.

src/Helpers/Excerpt.php
```diff
@@ -73,8 +74,14 @@ public function getExcerpt($length = 200, $includeTitle = false, $focus = null)
...
             $excerpt = '';
             array_walk($this->body, function ($value, $key) use (&$excerpt, $stripKeys) {
-                if (is_string($value) && !in_array($key, $stripKeys)) {
-                    $excerpt .= $value . ' ';
+                if (in_array($key, $stripKeys)) {
+                    return;
+                }
+                // We need non-empty strings that don't look like serialized JSON.
+                // Otherwise, Twig Markup is also OK.
+                if ((is_string($value) && !empty($value) && ($value[0]!='{')) ||
```
SvanteRichter: Perhaps we should also check for `($value[0] !== '[')` (some extensions store json arrays IIRC), and strict comparison would be nice even though it probably won't make a difference.

SvanteRichter (dismissed):

src/Helpers/Excerpt.php
bobdenotter: Good points both. Will do.

src/Helpers/Excerpt.php
bobdenotter: Ack.

bobdenotter:
Updated.

SvanteRichter (approved):

PR #7605, opened 2018-08-14
Fix 'generate Lorem Ipsum' dummy content

bobdenotter:
With new storage, the 'generate Lorem Ipsum content' didn't work. this PR fixes that.


SvanteRichter:
Any chance this could be solved by adding a toString method instead?

bobdenotter:
It has a `__toString`. This does make the `$alias = (string) $alias;` consistent. 

Without it, it'll break on this: 

screen shot 2018-08-14 at 20 32 20

I initially had `function resolveClassName(string $alias)`, but that's PHP 7+. I preferred to 'juggle' it once in the beginning, as opposed to 6 times in the entire method.

SvanteRichter:
Ah gotcha. I guess there are a few of these we can remove in 4.0.

SvanteRichter (approved):

PR #7615, opened 2018-08-16
[3.6] Feature/sortable repeaters

eduardomart:
- Repeater sets are now draggable/sortable
- Removed move up/down buttons
- Added kebab for Repeater headers as the 'sortable' indicator
- Uses the first field `type=text` for the headers


bobdenotter:
I incorporated your commit into #7617, and removed the commit where you reverted it. ;-)

I'm going to tweak that one, and merge.

PR #7617, opened 2018-08-16
Drag-sort for repeaters

bobdenotter:
- Removed move up/down buttons
- Added kebab for Repeater headers as the 'sortable' indicator
- Uses the first field type=text for the headers


bobdenotter (commented):
Thanks, @eduardomart 👍

PR #7619, opened 2018-08-18
Fix issue with html fields

eduardomart:
issue: When the repeaters contain a field `type: html`, this field's content will be blank after sorting.


bobdenotter (approved):
:+1:

PR #7621, opened 2018-08-18
Feature: drag-sort for content blocks

eduardomart:
* Blocks are now drag-sortable
* Added kebab icon to indicate drag-sortable-ness
* Added `icon` property
* Use first `type: text` field for the block header, if not use the block's `label`
* The new `icon` property also is displayed in the "Add new" button dropdown, falls back to the regular plus sign, if one is not defined
* Removed the move up/down buttons


bobdenotter (approved):
Like a charm! ✨

PR #7630, opened 2018-08-21
Small update to repeater headers

eduardomart:
Removed the `caption` option as we'll default to the repeater's label for the individual headers if there is no type: text field contained in the repeater.


bobdenotter (approved):
👍

PR #7636, opened 2018-08-26
[3.5 Hotfix] Fix for Select values inside repeaters inside templatefields

### LINKED ISSUE #7547: [BUG] Regression on select choices for Templatefields
This side effect seems to have been introduced via the fix in #7505 

Templatefields options don't match the same structure so the select choices need to be built in the same way that repeaters/blocks were changed.

Reproduction, config in theme.yml:

```
templatefields:
   index.twig:
       banner:
           label: Banners
           type: repeater
           fields:
               banner_title:
                   label: Title
                   type: text
               banner_position:
                   label: Position
                   type: select
                   values: { 'align-left':'Left', 'align-right':'Right' }
```

No select options showing in the editcontent screen.

credit to @mikenosko for the bug report.

rossriley:
Fixes #7547 

The values lookup needed to be adjusted to handle repeaters in a templatefields field. Also the repeater-group template here is adjusted so that it can receive a pre-built set of values rather than looking them up.

This will need to be looked at to simplify because we've essentially got two separate code paths to handle normal vs templatefields repeaters but we can do that as a follow-up.

### DIFF
--- a/app/view/twig/editcontent/_templatefields.twig
+++ b/app/view/twig/editcontent/_templatefields.twig
@@ -20,7 +20,7 @@
                 {% set templatelabelkey = templatefield.label ?: templatekey|capitalize %}
 
                 {# Ugly hack to be removed with Forms cutover … kittens are crying #}
-                {% if templatefield.type == 'select' %}
+                {% if templatefield.type in ['select', 'repeater', 'block'] %}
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Thanks, @rossriley 👍

PR #7656, opened 2018-09-11
Add ContentUserTrait to allow `record.user` stuff

### LINKED ISSUE #7643: [3.6] `record.user` is not set.
Under "legacy", `{{ dump(record.user) }}` and `{{ record.user.displayname }}` outputted the `user` details. Using new storage, it doesn't yet. 

New storage: 

screen_shot_2018-09-04_at_11_28_31

Legacy: 

screen_shot_2018-09-04_at_11_27_50

@rossriley Could you take a look at this?

bobdenotter:
Fixes #7643

### DIFF
--- a/src/Storage/Entity/Content.php
+++ b/src/Storage/Entity/Content.php
@@ -18,6 +18,7 @@ class Content extends Entity
     use ContentRouteTrait;
     use ContentTypeTrait;
     use ContentTypeTitleTrait;
+    use ContentUserTrait;
 
     /** @var string|Mapping\ContentType */
     protected $contenttype;
--- a/src/Storage/Entity/ContentUserTrait.php
+++ b/src/Storage/Entity/ContentUserTrait.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Bolt\Storage\Entity;
+
+trait ContentUserTrait
+{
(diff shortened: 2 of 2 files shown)


rossriley (approved):

rossriley:
cool :+1: 

I'm just noting this for reference, since it would be good to come back and improve in the future. I think it would be nicer for separation if we did on this what we did on some of the other `record.method` functionality and convert them into Twig filters. so in templates we could do `record|user` etc.

The reason for this is that we can work at reducing the dependencies needed to make an entity and move the functionality out of the record model. 

However, this is a fine way to start because when we do get those twig functions done we can just point the trait here to the twig function, as I've done elsewhere on `record.previous()` and some other.

bobdenotter:
@rossriley Yeah, we've had this conversation with @CarsonF as well, and I'm still on the fence. From the perspective of the code, it makes sense. We just need to be certain it'll make sense for users who mainly work on implementing sites. We need to make sure a change like that will make sense to them. In this case i'm not sure, because the `user` is _of_ the record. Meanwhile, with something like `next`, it's an operation _on_ the record. 

Let's discuss in detail when we get a base to work on for v4. :-)

PR #7661, opened 2018-09-12
[3.6] Don't display actual Bolt version as Extension's requirement when it is not set

JarJak:
Continuation of #7486.


JarJak:
At least those two extensions are affected (however I have no clue why recaptcha is displayed here, it is only a requirement of boltforms in its composer)

screen shot 2018-09-12 at 10 01 52

bobdenotter (approved):
👍

PR #7663, opened 2018-09-12
[3.6] Fix preview blocks

### LINKED ISSUE #7328: Preview with contentblocks shows no contentblocks

JarJak:
Fixes #7328

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -267,13 +267,12 @@ class Frontend extends ConfigurableBase
     private function fixBlockFieldsForPreview($record)
     {
         foreach ($record->contenttype['fields'] as $fieldSlug => $fieldSettings) {
-
             if ($fieldSettings['type'] === 'block') {
                 $fieldValue = $this->getBlockFieldValue($record, $fieldSlug);
                 if ($fieldValue) {
                     $newArray = [];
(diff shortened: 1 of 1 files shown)


bobdenotter:
This needs a bit of work.. Using a 'showcase' record with the standard 'base-2018' theme, it spits out an error. This might be because of a case I can catch in the `_sub_fields` blocks, _or_ in the code itself.  

screen shot 2018-09-12 at 19 09 01

bobdenotter (approved):
👍

PR #7664, opened 2018-09-12
Added method to login as specific user without password

ToBe998:
As discussed on Slack, the added `loinAsUser()` method enabled extension creators to override the login mechanics and connect bolt to various kinds of SSO systems. Since the logic for user management and authentication is currently (pre 4.0) not centralized enough, a lot of needed code would have to be duplicated to archive this. (And we certainly dont want to go there ;) )

I also renamed the `loginCheckPassword()` method to `loginCheckUser()` since the name is more fitting it's function, especially when password checks are now optional.

PS: The file is identical to 3.5 and could also be backported?

I can provide guides on how to do this if needed.

### DIFF
--- a/src/AccessControl/Login.php
+++ b/src/AccessControl/Login.php
@@ -73,7 +73,35 @@ class Login extends AccessChecker
         $this->getRepositoryAuthtoken()->deleteExpiredTokens();
 
         if ($userName !== null && $password !== null) {
-            return $this->loginCheckPassword($userName, $password, $event);
+            return $this->loginCheckUser($userName, $password, $event);
+        } elseif ($authCookie !== null) {
+            return $this->loginCheckAuthtoken($authCookie, $event);
+        }
+
+        $this->systemLogger->error('Login function called with empty username/password combination, or no authentication token.', ['event' => 'security']);
+        throw new AccessControlException('Invalid login parameters.');
+    }
+
+    /**
+     * Attempt to login a user with the given password. Accepts username or
+     * email.
+     *
+     * @param string             $userName
+     * @param AccessControlEvent $event
+     *
+     * @throws AccessControlException
+     *
+     * @return bool
+     */
+    public function loginAsUser($userName, AccessControlEvent $event)
+    {
+        $authCookie = $this->requestStack->getCurrentRequest()->cookies->get($this->authTokenName);
+
+        // Remove expired tokens
+        $this->getRepositoryAuthtoken()->deleteExpiredTokens();
+
+        if ($userName !== null) {
+            return $this->loginCheckUser($userName, null, $event, true);
         } elseif ($authCookie !== null) {
             return $this->loginCheckAuthtoken($authCookie, $event);
         }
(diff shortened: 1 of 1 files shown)


src/AccessControl/Login.php
```diff
@@ -91,7 +119,7 @@ public function login($userName, $password, AccessControlEvent $event)
      *
```
JarJak, line 120: phpdoc needs an update

JarJak (changes requested):

src/AccessControl/Login.php
```diff
@@ -83,15 +83,44 @@ public function login($userName, $password, AccessControlEvent $event)
...
+
+        $this->systemLogger->error('Login function called with empty username/password combination, or no authentication token.', ['event' => 'security']);
+        throw new AccessControlException('Invalid login parameters.');
+    }
+
+    /**
+     * Check a user login request for username/password combinations.
+     *
+     * @param string $userName
+     * @param string $password
```
JarJak: string|null

JarJak (changes requested):

src/AccessControl/Login.php
ToBe998: Are you sure you want this? That line is actually untouched by me.

JarJak: You have changed behavior of this method so now passing null as $password (as you do in line 104) is correct.

src/AccessControl/Login.php
```diff
@@ -83,15 +83,44 @@ public function login($userName, $password, AccessControlEvent $event)
...
+        // Remove expired tokens
+        $this->getRepositoryAuthtoken()->deleteExpiredTokens();
+
+        if ($userName !== null) {
+            return $this->loginCheckUser($userName, null, $event, true);
+        } elseif ($authCookie !== null) {
+            return $this->loginCheckAuthtoken($authCookie, $event);
+        }
+
+        $this->systemLogger->error('Login function called with empty username/password combination, or no authentication token.', ['event' => 'security']);
```
bobdenotter: Message needs to be changed so it's clear it's from `loginAsUser`, as opposed to a login with password.

src/AccessControl/Login.php
```diff
@@ -83,15 +83,44 @@ public function login($userName, $password, AccessControlEvent $event)
     }
 
     /**
-     * Check a user login request for username/password combinations.
+     * Attempt to login a user with the given password. Accepts username or
```
bobdenotter: Should be updated

src/AccessControl/Login.php
bobdenotter: Yes, it gets passed as `null` from `loginAsUser`.

bobdenotter (changes requested):
I've added some comments, and there's one more thing i think should be updated: 

```php
         if (!Blowfish::detect($userAuth->getPassword())) {
             $userEntity->setPassword($this->passwordFactory->createHash($password, '$2y$'));
        …
         }
```

This part should only be executed, it `$ignorePassword` is `false`.

src/AccessControl/Login.php
ToBe998: like this?

src/AccessControl/Login.php
ToBe998: like this?

bobdenotter (approved):
Works like a charm, ✨

PR #7665, opened 2018-09-12
[3.6+ tracker] setcontent_legacy: false

JarJak:
Looks like the New Storage seems to fully work, so let's enable it by default.


bobdenotter:
Looks like we have a few tests to fix yet. 😇 

To keep track: 

 - 2018-09-12: 30 failures + errors in PHPunit, 9 failures + errors in CodeCeption

JarJak:
- 2018-09-13: 28 failures + errors in PHPunit, 8 failures + errors in CodeCeption

rossriley:
I'm not really in favour of doing this in a point release. The idea of the switch was to allow us to switch over in 4.0 but allow the rest of the 3.x series to allow people to gradually transition.

There will be some intentional breakages between the two, otherwise we end up copying over the mistakes of the original storage where we need to have app dependency in the content entities etc if we switch over automatically outside of a major version then we tie our hands a lot more in what design decisions we can take as to breaking changes.

bobdenotter:
Maybe we shouldn't switch in the default, but this is still a good check to see if we're getting close to 100% coverage. :-)

> There will be some intentional breakages between the two, otherwise we end up copying over the mistakes of the original storage

Speaking of which.. In light of our recent changes in the core team, we should re-evaluate how much stuff we want to break going forward.. I'm leaning more and more to making a clean break and doing things in v4 the way we want, and do it as a proper SF 4 solution.

JarJak:
Keep

Wt., 13 lis 2018, 14:24: stale[bot] <notifications@github.com> napisał(a):

Issue #7667, opened 2018-09-13
[3.6] Omnisearch doesn't work in New Storage


bobdenotter:
Also showed up in #7665.

To see in action, go to `/async/omnisearch?q=lorem`

<img width="922" alt="screen shot 2018-09-13 at 07 59 10" src="https://user-images.githubusercontent.com/1833361/45469780-29f1f680-b72b-11e8-8b70-d3d781b8d9de.png">

bobdenotter:
Fixed in #7670

PR #7670, opened 2018-09-13
Fix search results (Frontend, Omnisearch, Async)

xiaohutai:
Fixes: #7667 

- Fixes Omnisearch in Async
- Fixes search results in Frontend search as well: it was always grouped per contenttype (but sorted within a contenttype group):
  - page/1
  - page/2
  - page/3
  - entry/1
  - entry/2
  - entry/3
  - showcase/1
  - showcase/2
  - showcase/3

The results between legacy and non-legacy are different, that may be because of a new/different search weighter.

For the code I looked at `Frontend`'s handling of search results. Not sure about the proper location for `getSortedResults` (see https://github.com/bolt/bolt/commit/d4e58ed815037debad013543c588d26acc6272c8). I'm kinda out of the loop with regards to the architecture.

Feedback more than welcome.

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -518,7 +518,7 @@ class Frontend extends ConfigurableBase
             $searchResult = $this->getContent($textQuery, $params);
 
             $result = [
-                'results' => $searchResult,
+                'results' => $searchResult->getSortedResults(),
                 'query'   => [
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Like a mofo! 👨‍🎤

PR #7672, opened 2018-09-13
[3.6] Fix new setcontent

JarJak:
What gets fixed:
- `where` parameters are taken into account
- order by (default order happens only if no `orderby` is set in setcontent)


JarJak:
I'm not sure if tests needs to be updated, or actual behavior (no way to override default order in setcontent) was expected...

bobdenotter (approved):
Looks good to me, as well! 👍

PR #7675, opened 2018-09-14
Use senderMail in outgoing mail

### LINKED ISSUE #6398: [RFC] Add default email variable to config
For email sending stuff some servers need a sending email address. 
If that is not set Bolt will send emails, but it will end up in some black hole without any error messages. 

This happens in the members extension if you don't set the `notification.name` and `notification.email` config vars. https://github.com/bolt/Members/pull/80
Boltforms also fails to send an email if no `from_email` is set.
The password reset email bypasses this by setting the sender to the email address of the account that requests the password.

I propose that we add a `default_email` to the global `config.yml` that all extensions can fall back to.
With that variable a lot of these problems can be prevented, and site admins only need to set that address once.

```yaml
# If you're trying out Bolt, just keep it set to SQLite for now.
database:
    driver: sqlite
    databasename: bolt

# The name of the website
sitename: A sample site

# The emailaddress that will be used by default for outgoing emails
default_email: webmaster@example.com

# The theme to use. Be sure to copy the {theme_path}/base-2016 folder, and
# change the name here accordingly.
```

If the `default_email` is not set the emailaddress of the first user with the 'root' role can be used.

The warning about the email configuration can also be amended to include sanity checks for this variable.

The `members` and the `boltforms` extension should probably be updated to fallback to that variable too :wink:

jadwigo:
This PR makes Bolt use general/mailoptions/senderMail for outgoing account emails.
It also explains the option in the config.yml.dist

This fixes #6398

### DIFF
--- a/app/config/config.yml.dist
+++ b/app/config/config.yml.dist
@@ -338,6 +338,9 @@ liveeditor: false
 #
 # Protip: If your webhost does not support SMTP, sign up for a (free) Sparkpost
 # account at https://www.sparkpost.com/pricing/ for sending emails reliably.
+#
+# The mail defaults use bolt@yourhostname with the site title as a default.
+# Override this with the senderName and senderMail fields
 
 # mailoptions:
 #     transport: smtp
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
:+1:

PR #7681, opened 2018-09-26
[Hotfix] directives should always be cleared before re-parsing

### LINKED ISSUE #7679: [3.6][BUG][New Storage] Setcontent returns single record even without returnsingle
Setcontent returns single record even without returnsingle **if run without params and after some returnsingle query**.
The record is not a singleton.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6 with new storage enabled

Looks like ContentQueryParser remembers directives between queries.

Reproduction
------------

### Steps to reproduce

There is more than zero Record entries in database.

{% setcontent record = 'records/1' returnsingle %}
{% setcontent records = 'records' %}
{{ dump(records) }}

### Expected result

QueryResultset (or any other iterable) with single record inside

### Actual result

Single record object

rossriley:
Fixes #7679 

When running a new query the flags always need to be reset. Fixes an issue where flags like `returnsingle` were sticking around between queries. Credit @JarJak for spotting this one.

### DIFF
--- a/src/Storage/Query/ContentQueryParser.php
+++ b/src/Storage/Query/ContentQueryParser.php
@@ -202,11 +202,11 @@ class ContentQueryParser
      */
     protected function parseDirectives()
     {
+        $this->directives = [];
+
         if (!$this->params) {
             return;
         }
-
-        $this->directives = [];
 
         foreach ($this->params as $key => $value) {
             if ($this->hasDirectiveHandler($key)) {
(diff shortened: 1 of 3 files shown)


bobdenotter:
Looks good to me.. 

@JarJak, can you confirm it's working for you too?

JarJak:
Fix works. It would be nice to add test covering this situation.

bobdenotter:
Still has test failures unfortunately, @rossriley

rossriley:
Yes, I know, this seems to be different now, codeception failure now, I'm not sure if it's related to this PR but I'm trying to reproduce locally.

bobdenotter (approved):
Thanks, @rossriley 👍

PR #7698, opened 2018-10-24
Stop stack being queried in _macro.twig when stack is disabled

benbriedis:
If assets are stored in an external repository (e.g. S3) then file accesses are slow and disabling the stack can greatly speed up the backend.  This fix removes a stack query that is run when the stack is disabled and causes a major slowdown.  (PS I'm new to pull requests and Bolt, sorry if I've done something wrong).


bobdenotter (approved):
Looks good to me. Thanks, @benbriedis 👍

PR #7699, opened 2018-10-24
[3.5] fix for incompatibility with latest passwordlib (origin and bolt fork)

ToBe998:
There has been a change in the latest PasswordLib package, both the original package and the bolt fork. Du a change i method signature, the extending legacy code in bolt now throws a warning.

```
ContextErrorException in PasswordLibFactory.php line 15:
Warning: Declaration of Bolt\Legacy\PasswordLib\PasswordLibFactory::createHash($password, $prefix = '$2a$') should be compatible with PasswordLib\Password\Factory::createHash($password, $prefix = '$2a$', array $options = Array)
```

This fix is actually only a backport for a change that is already in Bolt 3.6

### DIFF
--- a/src/Legacy/PasswordLib/PasswordLibFactory.php
+++ b/src/Legacy/PasswordLib/PasswordLibFactory.php
@@ -22,7 +22,7 @@ final class PasswordLibFactory extends Factory
     /**
      * {@inheritdoc}
      */
-    public function createHash($password, $prefix = '$2a$')
+    public function createHash($password, $prefix = '$2a$', array $options = array())
     {
         if ($prefix === false) {
             throw new \DomainException('Unsupported Prefix Supplied');
@@ -30,7 +30,7 @@ final class PasswordLibFactory extends Factory
         foreach ($this->implementations as $impl) {
             if ($impl::getPrefix() == $prefix) {
                 /** @var AbstractPassword $instance */
-                $instance = new $impl();
+                $instance = new $impl($options);
                 $instance->setGenerator($this->getGenerator());
 
                 return $instance->create($password);


bobdenotter:
Travis is not having a good day with this.. We'll need to see if it needs fixing, or if it's just Travis acting up.

bobdenotter:
@ToBe998 I'm not sure this patch it correct. Or even needed.. 

If you do `rm -rf composer.lock vendor; composer install` and then `composer show | grep passwordlib`.. Does it show this:

```
passwordlib/passwordlib            1.0.0-beta1        A Password Hashing Library
``` 

or this: 

```
bolt/passwordlib                   1.0.1    A Password Hashing Library
```

In case of the latter, what does `composer why bolt/passwordlib` say?

bobdenotter:
Pinging @tobe ;-)

ToBe998:
Here is a list of the situation.

**Library versions**
- passwordlib/passwordlib - master - **has $options**
- passwordlib/passwordlib - 1.0.0beta1 (last release) - none
- bolt/passwordlib - master - **has $options**
- bolt/passwordlib - 1.0.1 (last release) - **has $options**
- bolt/passwordlib - 1.0.0beta1 (last release) - none

**Bolt Requirements**
- Bolt 3.6.0 - has $options - requires bolt/passwordlib ^1.0 - ok
- Bolt 3.5.7 - none - requires passwordlib/passwordlib ^1.0@beta - ok

Which means ... i have no clue why i could not get it to work oO. I'm using 3.5.7 with it's original requires. I only changed it to debug this issue. I'll debug this a bit more and close the ticket if it turns out i did something stupid. Doing asap!

ToBe998:
I found the problem! If your project already has the `bolt/passwordlib` package installed (maybe because you tried 3.6 before and had to downgrade), composer will NOT remove that library and repace it with the proper `passwordlib/passwordlib` package.

I dont know the dependency resolver of composer to well, maybe it thinks they are both compatible.

Anyways, if you simply remove the wrong package and do a normal `composer update`, the proper package will get installed.

So ... not Bolt's fault ;)

PR #7705, opened 2018-10-30
Fix for field type blocks with translation extension

tiito78:
This block separately handles JSON content for repeaters and blocks. Prevent from getting "__internal" error on blocks when using translate extension.

### DIFF
--- a/src/Storage/Field/Type/RepeaterType.php
+++ b/src/Storage/Field/Type/RepeaterType.php
@@ -192,17 +192,38 @@ class RepeaterType extends FieldTypeBase
 
         // This block separately handles JSON content for Templatefields
         if (isset($data[$key]) && Json::test($data[$key])) {
-            $originalMapping[$key]['fields'] = $this->mapping['fields'];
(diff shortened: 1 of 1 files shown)


bobdenotter:
@tiito78 *bump* Did you see my question on Slack? Could you update the description of your PR's to detail what it is they fix?

doenietzomoeilijk:
I can confirm this (together with BoltTranslate/Translate#186) fixed the problem with translations for Block field types in a Bolt 3.6 + Translate 4.1.4 setup, where enabling translation for a Block field would result in errors thrown when you'd switch locales. In short: I had EN and NL locales, added a record for EN, saved it, switched to NL, edited something in the block field, saved, switched back to EN, error.

bobdenotter:
Thanks, @tiito78 (and @doenietzomoeilijk for testing) 👍

tiito78:
great @bob, and thanks a lot for all the hard work in making the best cms.... Hope I can help more...

bob:
You are welcome @tiito78

PR #7709, opened 2018-11-07
Performance improvements: No joins in query for pagers

bobdenotter:
As discovered by Czeslaw on Slack: 

> @bob previously i was too optimistic :slightly_smiling_face: turned out that te server was holding up some kind of temporary cache for the query and it returned the results from this cache :slightly_smiling_face: but - i found a proper solution, i've added `$queryCopy->resetQueryPart('join');` to line 131 of `Listing.php`
because that is basically the place where the [paging] occurs
this way i preserved all the paging behavior, but unbind the unnecessary relations
this works fine, and gives 1.5ms reaction

### DIFF
--- a/src/Storage/ContentRequest/Listing.php
+++ b/src/Storage/ContentRequest/Listing.php
@@ -129,6 +129,7 @@ class Listing
             $queryCopy->setMaxResults(null);
             $queryCopy->setFirstResult(null);
             $queryCopy->resetQueryPart('orderBy');
+            $queryCopy->resetQueryPart('join');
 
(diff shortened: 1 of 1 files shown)


rossriley:
Sorry, missed this PR, as you can see from the build failure it's too blunt to just remove joins from the queries because a lot of the content hydration depends on it, there's also cases like in taxonomy queries and relation queries where having the join there is directly needed for calculating the number of results in a set, so removing it here is going to cause issues.

That being said these queries should be about as quick as they can be so it would be interesting to have an example of what is slowing things down.

PR #7712, opened 2018-11-09
Handle Empty Search Parameter Error

### LINKED ISSUE #7711: New storage system can't handle empty keyword on search
Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6
| Install type                   |  any
| PHP version                 | any
| Web server                  | any
| For UX/UI issues        | -

Reproduction
------------

### Bug summary

If I call search page with empty keyword when content storage set to not use legacy system, it throws an error.

### Specifics

It's clear to me that empty search keyword should pass this test [here](https://github.com/bolt/bolt/blob/3.6/src/Storage/Query/QueryParameterParser.php#L54), but `$word` pattern has `+` modifier and we have zero words here.
I could send you a PR for this, but I am not sure replacing `+` modifier with `*` will solve the problem (in this specific issue it does) or it will make another elsewhere. I see that `$word` is used in almost every matcher.

### Steps to reproduce

- Set `setcontent_legacy: false` under `compatibility` key in `config.yml`
- Call search page without any keyword (ie. empty keyword)

### Expected result

Search listing should return all content with pager.

### Actual result

Error thrown:

```
QueryParseException in QueryParameterParser.php line 312: 
No matching value found for "%%"
```

rossriley:
Fixes #7711

Credit to @ntomka for spotting this one. Since the content search depends on having a value to search against we need to make sure we don't send empty strings to the search handler.

If we get one, we just return an empty resultset, site owners can then handle the error in frontend rather than throwing an exception.

### DIFF
--- a/src/Storage/Query/Handler/SearchQueryHandler.php
+++ b/src/Storage/Query/Handler/SearchQueryHandler.php
@@ -32,6 +32,11 @@ class SearchQueryHandler
             $query->setContentType($contentType);
 
             $searchParam = $contentQuery->getParameter('filter');
+            /** If we have an empty search filter then we need to return early */
+            if (empty($searchParam)) {
+                return $set;
+            }
+
             $query->setParameters($contentQuery->getParameters());
             $query->setSearch($searchParam);
 


ntomka:
Great! Thank you!

bobdenotter (approved):
Thanks, @rossriley

PR #7720, opened 2018-12-05
Fix setting up legacy Contenttype object

JarJak:
$this->app['storage']->getContenttype($contentType) can return object or array.


bobdenotter:
Looks good, but I don’t have my laptop with me at Symfonycon.. Is this urgent, or is it OK if i merge in after the weekend? :-)

JarJak:
@bobdenotter ok it is not urgent :) Have fun in Lisbon! 🍻

bobdenotter (approved):
👍

PR #7722, opened 2018-12-06
Fix filters for Content without taxonomy

JarJak:
Even if Content doesn't have a taxonomy, taxonomy filter's label was displayed.
Also a bit of code moves to make this file more clear ;)


bobdenotter (approved):
👍

PR #7725, opened 2018-12-11
Fixed asset url generation

ToBe998:
Before this change, the url of assets was generated through Package->getPath(), which relys on the master request inside the request stack. All of this is executed during initialization of the application inside the Applications constructor.

**Problem**
The request stack is empty during this execution phase as it is only pushed to the stack at start of `$app->run()`. THis means, the old method is working on an empty request and can not see any potential url prefixes from subfolder installs.

Interestingly, bolt's own assets are added inside the twig templates via `{{asset()}}`, which is executed during `$app->run()` and using a proper request.

**Solution**
The url generation has been moved from the `add` to the `processAsset` method. This method is executed during `$app->run()` and contains a filled request stack.

I also changed the key of the assets in 'add() ' from the now non existant `$url` to it's `$path`. It seems to be used only to check for uniqueness of assets and is working correctly there.

### DIFF
--- a/src/Asset/File/Queue.php
+++ b/src/Asset/File/Queue.php
@@ -62,10 +62,12 @@ class Queue implements QueueInterface
             $asset->setPackageName('extensions');
         }
 
+        $key = $asset->getPackageName()."/".$asset->getPath();
+
         if ($asset->getType() === 'javascript') {
-            $this->javascript[$asset->getPath()] = $asset;
+            $this->javascript[$key] = $asset;
         } elseif ($asset->getType() === 'stylesheet') {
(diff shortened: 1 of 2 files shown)


ToBe998:
Any news here or hint's on what to do to finally get this bugfix rolling?

ToBe998:
I had to remove the old test for url generation of assets. I changed it to test uniqueness of added assets as that's what the url was formerly used for. Also the real url test is not really possible without a fully faked request object and is most probably done is a completely different test. (It's not part of the tested Queue class anyways).

bobdenotter (approved):
Looks good to me, @ToBe998 👍

PR #7742, opened 2019-02-13
[3.x] Limit when values of a select is an entity

romulo1984:
The use of the limit, in a field of type select, only happens when the values are of type Yaml. This small change makes it possible to use threshold when the value is an entity.

A major problem that can occur with the lack of this feature is when a filter is not used and returns many results. Because there is no limit, the query may break the application and return a fatal error.

Usage
-------
```yaml
...
  fields:
    breaking_news:
      type: select
      values: entries/id,title
      keys: slug
      sort: -id
      limit: 10
```

If the property is not specified, the default limit of 500 is considered.

### DIFF
--- a/src/Form/Resolver/Choice.php
+++ b/src/Form/Resolver/Choice.php
@@ -146,6 +146,7 @@ final class Choice
 
         $filter = $field->get('filter');
         $filter['order'] = $field->get('sort');
+        $filter['limit'] = $field->get('limit') ?? 500;
         /** @var QueryResultset $entities */
         $entities = $this->query->getContent($contentType, $filter);
         if (!$entities) {
--- a/src/Form/Resolver/Choice.php
+++ b/src/Form/Resolver/Choice.php
@@ -115,7 +115,7 @@ final class Choice
      */
     private function getYamlValues(Bag $field)
     {
-        $values = array_slice($field->get('values', []), 0, $field->get('limit', DEFAULT_LIMIT), true);
+        $values = array_slice($field->get('values', []), 0, $field->get('limit', self::DEFAULT_LIMIT), true);
         if ($field->get('sortable')) {
             asort($values, SORT_REGULAR);
         }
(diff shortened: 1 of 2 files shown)


JarJak:
> If the property is not specified, the default limit of 500 is considered.
I don't think we should limit it by default. Also docs needs an update.

JarJak:
Limit is for maximum number of options to be selected, not for limiting number of choices.
However this issue could be fixed with using autosuggestion AND limit, so after entering more chars you can get what you need.

romulo1984:
@JarJak, I think you got the wrong one or the application works incorrectly. For yaml values, this limit is applied to limit the number of available options and not the number of selections.

https://github.com/bolt/bolt/blob/7a3a498123dd90e9854f75325a928a99fba5c5bf/src/Form/Resolver/Choice.php#L114-L122

JarJak:
>Use limit to limit the amount of items. Note that the default is 500, so if you have a long list, and you need to show them all to the user, raise this limit.

So yeah you are right, just the docs are not clear here :/

src/Form/Resolver/Choice.php
```diff
@@ -146,6 +146,7 @@ private function getEntityValues(Bag $field)
 
         $filter = $field->get('filter');
         $filter['order'] = $field->get('sort');
+        $filter['limit'] = $field->get('limit') ?? 500;
```
JarJak: $field->get('limit') has 500 as default value so ` ?? 500` is useless here.

Or if I am wrong and $field->get('limit') returns null, then 500 should be also used in `getYamlValues()`.

JarJak (changes requested):

src/Form/Resolver/Choice.php
romulo1984: @JarJak I have removed the coalesce (because it is not compatible with PHP 5.5.9) and kept 500 as default value. In my tests, I saw no limit in both cases, returning a very large result set.

With this change, the documentation will now have this information as true. :smile:

JarJak: Great! Can you set 500 as `DEFAULT_LIMIT` const?

JarJak:
Looks like tests need an update too :)

romulo1984:
@JarJak Up-to-date tests.

JarJak (approved):

JarJak:
Thank you @romulo1984 !

romulo1984:
Thank **you**, @JarJak 

When is the next release expected to go out?

Issue #7741, opened 2019-02-13
[RFC] Option for other attr on imagefields


ankedsgn:
Right now, the options for attributes on the imagefield are limited to `alt` and `title`.
I sometimes run into situations where I'd like more or other attributes. 

If I'm correct, the value of the imagefield is stored as JSON on the db, and the attributes are retrieved with `record.values.image.title` and `record.values.image.alt`.
 
So other then that it's not actually working at the moment, I don't think there would be a technical problem with implementing something like `record.values.image.customvalue` ?

Some thoughts on that?

sbonardt:
:+1: from me. I would like the option to add a 'caption' for instance

PR #7743, opened 2019-02-15
Allow custom attributes for image fields

ankedsgn:
Removing the boudaries that an attribute on an imagefield can only use 'title' of 'alt'.
Make way for custom attributes, such as multilingual captions and more.

Hope this can merge, and may help others!

Fixes: #7741

### DIFF
--- a/app/view/twig/editcontent/fields/_image.twig
+++ b/app/view/twig/editcontent/fields/_image.twig
@@ -54,7 +54,7 @@
 
 {% set attribute_fields = [] %}
 {% for attrib in option.attrib|default([]) %}
-    {% set aid = attrib|lower|replace({'alt text': 'alt'}) %}
+    {% set aid = attrib|lower|replace({' ': '_'}) %}
 
     {% if aid is not empty %}
         {% if aid == 'title' %}
@@ -62,7 +62,7 @@
         {% elseif aid == 'alt' %}
             {% set atitle = __('field.image.label.alt') %}
         {% else %}
-            {% set atitle = aid %}
+            {% set atitle = attrib %}
         {% endif %}
 
         {% set uid = key ~ '-' ~ aid %}
--- a/app/view/twig/editcontent/fields/_image.twig
+++ b/app/view/twig/editcontent/fields/_image.twig
@@ -56,11 +56,13 @@
 {% for attrib in option.attrib|default([]) %}
     {% set aid = attrib|lower|replace({'alt text': 'alt'}) %}
 
-    {% if aid in ['title', 'alt'] %}
+    {% if aid is not empty %}
         {% if aid == 'title' %}
             {% set atitle = __('general.phrase.title') %}
         {% elseif aid == 'alt' %}
             {% set atitle = __('field.image.label.alt') %}
+        {% else %}
+            {% set atitle = aid %}
         {% endif %}
 
         {% set uid = key ~ '-' ~ aid %}


jadwigo (dismissed):
The removal of the limit on the names of fields looks simple, and it makes 'data-something' attributes possible.

app/view/twig/editcontent/fields/_image.twig
```diff
@@ -56,11 +56,13 @@
...
     {% set aid = attrib|lower|replace({'alt text': 'alt'}) %}
 
-    {% if aid in ['title', 'alt'] %}
+    {% if aid is not empty %}
         {% if aid == 'title' %}
             {% set atitle = __('general.phrase.title') %}
         {% elseif aid == 'alt' %}
             {% set atitle = __('field.image.label.alt') %}
+        {% else %}
+            {% set atitle = aid %}
```
JarJak: {% set atitle = attrib %}

JarJak (changes requested):

JarJak:
Can you provide a test for it?

JarJak:
Also the docs needs an update.

app/view/twig/editcontent/fields/_image.twig
jadwigo: that would make that special 'alt text' case even weirder?

jadwigo:
> Can you provide a test for it?

is there a test for the old situation?

ankedsgn:
@JarJak Working on the docs right now

ankedsgn:
> Can you provide a test for it?

I've never made a test before, so at this moment no.

JarJak:
> > Can you provide a test for it?
> 
> is there a test for the old situation?

I don't know, does it matter?

app/view/twig/editcontent/fields/_image.twig
ankedsgn: My preference would be to delete the whole special 'alt text' case, but I have no clue why that was added there in the first case. It's not needed anymore now.

JarJak: I think the special 'alt text' is here for historical reasons. It is 3 years old and not documented so I think this replace can be removed too.

ankedsgn:
@JarJak Other then discussing what was there before or not; is there a manual / pointer / explanation about how to make tests for Bolt? I am a Front-end dev and other then bolt and Twig, I have no experience in writing those. Also, Was this mentioned in the PR manual and did I miss it? If not, worth adding it there too.

JarJak:
@anketwokings yes you are right, writing acceptance tests for Bolt 3 can be hard for frontend devs since it requires PHP knowledge (it uses Codeception as testing framework).

I will write this test, no problem.

In Bolt 4 those tests are written in Kakunin which is a JS thing, so writing tests should be simpler :).

ankedsgn:
> I will write this test, no problem.

That's great, thank you!

JarJak (approved):

JarJak:
Thank you @anketwokings !

bobdenotter:
Thanks, @anketwokings. Could you also add a sentence or two to the docs to explain the added functionality? https://docs.bolt.cm/3.6/fields/image

ankedsgn:
@bobdenotter Wasn't that already done in https://github.com/bolt/docs/pull/948 ?

bobdenotter:
@anketwokings Yes! Completely missed that one.. Never mind me.

PR #7745, opened 2019-02-26
[SECURITY FIX] filemanager file extensions

JarJak:
A logged in user could rename previosly uploaded file to any extension. Uploading malicious file and changing it's extension to .php could lead to remote code execution.

Now files can be renamed only to extensions that are allowed for upload.


bobdenotter (approved):
🚀

PR #7749, opened 2019-02-28
Log a warning instead of info when permissions are not configured

JarJak:


bobdenotter (approved):
Looks good to me, too! 👍

PR #7751, opened 2019-03-01
Allow lower and upper case file extensions

### LINKED ISSUE #7405: Can't upload 'foo.JPG' because of uppercase extension
Repro:
---
* goto file manager
* try to upload file with 'JPG' extension.
* it gives the error "can't upload because of extension" 👎 

## solutions
* check in one case (uppercase/lowercase)
* allow upload uppercase files .JPG
* rename file to lowercase extension after upload

JarJak:
Fixes #7405


bobdenotter (approved):
I know of a former teammember who was dead-set against this, but I think this change makes sense! 👍

PR #7660, opened 2018-09-11
[3.6][New Storage] Fix preview

JarJak:
Fully working fix for previewing content.

Fixes #7658

PR #7754, opened 2019-03-07
Add extra workaround for HTML-like and Image fields in preview

### LINKED ISSUE #7703: Previewing content using {{record.field}} displays html in output for text/editor field type
When previewing content using a template that outputs the the property/field of the record object (e.g. {{ record.body }} ), it encodes the html stored in the field value and outputs that.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6.0
| Install type                   |  tar.gz
| BC Break                     | yes/no
| PHP version                 | 5.6

Reproduction
------------
On clean install, create content for page content_type. Add bold/italic/newline etc to body field in editor.
Save and preview.

### Expected result
Body field HTML to be output as:
```
<p>test</p>

<p></p>

<p><strong>test</strong></p>

<p></p>
```

### Actual result
Body field HTML output as:

```
&lt;p&gt;test&lt;/p&gt;

&lt;p&gt;&lt;/p&gt;

&lt;p&gt;&lt;strong&gt;test&lt;/strong&gt;&lt;/p&gt;

&lt;p&gt;&lt;/p&gt;

&lt;p&gt;test&lt;/p&gt;
```

Doesn't occur in v3.5 for what I can see.

Thanks,
Mark

### LINKED ISSUE #7718: Error image field on preview page
I have field `banner` with type `image`.
default

On my page template `page.twig` I use this field:
default

In front site everything is ok. But when I click "Preview" button in backend
default
I get this error:
default

The problem is that the `image` type field is obtained in different ways.
When I do
`{{ dump(record.banner) }}`
in frontend I see
default
and in preview I see
default

This error began to manifest itself with version 3.6.0 and is present in 3.6.1 and 3.6.2.
In version 3.5.7 there was no such error - in both cases the field was returned as a string.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6

### LINKED ISSUE #7753: Preview displays HTML fields as escaped HTML.
Maybe duplicate of #7703 . 

This is happening since 3.6.0. HTML-like fields are passed into the preview as `string`, instead of as `Twig_Markup`, which makes Twig render them as extra escaped HTML. See screenshots: 

schermafbeelding_2019-03-07_om_10_42_59

schermafbeelding_2019-03-07_om_10_42_53

PR with fix upcoming!

bobdenotter:
This is a workaround to make "preview" behave more like "Normal view" of a record. 

It builds on #7660, and does a similar thing for HTML-like fields and Image fields. 

 - Fixes #7753 
 - Fixes #7703 
 - Fixes #7718

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -228,7 +228,6 @@ class Frontend extends ConfigurableBase
         }
 
         $this->fixBlockFieldsForPreview($content);
-
         $this->fixHTMLFieldsForPreview($content);
 
         $liveEditor = $request->get('_live-editor-preview');
@@ -294,7 +293,7 @@ class Frontend extends ConfigurableBase
      * an image. Like they are in a "real" page.
      *
      * This is needed to make previewing content with block fields fully working.
-     * See:
+     * See: https://github.com/bolt/bolt/issues/7753
      */
     private function fixHTMLFieldsForPreview($record)
     {
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -229,6 +229,8 @@ class Frontend extends ConfigurableBase
 
         $this->fixBlockFieldsForPreview($content);
 
+        $this->fixHTMLFieldsForPreview($content);
+
         $liveEditor = $request->get('_live-editor-preview');
         if (!empty($liveEditor)) {
             $jsFile = (new JavaScript('js/ckeditor/ckeditor.js', 'bolt'))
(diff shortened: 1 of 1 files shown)


src/Controller/Frontend.php
```diff
@@ -279,14 +281,39 @@ private function fixBlockFieldsForPreview($record)
...
 
         return $record;
     }
 
+    /**
+     * Helper function to make sure HTML-like fields are Twig_Markup, and image fields are
+     * an image. Like they are in a "real" page.
+     *
+     * This is needed to make previewing content with block fields fully working.
+     * See:
```
xiaohutai: See what? https://github.com/bolt/bolt/issues/7753?

bobdenotter: Thanks, fixed!

bobdenotter:
Green lights!

xiaohutai (approved):

virtadpt:
Thank you for fixing this!

PR #7759, opened 2019-03-08
Excerpt strip fields.

### LINKED ISSUE #7758: [RFC] Exerpt runtime strip keys
Hello,

`Bolt\Helpers\Excerpt` is great but it could be improved by adding `$stripKeys` as a fourth argument. This way, one could easily remove unwanted fields from a search result.

At the moment, the `$stripKeys` array is hardcoded [here](https://github.com/bolt/bolt/blob/3.6/src/Helpers/Excerpt.php#L58), a simple `array_merge` would do the job. An optional fourth argument wouldn't break any BC.

Keen on PR if this is approved 🙂

artggd:
Fixes #7758

Allow to strip unwanted fields out of the excerpt by adding a fourth argument to Excerpt and related runtime methods.

Doc update on its way.

### DIFF
--- a/src/Helpers/Excerpt.php
+++ b/src/Helpers/Excerpt.php
@@ -34,11 +34,11 @@ class Excerpt
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
-     * @param array|null        $stripFields
+     * @param array             $stripFields
      *
      * @return string|null
      */
-    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, $stripFields = null)
+    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, $stripFields = [])
     {
         $title = null;
         if ($includeTitle && $this->title !== null) {
@@ -52,10 +52,6 @@ class Excerpt
 
         if ($this->body instanceof LegacyContent) {
             $this->body = $this->body->getValues();
-        }
-
-        if (null === $stripFields) {
-            $stripFields = [];
         }
 
         if (!is_array($stripFields)) {
--- a/src/Helpers/Excerpt.php
+++ b/src/Helpers/Excerpt.php
@@ -34,11 +34,11 @@ class Excerpt
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
-     * @param array             $stripFields
+     * @param array|null        $stripFields
      *
      * @return string|null
      */
-    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, array $stripFields = [])
+    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, $stripFields = null)
     {
         $title = null;
         if ($includeTitle && $this->title !== null) {
@@ -52,6 +52,11 @@ class Excerpt
 
         if ($this->body instanceof LegacyContent) {
             $this->body = $this->body->getValues();
+        }
+
+        if (null !== $stripFields && !is_array($stripFields)) {
+            trigger_error(sprintf('Wrong type for "stripField" parameter. Expected array, got %s. Ignoring "stripField".', gettype($stripFields)), E_USER_DEPRECATED);
+            $stripFields = [];
         }
 
         if (is_array($this->body)) {
--- a/src/Helpers/Excerpt.php
+++ b/src/Helpers/Excerpt.php
@@ -54,7 +54,11 @@ class Excerpt
             $this->body = $this->body->getValues();
         }
 
-        if (null === $stripFields || !is_array($stripFields)) {
+        if (null === $stripFields) {
+            $stripFields = [];
+        }
+
+        if (!is_array($stripFields)) {
             trigger_error(sprintf('Wrong type for "stripField" parameter. Expected array, got %s. Ignoring "stripField".', gettype($stripFields)), E_USER_DEPRECATED);
             $stripFields = [];
         }
--- a/src/Helpers/Excerpt.php
+++ b/src/Helpers/Excerpt.php
@@ -54,7 +54,7 @@ class Excerpt
             $this->body = $this->body->getValues();
         }
 
-        if (null !== $stripFields && !is_array($stripFields)) {
+        if (null === $stripFields || !is_array($stripFields)) {
             trigger_error(sprintf('Wrong type for "stripField" parameter. Expected array, got %s. Ignoring "stripField".', gettype($stripFields)), E_USER_DEPRECATED);
             $stripFields = [];
         }
--- a/src/Helpers/Excerpt.php
+++ b/src/Helpers/Excerpt.php
@@ -34,10 +34,11 @@ class Excerpt
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
+     * @param array             $stripFields
      *
      * @return string|null
      */
-    public function getExcerpt($length = 200, $includeTitle = false, $focus = null)
+    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, array $stripFields = [])
     {
         $title = null;
         if ($includeTitle && $this->title !== null) {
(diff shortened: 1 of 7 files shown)


JarJak:
Could you add a test for it?

artggd:
Sure thing, fixing checks first :)

By the way, the CI fails because of a bad call of `content|excerpt` in `bolt/_base/_listing.twig` where `content.excerpt` should've been called.

`content.excerpt` and `content|excerpt` don't take the same arguments (`.excerpt` has `$includeTitle` as second argument). This didn't break in the past because the last argument was simply ignored. Now that I added another one, type hinting breaks.

Although this PR doesn't break BC, I'm afraid it could be dangerous in the end...

bobdenotter (dismissed):
> content.excerpt and content|excerpt don't take the same arguments (.excerpt has $includeTitle as second argument). 

This is because the `.excerpt` works on a `content` object only, but `|excerpt` is a generic filter, that works on strings, arrays or content-like objects. 

> Although this PR doesn't break BC, I'm afraid it could be dangerous in the end...

It seems to work fine, and I don't spot immediate issues.. 👍 

Are you worried about merging it in?

artggd:
Well, I understand the difference between `.excerpt` and `|excerpt` but it's easily confusable. Taking for proof there was a miscall in Bolt's core.

I reckon there might be quite an amount of projects miscalling `|excerpt` and passing the same args as `.excerpt`. Those will break on their next `composer up`.

Though I really need this feature, I'd say the merge could wait for the next major release. Unless you have any way of measuring the risk and/or are willing to accept it.

Your call ;)

artggd:
Or, maybe we could remove the type hint of `$stripFields` so that if a fourth arg is wrongly passed to `.excerpt` (miscall of `$focus` (string)) it won't break. But I'm not fond of this option, hacky...

JarJak:
Actually $stripFields param could be added only to Content method, because it is valid only with Content context.
I would leave excerpt filter as-is, then nothing will break for anyone :)

Docs needs an update, can you provide separate PR for them @artggd ?

artggd:
The filter is very versatile and is valid for `string`, `Content` or `array`. It still makes sense to have `$stripFields` in both filter and Content method IMO. 
The issue would remain anyway, it breaks when Content method is called with the filter params, not the other way round. Which is probably the most common case.

I'm thinking removing type hinting could actually be an ok solution with proper type control in the filter. It would also add some flexibility, allowing for usage of the filter like so 
```TWIG
{{ stuff|excerpt(stripFields="strip out those fields") }}
```
 Just need to explode `$stripFields` when it's a string.

If you guys are happy with this option, I'll make the update and add a couple of tests.

@JarJak I did provide a PR for doc update 😉👉 https://github.com/bolt/docs/pull/953

JarJak:
>The filter is very versatile and is valid for string, Content or array. 

Yes, you can use filter on content, but this is more like extra feature than normal usage. And since $stripFields makes sense only in Content context, I would leave this parameter only in Content method.

With that being said, IMO $stripKeys should also be moved into Content method, but that would introduce BC break.

However I won't block this PR because of my opinion :)
@bobdenotter what do you think?

JarJak (dismissed):

JarJak:
Oh, and I'm not happy with
```
{{ stuff|excerpt(stripFields="strip out those fields") }}
```
As an option. It doesn't make sense to me to allow such way of using this feature.

artggd:
Agreed, if we want to restrict $stripFields on Content only, then let's move out all the related logic to Content and leave the filter to only work with strings.  It'd be way more consistent.

What about merging this as a transitional step and keep strict typing for next major?

artggd:
> Oh, and I'm not happy with
> 
> ```
> {{ stuff|excerpt(stripFields="strip out those fields") }}
> ```
> As an option. It doesn't make sense to me to allow such way of using this feature.

Then $stripFields could simply be ignored if string and raise a warning (rather than a 500)

JarJak:
@artggd in next major (Bolt 4) we made a slightly different approach. Fields that are allowed in excerpt must implement Excerptable interface. Look for usages of https://github.com/bolt/four/blob/master/src/Entity/Field/Excerptable.php 
And as always, PRs are more than welcome!

artggd:
@JarJak This sounds very nice 🙂
Although I can see field values are stored in a json column, I don't know how you're going to achieve search without pulling out heavy indexing!
Digressing, this is not the place to discuss this but I'd be more than happy to help. I've had similar issue in the past and solved it with a different approach but I guess Bolt 4's architecture is already thought and written down.

JarJak:
@artggd we are open for all ideas and we can change everything before beta :)

artggd:
@JarJak I'll catch-up with you on https://github.com/bolt/four/issues/306 when I get some time :)

Short recap of this issue, we've got 4 options left:
1. Merging as is.
1. Removing `$stripFields` type hint and raise a warning if not an array.
1. Allowing `$stripFields` on Content method only and move the Content related logic out of the filter (inducing a BC break).
1. Allowing `$stripFields` on Content method only and keep Content related logic in the filter.

My personal fav being number 2

JarJak:
I'm ok with either 2 or 4. @bobdenotter ?

bobdenotter:
Nr. 2 has my preference! 👍

artggd:
Sweet then, working on it 👍

artggd:
Hey so, I'm still working but I've ran into a small coding issue.

I decided to log a message rather than '@trigger_error' because trigger_error doesn't show up in the debug toolbar unless it's a deprecated call. Also, it's a pain to unit test.

Problem is, `Excerpt` is not a service so I have to duplicate the same bit of code in the callers (which are services and can access the logger). The said bit of code tests the type of $stripFields and logs an error.

I'm really not happy with that.

Any recommendation? How'd you achieve that in a cleaner way?

JarJak:
Normally I would just throw an Exception there.

artggd:
I would too, but that was the point Nr. 2. Not throwing an exception so it doesn't break in case of a miscall 🙂

JarJak:
Then I think we can deprecate it.

Decorating every excerpt call with logger is not an option because there can always be n+1 call in extension or user's code.

src/Helpers/Excerpt.php
```diff
@@ -53,9 +54,16 @@ public function getExcerpt($length = 200, $includeTitle = false, $focus = null)
             $this->body = $this->body->getValues();
         }
 
+        if (null === $stripFields || !is_array($stripFields)) {
```
JarJak: default value triggers error?

JarJak (changes requested):

src/Helpers/Excerpt.php
artggd: I realize that this morning, update coming up now

artggd:
It should be ready to merge, I also added 3 tests on ContentTests.

src/Helpers/Excerpt.php
```diff
@@ -34,10 +34,11 @@ public function __construct($body, $title = null)
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
+     * @param array|null        $stripFields
      *
      * @return string|null
      */
-    public function getExcerpt($length = 200, $includeTitle = false, $focus = null)
+    public function getExcerpt($length = 200, $includeTitle = false, $focus = null, $stripFields = null)
```
JarJak: just set default value of $sripFields to [] here

JarJak (changes requested):

src/Helpers/Excerpt.php
```diff
@@ -34,10 +34,11 @@ public function __construct($body, $title = null)
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
+     * @param array|null        $stripFields
```
JarJak: remove null

JarJak (changes requested):

artggd:
Not sure what's wrong with the CI, the tests run smooth on my machine and it seems to fail on nothing related to my last dev.

JarJak (approved):

bobdenotter:
Tests are failing, a lot of: 

```
Error: Call to undefined method Twig\TemplateWrapper::getTemplateName()
```

Looks like this is introduced in this update: https://symfony.com/blog/twig-sandbox-information-disclosure

bobdenotter:
The issue with Twig is fixed. See: https://github.com/twigphp/Twig/issues/2892

artggd:
Great! Thanks @bobdenotter

bobdenotter:
@artggd Thank _you_ for the PR! 👍

PR #7762, opened 2019-03-22
[3.6] Remove unsupported guzzlehttp version

StevendeVries:
If you have version 5.* installed of guzzlehttp/guzzle Bolt will break due to vendor/bolt/bolt/src/Composer/Satis/PingService.php expecting GuzzleHttp\RequestOptions to exist which only is available in version 6.* of guzzlehttp/guzzle

This PR will remove the option to use 5.* and force you to use 6.* preventing the error.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "ext-xml": "*",
         "ext-zip": "*",
         "filp/whoops": "^2.0",
-        "guzzlehttp/guzzle": "^5.3.1 || ^6.2.1",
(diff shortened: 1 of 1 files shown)


bobdenotter (commented):
Thanks, @StevendeVries. 👍 

I had to check if Guzzle 6 supported PHP 5.5.9 (which is the minimum supported version for Bolt 3), but all seems well.

bobdenotter (approved):

PR #7763, opened 2019-03-26
[3.6] Make the bootstrap more robust

StevendeVries:
When using an old version of phpunit like `4.8.28` it does not provide a class called `PHPUnit\Framework\TestCase` required by `bolt/bolt/tests/phpunit/unit/BoltUnitTest.php`
The class is available in version 5.7. The dependancy with `phpunit/php-code-coverage` has also been updated.

Next to the phpunit version there is an issue when running the unittests, it can't seem to find the `vendor/autoload.php`. I've made the bootstrap a bit more defensive.
For context: We have the bootstrap located in: `vendor/bolt/bolt/tests/phpunit/bootstrap.php`

### DIFF
--- a/tests/phpunit/bootstrap.php
+++ b/tests/phpunit/bootstrap.php
@@ -2,13 +2,19 @@
 
 if (file_exists(__DIR__ . '/../../vendor/autoload.php')) {
     require_once __DIR__ . '/../../vendor/autoload.php';
-} else {
+} elseif (file_exists(__DIR__ . '/../../../vendor/autoload.php')) {
+    require_once __DIR__ . '/../../../vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../../vendor/autoload.php')) {
     require_once __DIR__ . '/../../../../vendor/autoload.php';
+} elseif (file_exists(__DIR__ . '/../../../../../vendor/autoload.php')) {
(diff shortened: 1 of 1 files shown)


StevendeVries:
This currently breaks being compatible with PHP 5.5.9, however the unittests didn't work properly with PHP 5.5.9 anyway due to the lower phpunit version.

bobdenotter:
Do you think you could get this fixed for PHP 5.5.9? For Bolt 3, we'd need to keep compatible with that version, i'm afraid.

StevendeVries:
@bobdenotter if you really want to keep supporting 5.5.9 I would suggest reverting this PR and rewrite `BoltUnitTest.php` to remove the dependancy on `PHPUnit\Framework\TestCase`

bobdenotter:
@StevendeVries We use SemVer, so we cannot drop PHP 5.5.9 support for Bolt 3. Only in a major release can we bump the minimum requirements. 

PHPUnit 5.7 should work on PHP 5.5.9 too, right?

StevendeVries:
Like stated in the ci: `phpunit/phpunit 5.7.0 requires php ^5.6` 😢

StevendeVries:
I'll remove the phpunit version increase to keep the bootstrap changes. Making it possible for you to merge.

bobdenotter:
Thanks, mate! 👍

bobdenotter (approved):

Issue #7764, opened 2019-03-26
Using `class: Foo\Bar` doesn't work in 3.6


bobdenotter:
You can define a `class` to use in contenttypes.yml, to override the Class used for a ContentType. Using `class: \Bundle\Site\MyContent`, you'd have:

```php
<?php

namespace Bundle\Site;

class MyContent extends \Bolt\Legacy\Content
{

}
```

And that should work. However, this is broken in 3.6: 

![Schermafbeelding 2019-03-26 om 14 32 58](https://user-images.githubusercontent.com/1833361/55001046-12e26a00-4fd4-11e9-937b-636b0437b2e5.png)

This is because `src/Legacy/Content.php` expects a `$app` when contructed:

```php
    public function __construct(Silex\Application $app, $contenttype = '', $values = [], $isRootType = true)
    {
        $this->app = $app;
        $this->isRootType = $isRootType;

```

While the builder in `src/Storage/Entity/Builder.php` doesn't pass it in: 

```php
    public function getEntity($entity = null)
    {
        if ($entity === null) {
            $class = $this->class;
            $entity = new $class();
        }
```

Either we change `Content.php` to magically conjure an `$app` if none is given, or we change `getEntity` to pass in the `$app`. Neither is very elegant, but we should fix it for backwards compatibility.

StevendeVries:
On the frontend I resolved the issue by using a custom Builder which adds `$app` to the exact class (using a classname comparison, nothing fancy but it works).
However the Backend (`/bolt`) still fails due to the fact the `\Bolt\Storage\EntityManager::getEntityBuilder` creates a new Instance instead of using the one provided from $app
![Screenshot 2019-04-03 at 11 39 20](https://user-images.githubusercontent.com/6906996/55469154-28274c00-5605-11e9-88c3-0efe6c91987d.png)

JarJak:
Shouldn't there be separate builder for legacy entities?

PR #7767, opened 2019-04-03
[3.6] WIP Use the EntityBuilder provided by  instead a new Instance

StevendeVries:
As mentioned in https://github.com/bolt/bolt/issues/7764 it's currently not possible to use a custom Entity/Builder for the Backend since it creates a new Instance. With this PR we are use the Builder provided in the StorageServiceProvider (see `src/Provider/StorageServiceProvider.php`)

### DIFF
--- a/src/Storage/EntityManager.php
+++ b/src/Storage/EntityManager.php
@@ -118,7 +118,7 @@ class EntityManager implements EntityManagerInterface
      */
     public function getEntityBuilder($className = null, ClassMetadata $classMetadata = null)
     {
-        $builder = new Entity\Builder($this->getMapper(), $this->getFieldManager());
+        $builder = $this->createEntityBuilder();
 
         if ($className !== null) {
             $builder->setClass($className);
@@ -129,6 +129,18 @@ class EntityManager implements EntityManagerInterface
         }
 
         return $builder;
+    }
+
+    /**
+     * @return Entity\Builder
+     */
+    private function createEntityBuilder()
+    {
+        if (! $this->builder) {
+            return new Entity\Builder($this->getMapper(), $this->getFieldManager());
+        }
+        $builderClass = get_class($this->builder);
+        return new $builderClass($this->getMapper(), $this->getFieldManager());
     }
 
     /**


bobdenotter:
It seems tests fail, if `$this->builder` isn't set yet. Maybe change it to: 

```php
    public function getEntityBuilder($className = null, ClassMetadata $classMetadata = null)
    {
        if (! $this->builder) {
            $this->builder = new Entity\Builder($this->getMapper(), $this->getFieldManager());
        }

        if ($className !== null) {
            $this->builder->setClass($className);
        }

        if ($classMetadata !== null) {
            $this->builder->setClassMetadata($classMetadata);
        }

        return $this->builder;
    }
```

StevendeVries:
Is there a reason for the storage.entity_builder not to be in the construct? Same goes for the field_manager and the collection_manager.

Making new Instances makes code less testable, I would rather inject the dependancy.

bobdenotter:
> Is there a reason for the storage.entity_builder not to be in the construct? 

I'm not 100% sure, but I expect the answer to that is "somewhere between organically grown legacy code" and "not entirely finished refactoring" ;-)

> Making new Instances makes code less testable, I would rather inject the dependancy.

The intent is to learn from this for v4, and do it better there 😇

StevendeVries:
Don't merge just yet, I want it to support more parameters if required.

StevendeVries:
I had to overwrite Storage to enable custom EntityBuilder with different parameters (a class reset might be a solution but seems inefficient to spend much time on legacy)

bobdenotter:
Shouldn't we try to get this in "as-is"? As you mention, it could do _more_, but it seems this is _better_ than it was, right?

StevendeVries:
If you believe it adds enough value as is, feel free to reopen and merge. For me I've used a different approach to solve my issue (only use a custom contenttype class for the frontend)

rossriley:
Looks like a good idea, I'll try and give it a try, just thinking additionally would there be any value to doing this approach.

```
$new = clone $this->builder;
$new->class = $className;

return $new;
```

Advantage there is you get an exact copy of the object that is built in the container only with the class name changed. It means in the future if we add more modifications to the container built version it will always reflect in these created objects.

bobdenotter:
Reopening.. 

Let me know your findings, @rossriley. You know this part of the code best. :-)

JarJak:
>Is there a reason for the storage.entity_builder not to be in the construct?

In current implementation, you call setters on Builder changing it's internal state. That's why it can't be injected from service container.

What I could propose is to move setting those values to builder's constructor, or if constructing it is heavy, add to it static method `forClass($className, $classMetadata)` which will do cloning and setting proper values.

PR #7768, opened 2019-04-03
Make sure `accept_file_types` is an INDEXED array, and disallow certain filetypes to be whitelisted

### LINKED ISSUE #7765: Post-3.6.6 upgrade: js error in /bolt/files
Flawlessly the site was upgraded from 3.4 to 3.6.6 recently. Ever since the backend shows a JS error in the files management section and doesn't allow uploading of files. When working on an entry of a content type, files can be uploaded properly - and they'll be put in YYYY-MM directory as defined in the default behavior. Enabling the debug log sadly didn't give me any clue. All that is logged when reloading the page is that the session cookie is renewed (`Generating authentication cookie — User ID: '1' Salt...`).

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6.6
| Install type                   | Composer install 
| BC Break                     | what is that?
| PHP version                 | 7.0.33-0ubuntu0.16.04.3
| Web server                  | Apache, version 2.4.18-2ubuntu3.8
| For UX/UI issues        | Chrome, MacOS, v. 73.0.3683.86

AppCFG:
```yaml
database:
    driver: mysql
    databasename: xxx
    username: xxx
    password: xxx
sitename: A sample site
theme: xxx
locale: de_DE
timezone: Europe/Berlin
maintenance_mode: false
maintenance_template: maintenance_default.twig
cron_hour: 3
favicon: ./favicon.ico
homepage: /
homepage_template: index.twig
add_jquery: false
recordsperpage: 10
caching:
    config: true
    templates: true
    request: false
    duration: 120
    authenticated: false
    thumbnails: true
    translations: true
changelog:
    enabled: false
thumbnails:
    default_thumbnail: [ 160, 120 ]
    default_image: [ 1000, 750 ]
    quality: 80
    cropping: crop
    notfound_image: view/img/default_notfound.png
    error_image: view/img/default_error.png
    save_files: true
    allow_upscale: false
    exif_orientation: true
    only_aliases: false
htmlcleaner:
    allowed_tags: [ div, span, p, br, hr, s, u, strong, em, i, b, li, ul, ol, mark, blockquote, pre, code, tt, h1, h2, h3, h4, h5, h6, dd, dl, dh, table, tbody, thead, tfoot, th, td, tr, a, img, address, abbr, iframe, caption ]
    allowed_attributes: [ id, class, style, name, value, href, src, alt, title, width, height, frameborder, allowfullscreen, scrolling, target, colspan, rowspan ]
accept_file_types: [ twig, gif, jpg, jpeg, png, pdf, epub, svg, json]
debug: false
debug_show_loggedoff: false
debug_permission_audit_mode: false
debug_error_level: 8181           
debug_error_use_symfony: false      
production_error_level: 8181 
debuglog:
    enabled: false
    filename: bolt-debug.log
    level: DEBUG
strict_variables: false
wysiwyg:
    images: true            
    anchor: false            
    tables: false            
    fontcolor: true          
    align: false             
    subsuper: false          
    embed: false             
    underline: false         
    ruler: false             
    strike: false            
    blockquote: false        
    codesnippet: false       
    specialchar: false       
    ck:
        autoParagraph: true  
        disableNativeSpellChecker: true 
        allowNbsp: false
google_api_key: "xxx"
liveeditor: false
mailoptions:
    transport: mail
    spool: false
cookies_use_remoteaddr: false
cookies_use_browseragent: false
cookies_use_httphost: false
cookies_lifetime: 1209600
cookies_domain: ".xxx.com"
hash_strength: 10
enforce_ssl: true
```

image

### LINKED ISSUE #7770: Can't upload files after upgrade to 3.6.6
After upgrade Bolt to 3.6.6 can't upload any file via File Management in backend.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.6.6
| Install type                   | Composer install
| For UX/UI issues        | Any browser

Reproduction
------------

Click to "Select files" button, select image file - button "Upload file" is inactive. I've tried different files with different extensions. In previous versions everything worked.
Also I've installed clear Bolt installation - the problem is actual.

Clipboard - 8 апреля 2019 г , 11-51

bobdenotter:
Fixes #7765. Fixes #7770.


bobdenotter:
Note: The important part of this change is to change the array from `associative` to `indexed`, so that later in the chain our JS knows how to `.join()` it. This broke when we introduced the uppercase/lowercase alternates.

src/Config.php
```diff
@@ -308,16 +308,22 @@ protected function parseGeneral()
...
-                    $general['accept_file_types'],
-                    array_map(function ($extension) {
-                        return strtolower($extension);
-                    }, $general['accept_file_types']),
-                    array_map(function ($extension) {
-                        return strtoupper($extension);
-                    }, $general['accept_file_types'])
+
+            // Just.. Say no to these.
+            $general['accept_file_types'] = array_diff($general['accept_file_types'], ['sh', 'asp', 'cgi', 'php', 'php3', 'ph3', 'php4', 'ph4', 'php5', 'ph5' ,'phtm', 'phtml']);
```
xiaohutai: 1. Isn't more transparent if there's only opt-in whitelist via `accept_file_types` in `config.yml`. Now there's a hard-coded blacklist in code that's not really apparent that it exists. If you really need these values I'd prefer defining these in `config.yml` too. But I don't see the point since a whitelist is most likely sufficient already.

If users do want to add `php`, isn't it really their fault/problem for opening the gates?

2. I prefer arrays declarations to be like this:
```
[
    'sh',
    'asp',
    'cgi',
    'php',
    'php3',
    'ph3',
    'php4',
    'ph4',
    'php5',
    'ph5',
    'phtm',
    'phtml',
]
```
(one per line, with trailing comma)

Otherwise, there's an space/comma inconsitency at `ph5' ,`
```diff
-'ph5' ,'phtm',
+'ph5', 'phtm',
```

xiaohutai (changes requested):
Looks good to me. At least fix that space and comma in the array declaration.

src/Config.php
```diff
@@ -308,16 +308,22 @@ protected function parseGeneral()
...
+
+            // accept uppercase and lowercase file extensions.
+            $general['accept_file_types'] = array_values(
```
JarJak: use collection (bolt has one)

JarJak (changes requested):

src/Config.php
JarJak: I agree that whitelist is better than blacklist

src/Config.php
```diff
@@ -308,16 +308,22 @@ protected function parseGeneral()
...
-                array_merge(
-                    $general['accept_file_types'],
-                    array_map(function ($extension) {
-                        return strtolower($extension);
-                    }, $general['accept_file_types']),
-                    array_map(function ($extension) {
-                        return strtoupper($extension);
-                    }, $general['accept_file_types'])
+
+            // Just.. Say no to these.
```
JarJak: Say yes to these and no to any others

JarJak (changes requested):

src/Config.php
bobdenotter: That's not feasible.. the entire idea about the `accept_file_types` is that it _is_ the whitelist of acceptible filetypes. We can impossibly list all the extensions that people _might_ want to put in their own lists. Simply not feasible. 

This actively _removes_ a few known bad actors, as a minor extra layer to guard against mishaps. Just like in our `allowed_tags` we let the user set which tags are allowed, but regardless we _remove_ `<script>` from that list. Same thing applies here.

src/Config.php
bobdenotter: @JarJak You could have done this when you introduced this bit of code in #7751. I'm fine with it the way it is now.

src/Config.php
bobdenotter: > If users do want to add php, isn't it really their fault/problem for opening the gates?

It's about escalation. Allowing this could possibly enable would-be attackers to exploit another non-critical issue, pair it with this, and gain RCE. This is just one small extra layer for a little bit of extra security.

If we want to get into semantics: This is *not* blacklisting, it is actively removing some items from the whitelist, effectively making the whitelist more strict.

I reformatted it, to look slightly more readable.

xiaohutai: I usually prefer one line per item just for adding/removing items not marking everything at the same time (also easier to `blame`). But in hindsight, that might be a bit overkill for this rather simple list.

I'll resolve this unless @JarJak finds this explode solution unacceptable 😂

src/Config.php
xiaohutai: (1)

@JarJak is correct, the resulting list is a list of file types that _are_ allowed. A better comment would be something like:

```
// Remove problematic file types
```

or maybe:

```
// Remove blacklisted file types
```

(2)

The logic is incomplete: it removes `php`, but not `PHP`. You can add `PHP` to `accept_file_types` and then upload `php` files.

bobdenotter:
I refactored the thing.. Took @xiaohutai's concerns into account so it filters correctly, and I've used a Collection.

src/Config.php
```diff
@@ -308,18 +309,22 @@ protected function parseGeneral()
...
-                    array_map(function ($extension) {
-                        return strtolower($extension);
-                    }, $general['accept_file_types']),
-                    array_map(function ($extension) {
-                        return strtoupper($extension);
-                    }, $general['accept_file_types'])
-                )
-            );
+
+            // To remove unacceptable / unwanted extensions from out list of Acceptable File Types
```
xiaohutai: `out` -> `the`

```diff
-// To remove unacceptable / unwanted extensions from out list of Acceptable File Types
+// To remove unacceptable / unwanted extensions from the list of Acceptable File Types
```

xiaohutai (dismissed):

src/Config.php
bobdenotter: Typo fixed, thanks.

src/Config.php
JarJak: (2) it can just do case-insensitive check

src/Config.php
JarJak: Yes, you are right.. I just didn't know Bolt have it's own Collection implementation when I did this :)

src/Config.php
bobdenotter: In my latest version of this PR, this is a non-issue, because it's handled correctly :-)

src/Config.php
bobdenotter: Ah, ok.. It's been taken care of now. :-)

src/Config.php
```diff
@@ -308,18 +309,22 @@ protected function parseGeneral()
...
+            $bag = $bag->map(function ($key, $ext) use ($removeFromAllowedFileTypes) {
+                if (!in_array(mb_strtolower($ext), $removeFromAllowedFileTypes)) {
+                    return mb_strtolower($ext);
+                }
+            })->clean();
+
+            // Set accept_file_types, with the previous bag _and_ uppercased alternatives.
+            $general['accept_file_types'] = $bag->merge($bag->map(function ($key, $ext) {
+                return mb_strtoupper($ext);
+            }))->toArray();
```
JarJak: However `.pHp` will still pass this firewall. `strtolower` should be done while checking extensions, not while building the list (maybe change it in separate PR?)

JarJak (changes requested):

src/Config.php
bobdenotter: it actually does that too. See line 319..

Example: 

Schermafbeelding 2019-04-05 om 13 00 51

Result: 

Schermafbeelding 2019-04-05 om 13 01 32

JarJak: So, if jpg is on this list, will `.jPG` be accepted too?

src/Config.php
```diff
@@ -308,18 +309,22 @@ protected function parseGeneral()
...
+
+            // To remove unacceptable / unwanted extensions from the list of Acceptable File Types
+            $removeFromAllowedFileTypes = explode('|', 'sh|asp|cgi|php|php3|ph3|php4|ph4|php5|ph5|phtm|phtml');
+
+            // Create a bag with lowercased extensions
+            $bag = Bag::from($general['accept_file_types']);
+            $bag = $bag->map(function ($key, $ext) use ($removeFromAllowedFileTypes) {
+                if (!in_array(mb_strtolower($ext), $removeFromAllowedFileTypes)) {
+                    return mb_strtolower($ext);
+                }
```
JarJak, line 323: else return null

JarJak (changes requested):

bobdenotter:
Ok, third attempt pushed: It now:

 - Doesn't duplicate the list, but
 - still allows upload of `test.jpg`, `test.JPG` and `test.JpG`, if `jpg` is in the "allowed" list.
 - It removes the unwanted extensions from the whitelist, disallowing upload of `test.php`, `test.PHP`, and `test.pHp`.
 - If the user (knowingly or unknowingly) modifies the `<input>` to allow `php` files, they're still scrubbed before being accepted as file. 

So, I think this should be GTG now.

xiaohutai (dismissed):
Looks good now 👍

JarJak:
>It removes the unwanted extensions from the whitelist, disallowing upload of test.php, test.PHP, and test.pHp.

Why is that? PHP files need to be explicitly allowed anyway, so why do we disallow a developer to allow them?

Also this change is not associated with any of mentioned issues, nor this PR's title. nor it's a security fix.

bobdenotter:
> Why is that? PHP files need to be explicitly allowed anyway, so why do we disallow a developer to allow them?

Because of a security issue that was responsibly disclosed to me. 

> Also this change is not associated with any of mentioned issues, nor this PR's title. 

Updated the title. 

> nor it's a security fix.

It might not seem like it, but it is indeed closing off a possible venue for escalation into RCE.

JarJak:
Anyway it must be mentioned in the docs.

src/Config.php
```diff
@@ -308,18 +309,22 @@ protected function parseGeneral()
...
+            $bag = Bag::from($general['accept_file_types']);
+            $bag = $bag->map(function ($key, $ext) use ($removeFromAllowedFileTypes) {
+                if (!in_array(mb_strtolower($ext), $removeFromAllowedFileTypes)) {
+                    return mb_strtolower($ext);
+                } else {
+                    return null;
+                }
+            })->clean();
+
+            // Set accept_file_types, with the previous bag _and_ uppercased alternatives.
```
JarJak: Outdated comment

JarJak (changes requested):

src/Config.php
bobdenotter: fixed

bobdenotter:
The `accept_file_types` wasn't in the proper docs, but i've added a description to `config.yml`.

app/config/config.yml.dist
```diff
@@ -234,7 +234,8 @@ htmlcleaner:
 #     autoconfirm: false
 
 # Define the file types (extensions to be exact) that are acceptable for upload
-# in either 'file' fields or through the 'files' screen.
+# in either 'file' fields or through the 'files' screen. Note that certain file-
+# types like `.php` are never acceptable, even if they are in this list.
```
JarJak: Specify exactly which are not accepted, "like php" is not enough.

JarJak (changes requested):

app/config/config.yml.dist
bobdenotter: Updated. :-)

JarJak (approved):

PR #7769, opened 2019-04-03
add information about how to set up custom ck styles

dadaxr:
By default the ckeditor build shipped with bolt support custom styles definition 
( ie the build contains the "styles combo" add, as we can see here : https://ckeditor.com/cke4/builder/9705ed1f5e13d331e24bd0e950b94e0e, this link is the one mentioned in bolt source code : src/vendor/bolt/bolt/app/src/lib/ckeditor/build-config.js ) 

But the default config.yml file does not mentioned it, and it's easy to think we have to create a bolt extension to add the required ckeditor addons (as it has already be done by @SahAssar  : https://discuss.bolt.cm/d/442-adding-ckeditor-plugins/9 )

This PR just add 2 config line to show it's quite easy to implement, and it could have save me few hours of investigation ;)


bobdenotter:
@dadaxr Thanks! :+1:

I took the liberty of adding in another line-break, and i'll merge it in!

bobdenotter (approved):

PR #7776, opened 2019-05-23
Make sure stripFields is set to array by default

StevendeVries:
Details
-------
Is this commit https://github.com/bolt/bolt/pull/7759/commits/70b3f2114b829642d4dca09af5d5661891ac9fad the default for stripFields is set to an array instead of `null`, without changing the calling methods to reflect the same change resulting in `trigger_error` foreach call without parameters which in turn can cause a 500 error if to many errors are triggered.

In the nginx/errorlog you would see:
`2019/05/23 12:26:35 [error] 1368#1368: *76 upstream sent too big header while reading response header from upstream,`

This PR make sure the calling functions are updated to reflect the new default value for stripFields.

### DIFF
--- a/src/Legacy/Content.php
+++ b/src/Legacy/Content.php
@@ -213,11 +213,11 @@ class Content implements \ArrayAccess
      * @param int               $length
      * @param bool              $includeTitle
      * @param array|string|null $focus
-     * @param array|null        $stripFields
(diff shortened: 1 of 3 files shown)


bobdenotter (approved):
Thanks, @StevendeVries

PR #7799, opened 2019-08-09
Fix taxonomy name collision

marcingajda:
Fix naming collisions for tags-like taxonomies with a certain name.

Details
-------

Currently fetching values for the tags field in content editor uses a `get` method on the `Content` class. This method takes the taxonomy name - for example `Foo` - and uses it to call a method named `getFoo`. If this method doesn't exists in object of `Content` then the magic method `__call` will find you trying to reach a taxonomy and it will return it for you. 

The problem is when you taxonomy name collides with an existing name in `Content` class. For example if your tags are named `related` then calling `getRelated` will invoke an existing method of `Content` class and will return something you don't want.

The solution I propose is to call a method of a taxonomies collection directly so we will always get the items we want. This fix is required only for tags. Groups and categories don't have this bug.


rossriley:
:+1: Looks good to me

bobdenotter (approved):
Looks good to me, too! 👍

PR #7805, opened 2019-08-20
Fixing lame ternary operator error in ImageRuntime.php

### LINKED ISSUE #7804: Thumbnail error - no "filename" defined
Problem appears when rendering twig with aliased thumbnail resolution:
`thumbnail(img, 'full')`

image

Boorj:
fixes #7804

### DIFF
--- a/src/Twig/Runtime/ImageRuntime.php
+++ b/src/Twig/Runtime/ImageRuntime.php
@@ -346,8 +346,11 @@ class ImageRuntime
             return $fileName;
         }
 
-        return isset($fileName['filename']) ? $fileName['filename']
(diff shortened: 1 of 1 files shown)


Boorj:
damn, that akward moment....

image

bobdenotter:
It's still using a nested ternary selector, which are _very_ hard to read due to them due to having weird associativity. 

Is it supposed to do this? 

```
if (isset($fileName['filename']) {
    return $fileName['filename'];
} 
if (isset($fileName['file']) {
    return $fileName['file'];
} 
return '';
```

If so, please change it to that. separate returns are much easier to read.

PR #7809, opened 2019-09-12
fix typo in AccessControlListener.php

Boorj:
Was it a typo or intentionally?

`https://github.com/bolt/bolt/blob/3.6/src/EventListener/AccessControlListener.php#L130`

See double PRE_DELETE:
```php
    public static function getSubscribedEvents()
    {
        return [
            StorageEvents::PRE_DELETE => 'onStorageEventPostSave',
            StorageEvents::PRE_DELETE => 'onStorageEventPreDelete',
        ];
    }
```


bobdenotter (approved):
Pretty sure that was a typo. Thanks!

PR #7812, opened 2019-10-18
[3.6] Export improvements.

doenietzomoeilijk:
- Bomb on slugless entities (since importing them causes issues)
- Properly handle relations (the storage part)


bobdenotter (approved):
Looks good, so far! Let's merge and onwards to the next quirk we encounter! 🚀

PR #7822, opened 2019-11-04
Fix tags in non-english languages and firefox caching selected options

### LINKED ISSUE #7659: Duplicate tags in tag field
When a post is saved with a tag that contains special characters, the 
tag selector upon subsequent page loads contains two different items for the same tag.

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.5
| Install type                   | Composer install
| For UX/UI issues        | Chrome 67

Reproduction
------------

### Bug summary

The problem occurs when an editor adds a tag to post, and that tag contains special characters – for instance, the word "bölt" (which isn't really a word). If they then reload the browser and enter the tag input field, they will see two separate copies of the same tag in the dropdown. It is possible to add them both to the input at the same time. When inspecting the options in the markup, I've found that the first copy has the tag slug as `value`, whereas the other one has the tag name as `value`.

### Specifics

  * URL: /bolt/editcontent/articles/XYZ#taxonomy
  * What version of Bolt are you using (down to the very last digit!): 3.5.4
  * What method did you use to install Bolt: composer
  * What browser and version you are using: Chrome 67
  * Please mention if you've checked it in other browsers as well: no

### Steps to reproduce

  1. Create a tagging taxonomy and a content type with a tag field.
  2. Create and save a post with the tag "bölt".
  3. Reload the page.
  4. Enter the tag input field and type "b".

### Expected result

There should be one row in the autocomplete list containing "bölt".

### Actual result

The "bölt" tag appears twice in the autocomplete list.

More observations
------------------

* When the bug appears, the hidden `<select>` contains two options for the same tag. Both use the tag name as the option label/content, but the first one has the tag slug as its `value`, while the other one has the tag name as its `value`.
* When the code in `fieldTags.js` loads tags from the backend, it only gets names, not slugs.
* When the code compares the existing tags in the `<select>` (which are prerendered server-side) with the tags that are fetched from the backend by JS, it uses the `value` attribute, which, again, contains slug values for existing tags.
* A similar problem seems to exist with the code that handles clicks in the tag cloud.
* Duplicates do not seem to stick in the database.

https://github.com/bolt/bolt/blob/204d8bd3c0bc8099569e5f6052aec851b7570b12/app/src/js/widgets/fields/fieldTags.js#L55-L60

image

SvanteRichter:
This fixes two issues:

1. Bolt saw the slug and the name of a tag as always equivalent, this makes them distinct
2. Firefox has a "feature" where it ignores `selected` if a different option has been selected before reload and the reload is not "hard". Since select2 takes over the actual input it didn't see the select as changed, so adding a tag and reloading would make it seem like the tag wasn't added, and then a subsequent save would remove the tag.

Fixes: #7659

### DIFF
--- a/app/src/js/widgets/fields/fieldSelect.js
+++ b/app/src/js/widgets/fields/fieldSelect.js
@@ -32,6 +32,11 @@
          */
         _create: function () {
             var select = this.element.find('select');
+            // Regigger firefoxes form cache kagigger
+            // See https://stackoverflow.com/questions/1479233/why-doesnt-firefox-show-the-correct-default-select-option
+            var selected = select.find('option[selected]');
+            select.find('option').removeAttr('selected');
+            selected.prop('selected', 'selected');
             var options = {
                 width: '100%',
                 placeholder: {
--- a/app/src/js/widgets/fields/fieldTags.js
+++ b/app/src/js/widgets/fields/fieldTags.js
@@ -35,11 +35,17 @@
                 tagcloud = this.element.find('.tagcloud'),
                 separators = [','],
                 tags = {};
+            // Regigger firefoxes form cache kagigger
+            // See https://stackoverflow.com/questions/1479233/why-doesnt-firefox-show-the-correct-default-select-option
+            var selected = taxonomy.find('option[selected]');
+            taxonomy.find('option').removeAttr('selected');
+            selected.prop('selected', 'selected');
 
             // Initialize the tag selector.
             if (!this.options.allowSpaces) {
                 separators.push(' ');
             }
+
             taxonomy.select2({
                 width: '100%',
                 tags: tags,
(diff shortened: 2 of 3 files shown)


bobdenotter (dismissed):
> adding a tag and reloading would make it seem like the tag wasn't added, and then a subsequent save would remove the tag

I've always seen this as a "unfortunate fact of life", but i'm happy to have a fix for it! 👍

bobdenotter:
There's one failure, on PHP 7.2 only, that says: 

Screenshot 2019-11-04 at 21 20 32

It's peculiar though, because I don't readily see how this change could cause that as a side effect.. 🤔

bobdenotter:
Still the same weird failure on 7.2 only.. I guess you don't see how this could be caused by your changes either, right?

SvanteRichter:
Nope, no idea... I'll try too find out.

SvanteRichter:
I can't find why my changes would lead to that failure...

bobdenotter (approved):

PR #7868, opened 2020-07-28
Fix ContextErrorException in PHP 7.4

### LINKED ISSUE #7864: [3.7] Bolt CMS 3 Incompatible with PHP 7.4 - ContextErrorException in ExceptionListener.php line 285
A brief description of the issue goes here.

Details
-------
A recent upgrade of Bolt from 3.6 to 3.7 required me to update my server's version of PHP from 7.0 to meet the requirements of all the Composer packages. I decided to go straight to PHP 7.4. After installing PHP 7.4 and running composer update, I got the following error on my homepage:

Screenshot 2020-07-24 at 11 26 06

Thanks to NGINX allowing you to easily specify a PHP version per application, I was able to try PHP 7.3, which then allowed the application to work without this exception occurring.

However, as everyone will be moving to PHP 7.4, I hope it's possible to make Bolt 3 work with PHP 7.4

|----------------- | ------
| Relevant Bolt Version | 3.7.1
| BC Break                     | yes
| PHP version                 | 7.4
| Web server                  | Nginx
| For UX/UI issues        | N/A

Reproduction
------------

### Bug summary

Bolt CMS version 3.7 throws a fatal exception when running under PHP 7.4

### Steps to reproduce

* Modify nginx to use php7.4-fpm instead of php7.3-fpm - error occurs
* Modify nginx to use php7.3-fpm - error does not occur

xiaohutai:
Fixes #7864

See: https://www.php.net/manual/en/migration74.other-changes.php#migration74.other-changes.ini

A `ContextErrorException` is thrown when `zend.exception_ignore_args` is set to `On`. The `args` key is not available.

A simple test-case:
```
ini_set('zend.exception_ignore_args', 'On');
trigger_error('test', E_USER_NOTICE );
```

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -282,7 +282,11 @@ class ExceptionListener implements EventSubscriberInterface
 
         $trace = $exception->getTrace();
         foreach ($trace as $key => $value) {
-            $trace[$key]['args_safe'] = $this->getSafeArguments($trace[$key]['args']);
+            // See: https://www.php.net/manual/en/migration74.other-changes.php#migration74.other-changes.ini
+            $trace[$key]['args_safe'] = [];
+            if (isset($trace[$key]['args'])) {
+                $trace[$key]['args_safe'] = $this->getSafeArguments($trace[$key]['args']);
+            }
 
             // Don't display the full path, trim 64-char hexadecimal file names.
             if (isset($trace[$key]['file'])) {


xiaohutai:
https://www.php.net/manual/en/ini.core.php#ini.zend.exception-ignore-args does say it's `Off` by default, so not entirely sure what's causing this behaviour.

bobdenotter (approved):
Good find!

PR #7871, opened 2020-08-03
Fixes slashes in directory names

xiaohutai:
- The 1st line allows slashes in parent's names.
- The 2nd line allows users to create "foo/bar/baz" in one go.


bobdenotter (approved):
👍

PR #7885, opened 2020-10-26
allow rel attribute by default

scaphandroid:
Allow rel attribute when sanitize HTML

As the ckeditor link options have 'rel' tags in there 'advanced' tab it would be a good idea to allow rel tag by default/ Because it's a bit confusing for the user to be allowed to add the tag but to not store it in the database.

(see discussion here --> https://boltcms.slack.com/archives/C094GL7ME/p1603187487310500)


bobdenotter (approved):
Thanks again, @scaphandroid 👍

PR #7894, opened 2021-01-10
FIXED: #7893 Fetching content by Taxonomy does not work if ContentType Name is different than ContentType Slug 

### LINKED ISSUE #7893: Fecthing content by Taxonomy does not work if ContentType Name is different than ContentType Slug
Taxonomy based content fetching does not work if **ContentType Slug** is different that **ContentType Name**.  my Bolt Version is 3.6 but bug exist in 3.7 too.

**To produce the bug:**

1. create a ContentType e.g. "articles" with slug "myarticles" 
2. Create a new record in "articles" and change a taxonomy category: "books" 
3. Now try fetching in a twig 
`{% setcontent articlelist = 'articles' where { categories:  'books' } %}`

**Result**
The twig fetch query will return 0 records

**Expected result**
It should have returned one record that we just created.

hrjt:
Issue #7893 Bug fixed

Fixes: #7893 
Fetching content by Taxonomy does not work if ContentType Name is different than ContentType Slug 

Details
-------
Reason for Bug: decodeContentQuery function builds query using ContentType Slugh whereas taxonomy mapping table uses ContentType Name to map. The query mismatch results 0 rows. After changing the query to lookup mapping by ContentType Name, query results are returned as expected.
   
Target Branch
------------------------

 * `release/3.7` — "stable" branch

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -1370,7 +1370,7 @@ class Storage
                             $this->parseWhereParameter($this->getTablename('taxonomy') . '.taxonomytype', $key),
                             $this->parseWhereParameter($this->getTablename('taxonomy') . '.slug', $value),
                             $this->parseWhereParameter($this->getTablename('taxonomy') . '.name', $value),
-                            $this->parseWhereParameter($this->getTablename('taxonomy') . '.contenttype', $contenttype['slug'])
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Hi @hrjt! Thank you for this fix! 👍

PR #7910, opened 2021-05-19
Improve Yaml export for compatibility with Bolt 4/5 Conimex import

bobdenotter:


src/Storage/Migration/Export.php
```diff
@@ -124,6 +130,7 @@ private function getRecords($contentTypeName, MutableBag $exportData, MutableBag
         $repo = $this->em->getRepository($contentTypeName);
         $metadata = $repo->getClassMetadata();
         // Get all the records for the ContentType
+        // $entities = $this->query->getContent($contentTypeName, ['limit' => 100, 'order' => '-id']);
```
nestordedios: I think this comment can be removed. For the rest everything looks good.

bobdenotter:
```suggestion
```

bobdenotter: Done!

nestordedios (approved):
Thanks @bobdenotter ! This looks good 👍

PR #7916, opened 2021-06-08
Fix querying by ContentType slug

I-Valchev:
Patches #7893, by allowing the `name` OR the `slug` to be used for querying.


bobdenotter:
🙈

PR #7925, opened 2021-07-14
fix bolt/bolt#7924 - php 7.4 incompatibility of legacy storage code

### LINKED ISSUE #7924: Bolt CMS 3.7.4 incompatibility with PHP 7.4 - Trying to access array offset on value of type bool
Composer installation of Bolt CMS 3.7.4 throws an error of 

> ContextErrorException in Storage.php line 1065:
> Notice: Trying to access array offset on value of type bool

when  trying to access the homepage if NGINX is configured to use php7.4-fpm. This occurs even using one of the default themes (i.e. base-2018).

The error does not occur with php-7.3-fpm.

Screenshot 2021-07-13 at 17 24 31

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.7.4
| Install type                   |  Composer
| BC Break                     | ?
| PHP version                 | 7.4
| Web server                  | NGINX
| For UX/UI issues        | N/A - all

Reproduction
------------

### Steps
Install Bolt CMS 3.7.4
Configure Nginx to process PHP files using the variable `fastcgi_pass unix:/var/run/php/php7.4-fpm.sock;`
Note the error shown above.
Configure Nginx to process PHP files using the variable `fastcgi_pass unix:/var/run/php/php7.3-fpm.sock;`
Note that the error does not occur.

### Expected result

Display the homepage

### Actual result

Error handler displays, as above.

Steve-MP:
Fix php7.4 incompatibility in legacy storage code.

src/Legacy/Storage.php line 1065 would try to test the value of an array member without first checking whether the variable contained the given array key. If the variable was empty, this would produce a falsy value prior to PHP7.4 but throw a fatal exception in PHP 7.4.

This PR adds a check that the $contenttype variable is populated and returns a default value if not, mimicking the outcome of the code if run in PHP < 7.4

Fixes bolt/bolt#7924

### DIFF
--- a/src/Legacy/Storage.php
+++ b/src/Legacy/Storage.php
@@ -1062,7 +1062,7 @@ class Storage
             $decoded['return_single'] = true;
             // if allow_numeric_slug option is set on contenttype, interpret number as slug instead of id
             $contenttype = $this->getContentType($decoded['contenttypes'][0]);
-            $field = ($contenttype['allow_numeric_slugs'] === true ? 'slug' : 'id');
+            $field = !empty($contenttype) ? ($contenttype['allow_numeric_slugs'] === true ? 'slug' : 'id') : 'id';
             $ctypeParameters[$field] = $match[2];
         } elseif (preg_match('#^/?([a-z0-9_(\),-]+)/search(/([0-9]+))?$#i', $textquery, $match)) {
             // like 'page/search or '(entry,page)/search'


bobdenotter (approved):
Hi @Steve-MP,  

I figured we had solved most 7.4 quirks by now, but apparently not. Thanks! 👍

PR #7926, opened 2021-07-14
Fix bolt/bolt#7902 PHP 7.4 exception on empty search

### LINKED ISSUE #7902: Search when querystring empty throws error on PHP 7.4
When trying to access the /search page on a Bolt install and not providing no search query string or an empty one, ie `/search?search=` or just `/search` it throws an 'ContextErrorException' on line 548 of Frontend.php and states `Trying to access array offset on value of type bool.`

Details
-------

|----------------- | ------
| Relevant Bolt Version | 3.7.2
| PHP version                 | 7.4.16
| Web server                  | Apache 2.4.41

Reproduction
------------

Setup a Bolt install and visit /search without providing a search query string and on < PHP 7.4 it shows the no results found page, switch to PHP 7.4 and instead it throws the following error.

### Bug summary

When trying to access the /search page on a Bolt install and not providing no search query string or an empty one, ie `/search?search=` or just `/search` it throws the below error.

```
Uncaught Exception: ContextErrorException .

ContextErrorException in Frontend.php line 548:
Notice: Trying to access array offset on value of type bool
```

Issue shows as line 548 of Frontend.php where it sets `->setCount($result['no_of_results'])`

### Steps to reproduce

Setup a Bolt install and visit /search without providing a search query string and on < PHP 7.4 it shows the no results found page, switch to PHP 7.4 and instead it throws the above error.

### Expected result

The search page renders and shows the message for no search results.

### Actual result

Screenshot attached. Only seems to be an issue in PHP 7.4. Have not tried anything higher as I'm not sure Bolt v3 supports anything higher than 7.4.

Screenshot 2021-03-12 at 10 36 43

Steve-MP:
PHP 7.4 throws an exception when someone tries to access an array item
on a variable of boolean type. An empty search string results in a "false"
value being allocated to the search results, rather than the expected array.
This commit adds a check that the results array
is populated before attempting to access an array member.

Fixes bolt/bolt#7902

### DIFF
--- a/src/Controller/Frontend.php
+++ b/src/Controller/Frontend.php
@@ -541,12 +541,14 @@ class Frontend extends ConfigurableBase
         if ($isLegacy) {
             $result = $this->storage()->searchContent($q, $contenttypes, $filters, $limit, $offset);
 
+            $resultcount = isset($result['no_of_results']) ? $result['no_of_results'] : 0;
+
             /** @var \Bolt\Pager\PagerManager $manager */
             $manager = $this->app['pager'];
             $manager
                 ->createPager($context)
-                ->setCount($result['no_of_results'])
-                ->setTotalpages(ceil($result['no_of_results'] / $pageSize))
+                ->setCount($resultcount)
+                ->setTotalpages(ceil($resultcount / $pageSize))
                 ->setCurrent($page)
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Thanks again, @Steve-MP 👍

Issue #7972, opened 2023-03-22
app/nut database:export outputs unexpected results for multi-contenttype select fields


simongroenewolt:
When running 

    app/nut database:export --file=export.yml --users

I noticed the output of many 'select' fields was a list of 'empty' fields like this:

```
routes:
    -
        slug: 4f33139afbd38339
        datecreated: '2020-10-01 10:05:41'
        datechanged: '2020-10-30 09:56:22'
        datepublish: '2020-10-01 10:04:57'
        datedepublish: ''
        ownerid: 3
        status: published
        templatefields: {  }
        spaces:
            - {  }
            - {  }
            - {  }
            - {  }
            - {  }
            - {  }
            - {  }
            - {  }
            ...
```

Details
-------

| Question                      | Answer
|----------------- | ------
| Relevant Bolt Version | 3.7.5
| Install type                   |  Composer install
| BC Break                     |  no
| PHP version                 | 7.4

Reproduction
------------

Create a contenttype with a specification like this:

```
routes:
    name: Routes
    singular_name: Route
    fields:
        spaces:
            type: select
            values: (contenttype1,contenttype2,contenttype3)/ref
            keys: slug
            autocomplete: true
            multiple: true
            sortable: true
            label: Ruimtes
```

Then add a couple of entries to the 'spaces' of contenttype1/2/3 and export using the `database:export` command.

It looks like the export command cannot handle select fields that have multiple contenttypes as 'target'. 

Database fields for a select field targeting a single type look like this:

> ["2b391388d903331e","76ecac4f0c7e1593","b3e554ba84186142","235fcaf57bb4adf7","5462482fe2c0c61e","57e5716ffe1b5d51","d917543c11b341a7","860efdd2fff5642f","b0f99f60982d23df","15cb8ca97ba84211","533c98e2c949b4cf","cf92292976629202"]

Database fields for a select field targeting multiple types look like this:

> ["reflections/87c7bf2b8ebce56a","connections/107c618c639e021a","rooms/be777b7bb2952d17","rooms/63e5e415afe9695a","rooms/c3987cee9c663574","rooms/df2e286b5b1f6aab","connections/c1d75a2c79b4dd29","connections/c51308d0bdd029bb","rooms/63ec3d3e42c894bc","rooms/789480da0181c5c1","rooms/74b4c2a6b6738583"]

[...]

PR #7973, opened 2023-06-27
[WIP] update/bugfix to Export.php that enables exporting select fields that reference multiple other contenttypes

simongroenewolt:
This PR updates the Export.php code so it uses a different way to look up (select) field content if it references other contenttypes. This was needed for my project to correctly export select fields that reference multiple content types. The current version would _not_ export any valid data for my select fields.

See https://github.com/bolt/bolt/issues/7972

I don't know if Bolt 3.7 is still being updated for this kind of stuff, and I don't have any experience using 3.7 myself so testing of this fix was limited to my current project only.

Labeled WIP just to warn this is not thoroughly tested and should not just be merged. Waiting for some discussion before marking it as a 'real' PR.

### DIFF
--- a/src/Storage/Migration/Export.php
+++ b/src/Storage/Migration/Export.php
@@ -217,36 +217,36 @@ final class Export
         if (is_array($values)) {
             $val = [];
             foreach ($values as $value) {
-                $reference = $contentTypeName . '/' . $value;
-                $val[] = $this->getContent($reference, $contentTypeName, (string) $value);
+                if (strpos($value, '/') !== false) {
+                    $reference = $value;
+                } else {
+                    $reference = $contentTypeName . '/' . $value;
+                }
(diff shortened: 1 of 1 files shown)


bobdenotter:
Thanks for the PR, @simongroenewolt. 

The change looks good at first glance, but i don't recall if we've ever bumped into this specific issue. I'll see if i can test it, and i'll release a minor update for Bolt 3.7.x. :-)

simongroenewolt:
@bobdenotter the content type that resulted in issues for me had a field like this:

```
routes:
    slug: routes
    name: Routes
    singular_name: Route
    fields:
        // ...
        spaces:
            type: select
            values: (rooms,reflections,connections)/{ref}
            keys: slug
            autocomplete: true
            multiple: true
            sortable: true
            label: Ruimtes
```

